PR #196, opened 2014-11-05
Renamed TransactionNamer.name class method.

fimmtiu:
The TransactionNamer class is rather poorly-behaved; it overrides the built-in `name` method with something completely different, thus preventing anyone from asking the class object what it's called. Observe:

```
1.9.3-p484 :001 > String.name
 => "String" 
1.9.3-p484 :002 > NewRelic.name
 => "NewRelic" 
1.9.3-p484 :003 > NewRelic::Agent::Instrumentation::ControllerInstrumentation::TransactionNamer.name
ArgumentError: wrong number of arguments (0 for 3)
```

Not a hypothetical problem; it bit me while I was instrumenting classes to find a memory leak. The fix is simple -- just rename the method.

### DIFF
--- a/lib/new_relic/agent/instrumentation/controller_instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/controller_instrumentation.rb
@@ -213,7 +213,7 @@ module NewRelic
         end
 
         class TransactionNamer
-          def self.name(txn, traced_obj, category, options={})
+          def self.txn_name(txn, traced_obj, category, options={})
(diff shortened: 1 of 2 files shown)


backus:
@benweint and @jasonrclark I opened a thread on the newrelic forums regarding the same issue for different classes: https://discuss.newrelic.com/t/fix-class-name-convention-violations-in-ruby-library/38060/1

Any chance you could fix these classes as well?

PR #198, opened 2014-11-18
prevent call on nil when transaction was closed

mikz:
Transaction can be closed from within the application (for example on Rack::Timeout error).
Then it tried to call `http_response_code` on nil.

> , [2014-11-18T12:26:43.036412 #28322] ERROR -- : app error: undefined method `http_response_code=' for nil:NilClass (NoMethodError)
> E, [2014-11-18T12:26:43.036617 #28322] ERROR -- : /home/bender/brain/shared/bundle/ruby/2.1.0/gems/newrelic_rpm-3.9.7.266/lib/new_relic/agent/instrumentation/middleware_tracing.rb:61:in `call'
> E, [2014-11-18T12:26:43.036738 #28322] ERROR -- : /home/bender/brain/shared/bundle/ruby/2.1.0/gems/unicorn-4.8.3/lib/unicorn/http_server.rb:576:in `process_client'

### DIFF
--- a/lib/new_relic/agent/instrumentation/middleware_tracing.rb
+++ b/lib/new_relic/agent/instrumentation/middleware_tracing.rb
@@ -57,7 +57,7 @@ module NewRelic
               result = target.call(env)
             end
 
-            if result.is_a?(Array)
+            if result.is_a?(Array) && state.current_transaction
               state.current_transaction.http_response_code = result[0]
             end
 


yuku-t:
Same problem here. I use the gem without touching objects in `NewRelic` module in my ruby code. 
I'm not sure this patch fixes my problem, but it is worth merging it because it does not change behavior in correct situation.

benweint:
Hi @yuku-t - we are planning to incorporate this change, but it's really only a safeguard against a cascade of an earlier failure. Specifically: if there's an exception during `Transaction.start` before we are able to set the `current_transaction`.

If you're hitting that case, there should be a bunch of backtraces an exception messages in your `newrelic_agent.log` file, and we'd very much like to see them if possible. Would you mind opening a ticket at [support.newrelic.com](https://support.newrelic.com) and attaching your `newrelic_agent.log` file, along with any other details about the exceptions you're seeing? Getting that information would help us get to the root cause of the failure here. Thanks!

benweint:
Hi @mikz - this change has been released as part of version 3.10.0 of the newrelic_rpm gem (I had to rebase against some of our more recent changes, so the commit hash changed, but you can see the commit [here](https://github.com/newrelic/rpm/commit/0801d0e26f3831c21df6004dd2dcb253785b8afb)). I'm going to close this out since the change has now been merged. Thanks again for submitting this!

PR #199, opened 2014-12-31
Add support for /system/lxproc on Illumos

jdelStrother:
Illumos-based systems have relatively poor fork performance, so shelling out to ps to read the RSS isn't great.  However, they support variant of `/proc` in `/system/lxproc`.  How about the attached?

### DIFF
--- a/lib/new_relic/agent/samplers/memory_sampler.rb
+++ b/lib/new_relic/agent/samplers/memory_sampler.rb
@@ -20,10 +20,7 @@ module NewRelic
           elsif platform =~ /linux/
             @sampler = ProcStatus.new
             if !@sampler.can_run?
-              ::NewRelic::Agent.logger.debug "Error attempting to use /proc/#{$$}/status file for reading memory. Using ps command instead."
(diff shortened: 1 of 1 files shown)


jdelStrother:
Hm, it sounds like it's actually only available on SmartOS at the moment.  They supply VMs here - http://wiki.smartos.org/display/DOC/Download+SmartOS

kwugirl:
Hi @jdelStrother! I've been going through and trying to update that various outstanding PRs on this repo--is this issue still a concern for you? If so, additional instructions or information on verifying this change would be greatly appreciated, as none of us on the team currently have experience with illumos or SmartOS, I believe.

jdelStrother:
Hi there - yeah, it's still a concern for us, though we've been running my
fork in production without issues.   What do you need help with for
verifying it?

kwugirl:
Cool, glad to hear that your fork has been running without any issues! When you said "it's actually only available on SmartOS at the moment," did you mean that `/system/lxproc` is only available on SmartOS of the various illumos distros?

tkrajcar:
Closing for now, no response.

PR #200, opened 2015-02-17
Initial support for Lotus Framework

joneslee85:
First attempt to get Lotus supported - WIP

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ automatically in Rails applications.
 * Ruby 1.8.7, REE, 1.9.x, 2.0.x, 2.1.x, 2.2.x
 * JRuby 1.6 and 1.7
 * Rubinius 2.x (Experimental support only)
-* Lotus 0.2 or later
+* Lotus 0.3.0 or later (Production Mode only)
 * Rails 2.1 or later for Production Mode
 * Rails 2.3 or later for Developer Mode
 * Sinatra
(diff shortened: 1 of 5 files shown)


joneslee85:
@benweint Lotus is employing Shotgun for code reloading and AFAIK this blocks NewRelic from working. Is there any plan to overcome this issue? Cheers

joneslee85:
It's good for review. We've been testing at Envato and things work great.

screen shot 2015-05-11 at 11 26 18 am


There are 2 failed build with JRuby, but seems irrelevant to my changes/

jasonrclark:
Hi @joneslee85, thanks for sending this along. Unfortunately, with Lotus still solidly pre-1.0, I don't think newrelic_rpm is ready to start supporting instrumentation for it.

The good news, though, is that from my investigation those agent changes shouldn't be strictly necessary for instrumenting Lotus. I set up a test app at https://github.com/jasonrclark/lotus-test, which successfully reported data to New Relic against the unmodified 3.12.0.288 version of the agent. This works thanks to our automatic Rack instrumentation. As long as New Relic gets loaded early enough, we can see web transactions.

The changes you made in the "framework" classes, aren't actually necessary for the agent to load, although admittedly it does miss out on defaulting to LOTUS_ENV. In the meantime I'd recommend setting RACK_ENV (or NEW_RELIC_ENV) to the same value during startup prior to requiring newrelic_rpm.

I imagine there's a lot more that Lotus instrumentation could do, including better transaction naming and model-level instrumentation. If you want to pursue that, I'd recommend making a third-party extension gem. We've got some docs at https://docs.newrelic.com/docs/agents/ruby-agent/frameworks/third-party-instrumentation to get you started, and I'm happy to answer other questions if that doesn't cover it.

Also, you are right that we don't work nicely with Shotgun. The level of effort to support a mode that's only used in development makes that unlikely to change anytime soon.

Thanks again, and let me know if I can help out in any of those other avenues.

PR #204, opened 2015-02-23
Added intrumentation for CAS variants of Dalli methods

zkamsler:
Dalli::Clients has variants of many of its normal methods that either
return or use the revision number from Memcached that can be used for
compare-and-swap operations. Instrumenting these methods ensures that
the information that NewRelic gets is not biases by their absence.

### DIFF
--- a/lib/new_relic/agent/instrumentation/memcache.rb
+++ b/lib/new_relic/agent/instrumentation/memcache.rb
@@ -77,6 +77,7 @@ DependencyDetection.defer do
       ::NewRelic::Agent.logger.info 'Installing Memcached instrumentation'
     end
     if defined? ::Dalli::Client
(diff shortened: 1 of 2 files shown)


jasonrclark:
Hi @zkamsler, thanks for the contribution.

It turns out we were revising the instrumentation on Memcache very significantly in the next release, so I've made sure that those methods are included. Unfortunately the code has moved a lot since then, so I won't be merging this, but will keep it open until I can point you to a new release with support for those methods.

Thanks again!

jasonrclark:
Hey @zkamsler! Release 3.11.0 of newrelic_rpm just went out which should incorporate code to cover these methods. Let us know if you see any problems with this. Thanks again for the contribution!

PR #205, opened 2015-03-20
Test on Ruby 2.2 and auto test on patched versions

nicolasleger:


jasonrclark:
Thanks @nicolasleger! I've pulled that into our private repo, and these changes should show up in an upcoming release. We use a different CI system internally, so hadn't noticed that Travis had lagged on 2.2. :+1:

PR #207, opened 2015-04-15
allow hostname to be overridden in config file

kitchen:
we have a bunch of servers named 'app1'. They all have unique FQDNs though. The problem is that in the NR web interface they are all just 'app1', and they also don't match up with our nrsysmond servers because those are properly configured to use the fqdn. This allows us to override the hostname ourselves, which we'll do when we generate the newrelic.yml file using puppet.

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1173,6 +1173,13 @@ module NewRelic
           :public       => true,
           :type         => Boolean,
           :description  => 'Disables installation of Grape instrumentation.'
+        },
+        :hostname => {
+          :default      => nil,
+          :allow_nil    => true,
+          :public       => true,
+          :type         => String,
+          :description  => 'Override automatic hostname detection.'
         }
       }.freeze
     end
--- a/lib/new_relic/agent/hostname.rb
+++ b/lib/new_relic/agent/hostname.rb
@@ -7,7 +7,10 @@ module NewRelic
     module Hostname
       def self.get
         dyno_name = ENV['DYNO']
-        if dyno_name && ::NewRelic::Agent.config[:'heroku.use_dyno_names']
+        hostname = NewRelic::Agent.config[:hostname]
+        if hostname
+          return hostname
+        elsif dyno_name && ::NewRelic::Agent.config[:'heroku.use_dyno_names']
           matching_prefix = heroku_dyno_name_prefix(dyno_name)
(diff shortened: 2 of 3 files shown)


kyanny:
Hi @kitchen,
As you may know, you can override app name by `NEWRELIC_APP_NAME` environment variable.
(according to [CHANGELOG](https://github.com/newrelic/rpm/blob/master/CHANGELOG), this feature has coming from v2.10.3).

kitchen:
Yes but this is host name, not app name.

-Jeremy

--
Insert PGP signature here.

kyanny:
I know, but as far as I know, New Relic shows app name in its web interface.

kitchen:
there are plenty of places where newrelic uses the hostname of the machine. For instance, we also use nrsysmond, which reports the machine as its fqdn. Because rpm will see "worker1.foo.bar.com" as "worker1" and nrsysmond will see it as "worker1.foo.bar.com", server monitoring fails to connect the two reports as coming from the same machine.

screen shot 2015-04-16 at 10 47 49 am

Above is our list of servers in our application (grouped by app_name)

screen shot 2015-04-16 at 10 49 01 am

and this is the list of applications running on one of those servers. You'll notice that it doesn't join up. This is because rpm isn't reporting the fqdn of the machine, only "worker2" (in this instance)

Allowing one to specify the hostname desired in the config file allows newrelic to work properly for us, and also removes a lot of ambiguity in that we have multiple machines with the same 'hostname' but different fully qualified domain names.

kitchen:
any update on this?

benweint:
Hey @kitchen, thanks for sending this along, and sorry for the delay in getting back to you about this!  I've been talking to a number of folks internally to try to figure out the best way of resolving this.

The bad news is that we're not going to take this change as-is. We definitely recognize that there's a lack of flexibility with our current data aggregation policies, and there are plans afoot to improve the situation, but there are a number of pieces involved throughout our whole software stack that we want to coordinate these changes with.

The good news is that we are planning a change in the near-term to address your specific use case. We're planning to add a setting to the Ruby agent to allow you to have the agent report a FQDN, instead of just the hostname part, which should disambiguate your servers. The nice part of this is that you'll be able to just blanket set this setting across all of your hosts (via the `newrelic.yml` file or an environment variable) without having to use a configuration management system or other technique to have the setting be different on each host.

I'm going to close out this PR, but we will ping you back here as soon as the FQDN setting is available in a released version of the agent.

PR #208, opened 2015-04-28
Test case to show if rescue_from is used in grape, request params aren't set.

akshah123:
Added test cases to demonstrate that newrelic_rpm doesn't capture params if grape controller uses rescue_from.

### DIFF
--- a/test/multiverse/suites/grape/grape_test.rb
+++ b/test/multiverse/suites/grape/grape_test.rb
@@ -100,6 +100,23 @@ class GrapeTest < Minitest::Test
       end
     end
 
+    def test_post_body_params_are_captured_with_error
+      with_config(:capture_params => true) do
+        assert_raises(GrapeTestApiError) do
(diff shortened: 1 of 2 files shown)


jasonrclark:
Hey @akshah123, thanks for passing that along!

We're in the final stages of a new agent version that changed a lot about parameter capture. I plugged your test in, updated it slightly for some internal agent changes, and everything looks green for 3.12.0.

I'll keep this open until the release is out so we can ping you. Would love to have you test once the new agent is available to confirm the problem you saw is resolved.

Thanks again for the PR!

akshah123:
@jasonrclark that's great news!  I will be sure to report back once I have had a chance to test out 3.12.0.

jasonrclark:
Hey @akshah123, we just released 3.12.0.288 of the agent which includes a modified version of the tests you provided, and which passes. Let us know how it goes when you get a chance to test it out, and thanks again for the contribution! :sparkles:

akshah123:
@jasonrclark looks to be working on test setup. I will report back if I see something different on production.  Appreciate the quick turn around.

PR #209, opened 2015-05-22
Add in-line comment fixture

thommahoney:
This commit adds a new fixture that demonstrates the expected behavior
using a query from the MySQL documentation page on comments:
https://dev.mysql.com/doc/refman/5.6/en/comments.html

### DIFF
--- a/test/fixtures/cross_agent_tests/sql_obfuscation/sql_obfuscation.json
+++ b/test/fixtures/cross_agent_tests/sql_obfuscation/sql_obfuscation.json
@@ -292,6 +292,18 @@
     "pathological": true
   },
   {
+    "name": "pathological/inline_comments",
+    "obfuscated": [
+      "SELECT ? ? + ?"
+    ],
+    "dialects": [
+      "mysql",
+      "postgres"
+    ],
+    "sql": "SELECT 1 /* this is an in-line comment */ + 1",
+    "pathological": true
+  },
+  {
     "name": "pathological/mixed_comments_and_quotes",
     "obfuscated": [
       "SELECT * FROM t WHERE ?"


drbrain:
In particular, we're making use of `/*! MySQL-specific code */` to enable MySQL-specific query optimizations.

benweint:
Thanks for reporting this!

This is unfortunately a difficult problem to solve.

The short version is that in order to mask out literal values in SQL queries, we need to be able to identify where strings start and end. Without knowledge of what quoting mode is in use by the database server (for MySQL, specifically the value of the `NO_BACKSLASH_ESCAPES` setting), it turns out to be impossible to reliably determine where string boundaries are in the presence of comments (which may include embedded string delimiters).

To work around this, we've taken the conservative approach of masking all text after the first comment-initiator sequence that we see in any query. As you've seen, this over-obfuscates in some cases.

Do your optimization-enabling comments *need* to be inline within the query rather than at the end?

There are alternate obfuscation heuristics we could use, but all of them come with tradeoffs, so we're hesitant to optimize for the inline-comment case unless it's something that's widely used.

Note that we do support using a custom SQL obfuscation strategy by calling `NewRelic::Agent::Database::Obfuscator.set_sql_obfuscator(...)`. This is a pretty old API and is not currently tagged as public, though it looks like it should be. [This comment](https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/database/obfuscator.rb#L27-L37) describes how to use it.

If you're certain that your inline comments will never contain embedded quote delimiters, you could probably achieve what you're after like this:

```ruby
NewRelic::Agent::Database::Obfuscator.instance.set_sql_obfuscator(:before) do |sql|
  sql.gsub(/\/\*(?:[^\/]|\/[^*])*?(?:\*\/|\/\*.*)/, '?')
end
```

kwugirl:
I'm going to close out this pull request since it seems there wasn't further discussion here, but do let us know if there's still a concern.

PR #210, opened 2015-05-28
include non-default ports in host

dwo:
This is a really ugly and incomplete way of achieving what I'm after, but I wanted to at least start some discussion about this.

At SK, we have a lot of internal services running off of one hostname, but on different, non-default ports. When the agent captures requests made through our HTTP Client (curb) for the External Services report, they all just get lumped under that single hostname.

This would allow one to look into requests to `http://host.name:1234` separately from requests to  `http://host.name:4321`.

Is there a neater way we could implement this across the board for all http clients?

### DIFF
--- a/lib/new_relic/agent/http_clients/curb_wrappers.rb
+++ b/lib/new_relic/agent/http_clients/curb_wrappers.rb
@@ -16,7 +16,7 @@ module NewRelic
         end
 
         def host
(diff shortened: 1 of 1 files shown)


dwo:
This is probably a pain, and I can probably work around this by configuring some more CNAMES.

PR #211, opened 2015-06-18
Integration with MongoDB Ruby Driver 2.1

durran:
This is for example only.

This is just a pull as a talking point on how I would envision the rpm gem would integration with the 2.1.0 Mongo Ruby driver once it is released. I have not written any tests yet, just wanted more to look at with the ongoing conversations we are having as to what this might look like.

### DIFF
--- a/lib/new_relic/agent/instrumentation/mongo.rb
+++ b/lib/new_relic/agent/instrumentation/mongo.rb
@@ -23,37 +23,9 @@ DependencyDetection.defer do
     end
   end
 
-  class CommandSubscriber
-
-    def started(event)
-      operations[event.operation_id] = event
-    end
-
-    def succeeded(event)
-      started_event = operations.delete(event.operation_id)
-      NewRelic::Agent.instance.transaction_sampler.notice_nosql_statement(
-        format_statement(started_event.command), event.duration
-      )
-    end
-
-    def failed(event)
-      started_event = operations.delete(event.operation_id)
-      # How does new relic want to handle failures?
-    end
-
-    private
-
-    def operations
-      @operations ||= {}
-    end
-
-    def format_statement(event)
-      "#{event.database_name}.#{event.command_name} #{event.command.inspect}"
-    end
-  end
-
   def install_mongo_command_subscriber
-    Mongo::Monitoring::Global.subscribe(Mongo::Monitoring::COMMAND, CommandSubscriber.new)
+    require 'new_relic/agent/instrumentation/mongodb_command_subscriber'
+    Mongo::Monitoring::Global.subscribe(Mongo::Monitoring::COMMAND, MongodbCommandSubscriber.new)
   end
 
   def install_mongo_instrumentation
--- a/lib/new_relic/agent/instrumentation/mongo.rb
+++ b/lib/new_relic/agent/instrumentation/mongo.rb
@@ -25,7 +25,10 @@ DependencyDetection.defer do
 
   def install_mongo_command_subscriber
     require 'new_relic/agent/instrumentation/mongodb_command_subscriber'
-    Mongo::Monitoring::Global.subscribe(Mongo::Monitoring::COMMAND, MongodbCommandSubscriber.new)
+    Mongo::Monitoring::Global.subscribe(
+      Mongo::Monitoring::COMMAND,
+      NewRelic::Agent::Instrumentation::MongodbCommandSubscriber.new
+    )
   end
 
   def install_mongo_instrumentation
--- a/lib/new_relic/agent/instrumentation/mongo.rb
+++ b/lib/new_relic/agent/instrumentation/mongo.rb
@@ -16,7 +16,44 @@ DependencyDetection.defer do
 
   executes do
     NewRelic::Agent.logger.info 'Installing Mongo instrumentation'
-    install_mongo_instrumentation
+    if NewRelic::Agent::Datastores::Mongo.is_monitoring_enabled?
+      install_mongo_command_subscriber
+    else
+      install_mongo_instrumentation
+    end
+  end
+
+  class CommandSubscriber
+
+    def started(event)
+      operations[event.operation_id] = event
+    end
+
+    def succeeded(event)
+      started_event = operations.delete(event.operation_id)
+      NewRelic::Agent.instance.transaction_sampler.notice_nosql_statement(
+        format_statement(started_event.command), event.duration
+      )
+    end
+
+    def failed(event)
+      started_event = operations.delete(event.operation_id)
+      # How does new relic want to handle failures?
+    end
+
+    private
+
+    def operations
+      @operations ||= {}
+    end
+
+    def format_statement(event)
+      "#{event.database_name}.#{event.command_name} #{event.command.inspect}"
+    end
+  end
+
+  def install_mongo_command_subscriber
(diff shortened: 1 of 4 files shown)


lib/new_relic/agent/instrumentation/mongo.rb
```diff
@@ -16,7 +16,44 @@
...
+      install_mongo_command_subscriber
+    else
+      install_mongo_instrumentation
+    end
+  end
+
+  class CommandSubscriber
+
+    def started(event)
+      operations[event.operation_id] = event
```
jasonrclark: Is there a reason that we'll need to hold onto the events between `started` and the finishing callbacks? Seems like there's a potential problem if, for whatever reason, the agent failed to see `succeeded` or `failed` for a command it had seen starting.

If all the necessary information is on the `event` at the end, could we just use it from there?

lib/new_relic/agent/instrumentation/mongo.rb
```diff
@@ -16,7 +16,44 @@
...
+    def succeeded(event)
+      started_event = operations.delete(event.operation_id)
+      NewRelic::Agent.instance.transaction_sampler.notice_nosql_statement(
+        format_statement(started_event.command), event.duration
+      )
+    end
+
+    def failed(event)
+      started_event = operations.delete(event.operation_id)
+      # How does new relic want to handle failures?
```
jasonrclark: We'd still probably want to do the normal timing, even for a failed command. If the command raises an exception elsewhere, our error tracing can catch that there. You'd want to still see it show up in the transaction's time spent.

lib/new_relic/agent/instrumentation/mongo.rb
```diff
@@ -16,7 +16,44 @@
...
+
+    def started(event)
+      operations[event.operation_id] = event
+    end
+
+    def succeeded(event)
+      started_event = operations.delete(event.operation_id)
+      NewRelic::Agent.instance.transaction_sampler.notice_nosql_statement(
+        format_statement(started_event.command), event.duration
+      )
```
jasonrclark: There's some good samples of the steps you'd want to actually take to record over in https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/instrumentation/active_record_subscriber.rb#L53-L83. You would want to use `NewRelic::Agent::Datastore::MetricHelper.metrics_for` instead of the ActiveRecord flavor, but overall the steps would be very similar.

lib/new_relic/agent/instrumentation/mongo.rb
```diff
@@ -16,7 +16,44 @@
...
+    end
+
+    private
+
+    def operations
+      @operations ||= {}
+    end
+
+    def format_statement(event)
+      "#{event.database_name}.#{event.command_name} #{event.command.inspect}"
```
jasonrclark: The existing instrumentation for 1.x has support for [controlling capture and obfuscation of the queries](https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/datastores/mongo/obfuscator.rb) based on a [config setting](https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/configuration/default_source.rb#L837-L850). Will probably want to behave according to those here as well.

jasonrclark:
Hey @durran! Left some thoughts on what you proposed. Assuming the subscriber gets all the information we want, this looks very plausible.

Am I correct in thinking these events are dispatched synchronously from the command execution?

lib/new_relic/agent/instrumentation/mongo.rb
durran: The completed event does not have the command that was executed, so it does need it. We guarantee that there will always be a completed or failed event for every started event that fires. The only exception would be an application crash, in which in this case would not matter.

durran:
Yes the events are dispatched synchonously. We haven't covered asnyc yet and would be in a later specification from us on how we see that handled.

lib/new_relic/agent/instrumentation/mongo.rb
jasonrclark: Ah, I wasn't recognizing that the events were different between the start and finish. That makes sense.

durran:
@jasonrclark I've updated and added unit test... The one last thing I might need some guidance on is how you want to approach the multiverse tests... They are very specific to previous versions of the driver and since the API changed I was curious on the best way you thought to handle it. It's not as easy as AR where everything goes from the Model level and thus API changes don't matter. :)

tkrajcar:
@durran Yeah, looking at the existing `mongo` multiverse suite, I think in this case it makes sense to leave that alone, since we're not planning to drop that support, and start a new mongo2 suite with a reasonable amount of end-to-end coverage of the subscriber behavior. We like to not do that when we can avoid it, but in this case it makes sense to me!

mwear:
@durran  I'll pull this into our private development repo and begin setting up a multiverse suite for it.  If I run into any trouble or have questions I'll get in touch with you.  Thanks for your work this!

mwear:
@durran 

I've pushed up a branch that contains your changes and multiverse tests for Mongo 2 that has been rebased against master here: https://github.com/newrelic/rpm/tree/RUBY-1498-mongo-multiverse. A number of the tests are failing as the subscriber does not generate the metrics that we expect.  The failures are generally the result of one of three scenarios.

* The method indicated by the subscriber is not the method invoked in the driver for many of the cases.  For example calls to `insert_one` and `insert_many` both map to `insert` which doesn't match either the methods invoked by the user. It would be ideal if the `command_name` on the event mapped to the actual method name.
* The instrumentation determines the collection from `event.command.values.first`, which seems a little fragile, and is not always the collection.  For example, when calling `find`, `event.command.values.first` is the value of the query.
* Many of the test failures are due to the value that `format_statement` passes to `NewRelic::Agent.instance.transaction_sampler.notice_nosql_statement`. A hash as formatted by `NewRelic::Agent::Datastores::Mongo::StatementFormatter should be passed to `NewRelic::Agent.instance.transaction_sampler.notice_nosql_statement` rather than the string that is currently being generated.  Take a look at this file for more details: https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/datastores/mongo/statement_formatter.rb

Addressing the issues above should get the multiverse tests passing.  Here are some tips on running tests from mulitverse.

To run the mongo suite only, type:

```
bundle exec rake test:multiverse[mongo]
```

To run only the mongo 2 tests, type:

```
bundle exec rake test:multiverse[mongo,env=0]
```

To run a specific test from mongo 2, type:

```
bundle exec rake test:multiverse[mongo,env=0,name=test_records_metrics_for_insert_one]
```

To run the suite with pry available (note you will still need to add `require 'pry'; binding.pry`:

```
bundle exec rake test:multiverse[mongo,debug]
````
Please pull this into your fork and get the multiverse tests passing.  If you have any questions let us know.

durran:
Thanks - I'll pull the branch down and get to it. :)

kwugirl:
Hey @durran would it be all right to close this particular PR, since it looks like you finished up the work in https://github.com/newrelic/rpm/pull/214? Just wanted to make sure you didn't have other plans for what you have in here, but it sounded like you pushed this up to get initial feedback which you did receive from Matt.

durran:
@kwugirl Yes this can be closed - sorry. :)

kwugirl:
Not at all, I was just going through and wanted to clean up our open PR list a little bit. Thank you!

PR #212, opened 2015-07-27
Make possible to auto-inject script without body tag

vitalikdanchenko:
On my opinion an absence of body tag in beginning of the source shouldn't skip another suitable locations to inject script for browser monitoring. This often make unavailable to use `auto_instrument`, when project contains a lot of inline JS in head of document to speed up page rendering.

I also modified a log message by merging them, because one of them was never used. I hope my variant will fit you.

### DIFF
--- a/lib/new_relic/rack/browser_monitoring.rb
+++ b/lib/new_relic/rack/browser_monitoring.rb
@@ -83,27 +83,23 @@ module NewRelic::Rack
       # Only scan the first 50k (roughly) then give up.
       beginning_of_source = source[0..SCAN_LIMIT]
 
-      if body_start = find_body_start(beginning_of_source)
-        meta_tag_positions = [
-          find_x_ua_compatible_position(beginning_of_source),
-          find_charset_position(beginning_of_source)
-        ].compact
+      meta_tag_positions = [
+        find_x_ua_compatible_position(beginning_of_source),
+        find_charset_position(beginning_of_source)
+      ].compact
 
-        if !meta_tag_positions.empty?
-          insertion_index = meta_tag_positions.max
-        else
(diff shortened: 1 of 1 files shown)


lib/new_relic/rack/browser_monitoring.rb
```diff
@@ -83,27 +83,23 @@ def autoinstrument_source(response, headers, js_to_inject)
...
+        find_x_ua_compatible_position(beginning_of_source),
+        find_charset_position(beginning_of_source)
+      ].compact
 
-        if insertion_index
-          source = source[0...insertion_index] <<
-            js_to_inject <<
-            source[insertion_index..-1]
-        else
-          NewRelic::Agent.logger.debug "Skipping RUM instrumentation. Could not properly determine location to inject script."
```
vitalikdanchenko: It was never showed

kwugirl:
Hi @vitalikdanchenko thanks for this contribution, and apologies it took a little while for us to get back to you! 

I don't think we'll want to remove the scanning for a body tag for this instrumentation as this was settled on as the result of collaboration with the Browser team, to take care with our customers' apps. Is it possible for your inline JS to be at the bottom of the page instead?

I'm curious about the log message though--wouldn't that still come up in your case where there isn't a body tag in the first 50k lines? It would only appear if you've set your logs to the debug level, but it seems like it's possible for that code to get there--unless you're concluding that `insertion_index` never evaluates to false? If it does look like that message would never get hit, please let me know and perhaps we can do some code cleanup there at least.

kwugirl:
Hi @vitalikdanchenko -- I'm going to go ahead and close this PR, but thanks again for your contribution, and please do let us know of other ideas you come across for improving the agent.

PR #213, opened 2015-08-14
Don't use a pager when running git log.  

johnnaegle:
With many commits and no environment or config variable set, deploys will hang trying to report deployments and change sets to new relic.  To finish the deploy, you have to login to the server and kill the git process.  Then the deployment can complete.


tkrajcar:
Thanks for the contribution @johnnaegle! We'll get this pulled into our internal repo and tested. The PR will stay open here until it's released, but it's being reviewed, don't worry. :)

tkrajcar:
@johnnaegle This was released today in 3.13.1.300. Thanks for the contribution!

PR #214, opened 2015-08-18
RUBY-1498: Event publishing with Mongo Ruby driver 2.1.0

durran:
@mwear This addresses most issues at stated here https://github.com/newrelic/rpm/pull/211#issuecomment-132009360

Responses to those questions:

1. `event.command.values.first` is guaranteed by our spec to always be the command name since `BSON::Documents` are ordered - it was an issue in the beta version of the Ruby driver, which is fixed in master and will be in the RC relase. I pointed the multiverse tests at master now instead of the beta.

2. The command that is issued will not always match the event name, as this is low level monitoring and we wanted to avoid any sort of method tracing in order to not break future the agent in future API changes to the driver. We also deemed it more important for a user to know exactly what was happening in the database... Ie a `find` can be a `find`, several `getmore` commands and a `killcursors` command... An `insert_many` can be split up into several `insert` commands, etc. This is the same across all of our drivers now in compliance with the spec.

3. I switched to use the statement formatter but there is one remaining failing test because it uses a whitelist approach instead of a blacklist one. Right now it's going to filter out the command name and collection name from the statements because they are not included in the list. This will include *all* commands sent to the database and any future additional commands. If you agree - I'd like to change this.

### DIFF
--- a/lib/new_relic/agent/datastores/mongo/event_formatter.rb
+++ b/lib/new_relic/agent/datastores/mongo/event_formatter.rb
@@ -21,7 +21,7 @@ module NewRelic
             result = {
               :operation => command_name,
               :database => database_name,
-              :collection => command.values.first
+              :collection => command.values[0]
             }
 
             command.each do |key, value|
--- a/lib/new_relic/agent/datastores/mongo/event_formatter.rb
+++ b/lib/new_relic/agent/datastores/mongo/event_formatter.rb
@@ -19,9 +19,9 @@ module NewRelic
             return nil unless NewRelic::Agent.config[:'mongo.capture_queries']
 
             result = {
-              'operation' => command_name,
-              'database' => database_name,
-              'collection' => command.values.first
+              :operation => command_name,
+              :database => database_name,
+              :collection => command.values.first
             }
 
             command.each do |key, value|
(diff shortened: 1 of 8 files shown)


mwear:
@durran thanks for the work.  I understand both 1 and 2 from the list above and am fine with those explanations.  As for 3, we don't have to use the `StatementFormatter` for this instrumentation, but we need to be able to transform the events from the `CommandSubscriber` into a statement equivalent to what `StatementFormatter` would give us from the old mongo payloads.  If it'd be easier to make a `NewRelic::Agent::Datastores::Mongo::EventFormatter` that does this work, then that's fine.  I'm also open to any other way you would like to solve that problem.  Let me know if you have any questions.

durran:
@mwear I think that is a good idea for #3 - I'll work on that and add to this pull tomorrow - something along the lines of a `NewRelic::Agent::Datastores::Mongo::EventFormatter` as you suggested.

durran:
@mwear Next question. I added the EventFormatter, but wanted to know if in the output the `operation`, `database` and `collection` keys need to be symbols. Right now they are strings.

https://github.com/newrelic/rpm/pull/214/files#diff-0203b51aa1cdea5921861f50ebffc582R21

mwear:
@durran This all looks really good. I think we are fine with string keys for `operation`, `database`, and `collection` as this data is ultimately serialized as JSON when sending it over the wire. I'll verify that in end-to-end testing and on the off chance it does need to be changed it's something I can handle.  Thanks so much for all the work on this.  I'm going to merge this PR.  If I run into any issues I'll be in touch with with you, otherwise thanks again!

durran:
@mwear Great! Let me know if there is anything else you need from me.

PR #216, opened 2015-08-24
Rubinius at_exit handlers

packrat386:
Rubinius can support the `at_exit` block used by `install_exit_handler`.  There is no need to prevent this behavior.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,9 @@ rvm:
   - 2.2
   - jruby-19mode
   - jruby-18mode
+  - rbx-2.5.8
 
+matrix:
+  allow_failures:
+    - rvm: rbx-2.5.8
+      
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,8 @@ rvm:
   - 2.2
   - jruby-19mode
   - jruby-18mode
-  - rbx-2.5.8
+  - rbx-2
 
 matrix:
   allow_failures:
-    - rvm: rbx-2.5.8
-      
+    - rvm: rbx-2
--- a/lib/new_relic/agent/agent.rb
+++ b/lib/new_relic/agent/agent.rb
@@ -370,7 +370,6 @@ module NewRelic
           def should_install_exit_handler?
             (
               Agent.config[:send_data_on_exit]                  &&
-              !NewRelic::LanguageSupport.using_engine?('rbx')   &&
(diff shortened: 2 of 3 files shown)


kwugirl:
Hi @packrat386 thanks for these contributions! We took a look at the history of excluding Rubinius from installing that exit handler and it seems to go pretty far back. Has there been a recent change to Rubinius that supports this or were we perhaps relying on outdated documentation? It'd be great to see confirmation that Rubinius can support the `at_exit` block so we know to look there in the future. 

Also, I'm a little confused by the 2nd commit, "Added rbx to travis as an allowed failure." What was the intent of added Rubinius to Travis, only to set it as an allowed failure?

packrat386:
Adding rbx to travis is to show that all of the tests pass on the most recent version after I made my changes. I added it as an allowed failure because the README says that rbx support is experimental, and I wasn't sure if you wanted to commit to having it as part of your test suite (considering it isn't already there).

As for whether the `at_exit` block is handled correctly in rbx, the tests pass, which should hopefully indicate that the library is working as intended even after adding `at_exit` blocks in an rbx environment. I can find no documentation anywhere of rbx ever *not* supporting `at_exit` blocks since as far as I am aware that haven't been any changes to that feature of the language recently.

kwugirl:
@packrat386 ah ok thanks for the additional explanation! I haven't been on the Ruby Agent team for a;; that long just yet, so I'll have to do a bit more research with the team, but thanks again for the contribution, and we'll let you know if this makes it in to a future release of the agent.

kwugirl:
@brixen, this is the PR that you mentioned to me last night at the PDX.rb meetup, right? And it has your stamp of approval?

brixen:
@kwugirl yes, it is. The only (very minor) change I'd suggest is using `rbx-2` instead of `rbx-2.5.8` because we release often and we highly encourage people to be on the current stable release. If you want, I can send a separate PR for that or @packrat386 could update this one.

Thanks!

kwugirl:
Cool, thanks @brixen and @packrat386 for the update. We'll get this pulled into our internal repo, though the PR will stay open here until it's released.

tkrajcar:
This PR was merged and released in 3.13.1.300 this morning. Thanks @packrat386 for the contribution!

brixen:
Thanks, @tkrajcar!

PR #218, opened 2015-09-08
Prefix support for Heroku Dyno names

stigchristian:
Useful when having multiple Heroku Deployments reporting to the same New Relic application.

### DIFF
--- a/lib/new_relic/agent/hostname.rb
+++ b/lib/new_relic/agent/hostname.rb
@@ -10,7 +10,12 @@ module NewRelic
         if dyno_name && ::NewRelic::Agent.config[:'heroku.use_dyno_names']
           matching_prefix = heroku_dyno_name_prefix(dyno_name)
           dyno_name = "#{matching_prefix}.*" if matching_prefix
-          dyno_name
+          dyno_prefix = ENV['DYNO_PREFIX']
+          if dyno_prefix
+            dyno_prefix + dyno_name
+          else
+            dyno_name
+          end
         else
           Socket.gethostname
         end


tkrajcar:
Hi @stigchristian, thanks for the contribution! Is `DYNO_PREFIX` an environment variable that is set by Heroku? I've never seen it before and don't see it in their docs.

stigchristian:
Hi,

the variable is not defined anywere and has to be set manually, it could probably be a "NEW_RELIC_x" optional variable. However, it will provide our use-case great flexibility and insights.

--Stig

Sendt fra min iPhone

tkrajcar:
Hi @stigchristian, got it! Thanks for the details. We actually have manual host name overrides on our short-term roadmap, which, as you pointed out, will use a `NEW_RELIC_` prefixed variable, and have full docs and test coverage. I'm going to close this for now, but will make a note to make sure and ping this PR when we release the feature, hopefully soon. In the meantime, feel free to run your fork of the agent - it shouldn't be too many versions away. Thanks! Tim

stigchristian:
Ok, great! Is it planned to support Heroku-like naming scheme (i.e. name_prefix + incremental_id)? 

The Heroku ENV-variables are global for the entire deployment, and it is neccessary to differentiate between the dynos.

--Stig 

Sendt fra min iPhone

tkrajcar:
Good to know - we'll certainly consider that use case! Some work needs to be done to ensure we have feature parity with the other New Relic agents, but we can certainly look into it.

PR #220, opened 2015-09-18
Don't block running newrelic agent controller instrumentation on rails 1.2

costi:
The controller instrumentation still works on rails 1.2, even though it's not supported by Newrelic. Actually tested on a production rails 1.2 app.

The dependency check for controller instrumentation should do the same thing like the [view rendering instrumentation] (https://github.com/newrelic/rpm/blob/f1c5b4296531ef39f76fef90744c46b4d3398d47/lib/new_relic/agent/instrumentation/rails/action_controller.rb#L10-L14) here:  
     
       defined?(Rails::VERSION::STRING) && Rails::VERSION::STRING =~ /^(1\.|2\.0)/ # Rails 1.* - 2.0

### DIFF
--- a/lib/new_relic/agent/instrumentation/rails/action_controller.rb
+++ b/lib/new_relic/agent/instrumentation/rails/action_controller.rb
@@ -86,11 +86,11 @@ DependencyDetection.defer do
   end
 
   depends_on do
-    defined?(Rails) && Rails::VERSION::MAJOR.to_i == 2
+    defined?(Rails) && (1..2) === Rails::VERSION::MAJOR.to_i
   end
 
   executes do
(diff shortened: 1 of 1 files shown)


tkrajcar:
Hi @costi, thanks for the contribution. Unfortunately, there's a big difference between "happens to work" and "fully supported" - for us to fully support Rails 1.x again, we would have to be continuously testing all of the agent's features on Rails 1.x, have our support team ready to handle tickets, have supporting documentation, etc. That's just not going to happen for software that's many years out of support and represents a tiny fraction of our market. You're welcome to make local modifications to the agent and run it, and I'm glad that it's working for you!

PR #221, opened 2015-10-23
check for disabled_middleware_instrumentation first

jessesanford:
When using a Mock/Stub/Fake ::Rails class that does not implement the VERSION this will fail unnecessarily even if the config file has disabled_middleware_instrumentation set to true. The re-ordering allows everything to work correctly.

### DIFF
--- a/lib/new_relic/agent/instrumentation/rails_middleware.rb
+++ b/lib/new_relic/agent/instrumentation/rails_middleware.rb
@@ -8,11 +8,11 @@ DependencyDetection.defer do
   named :rails_middleware
 
   depends_on do
-    defined?(::Rails) && ::Rails::VERSION::MAJOR.to_i >= 3
+    !::NewRelic::Agent.config[:disable_middleware_instrumentation]
   end
 
   depends_on do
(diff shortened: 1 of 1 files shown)


kwugirl:
@jessesanford apologies on the very delayed response here, but this looks fine, if it helps you (out of curiosity, when are you using a mock ::Rails class? in testing?).  I'm pulling this into our internal repo, though the PR will stay open here until it's released. Thanks for the contribution!

jessesanford:
There was a time when I was using a mock rails class to create a standard interface for certain internal libs that had not been refactored to remove their rails dependencies. I imagine it would also happen in testing. However I don't know why someone would be using newrelic in their testing environments?

kwugirl:
@jessesanford ah ok, that makes sense, thanks for the extra detail. Yeah, we try to blacklist certain rake tests and other executables so that the agent isn't started up in a test environment, but it's a best effort kind of thing.

kwugirl:
This PR was merged and released in 3.15.0.314 this morning. Thanks @jessesanford for the contribution!

PR #223, opened 2015-12-08
Coerce integers attributes into strings

sax:
Avoids imprecise float conversion when tracking bigint values. Without conversion to strings, the values are coerced in a fashion that modifies the value of the integer.

For instance, with a source value of 129051960153211241, the value as seen in NewRelic is changed to 129051960153211250.

### DIFF
--- a/lib/new_relic/coerce.rb
+++ b/lib/new_relic/coerce.rb
@@ -46,8 +46,10 @@ module NewRelic
 
     def scalar(val)
       case val
-      when String, Integer, TrueClass, FalseClass, NilClass
+      when String, TrueClass, FalseClass, NilClass
         val
+      when Integer
+        val.to_s
       when Float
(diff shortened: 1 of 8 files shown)


tkrajcar:
Hi @sax, sorry it took so long to get back to you. Unfortunately we can't do this; it would break numeric operators in our Insights product. It would be great to get a better understanding of what you're putting in with such large values, though, maybe there's another way to support it. Maybe you can open a ticket at http://support.newrelic.com/ with more detail

sax:
There are two main reasons for us to use bigint column types. The first is that we have a table that actually has more records than fit into the maximum value of int. The other is where we use a sharding strategy that relies on bigints. 

I'm sure that others face this problem, even if they don't yet know it. When a bigint is recorded by newrelic error tracking the value shown will be wrong.

PR #224, opened 2015-12-23
Removing the link to the app role

gerr1t:
Deployments that don't use the :app role are failing because of the fact that this job is fixed to the :app role. Why is there even a link to the app role? What does it have to do with a role? We just need a single deployment message to NewRelic.


gerr1t:
Never mind, I realise that it uses it to get the revision of the previous build, and with that the changes.

PR #227, opened 2016-01-20
Replace alias_method with Module.prepend

pschambacher:
This is work in progress but before moving further I'd like to be sure this PR would be accepted.
We're facing an issue at the moment since we replaced `alias_method_chain` with `Module.prepend` in some of ours gems (active_record_shards for instance). Mixing `alias_method_chain` and `Module.prepend` tend to provoke `stack level too deep` errors.

This PR aims to remove `alias_method` (used in the alias_method_chain way) in favor of `Module.prepend`.

To achieve this goal, Ruby < 2.0 need to be abandoned. IMHO this is fine since 1.8 and 1.9 aren't supported anymore.

If people are OK with this PR I'll keep going and remove all the other `alias_method` calls.
@jasonrclark, @benweint ?

Easier to review [without whitespaces](https://github.com/newrelic/rpm/pull/227/files?w=0)


tkrajcar:
Hey @pschambacher, thanks for checking in before going farther, and I'm glad you did; unfortunately we still need to support 1.8 and 1.9 as of the moment. We will certainly be moving towards dropping 1.8.7/REE/1.9.2 support soon, but probably won't drop 1.9.3 support at that time, either.

In your specific case it may make sense to continue to run a fork of the agent that has the changes you need.

pschambacher:
Hey Tim

Yeah I fear that we'll have to run on this fork because we can't run anything at the moment.
I'll leave this PR open once done so when it's fine to drop 1.9.3 you can merge.

Kulgar:
That would be a +1 for me. There is a bad interaction between New Relic and Switchman in one of our projects. See this issue: https://github.com/instructure/canvas-lms/issues/752

We have a paid account with New Relic but we no longer can run it in our main project because of that. That is really bad. :(

@pschambacher : I will definitely be interested in your fork!

pschambacher:
@Kulgar FYI I reported the bug on Ruby: https://bugs.ruby-lang.org/issues/11120
That said, since we're master of our own systems we just decided to revert our `Module.prepend` changes and go with `alias_method` (see [this PR](https://github.com/zendesk/active_record_shards/pull/57)).

The main problem is that the ruby core team probably won't fix this problem. So we have to dodge bullets until everybody uses `Module.prepend`.

That said, the bug happens if you use `alias` **after** using `Module.prepend` with the same method. The other way around works. So if you still want to use NewRelic you could use this fork (at your own risk, tests are green and preliminary tests on our end were conclusive but it wasn't tested against our production code) or you need to force NewRelic to load **first** before any gem has a chance to call `Module.prepend`.

Kulgar:
@pschambacher : Thanks a lot for those information! I think I'll go with the second solution. Or at least I'll try :-) It does sound like a really interesting quick win :wink:

Kulgar:
Well, unfortunately none of the solutions worked :( 
Maybe I'll just hack the gem myself for my own needs. 

Thanks anyway :)

dawidof:
Could someone make it working on rails 5?

aguynamedben:
We're also paying customers of 3 years, and New Relic isn't usable with Rails 5 for us. We're commenting it out and I guess we'll email support to disable billing for a while...

aguynamedben:
@tkrajcar I don't see why New Relic can't review this, merge it, and create a new major release (4.0) that drops support for Ruby < 2.0, and continue to maintain a 3.x.x version for a while for customers running Ruby < 2.0.

Support for Ruby version 1.9.3 ended February 23, 2015...

kwugirl:
Hi everyone--wanted to share that last week we released a version of the agent that uses `Module#prepend` for ActiveRecord 5 instrumentation! See the release notes for [v3.17.2](https://docs.newrelic.com/docs/release-notes/agent-release-notes/ruby-release-notes/ruby-agent-3172327). This should help some of the folks that were seeing the `stack level too deep` errors in Rails 5 apps, since we think most of those were from the AR5 instrumentation.

Additionally, we published a blog post on this topic: [Resolving `Module#prepend` and `alias_method` Conflicts Involving the New Relic Ruby Agent](https://blog.newrelic.com/2016/12/15/ruby-agent-module-prepend-alias-method-chains/). It's largely intended as an introduction to why those errors come up, but also covers some of our current thinking. We would like to be able to install instrumentation using either `Module#prepend` _or_ `alias_method`, so we're closing this pull request in favor of that strategy instead. Thanks again however for bringing the issue up!

PR #230, opened 2016-02-18
Fix delayed jobs sampler query in AR3

rdokov:
This fixes a regression introduced by 1e457bb7

At least in postgresql `count(false)` returns 1, so the query that was
generated was always counting everything in the `delayed_jobs` table and
not just the non-null records as it is meant to do.

### DIFF
--- a/lib/new_relic/agent/samplers/delayed_job_sampler.rb
+++ b/lib/new_relic/agent/samplers/delayed_job_sampler.rb
@@ -51,7 +51,7 @@ module NewRelic
 
         def count(query)
           if ::ActiveRecord::VERSION::MAJOR.to_i < 4
-            ::Delayed::Job.count(query)
+            ::Delayed::Job.count(conditions: query)
           else
             ::Delayed::Job.where(query).count
           end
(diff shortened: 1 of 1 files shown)


kwugirl:
Thanks for the contribution and pointing this out @rdokov! I think the test failures are because we currently include 1.8.7 in our support, so still need to use hashrocket syntax: `::Delayed::Job.count(:conditions => query)` should work, I think.

rdokov:
Yes, my bad for using the new hash syntax. I've fixed that, but one of the builds still failed and I can't really figure out why. (it just seems like a random timeout error).

kwugirl:
@rdokov Thanks for changing the syntax. We re-ran that build and it's all green now, so this is looking good! We'll look into pulling this into our internal repo and update this PR later on.

kwugirl:
@rdokov this was released in [3.15.1.316](https://docs.newrelic.com/docs/release-notes/agent-release-notes/ruby-release-notes/ruby-agent-3151316) today (Github doesn't mark this PR as merged and closed it automatically because we have to rebase it against our internal dev branch as part of our process, so it ended up having a different SHA). Thank you again for the contribution!

PR #231, opened 2016-03-09
Make it possible to define a newrelic_role for deployment with capistrano 3

NielsKSchjoedt:
This will make it possible to have the deployment notification sent
from the hosts assigned to the role.

Why?
The current setup requires you to have the new relic config .yml with production license keys straight in the source code. In many cases, the "real" config will be provisioned using something like Ansible or Chef  and will only be available on the production hosts.

Using a method like the suggested, it would be possible to do something like: 
```ruby
#deploy.rb

set :newrelic_role, :primary_web_server
```  

And then in:

```ruby
# environments/production.rb

server 'srv-01.example.com', user: fetch(:deploy_user), roles: %w{web app, primary_web_server}

```

This would also enable the notification to go to different app_names, if more hosts has the role assigned and the config .yml is different on them (which it is in our case, because we want to completely separate background processing stats from web stats)


lib/new_relic/recipes/capistrano3.rb
```diff
@@ -9,46 +9,56 @@
...
+      end
+    else
+      run_locally do
+        notice_deploymet
+      end
+    end
+  end
 
-      require 'new_relic/cli/command.rb'
+  def notice_deploymet
```
kwugirl: Hi @NielsKSchjoedt thanks for the contribution, it does seem like this will be useful! We were thinking that perhaps it might be a bit confusing to see a method name here that's so similar to the task name. Perhaps something like `upload_deployment` or `send_deployment` or some other variation? Open to your ideas.

kwugirl:
Additionally, we pushed out a 3.15.1 release yesterday so if you rebase against that, at least some of the test failures should clear up a bit, I believe.

kwugirl:
Hi @NielsKSchjoedt wanted to see if you thought you'd get a chance to incorporate some of the feedback here and rebase against the latest release?

NielsKSchjoedt:
Done.

kwugirl:
Awesome, thanks @NielsKSchjoedt ! Pulling this in to our internal dev branch now. I'll update this PR when it's released.

kwugirl:
@NielsKSchjoedt this was just released this morning in 3.16.0. Thanks again for the contribution!

PR #233, opened 2016-03-20
Add transaction support for padrino 0.13

fnordfish:
Padrino 0.13 started to use mustermann as routing engine. Therefore, the method `route_eval` might not get called anymore and the `route_obj` is added later to the `request`.
This instrumentation hooks into the new `invoke_route` method which came with padrino 0.13.

This addresses padrino/padrino-framework#2002 and https://discuss.newrelic.com/t/missing-route-details-in-transaction-log/34989

### DIFF
--- a/lib/new_relic/agent/instrumentation/padrino.rb
+++ b/lib/new_relic/agent/instrumentation/padrino.rb
@@ -27,6 +27,42 @@ DependencyDetection.defer do
 
       alias dispatch_without_newrelic dispatch!
       alias dispatch! dispatch_with_newrelic
+
+      # Padrino 0.13 mustermann routing
+      if private_method_defined?(:invoke_route)
+        include NewRelic::Agent::Instrumentation::Padrino
+
+        alias invoke_route_without_newrelic invoke_route
+        alias invoke_route invoke_route_with_newrelic
+      end
+    end
+  end
+end
+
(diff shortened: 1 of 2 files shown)


fnordfish:
Sorry for the failing Travis builds. I've checked some and it looks like they where failing from log error logs:

```
...
Finished tests in 46.156139s, 55.5722 tests/s, 386.5141 assertions/s.

2565 tests, 17840 assertions, 0 failures, 0 errors, 0 skips
...
The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).

The build has been terminated.
```

RUN-CMD:
Thank you for putting this together! While we are reviewing it, could you please point us toward the relevant changes introduced in padrino that switched to the new routing engine?

fnordfish:
The relevant changes have been made in the `Padrino::Routing::InstanceMethods#route!` method inside `padrino-core/lib/padrino-core/application/routing.rb` (https://github.com/padrino/padrino-framework/compare/0.12.5...0.13.1#diff-c2f51b1c75eae89ededcaca57ff6e0d2L928) which used to call `route_eval` for a matching route.

kwugirl:
This has now been released in 3.15.2! Thank you @fnordfish

PR #234, opened 2016-03-25
Restore older DataMapper support after password obfuscation fix.

blaet:
DataMapper versions below v0.10.0 do not implement the #options accessor. Instead, the adapter name (scheme) has to be fetched from the URI object. 

Fetching the URI this deeply nested is needed to support both DataObjectsAdapter and DataMapper::Collection/Model instances.

### DIFF
--- a/lib/new_relic/agent/instrumentation/data_mapper.rb
+++ b/lib/new_relic/agent/instrumentation/data_mapper.rb
@@ -147,7 +147,8 @@ module NewRelic
                 strategy = NewRelic::Agent::Database.record_sql_method(:slow_sql)
                 case strategy
                 when :obfuscated
-                  statement = NewRelic::Agent::Database::Statement.new(e.query, :adapter => self.options[:adapter])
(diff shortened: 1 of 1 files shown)


kwugirl:
Hi @blaet thanks for the contribution! I didn't know that about DataMapper versions <0.10.0. We'll pull that in to our dev branch.

Just as an FYI, I believe there are security issues with those versions of DM, namely that they will not include this fix against showing the display connection password (https://github.com/datamapper/do/commit/5e05356dcc7cc3efbd77be33f87afdaaf8a9957a) that was released in [0.10.8](https://github.com/datamapper/do/blob/master/data_objects/ChangeLog.markdown#0108-2012-02-10) in 2012.

blaet:
Thanks for the heads-up @kwugirl! Though in my use case, this issue has been handled. :smiley_cat:

kwugirl:
This has now been released in 3.15.2! Thank you @blaet

blaet:
Awesome news! Thanks @kwugirl!

PR #235, opened 2016-04-04
Fix: deprecation warning in Grape 0.16.x.

dblock:
In Grape 0.16.x `route_` methods are no longer prefixed. See https://github.com/ruby-grape/grape/blob/master/UPGRADING.md#replace-rack-mount-with-new-router.

### DIFF
--- a/lib/new_relic/agent/instrumentation/grape.rb
+++ b/lib/new_relic/agent/instrumentation/grape.rb
@@ -28,15 +28,29 @@ module NewRelic
           Transaction.set_default_transaction_name(txn_name, :grape, node_name)
         end
 
-        def name_for_transaction(route, class_name)
-          action_name = route.route_path.sub(FORMAT_REGEX, EMPTY_STRING)
-          method_name = route.route_method
+        if defined?(Grape::VERSION) && VersionNumber.new(::Grape::VERSION) >= VersionNumber.new("0.16.0")
+          def name_for_transaction(route, class_name)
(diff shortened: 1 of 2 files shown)


dblock:
Build is failing with these:

```
Installing mime-types-data 3.2016.0221
Gem::InstallError: mime-types-data requires Ruby version >= 2.0.
```

I don't think this is me.

kwugirl:
Hi @dblock -- yes, that mime-types-data failure is a test dependency we've had to fix in our dev branch. This looks good otherwise though, thanks for the contribution! I'll pull it in to our internal repo.

dblock:
👍 

I also did wonder whether it's better to build two separate plugins for different versions of Grape, or some kind of inheritance here. Curious what you think.

kwugirl:
@dblock I don't think we have a specific policy around when there should be new instrumentation written for different versions, but in practice it's usually only if there have been particularly significant changes, like a few months back with the Mongo 2 driver. Mostly it's been for different Rails and ActiveRecord versions.

kwugirl:
This has now been released in 3.15.2! Thank you @dblock

supermario:
For anyone else who's ended up here confused as to why you're still seeing these warnings despite being on the latest version of rpm – it's likely you're using `Bundler.require` and `newrelic_rpm` is being loaded before `grape`.

Can be resolved by requiring `newrelic_rpm` after `grape` in the Gemfile.

kwugirl:
Oh that's great to know, thank you @supermario !

PR #239, opened 2016-04-15
Use the new Grape helpers to avoid the deprecation warnings

alfonsocora:
[`route_xxx` methods have been deprecated in version 0.15.1 of Grape](https://github.com/ruby-grape/grape/blob/master/UPGRADING.md) and are now throwing deprecation warnings.

### Changes

* Use `path` instead of `route_path`
* Use `request_method` instead of `route_method`
* Use `version` instead of `route_version`


alfonsocora:
Ah, nvm, this is a dupe of #235

kwugirl:
Thanks for giving it a shot regardless @alfonsocora! :dancers:

PR #240, opened 2016-04-27
Fixes deprecated grape methods usage

gottfrois:
Latest grape version deprecated `route_` methods in favor of directly calling method on the `route` instance.


kwugirl:
@gottfrois thanks for looking into this! I believe https://github.com/newrelic/rpm/pull/235 address this issue and was included in yesterday's 3.15.2 release. If you upgrade to using the latest version of the gem, does that resolve your issue?

gottfrois:
Oh great, should have looked at closed PR first :) i'll try upgrading and see, thx!

PR #243, opened 2016-05-25
Allow oracle_enhanced adapter to obfuscate properly

drinks:
The oracle_enhanced ActiveRecord driver is a popular choice among those using Oracle with Rails, and is currently [mapped to the product name 'Oracle'](https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/instrumentation/active_record_helper.rb#L201) in this project. However, the only adapter name supported for obfuscation with Oracle is `:oracle`. This PR allows the `:oracle_enhanced` adapter to use the oracle obfuscator path, turning query dumps like this:

```SQL
"SELECT  ?.* FROM ?  WHERE ?.? = ? AND ROWNUM <= ?"
```

into this:

```SQL
"SELECT  \"MYTABLE\".* FROM \"MYTABLE\"  WHERE \"MYTABLE\".\"MYCOLUMN\" = ? AND ROWNUM <= ?"
```

I'm happy to add tests, but don't see anything currently around the oracle dialect. If you'd like to see some, just point me to the right place and I'll have at it.

Thanks for reading!

### DIFF
--- a/lib/new_relic/agent/database/obfuscation_helpers.rb
+++ b/lib/new_relic/agent/database/obfuscation_helpers.rb
@@ -44,7 +44,8 @@ module NewRelic
           :postgres => /'|\/\*|\*\/|\$(?!\?)/,
           :sqlite => /'|\/\*|\*\//,
           :cassandra => /'|\/\*|\*\//,
-          :oracle => /'|\/\*|\*\//
+          :oracle => /'|\/\*|\*\//,
+          :oracle_enhanced => /'|\/\*|\*\//
         }
 
         PLACEHOLDER = '?'.freeze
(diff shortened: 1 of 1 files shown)


kwugirl:
@drinks thank you for this contribution, and especially the detailed and clear writeup! I had missed that `"oracle_enhanced" => "Oracle"` mapping when I was working on that obfuscation update so I'm glad you've fixed that now. I'll pull this in to our internal repo. 

As purely FYI, the tests we have around obfuscation are done through https://github.com/newrelic/rpm/blob/master/test/new_relic/agent/database/sql_obfuscation_test.rb and https://github.com/newrelic/rpm/blob/master/test/fixtures/cross_agent_tests/sql_obfuscation/sql_obfuscation.json -- but those are mainly for verifying behavior across the different language agents that New Relic produces. Assuming there are no known differences in the queries we'd pick up from `oracle_enhanced` vs `oracle` database adapters, this seems fine as is to me.

drinks:
Thanks for the review, @kwugirl! What needs to happen to get this merged and a new release cut?

kwugirl:
@drinks nothing more needed on your end, just ours. I have the pull request all set up and tests passing on our internal Github Enterprise repo, so I'll bug one of my teammates to merge that. Then, it should be part of the next release, which should be in the next couple of weeks. Hope that works!

kwugirl:
@drinks this was just released this morning in 3.16.0. Thanks again for the contribution!

PR #244, opened 2016-06-30
Fix incompatibility issue with activerecord-turntable

phamvanhungmixi:
I got incompatibility issue when using `activerecord-turntable`.
The problem is `activerecord-turntable` override log method of AbstractAdapter for logging current shard name. `activerecord-turntable` is loaded after `rpm`, and it override log method of `rpm`. Therefore, New Relic `rpm` can't collect mysql metric.
https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/instrumentation/active_record.rb#L32
https://github.com/drecom/activerecord-turntable/blob/master/lib/active_record/turntable/active_record_ext/abstract_adapter.rb#L10

I want to resolve the issue without breaking anything by load `activerecord-turntable` before `rpm` when using `activerecord-turntable`

### DIFF
--- a/lib/new_relic/agent/instrumentation/active_record.rb
+++ b/lib/new_relic/agent/instrumentation/active_record.rb
@@ -22,6 +22,9 @@ module NewRelic
           end
 
           ::ActiveRecord::ConnectionAdapters::AbstractAdapter.module_eval do
+            if defined?(::ActiveRecord::Turntable)
+              include ::ActiveRecord::Turntable::ActiveRecordExt::AbstractAdapter
+            end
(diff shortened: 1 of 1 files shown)


kwugirl:
Hi @phamvanhungmixi! I've been looking over the support ticket you filed with our support team and based on my test app with Rails + `activerecord-turntable` + `newrelic_rpm`, the loading order there seems to be as desired for that `log` method overrided by `activerecord-turntable`. We'd like to avoid special casing as `newrelic_rpm` should be set up to try to be the last library loaded, though I understand something is changing that in your fork, perhaps due to the modifications to work with Padrino. We have a different suggested workaround that will hopefully address your situation!

PR #246, opened 2016-07-13
Override Sequel::Database#log_connection_yield instead of #log_yield

munshkr:
Version 4.35 changes the way Sequel logs queries by defining a new function `Sequel::Database#log_connection_yield` and making the old `#log_yield` delegate to the new one. The problem is that all Sequel adapters now use `#log_connection_yield`, so NewRelic instrumentation is never executed and fails silently.

This changes NewRelic Sequel extension so that it overrides `#log_connection_yield` if defined, or `#log_yield`, to keep it backwards-compatible with older versions of Sequel.

See commit https://github.com/jeremyevans/sequel/commit/aab4d16f842256211420755648f242e768d4834c

I was not sure how to include a test for this, but I used [this script](https://gist.github.com/munshkr/94925d4a848e72788967ddef72e01cc9) to make sure it works on new and old versions of Sequel.

### DIFF
--- a/lib/sequel/extensions/newrelic_instrumentation.rb
+++ b/lib/sequel/extensions/newrelic_instrumentation.rb
@@ -32,8 +32,30 @@ module Sequel
   #
   module NewRelicInstrumentation
 
+    def self.extended(db)
+      if db.respond_to?(:log_connection_yield)
+        db.extend(LogConnectionYield)
+      else
+        db.extend(LogYield)
+      end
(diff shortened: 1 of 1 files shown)


kwugirl:
Hi @munshkr -- thank you for taking a look at this problem! We'd received a report of it a couple weeks ago and so have a fix for it on our dev branch internally and expect it to be part of the next release.

PR #247, opened 2016-07-14
DataMapper::Adapters::DataObjectsAdapter objects do not support #uri.

postmodern:
* Use `adapter.options[:adapter]` instead, which returns the adapter name as
  a String (ie: `"postgres"`).
* http://www.rubydoc.info/gems/dm-do-adapter/DataMapper/Adapters/DataObjectsAdapter
* http://www.rubydoc.info/gems/dm-core/DataMapper/Adapters/AbstractAdapter#options-instance_method

### DIFF
--- a/lib/new_relic/agent/instrumentation/data_mapper.rb
+++ b/lib/new_relic/agent/instrumentation/data_mapper.rb
@@ -147,7 +147,16 @@ module NewRelic
                 strategy = NewRelic::Agent::Database.record_sql_method(:slow_sql)
                 case strategy
                 when :obfuscated
-                  adapter_name = self.respond_to?(:options) ? self.options[:adapter] : self.repository.adapter.options[:adapter]
+                  adapter_name = if self.respond_to?(:options)
+                    self.options[:adapter]
+                  else
+                    if self.repository.adapter.respond_to?(:options)
+                      self.repository.adapter.options[:adapter]
+                    else
+                      # DataMapper < 0.10.0
+                      self.repository.adapter.uri.scheme
+                    end
+                  end
                   statement = NewRelic::Agent::Database::Statement.new(e.query, :adapter => adapter_name)
                   obfuscated_sql = NewRelic::Agent::Database.obfuscate_sql(statement)
                   e.instance_variable_set(:@query, obfuscated_sql)
--- a/lib/new_relic/agent/instrumentation/data_mapper.rb
+++ b/lib/new_relic/agent/instrumentation/data_mapper.rb
@@ -147,7 +147,7 @@ module NewRelic
                 strategy = NewRelic::Agent::Database.record_sql_method(:slow_sql)
                 case strategy
                 when :obfuscated
-                  adapter_name = self.respond_to?(:options) ? self.options[:adapter] : self.repository.adapter.uri.scheme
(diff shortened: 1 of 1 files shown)


postmodern:
This was causing `NoMethodError: undefined method `uri' for #<DataMapper::Adapters::PostgresAdapter:...>` when a `DataObjects::IntegrityError` was triggered due to a SQL unique index constraint.

kwugirl:
@postmodern thanks for this contribution! Do you happen to know if this works for older versions of DataMapper as well? That `.uri.scheme` came from https://github.com/newrelic/rpm/pull/234 I believe.

postmodern:
It appears the switch from `#uri` to `#options` occurred in [dm-core 0.9.9](http://www.rubydoc.info/gems/dm-core/0.9.9/DataMapper/Adapters/AbstractAdapter#uri-instance_method) -> [dm-core 0.10.0](http://www.rubydoc.info/gems/dm-core/0.10.0/DataMapper/Adapters/AbstractAdapter#options-instance_method). Does newrelic_rpm need to support DataMapper < 0.10.0?

kwugirl:
@postmodern oh cool, thanks for looking that up. I think officially we list newrelic_rpm as supporting DataMapper ~>1.0...but I think we probably figure if it can be made to work for the earlier versions, we're ok to accept that.

Since from https://github.com/newrelic/rpm/pull/234 it looks like at least @blaet is using DataMapper < 0.10.0, would you be open to updating this PR so that both your and their cases are covered? Thank you!

blaet:
@kwugirl @postmodern I'm indeed still working with some systems running DataMapper 0.9.9. So if keeping it compatible is not too much of a hassle, that would be great.

postmodern:
Updated to also support `adapter.uri`. Tried to match the indentation style of `if/else` blocks that return values.

postmodern:
@kwugirl ping. The necessary changes are in.

lib/new_relic/agent/instrumentation/data_mapper.rb
```diff
@@ -147,7 +147,16 @@ def self.method_body(clazz, method_name, operation_only)
...
                 when :obfuscated
-                  adapter_name = self.respond_to?(:options) ? self.options[:adapter] : self.repository.adapter.uri.scheme
+                  adapter_name = if self.respond_to?(:options)
+                    self.options[:adapter]
+                  else
+                    if self.repository.adapter.respond_to?(:options)
+                      self.repository.adapter.options[:adapter]
+                    else
+                      # DataMapper < 0.10.0
+                      self.repository.adapter.uri
```
kwugirl: I think this should be `self.repository.adapter.uri.scheme`?

postmodern:
Uh embarrassing. Amended the commit.

kwugirl:
Nah thanks for making all those changes, I do appreciate it! I'm about to leave on a two-week trip so was a little slow in getting feedback to you while I wrapped stuff up, sorry about that. I'll pull this in to our internal repo and make it part of the next release!

kenichi:
This has been released in [3.16.1](https://github.com/newrelic/rpm/blob/master/CHANGELOG#L32-L36)! Thanks @postmodern

PR #248, opened 2016-07-25
Fix versioning for cascading Grape routes

gmassanek:
We use Grape's ability to cascade versions, similar to [this post](http://code.dblock.org/2013/07/19/evolving-apis-using-grape-api-versioning.html). When doing so, the NR grape instrumentation has two issues:

1. If there is only a version of an endpoint in the earliest versions, the transaction name will include all the versions it applies to, ex: `["1.0", "2.0", "3.0"]: /my/endpoint (GET)`
2. If there is a newer version, based on some nuance of Grape cascades, the transaction name will always be the most recent.

The changes in this PR start using the `api.version` key that Grape puts into the rack env once it decides which route/version is actually going to respond to the call. I've tested locally and this fixes the issues in our grape app.

Note: I believe the tests for the `Accept-Version` header were guarded by an overly ambitious version guard and were not running (they were looking for v 4.0.0 of grape, but the latest is 0.16). I've updated so these run if the grape version >= 0.16.0.

### DIFF
--- a/lib/new_relic/agent/instrumentation/grape.rb
+++ b/lib/new_relic/agent/instrumentation/grape.rb
@@ -11,43 +11,46 @@ module NewRelic
         extend self
 
         API_ENDPOINT   = 'api.endpoint'.freeze
+        API_VERSION    = 'api.version'.freeze
         FORMAT_REGEX   = /\(\/?\.[\:\w]*\)/.freeze # either :format (< 0.12.0) or .ext (>= 0.12.0)
         VERSION_REGEX  = /:version(\/|$)/.freeze
         EMPTY_STRING   = ''.freeze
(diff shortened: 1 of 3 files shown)


kenichi:
@gmassanek thank you for the PR! We are looking into incorporating this into the next release.

gmassanek:
@kenichi many thanks. Do you have any information on the typical release cycle?

kenichi:
@gmassanek we generally release about once a month, so one should be coming soon. I will try to let you know when we have a hard date set.

kenichi:
This has been released in [3.16.1](https://github.com/newrelic/rpm/blob/master/CHANGELOG#L10-L16)! Thanks @gmassanek

gmassanek:
Great news! I just tested it out and it works great. Thanks for bringing the change in.

PR #250, opened 2016-08-23
Do not install_newrelic_job_tracer into Delayed::Job if it has been done already

bonybrown:
When using multiple Delayed::Worker instances in the same process, the Delayed::Job class gets patched each time the worker is created.
Small patch to ensure it is only done once.

### DIFF
--- a/lib/new_relic/agent/instrumentation/delayed_job_instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/delayed_job_instrumentation.rb
@@ -98,9 +98,11 @@ DependencyDetection.defer do
         NewRelic::DelayedJobInjection.worker_name = worker_name
 
         if defined?(::Delayed::Job) && ::Delayed::Job.method_defined?(:invoke_job)
-          ::NewRelic::Agent.logger.info 'Installing DelayedJob instrumentation [part 2/2]'
-          install_newrelic_job_tracer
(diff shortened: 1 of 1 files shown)


kwugirl:
@bonybrown thanks for the contribution! Could you tell us a bit more about the process you're running that's spinning up multiple workers? Curious if there's a root cause that we'd want to address.

bonybrown:
Oh, forgot to mention, that without the patch, the multiple instrumentation of Delayed::Job causes a "stack level too deep" exception every time a job is attempted, and the job errors out.
Looks like the code endlessly recurses around the instrumented and alias methods of Delayed::Job

kwugirl:
Thanks for the additional details @bonybrown! tbh we had thought our agent might not work on Windows...but glad to pull this in since it seems to help out with your situation. I'll open a PR with your change against our internal repo and will update you here when it's part of the next release.

akshah123:
I am seeing similar error on our ubuntu environment.  Not sure if it is related but the issue started happening with Rails 5.0 and agent version 3.16.0

kwugirl:
@akshah123 oh interesting. Does this patch fix the issue for you? If not, we may want to gather additional details of your case through a support ticket.

akshah123:
@kwugirl i think my problem is unrelated as we are starting two separate processes. I have created a support ticket for my issue.

kenichi:
This has been released in version 3.16.3! Thanks, @bonybrown

PR #253, opened 2016-10-16
'ignore_url_regexes' should prevent all instrumentation

diego-silva:
Giving the name of the config: `rules.ignore_url_regexes` I was expecting and under the impression that it would prevent ALL kind of data from being collected from the specified url's.
Imagine my surprise when I realized that my `Browser` data was wrong because this configuration would only suppress the reporting of APM data...

Trying to search for a better documentation or a simple way to prevent ALL data from being reported to New Relic I found out that there was no option for doing that.

What do you guys think of making the `ignore_url_regexes` to really ignore every kind of instrumentation by not adding the browser monitoring as well?

### DIFF
--- a/lib/new_relic/rack/browser_monitoring.rb
+++ b/lib/new_relic/rack/browser_monitoring.rb
@@ -53,9 +53,19 @@ module NewRelic::Rack
       NewRelic::Agent.config[:'browser_monitoring.auto_instrument'] &&
         status == 200 &&
         !env[ALREADY_INSTRUMENTED_KEY] &&
+        !is_route_ignored?(env) &&
         is_html?(headers) &&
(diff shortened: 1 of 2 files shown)


kwugirl:
Hi @diego-silva thank you very much for bringing this inconsistency to our attention! I can definitely see how the way the agent works now would be unexpected for your browser instrumentation. I'm closing this PR because we think we'd want to address the issue at a deeper level* but we've discussed it on the team and I've created a ticket in our backlog to track this as a bug to be fixed. Thanks again for letting us know and giving it a shot!

*in short, there's a concept of ignoring transactions that we think should be fixed to make sure browser data is ignored if the transaction is ignored, and this config setting does affect the list of transactions to be ignored.

PR #261, opened 2017-01-20
Adds support for array ENV variables

mezis:
The [documentation](https://docs.newrelic.com/docs/agents/ruby-agent/installation-configuration/ruby-agent-configuration#attributes) mentions several options that take an array value.
Readers would expect those to be settable through the environment, but this is unsupported; e.g. this doesn't work:

```# .env
NEW_RELIC_ATTRIBUTES_INCLUDE=jobs.sidekiq.args.*,jobs.resque.args.*
```

This diff adds support for setting array options through comma-separate environment variables.

### DIFF
--- a/lib/new_relic/agent/configuration/environment_source.rb
+++ b/lib/new_relic/agent/configuration/environment_source.rb
@@ -94,6 +94,8 @@ module NewRelic
             elsif value != nil
               self[config_key] = true
             end
+          elsif type == Array
+            self[config_key] = value.split(',')
           else
             ::NewRelic::Agent.logger.info("#{environment_key} does not have a corresponding configuration setting (#{config_key} does not exist).")
             ::NewRelic::Agent.logger.info("Run `rake newrelic:config:docs` or visit https://newrelic.com/docs/ruby/ruby-agent-configuration to see a list of available configuration settings.")
--- a/test/new_relic/agent/configuration/environment_source_test.rb
+++ b/test/new_relic/agent/configuration/environment_source_test.rb
@@ -82,6 +82,12 @@ module NewRelic::Agent::Configuration
       end
     end
 
+    def test_arrays_are_applied
+      ENV['NEW_RELIC_ATTRIBUTES_INCLUDE'] = 'jobs.sidekiq.args.*,jobs.resque.args.*'
+      source = EnvironmentSource.new
+      assert_equal %w| jobs.sidekiq.args.* jobs.resque.args.*  |, source[:'attributes.include']
+    end
+
     def test_set_log_config_from_environment
       ENV['NEW_RELIC_LOG'] = 'off/in/space.log'
       source = EnvironmentSource.new


kwugirl:
Hi @mezis ! Thanks so much for taking a look into this issue. Would you be able to be describe a bit more the circumstances in which you're seeing that comma-separated values for environment variables are _not_ being transformed into arrays? Based on https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/configuration/manager.rb#L114 it seems like those values should get transformed into arrays eventually. If I add the following as a test:
```ruby
def test_should_transform_value_from_env
  ENV['NEW_RELIC_ATTRIBUTES_INCLUDE'] = 'jobs.sidekiq.args.*,jobs.resque.args.*'
  manager = NewRelic::Agent::Configuration::Manager.new

  assert_equal ['jobs.sidekiq.args.*','jobs.resque.args.*'], manager[:"attributes.include"]
ensure
  ENV.delete('NEW_RELIC_ATTRIBUTES_INCLUDE')
end
```
to https://github.com/newrelic/rpm/blob/master/test/new_relic/agent/configuration/manager_test.rb that test seems to pass with the agent code as-is. I also started up a Rails app with `NEW_RELIC_ATTRIBUTES_INCLUDE=asdf,asdf bundle exec rails` and after the app had started up, `puts "#{NewRelic::Agent.config[:"attributes.include"]}"` outputs `['asdf','asdf']` which seemed to indicate it's being read as an array.

I did see that if I looked for that config value in the agent log, the fact that the value has been transformed doesn't show up there, but wasn't sure if that was what you were looking at too? 

I didn't test actually capturing both sidekiq and resque job arguments at the same time after setting the environment variable, so let me know if the issue you're seeing is that one specifically or seems to be broadly applicable to comma-separated values for environment variables that should be transformed into arrays.

kwugirl:
Hi @mezis @gregbeech @kplattret -- would love to hear whether you're still seeing issues around this functionality. If I don't hear back otherwise, I'll plan to close this PR in a couple days.

aleksandrs-ledovskis:
I can confirm that using regular string with comma-separated values gets processed correctly by New Relic gem.

Specifically, I was after `rake.tasks` configuration via `NEW_RELIC_RAKE_TASKS=.+`, and while indeed there appears

```
INFO : NEW_RELIC_RAKE_TASKS does not have a corresponding configuration setting (rake.tasks does not exist).
```

message in log file, switching verbosity to `debug` yields extra information, that confirms Rake task instrumentation is enabled:

```
DEBUG : Instrumenting Rake tasks: [/.+/]
```

---

As a possible improvement, I would suggest improving logic when `INFO` message is shown.

kwugirl:
@aleksandrs-ledovskis I'm no longer working at New Relic but I don't see current members of that team on this PR, so wanted to let you know that if you wanted to submit that improvement suggestion, you'll probably want to reach out to the New Relic support team.

PR #262, opened 2017-01-23
deployment notification - run only on primary :newrelic_role

pangkalizer:
when assigned role to :newrelic_role variable, only run deployment notice on primary

### DIFF
--- a/lib/new_relic/recipes/capistrano3.rb
+++ b/lib/new_relic/recipes/capistrano3.rb
@@ -10,7 +10,7 @@ namespace :newrelic do
   desc "Record a deployment in New Relic (newrelic.com)"
   task :notice_deployment do
     if fetch(:newrelic_role)
-      on roles(fetch(:newrelic_role)) do
+      on primary fetch(:newrelic_role) do
         send_deployment_notification_to_newrelic
       end
     else


kwugirl:
Hi @pangkalizer thanks so much for taking a look at this recipe! I have to confess that I don't know very much about how to use Capistrano and as a result was wondering if it's possible to have a setup where you do _not_ have a particular server designated as the primary? I found [this documentation](http://capistranorb.com/documentation/advanced-features/properties/) which says:

> By convention the first server in a role declaration is taken to be the primary, but the :primary property will not actually be set in such a case.

In that situation, is there a risk that no deployment notices would be sent up at all?

pangkalizer:
hi @kwugirl, as cap documentation says,  capistrano will elect first server on the role as `primary` when none has been explicitly set.. ref [fetch_primary](https://github.com/capistrano/capistrano/blob/5fa5fe1a9c9f7992550f342eec850ca499f1dd10/lib/capistrano/configuration/servers.rb#L52). there even some cases when don't server server roles at all specially on dynamic/auto scale scenario.  the edit was to ensure that newrelic deploy notice will only run once on selected role.

kwugirl:
Oh cool, thanks for the reference to the code for `fetch_primary` @pangkalizer. What's happening now when the deploy notice happens multiple times on a selected role? I think you can view the deployments page in the New Relic UI by server, so would you end up only seeing deployment notices for the server that had been designated as the primary...? Not sure whether Capistrano server objects map to what we report as servers in the New Relic UI.

pangkalizer:
To be honest, I haven't tried that config yet. I was hoping you can shed light on that matter actually. At the moment, I don't have a test environment to check that setting. 

However just purely looking at the code. I don't see any hint that it is limiting the call to NewRelic API endpoint to just once. Unless the API make some unique filtering on the server side, it would be too late coz the call has already been made, unnecessarily.
https://github.com/newrelic/rpm/blob/master/lib/new_relic/cli/commands/deployments.rb#L65

We are also using Rollbar for error reporting and a similar implementation could be seen here.
https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/tasks/rollbar.cap#L9

kwugirl:
I looked through the code again and I think it's actually designed _not_ to limit the call to the New Relic API just once per deployment to a series of servers, because the deployment notice includes information about which host it's happening on: https://github.com/newrelic/rpm/blob/master/lib/new_relic/cli/commands/deployments.rb#L71 This allows you to view deployments by server in the UI.

Given that, I'm not sure this change makes sense. Please do let us know if you have a case that overturns any of our assumptions here, though. I'll leave this PR open for a few more days for further comments but then close it out next week if there aren't any.

pangkalizer:
I think you're confusing the "deployment notice" here, it's a notice sent to NewRelic API when a new version of the app is deployed to a given app environment. normally, the deployment notice will be coming from a computer where `cap deploy` was run (aka `run_locally`). however, when we set `:newrelic_role` in capistrano, the deployment will run remotely on servers having that role. if you have 3 servers on `:app` role, it will run the "deployment notice" call on each of these servers. 

we don't want 3 deployment notice here but just one. the change ensures that it will only happen once using `:primary` server on selected role.

kwugirl:
Yup, that's the same deployment notice I was talking about. I believe that having 3 deployment notices sent up is the intended behavior, because it's 1 notice per server. If you only sent up 1, when you view the UI broken down by server, it would look as though you only had a deployment on one of your 3 servers, rather than on all 3.

kwugirl:
Hi @pangkalizer I'm going to go ahead and close this PR as I believe the current setup is intentional. If you have feedback for us on how to improve our documentation around this or how these markers are displayed to you in the UI, we'd love to hear it! https://support.newrelic.com/ is probably a good location to start for that. Thanks again for looking into the code here!

PR #263, opened 2017-02-06
Use Hash#fetch instead of Hash#default_proc

plus3x:
Hash#default_proc can't be marshalled, it brokes chaching.

### DIFF
--- a/lib/new_relic/agent/transaction_metrics.rb
+++ b/lib/new_relic/agent/transaction_metrics.rb
@@ -10,11 +10,11 @@
 module NewRelic
   module Agent
     class TransactionMetrics
-      DEFAULT_PROC = Proc.new { |hash, name| hash[name] = NewRelic::Agent::Stats.new }
+      DEFAULT_STATS = NewRelic::Agent::Stats.new
 
       def initialize
-        @unscoped = Hash.new(&DEFAULT_PROC)
-        @scoped   = Hash.new(&DEFAULT_PROC)
+        @unscoped = {}
+        @scoped = {}
       end
 
       # As a general rule, when recording a scoped metric, the corresponding
@@ -38,7 +38,7 @@ module NewRelic
       end
 
       def [](key)
-        @unscoped[key]
+        @unscoped.fetch(key, DEFAULT_STATS)
       end
 
       def each_unscoped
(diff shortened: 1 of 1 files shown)


kwugirl:
Hi @plus3x ! Can you tell me a bit more about what caching is being broken here? Do you have a test case handy for where the current code is breaking?

plus3x:
Simple way to reproduce:

kwugirl:
Thanks for the example, @plus3x. Can you share more about in what context you're needing to do this `Marshal.dump`?

plus3x:
```ruby
NewRelic::Agent::Transaction.wrap(NewRelic::Agent::TransactionState.tl_get, nil, nil) do
  Rails.cache.fetch('key') { Typhoeus.get 'https://google.com' }
end
#=> ETHON: performed EASY effective_url=https://google.com/ response_code=302 return_code=ok total_time=0.020555
#=> TypeError: can't dump hash with default proc
```
In my case it's  generates automatically on controller action.

kenichi:
Hello @plus3x, thanks again for your response.

In the above example, the value being stored in the cache is the return of `Typhoeus.get`, which is a `Typhoeus::Response` instance. This object shouldn't have any references to a `NewRelic::Agent::TransactionMetrics` instance; if that is what you are seeing, please let us know. In my investigations, I found what I believe you are seeing:

```
    591: def save_block_result_to_cache(name, options)
    592:   result = instrument(:generate, name, options) do
    593:     yield(name)
    594:   end
    595:
 => 596:   write(name, result, options)
    597:   result
    598: end

[1] pry(#<ActiveSupport::Cache::MemoryStore>)> result.class
=> Typhoeus::Response
[2] pry(#<ActiveSupport::Cache::MemoryStore>)> Marshal.dump(result)
TypeError: can't dump hash with default proc
from (pry):2:in `dump'
```

In this case, I believe the default_proc you are running into is from [typhoeus](https://github.com/typhoeus/typhoeus/blob/master/lib/typhoeus/response/header.rb#L87-L96). Again, please let us know if you have tracked this down to `TransactionMetrics` in some other way.

kwugirl:
Hi @plus3x based on the information we have so far, it looks like this change would not fix the issue you're having, so we'll be closing this PR. If new information comes to light though, please do let us know!

PR #264, opened 2017-03-02
Support NEW_RELIC_ENV when Rails is detected

andreacfm:
I have a use case where I need to boot the agent in a Rails app suing a different new_relic env. While the app boot in production I need to configure the agent under the qa environment. Using the NEW_RELIC_ENV is the most easy way to do this but looks like when rails is detected teh env var is not considered.

Is this a mergeable approach?

### DIFF
--- a/lib/new_relic/control/frameworks/rails.rb
+++ b/lib/new_relic/control/frameworks/rails.rb
@@ -12,7 +12,7 @@ module NewRelic
       class Rails < NewRelic::Control::Frameworks::Ruby
 
         def env
-          @env ||= RAILS_ENV.dup
+          @env ||= ( ENV['NEW_RELIC_ENV'] || RAILS_ENV.dup )
         end
 
         # Rails can return an empty string from this method, causing
--- a/lib/new_relic/control/frameworks/rails3.rb
+++ b/lib/new_relic/control/frameworks/rails3.rb
@@ -16,7 +16,7 @@ module NewRelic
       class Rails3 < NewRelic::Control::Frameworks::Rails
 
         def env
-          @env ||= ::Rails.env.to_s
+          @env ||= ( ENV['NEW_RELIC_ENV'] || ::Rails.env.to_s )
         end
 
         def rails_root


JangoSteve:
I just came across this same issue, thought I was going crazy because [the documentation says the order of preference for the Ruby agent is](https://docs.newrelic.com/docs/agents/ruby-agent/configuration/ruby-agent-configuration):

> The order of the environment variables the Ruby agent looks for are:
> 1. NEW_RELIC_ENV
> 2. RUBY_ENV
> 3. RAILS_ENV
> 4. RACK_ENV

However, every time I booted the app, I could see in the newrelic logs that it would completely ignore the `NEW_RELIC_ENV` variable.

However, I see that when you have a Rails app (Rails 3 or greater), it inserts these into the top:

```
1. ::Rails.env.to_s
2. RAILS_ENV.dup
```

And since `Rails.env` wraps `ENV["RAILS_ENV"]` and `ENV["RACK_ENV"]` ([see source](https://github.com/rails/rails/blob/807b0cb9935e2bf602e36d7a232e1676af906169/railties/lib/rails.rb#L70)), the above essentially moves those to the top, making the actual order of precedence _not at all what is stated in the docs_. For a Rails app, it's actually:

```
1. RAILS_ENV
2. RACK_ENV
```

If I had to guess, I'd say that this is unintended as a result of the inheritance pattern of `NewRelic::Control::Frameworks::Rails5 < NewRelic::Control::Frameworks::Rails4 < NewRelic::Control::Frameworks::Rails3 < NewRelic::Control::Frameworks::Rails < NewRelic::Control::Frameworks::Ruby`, in that the `Ruby` class has the right behavior, but then the `#env` method in the `Rails` classes take precedence.

I'd say the better fix would be to rewrite the `#env` methods in the `Rails3` and `Rails` classes to something like:

```
def env
  @env ||= super || ::Rails.env.to_s
end
```

However, that wouldn't work, because then if no env variables are set, you'd always end up with `"development"` since that's last value in the `Ruby#env` control which is always set.

Frankly, I don't fully understand why the `::Rails.env.to_s` is needed at all! It just wraps the env variables you're already looking at, and in the same order, even back in Rails 3.0 ([see source](https://github.com/rails/rails/blob/3-0-stable/railties/lib/rails.rb#L79)).

I think the _best_ fix to all this would be to remove the `#env` method entirely in the `NewRelic::Control::Frameworks::Rails` and `NewRelic::Control::Frameworks::Rails3` classes, and just let the default `NewRelic::Control::Frameworks::Ruby#env` method do the work.

On a side note, even the code in the `NewRelic::Control::Frameworks::Ruby` class doesn't match the documentation, as the docs don't mention anything about defaulting to "development". Even with the fix mentioned above, the docs should be changed to reflect the default value as well:

> The order of the environment variables the Ruby agent looks for are:
> 1. NEW_RELIC_ENV
> 2. RUBY_ENV
> 3. RAILS_ENV
> 4. RACK_ENV
> 5. "development"

kwugirl:
Hi @andreacfm and @JangoSteve -- thank you both very much for looking into this issue and proposing changes for how we can address it. @JangoSteve your writeup was most impressively thorough and definitely helped us navigate through a section of our codebase that those of us currently on the team hadn't spent much time with before! So to address your points:

1. Based on your feedback, we've updated our documentation on configuring the agent for different environments (https://docs.newrelic.com/docs/agents/ruby-agent/configuration/ruby-agent-configuration) to call out the current behavior difference for Rails vs. non-Rails apps. We also tried to call out that final fallback for non-Rails apps is `development` (though it's not in the numbered list since the list is described as environment variables and `development` isn't an environment variable...though maybe we should change the title of the list?) Please let us know if this is clearer than what we'd had previously, since we'll want this documentation on what to expect up until the recently release 4.0.0 version. We certainly want to reduce confusion more than we cause it!

2. Looking into the history of the code a bit, this area of the code does seem like it's been tacked onto over time. Thanks again @JangoSteve for doing all that research into even older Rails source code to see what we actually end up calling there, that was very helpful.

3. We've decided to pull in this set of changes over https://github.com/newrelic/rpm/pull/265 which should be part of the next release. @JangoSteve makes an good case for simplifying the behavior via taking advantage of the inheritance that's already set up in there, but going up to `NewRelic::Control::Frameworks::Ruby` also means adding in `RUBY_ENV` prior to `RAILS_ENV` of variables to check. We thought it would be safer to more directly target the desired new functionality here of being able to use either just `NEW_RELIC_ENV` or `RAILS_ENV`. 

Please let me know if there are any further questions!

JangoSteve:
That's a shame, it took a good amount of extra time investment to navigate the test suite and get the actual automated tests working which prove the other pull request works. I guess we'll just keep using the forked version that's better tested for this functionality.

I would also consider the behavior you're describing in (3) to be yet _another_ bug. Why would the Ruby hook look at `ENV["RAILS_ENV"]` at all? That seems unexpected, especially since the env method gets overridden for Rails apps anyway. In other words, the issue you're describing with (3) could be fixed by deleting even more code, just delete `|| ENV["RAILS_ENV"]` from the `Ruby#env` method.

If it were me, I'd go with #265 with the issue fixed and tested, and just update it to remove the check for `ENV["RAILS_ENV"]` from `Ruby#env`, and add that instead to `Rails#env`. Then you get tests, clearer code that matches the documentation, and less indirection.

We'll probably just update our fork with the change above and keep using that one.

JangoSteve:
A few additional thoughts...

These two statements in your response seem to be at odds with each other:

> Looking into the history of the code a bit, this area of the code does seem like it's been tacked onto over time.

> We thought it would be safer to more directly target the desired new functionality here of being able to use either just `NEW_RELIC_ENV` or `RAILS_ENV`.

So, the entire problem which led to this bug is that the code has been tacked onto over time without any sort of tests, and the solution is to tack on _more code, once again_, without any tests. How is that safer? If what you mean is that tacking on more code is preferable to any actual refactoring due to the lack of tests, that makes sense.

But I don't think it's helpful to speculate that something _could_ break in legacy apps due to not understanding how those legacy applications, which this code integrates with, actually work. Especially not when you can literally look at the source code, even through the legacy versions, to understand how the underlaying applications you're integrating with actually behave.

As a thought exercise, lets try building up the [untested] intended behavior as described in the updated documentation, but with an understanding of how the applications work that we want to support. From the current codebase, I'd say it's:

* Rails (let's go back to the oldest version labeled in the Rails Github repo, which is [1.2 stable](https://github.com/rails/rails/tree/1-2-stable), which I'm going to assume is what the `NewRelic::Control::Frameworks::Rails` class is targeting)
* Rack (since there are no version-specific Rack, Sinatra, etc. files, I'm going to assume we're just targeting the latest version for these)
* Sinatra
* And just plain old Ruby

## Rails

### 1.2 stable

Uses the `RAILS_ENV` constant. [This is set in the Railties initializer](https://github.com/rails/rails/blob/1-2-stable/railties/lib/initializer.rb#L10), and it looks at `ENV["RAILS_ENV"]` or defaults to "development":
```
RAILS_ENV = (ENV['RAILS_ENV'] || 'development').dup unless defined?(RAILS_ENV)
```

Based on the above, it looks like this is what the `NewRelic::Control::Frameworks::Rails` class is targeting since it looks for the `RAILS_ENV` constant.

However, [even Rails 1.2 had the `Rails.environment` method](https://github.com/rails/rails/blob/1-2-stable/railties/lib/initializer.rb#L562-L564), which simply just returned `RAILS_ENV`:

```
    def environment
      ::RAILS_ENV
    end
```

[Rails 2.0 was the same as 1.2](https://github.com/rails/rails/blob/2-0-stable/railties/lib/initializer.rb#L12).

### 2.1

[Rails 2.1 introduced `Rails.env`](https://github.com/rails/rails/blob/2-1-stable/railties/lib/initializer.rb#L38-L41), but [kept `Rails.environment` the same as v1.2-2.0](https://github.com/rails/rails/blob/2-1-stable/railties/lib/initializer.rb#L727-L729). The new `Rails.env` still looks at the same `RAILS_ENV` constant, which still behaves the same as always, but wraps it as a `ActiveSupport::StringInquirer`:

```
    def env
      require 'active_support/string_inquirer'
      ActiveSupport::StringInquirer.new(RAILS_ENV)
    end
```

All this does is let's you query the `==` status of a string as a method (e.g. for an instance of StringInquirer, `s == "production"` is equivalent to `s.production?`). No other magic here. Since the current implementation in rpm is to turn `Rails.env`

### 3.0 and beyond

From 3.0 onward, [`RAILS_ENV` is deprecated in favor of `Rails.env`](https://github.com/rails/rails/blob/92703a9ea5d8b96f30e0b706b801c9185ef14f0e/guides/source/3_0_release_notes.md#railties), and [the `Rails.env` method now additionally looks at `ENV["RACK_ENV"]`](https://github.com/rails/rails/blob/3-0-stable/railties/lib/rails.rb#L79-L81) instead of the `RAILS_ENV` constant:

```
    def env
      @_env ||= ActiveSupport::StringInquirer.new(ENV["RAILS_ENV"] || ENV["RACK_ENV"] || "development")
    end
```

### Rails summary

We can just use `RAILS_ENV` for Rails 1.2-2.0, and for everything else, we use `Rails.env`. Since we need to provide the ability to override, we'd prepend these with a lookup for `ENV["NEW_RELIC_ENV"]`.

For Rails at least, it does seem like this pull request accomplishes this.

## Rack

[Rack just uses `ENV["RACK_ENV"]` and defaults to "development"](https://github.com/rack/rack/blob/1482870b381b1517305c8c0aac05feb4d5b3aa6a/lib/rack/server.rb#L205):

```
environment  = ENV['RACK_ENV'] || 'development'
```

It seems odd there's not a Rack class in rpm, but as it is, the standard Ruby can look up the `ENV["RACK_ENV"]`. By this logic though, Rails could do the same, and doesn't need the `env` method overridden at all in the separate Rails control classes.

## Sinatra

[Sinatra actually uses `ENV["APP_ENV"]` or `ENV["RACK_ENV"]` (converted to a symbol) or defaults to `:development`](https://github.com/sinatra/sinatra/blob/379e360f5db78352acbf1263ab290b219f2d1df2/lib/sinatra/base.rb#L1793):

```
set :environment, (ENV['APP_ENV'] || ENV['RACK_ENV'] || :development).to_sym
```

Oddly enough, neither the Sinatra control class or the Ruby class in rpm looks at `ENV["APP_ENV"]`, so this would seem to be a bug with the Sinatra integration.

## Leaking environment variables across apps

One last point to consider is whether or not the rpm library is concerned with environment variables leaking across apps on the same server. For example, if I have a server with both a Rails app and a Sinatra app running on two different ports, I may consider it safe to include `RAILS_ENV=production` and `APP_ENV=staging` in my global environment on the server, since any non-Rails app would never look at `ENV["RAILS_ENV"]` and I know Rails doesn't look at `ENV["APP_ENV"]`.

However, if I have the New Relic rpm gem installed in both apps, my Sinatra app is going to keep reporting its metrics with my Rails app's environment!

(OK, so we've found bug no. 2, in addition to bug no. 1 above where the Sinatra integration doesn't use the `ENV["APP_ENV"]` like Sinatra apps do)

## Summary

### Better Approach - Option 1

If the rpm gem is not concerned with leaking environment variables between apps, then we don't need to override the `env` method in any of the `Rails` control classes, and we could just update the base `Ruby` control class's method to be this:

```
        def env
          @env ||= ENV['NEW_RELIC_ENV'] || ENV['RAILS_ENV'] || ENV['APP_ENV']
                   ENV['RUBY_ENV'] || ENV['RACK_ENV'] || 'development'
        end
```

Notice in the above, I moved `RAILS_ENV` to the front of the line, since if it's a Rails app, the Rails app's environment is always computed by the environment variables, `RAILS_ENV`, `RACK_ENV`, and defaults to "development". So, in the above, I've preserved that order (and since we're not worried about environment variables leaking across apps, we don't care that `APP_ENV` and `RUBY_ENV` are in the middle, because they just won't be defined as far as we're concerned). But also, we don't want something generic like `RUBY_ENV` potentially overriding our more specific `RAILS_ENV` variable, so that's why we changed that order.

We also added `APP_ENV` above to take care of Sinatra apps.

And now none of our framework control classes need to override the `env` method, because we've already duplicated the logic of all of them in our base `Ruby` control instance method, so we could remove those entirely (which is what I did in #265).

### Better Approach - Option 2

However, if we _are_ worried about variables leaking, then this pull request still exhibits those issues, because we'd want to override the `env` method for both Sinatra and Rails (this pull request only keeps it overridden for Rails), _and_ we'd want to remove the Rails-specific `ENV["RAILS_ENV"]` from the `Ruby#env` method, and make it this instead:

```
        def env
          @env ||= ENV['NEW_RELIC_ENV'] || ENV['RACK_ENV'] ||
                   ENV['RUBY_ENV'] || 'development'
        end
```

Also, note that in the above, I switched the order of `RACK_ENV` and `RUBY_ENV`, because every Rack app is Ruby, but not every Ruby app is Rack, so we should put the more specific `RACK_ENV` first so that the more generic `RUBY_ENV` doesn't potentially hi-jack our env setting in our Rack app.

And then we'd want to keep the changes to the Rails and Rails3 control classes from this pull request.

But we'd also want to add the correct env method to the Sinatra control class:

```
        def env
          @env ||= ENV['NEW_RELIC_ENV'] || ENV['APP_ENV'] || ENV['RACK_ENV'] || 'development'
        end
```

andreacfm:
@JangoSteve @kwugirl Please consider my code as a proposal. I had to fix this issue in a fast way and this did the trick. I am using the fork and works as expecetd. I am opened to every contribution you may suggest that could be mergeable into the next release.

sairam:
Hi there, 

I came across from the documentation similar to the 2nd comment in the thread. 

Thought I would submit a PR, but came across this one. Any plans to merge this soon?

The project I am working on needs this. (we run with RAILS_ENV=production) on Staging as well as Production boxes.

kwugirl:
Good timing @sairam--we actually just merged it in on our internal repo yesterday (there was some last minute back and forth on whether to move forward with this feature) so it should be in the next release, which we are planning for early April. 

Out of curiosity though, I'd be interested to learn a bit more about why it is that you would have different Rails environments versus the environments you'd want the Ruby agent to boot up in for monitoring? Is there a particular restriction on creating additional Rails environments to match the different Ruby agent environments?

sairam:
I am working on a 2+ year old code base (there is existing infrastructure setup with these assumptions).

I think the assumption was to have an exact configuration apart from ENV variables to be modified across `staging` and `production` in the infra setup. Personally, I find this approach cleaner than having a separate environment name. 

We modified `AIRBRAKE_ENV` and `CLOUDINARY_ENV` to use `staging` and wanted to see if newrelic had the same.

sairam:
@kwugirl came across NEW_RELIC_APP_NAME which solves my problem.

kenichi:
This has been released in version 4.1.0! Thanks, @andreacfm!

PR #265, opened 2017-03-09
Fixed issue with NEW_RELIC_ENV environment variable not being used for Rails apps

JangoSteve:
(Includes test)

See comment in related pull request, and why this pull request would be my preferred way of fixing:

https://github.com/newrelic/rpm/pull/264#issuecomment-285362792


kwugirl:
Thanks again @JangoSteve for looking into this--see https://github.com/newrelic/rpm/pull/264#issuecomment-288552069 for discussion on why we chose the approach we did. tl;dr we're prioritizing minimizing scope of change to reduce risk of unintended effects over cleaner/more unified code.

PR #266, opened 2017-03-29
Fix git scm detection on capistrano3 recipe

mengqing:
Prior to capistrano v3.7.0, scm can be fetched through Capistrano::DSL

Since capistrano v3.7.0, scm has been deprecated and replaced by plugin system, in which
scm will simply return nil.

An alternative solution would be to check if scm specific rake task is present, eg: "git:check", "hg:check", etc

### DIFF
--- a/lib/new_relic/recipes/capistrano3.rb
+++ b/lib/new_relic/recipes/capistrano3.rb
@@ -34,7 +34,7 @@ namespace :newrelic do
       user        = fetch(:newrelic_user)
       license_key = fetch(:newrelic_license_key)
 
-      unless scm == :none
+      if (respond_to?(:scm_plugin_installed?) && scm_plugin_installed?) || (scm.present? && scm != :none)
         changelog ||= lookup_changelog
         rev       ||= fetch(:current_revision)
       end
@@ -69,7 +69,7 @@ namespace :newrelic do
 
     debug "Retrieving changelog for New Relic Deployment details"
 
-    if scm == :git
+    if Rake::Task.task_defined?("git:check")
       log_command = "git --no-pager log --no-color --pretty=format:'  * %an: %s' " +
                     "--abbrev-commit --no-merges #{previous_revision}..#{current_revision}"
       `#{log_command}`


kwugirl:
Hi @mengqing thanks for submitting this fix! I'm not very familiar with Capistrano but have been reading the documentation and came across https://github.com/capistrano/capistrano/blob/master/UPGRADING-3.7.md which covers what you said about Capistrano v3.7.0+ replacing the scm variable with a plugin system instead. I could also see that `Rake::Task.task_defined?("git:check")` only returns true after:
```
require "capistrano/scm/git"
require "capistrano/deploy"
install_plugin Capistrano::SCM::Git
```
have all been run (I'm just testing in an irb console).

However, I'm not quite sure how to verify whether you can also run `Rake::Task.task_defined?("git:check")` and get the desired result of `true` if you are still on Capistrano v3.6.1 or earlier? Do you have any additional guidance on that part? I'd like to be able to maintain compatibility across the different 3-series versions of Capistrano with this recipe.

Additionally, if this recipe is going to be updated to move away from using the scm variable, I think there's another check in this file that it would be great to see updated as well.

Thanks!

mengqing:
Hi, I'm wondering if this PR is still going to be accepted. Sorry for the long delay until I recently checked my PR lists.

@kwugirl I'm no longer using newrelic in my current work place, but to answer your question, prior to Capistrano v3.7, the rake task simply get required based on the `scm` setting set in the capistrano deploy.rb

eg:

```
# deploy.rb
set :scm, :git
```

and that will load the relevant rake task, https://github.com/capistrano/capistrano/blob/v3.6.1/lib/capistrano/git.rb#L1

As to your second question, I have updated the code to check for both prior to v3.7 and after v3.7

kwugirl:
@mengqing I'm no longer working at New Relic myself, but @mwear or @danascheider can take a look I'm sure. Good luck!

danascheider:
@mengqing Sorry to take so long with this! I'm taking a look at it, although I will also have to familiarize myself more with Capistrano before I can really review this effectively. Thank you for your patience!

undees:
Hello, @mengqing! I'd like to echo the words of thanks from @kwugirl and @danascheider, and to let you know that we're still investigating your PR for possible inclusion in the agent.

undees:
Hi @mengqing. We wanted to let you know that your changes have been integrated into version 5.0.0 of the agent, which was released this morning. Thank you for your contribution!

PR #267, opened 2017-04-30
monitoring should default to off in test environment

notEthan:
monitor_mode should default to false when newrelic's test_mode is true. this is consistent with the default values generated for newrelic.yml: https://github.com/newrelic/rpm/blob/4.1.0.333/newrelic.yml#L33-L36 

test_mode should default to true when env == 'test'.

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -272,7 +272,9 @@ module NewRelic
           :description => 'Specify the <a href="https://docs.newrelic.com/docs/apm/new-relic-apm/installation-configuration/name-your-application">application name</a> used to aggregate data in the New Relic UI. To report data to <a href="https://docs.newrelic.com/docs/apm/new-relic-apm/installation-configuration/using-multiple-names-app">multiple apps at the same time</a>, specify a list of names separated by a semicolon <code>;</code>. For example, <code>MyApp</code> or <code>MyStagingApp;Instance1</code>.'
         },
         :monitor_mode => {
-          :default => value_of(:enabled),
+          :default => Proc.new {
+            NewRelic::Agent.config[:enabled] ? NewRelic::Control.instance.env != 'test' : false
+          },
           :public => true,
           :type => Boolean,
           :allowed_from_server => false,
--- a/test/new_relic/agent/agent_test.rb
+++ b/test/new_relic/agent/agent_test.rb
@@ -19,7 +19,7 @@ module NewRelic
         @agent.agent_command_router.stubs(:new_relic_service).returns(@agent.service)
         @agent.stubs(:start_worker_thread)
 
-        @config = { :license_key => "a" * 40 }
+        @config = { :license_key => "a" * 40, :monitor_mode => true }
         NewRelic::Agent.config.add_config_for_testing(@config)
       end
 


notEthan:
I haven't been able to reproduce the remaining failures locally. if the maintainers have any advice or feedback that would be appreciated.

mwear:
Hi @notEthan, I was wondering if you could explain what issue you are trying to solve with this PR? `test_mode` was included as a substitution for `developer_mode` which was removed during this comiit: https://github.com/newrelic/rpm/commit/45f40d042c7025c32412884031ad14c53a13c3c3. This is meant to be a temporary change and `test_mode` is likely to be removed from the agent altogether. This is meant to be internal to the agent and is only here to facilitate running internal tests.

notEthan:
the issue is that if newrelic.yml is not present, then newrelic defaults to enabled in test environment and makes requests to newrelic servers, when it should be disabled. we have newrelic.yml generated by our deployment process and not stored in our git repo. 

I used `test_mode` since it seemed like an appropriate place to put this even though it seems to be otherwise unused. if that setting isn't staying around, then the default proc for `monitor_mode` could look at the `env` instead to decide whether to enable.

mwear:
If your goal is to set `monitor_mode` to false without having a `newrelic.yml` file, this can be done by setting an environment variable. You can set `NEW_RELIC_MONITOR_MODE=false` when running your tests. For situations where you do have a `newrelic.yml` these options can be set per environment, or using erb for more control. Since the future of test_mode is uncertain, this is probably your best way forward for now. Let us know if this works for you.

notEthan:
sure, I know there are ways I can turn off monitor mode. the point is that it's better for the gem to apply sensible defaults, though. having its default behavior match the defaults generated in newrelic.yml is the drive of this PR. I don't care if it uses test_mode or just sets monitor_mode directly - I have updated the PR to not touch test_mode given its uncertain fate.

mwear:
Hi @notEthan. This PR has quite a few test failures. Are you still working on this or should I close it out?

notEthan:
sounded like it might not be wanted so I did not put more into it - if it will be accepted with tests passing I will get them passing, but as I noted above I was not having success in producing those test failures locally. I can put some time in to get them passing, but if you have some advice on what might be the issue that'd be much appreciated.

mwear:
Hi @notEthan. It's not that the work isn't wanted, it's mainly that these changes need to play well with our testing environment.  Speaking of which, you may notice that there are a number of suites that run on the Travis build. We have our unit tests, which are invoked using `bundle exec rake test` and we have our multiverse suite (integration tests), invoked using `bundle exec rake test:multiverse`. On Travis, units and multiverse will be run on a number of ruby versions. In addition multiverse is divided into a number of suites. For the sake of your changes, you should the run the unit tests and multiverse on at least one version of ruby locally before pushing up your changes. Multiverse does not lock down all of the dependencies for individual suites, as a result, sometimes the release of a new gem version can cause failures and we have to add special case code to handle the situations that arise there. I would recommend rebasing your current work and trying to get a clean pass through the unit tests and multiverse.

travisofthenorth:
+1

travisofthenorth:
@mwear sorry if this isn't the place, but it seems topical. Is there a way to change settings at runtime? I've been playing around with 4.2, and it seems like the only way to do this is to reset the config cache, grab a config source (instance variable), and set it manually. Am I missing a much easier way to do this?

For context: a lot of my apps use a boilerplate gem which includes NewRelic. We're trying to default `test_mode` in dev/test like @notEthan , but the NR code is already required by the time we can inject ourselves in there.

mwear:
@travisofthenorth: Preventing agent startup programmatically is going to be challenging as it begins starting in a background thread when the gem is required. Therefore attempts at preventing startup programmatically will be too late. The best option pass an environment variable that will prevent startup in the first place. Passing`NEW_RELIC_MONITOR_MODE=false` will achieve what you're looking for.

Since this PR has been idle for quite some time I am going to close it. We'll gladly accept of a working version if anyone would like to contribute it.

PR #268, opened 2017-05-16
Stop packaging test files with the gem

dimko:
It reduces the size of the gem by half, from 645K to 294K. The default for new gems (created by bundler) is not to include test files. See bundler/bundler#3207


kenichi:
Thank you @dimko, we will be pulling this into our 4.2.0 release. Stay tuned!

kenichi:
This has been released in version 4.2.0! Thanks, @dimko!

PR #269, opened 2017-05-16
Check Rails::VERSION instead of just Rails

AlexRiedler:
Sometimes gems will introduce the module 'Rails' just to get around internal works (sucks but its true);

Since this is trying to access something that could or could not be defined; why not just check the entire tree instead for safety reasons.

Anyone that actually has Rails should have the VERSION field.

FIXES: this stops error collecting if Rails module is defined; but Rails::VERSION is not.

### DIFF
--- a/lib/new_relic/noticed_error.rb
+++ b/lib/new_relic/noticed_error.rb
@@ -165,7 +165,7 @@ class NewRelic::NoticedError
       @exception_class_name = UNKNOWN_ERROR_CLASS_NAME
       @message = NIL_ERROR_MESSAGE
     else
-      if defined?(Rails) && Rails::VERSION::MAJOR < 5 && exception.respond_to?(:original_exception)
+      if defined?(Rails::VERSION) && Rails::VERSION::MAJOR < 5 && exception.respond_to?(:original_exception)
         exception = exception.original_exception || exception
       end
       @exception_class_name = exception.is_a?(Exception) ? exception.class.name : UNKNOWN_ERROR_CLASS_NAME


kenichi:
@AlexRiedler thank you for the contribution! There are a few more places this constant is checked, was there something specific about this one that was causing you problems vs. the others? Are you only concerned with it not collecting errors? Also, can you give us an example or two of a gem that defines `Rails` outside of rails?

AlexRiedler:
@kenichi The reason why is cause error collection fails; but error counts are still collected (very confusing experience). I am not sure why the others do not seem to be affected by this but this is the only one that affected me.

Hmmm I am using rails-html-sanitizer; but I just read now that you should not use it outside Rails and just use Loofah. Maybe I can avoid this in my codebase directly (seems to be a dependency issue on my end).

Maybe this change is not needed then; although I feel like if rails framework is off this shouldn't happen? hmmm, should that be addressed instead? (otherwise I think closing should be fine!)

danascheider:
@AlexRiedler I think the change is not needed, so I'm closing this. Please let us know if you need help with anything else :)

TheWudu:
Hi!

We just encountered the same problem. As a lot of gems, not only the rails-html-sanitizer define the rails module (for whatever reason), it would be nice if that check works properly. 

The check itself only needs the `Rails::VERSION`, so it would be proper to only check for `Rails::VERSION`. 

But @danascheider I would also vote for adapting it at all the places where it is used. 

Thanks in advance.

mwear:
Thanks for the contribution @AlexRiedler. Your changes are part of Ruby Agent version 4.7.0.339 which was released on 12/19.

PR #270, opened 2017-05-23
Verify uri.host before downcasing to prevent NilClass error

rafaelv90:
# Overview
Normalization of external hostnames was introduced in `v4.1.0`. One of the steps of this normalization is downcasing external hostnames.

## Problem
In case of an invalid URL (user entered information or specs), the hostname can be `nil`. Then calling `downcase!` in such object would result in `NoMethodError: undefined method 'downcase' for nil:NilClass`.

## Proposed Solution
A simple `nil` check before trying to downcase the hostname would solve this, since if no hostname is available, there's no need for using `downcase!`

### DIFF
--- a/lib/new_relic/agent/http_clients/uri_util.rb
+++ b/lib/new_relic/agent/http_clients/uri_util.rb
@@ -37,7 +37,7 @@ module NewRelic
               uri = ::URI.parse(url)
             end
           end
-          uri.host.downcase!
+          uri.host.downcase! unless uri.host.nil?
           uri
         end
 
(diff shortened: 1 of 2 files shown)


kenichi:
@rafaelv90 thank you! that is indeed a good check to have there. what do you think about adding a test for this case in test/new_relic/agent/http_clients/uri_util_test.rb ?

rafaelv90:
@kenichi Nice! Will submit the commit as soon as I can!

rafaelv90:
@kenichi Do you think it's good now?

danascheider:
@rafaelv90 We've got some test failures that we're having trouble reproducing locally, so we're working on figuring out the source of that. Most likely it has to do with incompatible gem versions - something new got released and broke it. So once we figure that out it should be good to go. 👍

danascheider:
This has been accepted, we rebased it to fix the failures and it's being merged in our private repo - we'll release it with our next release. Thanks @rafaelv90!

danascheider:
This has been released in version 4.3.0! Thanks, @rafaelv90!

PR #271, opened 2017-06-06
Requires Datastores in Agent.

kevingriffin:
Previously this module was being defined implictly by
other classes in its namespace, without the methods being defined.


danascheider:
@kevingriffin This has been accepted and will be merged pending test passage after rebase. The failures were due to a gem update and were fixed in our last release, so that shouldn't be an issue. Your changes will be included in our next release. Thanks for the contribution!

danascheider:
This has been released in version 4.3.0! Thanks, @kevingriffin!

PR #272, opened 2017-06-29
Support Postgis ActiveRecord adapter as Postgres

vvondra:
The postgis adapter is wraps the original postgres adapter https://github.com/rgeo/activerecord-postgis-adapter/blob/master/lib/active_record/connection_adapters/postgis_adapter.rb#L30

Right now APM falls back to ActiveRecord instead of Postgres


vvondra:
Tested for us in production on a gem installed from the fork branch, works without issues 👌

mwear:
@vvondra Thanks for the contribution. It will appear in the next release of the agent.

danascheider:
This has been released in version 4.3.0! Thanks, @vvondra!

PR #273, opened 2017-09-01
Support slow SQL traces with postgis adapter

aripollak:
Explicit PostGIS support was recently added, but slow SQL traces still weren't working because postgis didn't translate to a supported adapter in this part of the code. This fixes that.


kenichi:
@aripollak thank you, we will be pulling this in to our 4.5 release. stay tuned!

kenichi:
This has been released in version 4.5.0! Thanks, @aripollak!

PR #274, opened 2017-10-11
Monitor mode default false in development/test

travisofthenorth:
This piggy-backs off of the work in this PR: https://github.com/newrelic/rpm/pull/267

This attempts to set a sensible default for `monitor_mode`, i.e. it should be disabled in development/test. There are cases where it is cumbersome to control Newrelic integration with yml/env configuration. In my specific case, we use a boilerplate gem in all of our apps that includes this one, so ideally we would be able to avoid NR configuration in each new app.

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -272,7 +272,9 @@ module NewRelic
           :description => 'Specify the <a href="https://docs.newrelic.com/docs/apm/new-relic-apm/installation-configuration/name-your-application">application name</a> used to aggregate data in the New Relic UI. To report data to <a href="https://docs.newrelic.com/docs/apm/new-relic-apm/installation-configuration/using-multiple-names-app">multiple apps at the same time</a>, specify a list of names separated by a semicolon <code>;</code>. For example, <code>MyApp</code> or <code>MyStagingApp;Instance1</code>.'
         },
         :monitor_mode => {
-          :default => value_of(:enabled),
+          :default => Proc.new {
+            NewRelic::Agent.config[:enabled] ? NewRelic::Control.instance.env != 'test' : false
+          },
           :public => true,
           :type => Boolean,
           :allowed_from_server => false,
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -273,7 +273,8 @@ module NewRelic
         },
         :monitor_mode => {
           :default => Proc.new {
-            NewRelic::Agent.config[:enabled] ? NewRelic::Control.instance.env != 'test' : false
+            development_or_test =  %w(test development).include?(NewRelic::Control.instance.env)
+            NewRelic::Agent.config[:enabled] ? !development_or_test : false
           },
           :public => true,
           :type => Boolean,
--- a/test/agent_helper.rb
+++ b/test/agent_helper.rb
@@ -408,7 +408,8 @@ def find_all_nodes_with_name_matching(transaction_sample, regexes)
 end
 
 def with_config(config_hash, at_start=true)
-  config = NewRelic::Agent::Configuration::DottedHash.new(config_hash, true)
+  defaults = { :monitor_mode => true }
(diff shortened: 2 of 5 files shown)


travisofthenorth:
@notEthan wasn't kidding about those test failures 😆

mwear:
@travisofthenorth thanks for working on this. As you can see this is not as straightforward of a task as it appears on the surface. Sadly, we will not be able to accept a pull request that disables `monitor_mode` in the development environment. A number of users use NewRelic in development.

There may be other ways to solve your issue. Out of curiosity, how do you run tests for your application? The reason that I ask, is that the agent checks certain blacklists to see if it should start up or not. You can see that logic [here](https://github.com/newrelic/rpm/blob/2bc79e102185861ff078fd15ccbf31a32f088b8a/lib/new_relic/agent/autostart.rb#L26-L30). Generally the agent is prevented from running during testing due to the `in_black_listed_rake_task?` check. For example, `test` and `spec` are both in that list.

The list of rake tasks we check is extensive, but it's possible we may be missing one if you run your tests using rake. It's also possible you have a custom task we don't recognize. You can see the full list of tasks we check for [here](https://github.com/newrelic/rpm/blob/2bc79e102185861ff078fd15ccbf31a32f088b8a/lib/new_relic/agent/configuration/default_source.rb#L193-L242). If we are missing a common task it might just be easier to make PR to add it in. If you have a custom task, you are able to specify your own list of blacklisted rake tasks in your `newrelic.yml`. The configuration option is: [`autostart.blacklisted_rake_tasks`](https://github.com/newrelic/rpm/blob/2bc79e102185861ff078fd15ccbf31a32f088b8a/lib/new_relic/agent/configuration/default_source.rb#L393-L398). You could do something like the following in your `newrelic.yml`:

```
autostart:
  blacklisted_rake_tasks: <%= ::NewRelic::Agent::Configuration::AUTOSTART_BLACKLISTED_RAKE_TASKS + ",mytask" %>
```
We also have [`autostart.blacklisted_constants`](https://github.com/newrelic/rpm/blob/2bc79e102185861ff078fd15ccbf31a32f088b8a/lib/new_relic/agent/configuration/default_source.rb#L379-L384) and [`autostart.blacklisted_executables`](https://github.com/newrelic/rpm/blob/2bc79e102185861ff078fd15ccbf31a32f088b8a/lib/new_relic/agent/configuration/default_source.rb#L386-L392) that work similarly.

It sounds as though you are not interested in using environment variables, however, there are some changes you could make to your newrelic.yml. One option is to override the monitor_mode setting for the test environment. It would look something like this:

```
---
common: &default_settings
  license_key: XXXXXX
  enabled: true
  app_name: My App
  monitor_mode: true

production:
  <<: *default_settings
development:
  <<: *default_settings
test:
  <<: *default_settings
  monitor_mode: false
```

Another option is to use ERB in the newrelic.yml:

```
  license_key: XXXXXX
  enabled: true
  app_name: My App
  monitor_mode: <%= some_condition ? true : false %>
```
Let us know if any of this options will work for you. If not, feel free to continue working through this PR.

travisofthenorth:
> Sadly, we will not be able to accept a pull request that disables monitor_mode in the development environment

@mwear not even as a default? I would think that in dev, you would expect to set a flag to enable NR integration, not the other way around. I was pretty surprised to discover that I was sending stats to NR from my laptop just by virtue of having the gem loaded.

As I mentioned, we use a boilerplate gem that includes NR, so ideally it would require _no_ configuration (no ENV vars, no newrelic.yml) for apps using the boilerplate gem. And for the most part this works...the defaults get us everything we need in staging/prod (we have throw the app name and license key in the ENV, but this is acceptable).

So I am really targeting `development` for my needs. I included `test` as well to try to address @notEthan's concerns.

mwear:
@travisofthenorth disabling New Relic in development will be an unexpected change for users who currently rely on it and we will not be able to accept a PR making that change. 

Sorry for the confusion, but I thought you were looking for a way to disable the agent in your test environment. You should be able to put ERB with a conditional in your newrelic.yml to turn off the agent in the development environment. I understand that you have many apps that are based of a template, but the template could include that modification to the newrelic.yml file.

PR #276, opened 2017-11-22
Use lazy load hooks to hook on ActionController::Base

Edouard-chin:
👋 

Use lazy load hooks to hook on ActionController::Base

- Calling directly ActionController::Base affects the initialization process (one of them being one that controls ActionController configuration)
- As a result setting configuration inside an initializer (such as the asset_host), doesn't have any effect

cc/ @rafaelfranca @haani

### DIFF
--- a/lib/new_relic/agent/instrumentation/rails5/action_controller.rb
+++ b/lib/new_relic/agent/instrumentation/rails5/action_controller.rb
@@ -21,13 +21,12 @@ DependencyDetection.defer do
   end
 
   executes do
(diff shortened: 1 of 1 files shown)


danascheider:
Hi @Edouard-chin! Thank you for taking the time to put together this PR. We're accepting it and it will be included in our upcoming version 4.7 release. We really appreciate your help making the agent better for everyone!

Edouard-chin:
Thanks !

PR #277, opened 2017-12-08
Fix test errors caused by 585f07e

jrmhaig:
When upgrading to use `newrelic_rpm` version 4.5.0.337 our unit tests result in a large number of errors mostly like:

```
ActiveRecord::RecordInvalid: Validation failed: Branch must exist
```

There are a few others but this error is by far the largest in number. I have identified the commit that causes this (https://github.com/newrelic/rpm/commit/704fe528c245b90acfd71358a6b3928b8076372e) and I notice that the code that is moved from the end to the start of the block is also moved outside of the `ActiveSupport.on_load(:active_record) do` block. These lines have subsequently been changed into a single line:

```
::NewRelic::Agent::PrependSupportability.record_metrics_for(::ActiveRecord::Base, ::ActiveRecord::Relation)
```

If I move this back inside the `on_load` block then all our tests run without error.

I am raising a PR as I am unable to create an issue. I realise that this may not be a complete fix as other things may be affected but it prevents the test errors we are seeing so it is at least moving towards a fix.

### DIFF
--- a/lib/new_relic/agent/instrumentation/active_record_5.rb
+++ b/lib/new_relic/agent/instrumentation/active_record_5.rb
@@ -79,12 +79,11 @@ DependencyDetection.defer do
   end
 
   executes do
-    ::NewRelic::Agent::PrependSupportability.record_metrics_for(::ActiveRecord::Base, ::ActiveRecord::Relation)
-
     ActiveSupport::Notifications.subscribe('sql.active_record',
       NewRelic::Agent::Instrumentation::ActiveRecordSubscriber.new)
 
     ActiveSupport.on_load(:active_record) do
+      ::NewRelic::Agent::PrependSupportability.record_metrics_for(::ActiveRecord::Base, ::ActiveRecord::Relation)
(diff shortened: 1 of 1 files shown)


bnaysakya:
@jrmhaig coming from [this](https://discuss.newrelic.com/t/newrelic-rpm-throws-error-in-consume-message-headers/52736) issue, I downgraded newrelic_rpm to **4.3.0.335** but still I'm getting:
```
ERROR : Error in consume_message_headers
ERROR : NoMethodError: undefined method `[]' for nil:NilClass
```
But, using **newrelic_rpm** to **3.16**  solved this issue for me.

danascheider:
@jrmhaig Thank you so much for putting in the time and effort to make New Relic better for everyone! We've merged your code and are going to be releasing it in version 4.7. 

Incidentally, the `NoMethodError`s you're seeing are not related to this issue at all - they're related to a bug in our RabbitMQ instrumentation that will also be fixed in the upcoming release.

Thanks again for your work on this!

mwear:
Thanks for the contribution @jrmhaig . Your changes are part of Ruby Agent version 4.7.0.339 which was released on 12/19.

PR #279, opened 2017-12-23
Initialize NewRelic before Rails' config-initializers

tonyta:
Before, the initializer for NewRelic did not specify any ordering. In some situations, Rails will order its `load_config_initializers` before NewRelic's `newrelic_rpm.start_plugin`. This is a problem if NewRelic is called in a config-initializer (e.g. when defining method tracers) and will output the following warning:

```
Agent unavailable as it hasn't been started.
```

This change ensures that `newrelic_rpm.start_plugin` always gets loaded before config-intiializers are.

The problem is detailed and the fix is demonstrated here: https://github.com/tonyta/newrelic_warning/pull/1

---

An unrelated dependency loading issue is fixed in https://github.com/newrelic/rpm/pull/279/commits/cc9f74a53c1f2f3d9a4d05a48418b15dfcf22952 to allow tests to pass.

### DIFF
--- a/lib/newrelic_rpm.rb
+++ b/lib/newrelic_rpm.rb
@@ -37,7 +37,7 @@ elsif defined?(Rails::VERSION)
     module NewRelic
       class Railtie < Rails::Railtie
 
-        initializer "newrelic_rpm.start_plugin" do |app|
+        initializer "newrelic_rpm.start_plugin", before: :load_config_initializers do |app|
           NewRelic::Control.instance.init_plugin(:config => app.config)
         end
       end
--- a/test/multiverse/suites/sidekiq/test_worker.rb
+++ b/test/multiverse/suites/sidekiq/test_worker.rb
@@ -2,6 +2,8 @@
 # This file is distributed under New Relic's license terms.
 # See https://github.com/newrelic/rpm/blob/master/LICENSE for complete details.
 
+require_relative "sidekiq_server"
+
 class TestWorker
   include Sidekiq::Worker
 


tonyta:
The changes between https://github.com/newrelic/rpm/pull/279/commits/4b3c20034093961c0e9c48f073fadb72d857f8fe and https://github.com/newrelic/rpm/pull/279/commits/cc9f74a53c1f2f3d9a4d05a48418b15dfcf22952 don't seem to account for the new test failures (which might be related to [issues on the Rubygems server](https://status.rubygems.org/incidents/hcqztxftbpj8)). The failures are likely nondeterministic and could probably just be retried. ¯\\_(ツ)_/¯

Hope this helps!

tonyta:
Hi @mwear! Anything blocking this? Anything I can do to help?

mwear:
Hi @tonyta. Thanks for checking in. We've pulled this branch into our private repo and are currently discussing it. We are mostly in favor of the change as it makes the point at the which the agent is started better defined. There is a small amount of concern about changing things around agent startup, especially if there are users relying on some of the idiosyncrasies around how it is currently initialized. That's mainly the source of our discussions for now.

We'll keep you updated on the status. Also, I wanted to give a brief run down of how we handle public PRs as it is likely a slightly different workflow from other projects you may have contributed to. We do not merge PRs directly to the public repo. We first pull them into our private repo, discuss the changes, and if we accept them, we'll merge it into our development branch on our private repo. At that point we would put an `accepted` label on the public PR. Your changes would make it to the public repo when we push out our next release. 

Thanks for your submission and we'll be in touch if we need anything.

mwear:
Hi @tonyta. Thanks for the contribution and the patience. This will be going into our next release. While we don't have an exact date for you, I'd expect to see it in the next couple of weeks. Thank you!

mwear:
Hi @tonyta. I wanted to let you know that your changes have been integrated into version 4.8.0 of the agent, which was released this morning. Thank you for your contribution!

tonyta:
Thanks @mwear! 💛💚💙💜❤️

PR #280, opened 2018-02-23
Remove type attribute from script tags

Spharian:
`type="text/javascript"` is optional for the `<script>` tag.

The [W3C Validator](https://validator.w3.org) reports the following results:
```
Warning: The `type` attribute is unnecessary for JavaScript resources.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,9 +130,6 @@ If you can't find what you're looking for there, reach out to us on our [support
 site](http://support.newrelic.com/) or our [community forum](http://forum.newrelic.com)
 and we'll be happy to help you.
 
-Also available is community support on IRC: we generally use #newrelic
-on irc.freenode.net
-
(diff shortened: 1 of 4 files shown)


undees:
Hi, @Spharian. Thank you for this pull request! We're reviewing it for possible inclusion in an upcoming agent release. We'll post an update here as soon as we know which release we'll be targeting.

cbhp:
@undees, I also please need this PR for being fully HTML5 compliant.
Using `<script type="text/javascript">` is deprecated. Using `<script>` is enough.

Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-type

rachelleahklein:
Hello, @Spharian. We wanted to let you know your change has been integrated into version 5.1.0 of the agent, which was released today. Thank you again for contributing!

PR #281, opened 2018-02-28
Add webpacker:compile to blacklisted tasks

claudiob:
Webpacker ships with Rails 5 and the task of compiling assets with webpack is pretty common.
Just like `assets:precompile` is a member of `AUTOSTART_BLACKLISTED_RAKE_TASKS`,
I suggest that `webpacker:compile` should be as well


undees:
Hi, @claudiob.

Thank you for preparing this PR! We're taking a look to see if we can fit this into a future release.

rachelleahklein:
Hi, @claudiob. We wanted to let you know your change has been integrated into version 5.1.0 of the agent, which was released today. Thank you again for contributing!

PR #283, opened 2018-04-27
Add frozen_string_literal pragma

oniofchaos:
I was not able to add to the following 4 files due to test failures I could not figure out
```
lib/new_relic/agent/transaction.rb
test/new_relic/agent/audit_logger_test.rb
test/new_relic/agent/database_test.rb
test/new_relic/agent/error_trace_aggregator_test.rb
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'rubygems'
 require 'rake/testtask'
 require 'yard'
--- a/init.rb
+++ b/init.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # encoding: utf-8
 # This file is distributed under New Relic's license terms.
 # See https://github.com/newrelic/rpm/blob/master/LICENSE for complete details.
--- a/install.rb
+++ b/install.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # encoding: utf-8
 # This file is distributed under New Relic's license terms.
 # See https://github.com/newrelic/rpm/blob/master/LICENSE for complete details.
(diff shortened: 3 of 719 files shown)


oniofchaos:
Can anyone offer assistance on debuggging the test failures. Looks like it's either `
No output has been received in the last 30m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated` or it's ordering problems on a couple hashes.

undees:
Hi, @oniofchaos! Thank you for this contribution! We've spent some time today investigating flipping on the frozen string literals switch across the entire agent. A lot of our files will behave the same way with the setting on (and in fact we freeze a lot of string literals ourselves in the agent code).

Unfortunately, there are several places (including, as you point out, in `transaction.rb` but also in `obfuscator.rb` and others) where we're going to have to make some bigger changes before the code is frozen-string-ready. It will be a while before we've got the bandwidth to resume the investigation and implement the needed changes.

I'm going to close the PR for now, but I hope we can eventually revisit this isssue, especially considering that frozen strings are likely to be the default in Ruby 3. I want to reiterate our thanks for your starting the discussion of this upcoming change in Ruby.

oniofchaos:
Any thoughts about doing it in smaller chunks, ignoring the sections that you mentioned that need more love?

undees:
@oniofchaos We talked a little about different approaches—like taking on a smaller group of files at a time. We've likely attained most of the performance benefit of frozen strings by explicitly freezing literals explicitly throughout the code base. The frozen-by-default situation will be important to address before Ruby 3.0 comes out, but we don't quite have the bandwidth to do the investigation just yet.

PR #284, opened 2018-07-13
Enable users to customize attributes sent to transaction tracer.

nandosousafr:
## The problem

There are some cases that the size of the job argument is a big data structure.

Example:

Let's say that your worker process 1000 mailing events per execution.

```ruby
class MailingEvent::Processor
  include Sidekiq::Worker

  def perform(events)
    ....
  end
end
```
Now, if you want to analyze this worker with Transaction Tracer the page freezes trying to load all attributes.

 ## Solution

This PR enables users to customize these attributes. example:

```ruby
class MailingEvent::ProcessorWorker
  include Sidekiq::Worker

  def perform(events)
    ....
  end

  def newrelic_attibutes(args)
    {
       number_of_items: args.first.count
    }
  end
end
```

### DIFF
--- a/lib/new_relic/agent/instrumentation/sidekiq.rb
+++ b/lib/new_relic/agent/instrumentation/sidekiq.rb
@@ -26,8 +26,14 @@ DependencyDetection.defer do
           self.class.default_trace_args(msg)
         end
 
+        attributes = if worker.respond_to?(:newrelic_attributes)
+                       worker.newrelic_attributes(msg['args'])
+                     else
+                       msg['args']
+                     end
+
         perform_action_with_newrelic_trace(trace_args) do
-          NewRelic::Agent::Transaction.merge_untrusted_agent_attributes(msg['args'], :'job.sidekiq.args',
+          NewRelic::Agent::Transaction.merge_untrusted_agent_attributes(attributes, :'job.sidekiq.args',
             NewRelic::Agent::AttributeFilter::DST_NONE)
 
           yield


undees:
Hi, Nando. Thank you for this PR! I'll take a look at it with the Ruby Agent team.

henrich-m:
Hi @undees, any news in this topic ?

undees:
What a coincidence! We were just talking about this PR as a team yesterday. 🙂  We think there may be ramifications to storing this information in the agent attributes—we've scheduled time to look at the code on Friday as a team. We can update you here once we have more information.

undees:
Hello, friends. We've looked through this code as a team, and there are some non-obvious consequences to the way agent attributes are ferried through our system. This particular bucket of keys and values isn't the best-suited for arbitrary user data. Sadly, I'm going to need to close this pull request.

## Alternatives

The good news is that, for the use case you describe above, there is a way to keep your transaction attributes manageable. The `attributes.include` and `attributes.exclude` configuration options will let you omit specific Sidekiq job arguments from the transaction attributes.

For example, if you have a Sidekiq job where the second argument may contain a lot of data:

```ruby
class VeryGoodWorker
  include Sidekiq::Worker

  def perform(small_data, big_data)
    # ...
  end
end
```

...you can keep the `small_data` argument on the Transactions page but leave out the `big_data` argument by adding the following top-level entries to your `config/newrelic.yml`:

```yaml
attributes.include: job.sidekiq.args.*
attributes.exclude: [job.sidekiq.args.1]
```

The arguments nest arbitrarily deeply (e.g.,`job.sidekiq.args.1.0` would match the first item in the `big_data` array above) and support wildcards. Sometimes the easiest way to figure out the Sidekiq key for an argument is to run a query like the following one in Insights:

```sql
SELECT keyset() FROM Transaction WHERE appName = 'Your App Name'
```

...and look for entries beginning with `job.sidekiq`:

insights__home

### Custom Attributes

If skipping a job argument outright is too coarse-grained, custom attributes will let you specify the exact data you want to include. For the use case you pointed out of recording the size of a collection, you could do so with a line like the following in your job's `perform` method:

```ruby
::NewRelic::Agent.add_custom_attributes(number_of_events: big_data.count)
```

This value would then show up both in Insights and on the Transactions page.

Thank you again for taking the time to write up your proposal. I hope one of the approaches here will be a quicker path to your goal.

PR #285, opened 2018-08-23
Use Delayed::Job#display_name when available

printercu:
Hi!

This method is available already for 9 years: https://github.com/collectiveidea/delayed_job/blame/master/lib/delayed/performable_method.rb#L17

I can remove fallback to custom implementation if it works for you.

Actually I want this changes, because we have some patches for delayed job, which modify `display_name` as well.

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -761,6 +761,13 @@ module NewRelic
           :allowed_from_server => false,
           :description => 'If <code>true</code>, disables the use of GC::Profiler to measure time spent in garbage collection'
         },
+        :'delayed_job.use_display_name' => {
+          :default => false,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => 'Use Delayed::Job#display_name for naming transactions.'
+        },
         :'sidekiq.capture_params' => {
           :default => false,
           :public => true,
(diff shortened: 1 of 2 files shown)


printercu:
This may be separate thread, howeret it's related: 
WDYT about having `newrelic_trace_args` and company similar to Sidekiq instrumentation: https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/instrumentation/sidekiq.rb#L47 ?

I can add commits for this or prepare separate PR.

undees:
Hi, @printercu. Thank you for this proposal!

We like the idea of having the display name available for use in the metric. Our only concern is the possibility of metric names changing unexpectedly for customers. Would you be willing to gate this behavior behind a configuration setting, as in:

```
if NewRelic::Agent.config[:'delayed_job.use_display_name'] && payload_object.respond_to?(:display_name)
```

(This would require an addition to `default_source.rb` to add this new setting.)

Would you feel comfortable incorporating that extra check into your PR?

We haven't closely looked at the question of mimicking `newrelic_trace_args`; I like your idea of opening a separate PR for that propsal.

printercu:
Hi! Please check the last commit, if I got everything right with option definition.

I've also removed `respond_to?` condition, because new option is disabled by default, and I assume that users will enable it if they have supported version of dj.

undees:
Hi @printercu. Sorry for the long wait since our initial investigation into this PR.

We've made some changes to how agent attributes are used in the agent code, and we're working on bigger changes to the tracing API that may affect what you want to do here. So, we're closing this PR and #287 for now in the hopes that we can find a way to get the result you want here once we're done.

Thanks for your patience!

PR #287, opened 2018-09-14
Removed extra nesting level in delayed_job_instrumentation.rb

printercu:
Following #285 

Before adding `newrelic_trace_args` for Delayed::Job, I propose to remove extra nesting level to keep code readable in next pr.

### DIFF
--- a/lib/new_relic/agent/instrumentation/delayed_job_instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/delayed_job_instrumentation.rb
@@ -83,11 +83,21 @@ DependencyDetection.defer do
     defined?(::Delayed) && defined?(::Delayed::Worker) && !NewRelic::Agent.config[:disable_dj]
   end
 
+  depends_on do
+    # Because init_plugin runs dependency detection again.
+    !Delayed::Worker.method_defined?(:initialize_with_new_relic)
+  end
+
   executes do
     ::NewRelic::Agent.logger.info 'Installing DelayedJob instrumentation [part 1/2]'
   end
 
   executes do
+    unless defined?(::Delayed::Job) && ::Delayed::Job.method_defined?(:invoke_job)
+      NewRelic::Agent.logger.warn("Did not find a Delayed::Job class responding to invoke_job, aborting DJ instrumentation")
+      next
+    end
+
     Delayed::Worker.class_eval do
       def initialize_with_new_relic(*args)
         initialize_without_new_relic(*args)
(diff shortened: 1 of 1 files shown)


lib/new_relic/agent/instrumentation/delayed_job_instrumentation.rb
```diff
@@ -83,11 +83,21 @@ def method_name(payload_object)
...
+
   executes do
     ::NewRelic::Agent.logger.info 'Installing DelayedJob instrumentation [part 1/2]'
   end
 
   executes do
+    unless defined?(::Delayed::Job) && ::Delayed::Job.method_defined?(:invoke_job)
+      NewRelic::Agent.logger.warn("Did not find a Delayed::Job class responding to invoke_job, aborting DJ instrumentation")
+      next
+    end
```
printercu: I've not seen such warnings in other plugins. Should I just move it to `depends_on` block without calling a logger?

undees:
Hi, @printercu.

Thank you for this cleanup proposal! The `newrelic_trace_args` is mainly used for naming transactions—it might not be suitable as a general-purpose data bucket. Could we chat for a bit about your use case?

I'll be back in the office on Tuesday. I work in the US Pacific time zone (GMT-7) but can chat outside work hours as long as it's not the middle of the night 😁. Is there a time that will work for you, and do you have a preferred chat technology?

printercu:
Hi! Sure, we can chat.

I've just prepared [pr](https://github.com/printercu/rpm/pull/1) in local repo with supposed next changes, to make it more clear. In this way every payload class is responsible for its transaction naming, so it's more flexible rather than current Naming module which supports only PerformableMethod.

Along with naming transactions, `newrelic_trace_args` is useful for setting custom parameters with `:params`. While we don't have custom jobs we want to rename transactions for, we would like to track `object.id` and some other data for each `PerformableMethod` jobs. For now it's possible with `add_custom_attributes`, but it requires patching `invoke_job_without_new_relic`, which is not obvious. Patching`invoke_job` will not work because it's outside of newrelic transaction. `newrelic_trace_args` looks better, because it has more suitable naming and is separated from performing job logic.

WDYT?

undees:
Hi @printercu. Sorry for the long wait since our initial investigation into this PR.

We've made some changes to how agent attributes are used in the agent code, and we're working on bigger changes to the tracing API that may affect what you want to do here. So, we're closing this PR and #285 for now in the hopes that we can find a way to get the result you want here once we're done.

Thanks for your patience!

PR #289, opened 2018-11-06
Truncate backtraces over 1MB

PatrickTulskie:
The current backtrace truncate logic currently removes important frames in the backtrace on larger applications (multiple middlewares, callbacks, gems, etc). This bit me pretty hard when I was trying to debug an app that was on fire but I was missing important frames from the backtrace in the error logs in APM.

I spoke with support about this and they informed me that I could monkey patch the gem so that it supports larger stack traces so we do not lose important information. They told me that stack traces over 1MB will be dropped by the APM API.

This PR removes the backtrace frame cap and limits overall size of the backtrace to 1MB. Like the previous implementation it removes the middle of the backtrace since that's where you're most likely to see an infinite loop or something anyway.

I updated the tests and I think I got everything working but if you need more to make this happen, lemme know and I'll get it done.

### DIFF
--- a/lib/new_relic/agent/error_collector.rb
+++ b/lib/new_relic/agent/error_collector.rb
@@ -13,9 +13,9 @@ module NewRelic
       # made configurable in the future. This is a tradeoff between
       # memory and data retention
       MAX_ERROR_QUEUE_LENGTH = 20
-      # Maximum number of frames in backtraces. May be made configurable
-      # in the future.
-      MAX_BACKTRACE_FRAMES = 50
+      # Traces need to be under 1 MB to be accepted by the APM API
+      # Less than a real MB, but gives us some buffer for truncate notice
+      MAX_BACKTRACE_SIZE = 1000000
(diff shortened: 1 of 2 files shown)


rachelleahklein:
Thanks for submitting this pull request, @PatrickTulskie! We are evaluating it for possible inclusion into a future version of the Ruby agent.

PatrickTulskie:
@mwear Ah okay I was under the impression that the stack trace itself was limited to 1 MB (which seems huge) from my quick convo with support, but it makes sense that it's for all of them during a harvest cycle post compression. I guess depending on compression, repetitive stack traces should shrink considerably.

Anyhow, yeah I can totally knock out the code to make the stack trace lines configurable with a default of 50. I'll reference this PR and open up a fresh one next week with the changes.

mwear:
@PatrickTulskie: Thank you for your understanding and willingness to work on this issue!

PatrickTulskie:
Closing this out. New approach is in #291

PR #290, opened 2018-11-25
Add Rails 6 support

jacobbednarz:
Updates the agent configuration to detect rails 6 and allow running on
the alpha rails version.

### DIFF
--- a/lib/new_relic/control/frameworks/rails6.rb
+++ b/lib/new_relic/control/frameworks/rails6.rb
@@ -1,0 +1,14 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/rpm/blob/master/LICENSE for complete details.
+
+require 'new_relic/control/frameworks/rails4'
+
+module NewRelic
+  class Control
+    module Frameworks
+      class Rails6 < NewRelic::Control::Frameworks::Rails4
+      end
+    end
+  end
+end
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -95,6 +95,8 @@ module NewRelic
                 :rails4
               when 5
                 :rails5
+              when 6
+                :rails6
               else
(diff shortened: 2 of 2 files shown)


rachelleahklein:
Thank you for your submission, @jacobbednarz! We are reviewing it for possible inclusion in a future version of the Ruby agent.

lib/new_relic/control/frameworks/rails6.rb
```diff
@@ -0,0 +1,14 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/rpm/blob/master/LICENSE for complete details.
+
+require 'new_relic/control/frameworks/rails4'
+
+module NewRelic
+  class Control
+    module Frameworks
+      class Rails6 < NewRelic::Control::Frameworks::Rails4
```
radar: Should this perhaps inherit from Rails5?

radar:
@rachelleahklein Hi! I was wondering what the status of this is? Rails 6 is due to be released in April and it would be good to have a version of `newrelic_rpm` that works with the beta releases that are already out there.

mwear:
@jacobbednarz Thank you for your contribution, this was included in our 6.0 release that went out this morning. We don't merge PRs directly from the public repo. Rather, we pull them into our private repo and integrate the changes there. We end up closing the public PRs, but your commits and attribution are retained.

@radar You are correct that Rails6 should inherit from Rails5. We'll get that resolved in the 6.1 release. This PR doesn't add full Rails 6 support. It only makes the agent aware of Rails 6 so it doesn't log an error message. While Rails 6 support is on our roadmap, the work isn't scheduled until closer to the release date. We can look at what it will take to get preliminary support out faster given that you and @asaaki have expressed interest.

lib/new_relic/control/frameworks/rails6.rb
jacobbednarz: Definitely, apologies on this. It should be I've copy pasta'd wrong.

PR #291, opened 2018-12-03
Make max_backtrace_frames configurable

PatrickTulskie:
This makes max_backtrace_frames configurable via YAML and other agent config stuff instead of hardcoding it to 50 frames.

This is in response to feedback I got on #289.

Notes:

1. Default number of frames is still 50
2. I left the ability to override the number of frames in the truncate method's parameters. Since this is now configurable, I was going to change the tests over to just use the config but I wasn't sure if there was something else that was relying on that.

If you'd like me to, I can remove the parameter override and just update the tests to use `with_config` to make sure the coverage is the same.

Please let me know if there are additional changes or tests needed to get this in.

### DIFF
--- a/config.dot
+++ b/config.dot
@@ -18,6 +18,7 @@ digraph AgentEnabled {
   "[error_collector.capture_source]"
   "[error_collector.enabled]"
   "[error_collector.ignore_errors]"
+  "[error_collector.max_backtrace_frames]"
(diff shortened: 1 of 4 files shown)


undees:
Thank you, @PatrickTulskie! We're looking at this change for possible inclusion in a future Ruby Agent version.

PR #293, opened 2018-12-28
Add grape 1.2+ support

ohbarye:
Hello maintainers. 

I hope you would have a look at this when you have time because our team is facing a problem that we can't see detailed metrics in the New Relic web. 🙏 

## Problem

In short, `newrelic_rpm` does not work with `grape` v1.2.0 or posterior. I could see only metrics for `Foo#call` methods in the dashboard. It seems that the agent fails to track environments.

image

**My environment**

- newrelic_rpm 5.6.0.349
- grape 1.2.2
   - 📝 When I tried to downgrade `grape` to `1.1.0`, it worked well.
- rails 4.2.9 (Seems nothing relevant to this problem)
- Ruby 2.5.0 (Seems nothing relevant to this problem)

## Cause

`grape` 1.2.0 brought some breaking changes and it affects `newrelic_rpm`. Those changes are the two below.

1. The class `Grape::API` no longer refers to an API instance, rather, what used to be `Grape::API` is `Grape::API::Instance`
2. How to get the class name is changed. `name` => `base.name`

ref: https://github.com/ruby-grape/grape/blob/c20a73ac1e3f3ba1082005ed61bf69452373ba87/UPGRADING.md#upgrading-to--120

## Solution

I just made a patch around `::NewRelic::Agent::Instrumentation::GrapeInstrumentation` to follow the changes.

## Testing

At first, I confirmed `grape` v1.2.0 breaks the latest `newrelic_rpm`:

1. I just tweaked `test/multiverse/suites/grape/Envfile` so that it executes tests against `grape` v1.2.0.
2. I run `bundle exec rake test:multiverse[grape]`.
   - Tests against `1.2.0` failed. (There were too long logs that I wasn't able to paste them here.)

Next, I made a change to `lib/new_relic/agent/instrumentation/grape.rb`, and then I run `bundle exec rake test:multiverse[grape]` again. The tests passed now!

At last, I tested my forked version so that I could confirm it would wholly work well.

1. I added `gem 'newrelic_rpm', git: 'https://github.com/ohbarye/rpm.git', branch: 'grape-1.2'` into my Gemfile, then run `bundle install`.
   - Of course, I kept other library versions as is. (grape 1.2.0 etc.)
2. I deployed the Rails application and manipulated so that it sends metrics to NewRelic.
   - Finally, I was able to see the detailed metrics like below. That's what I wanted.

image

### DIFF
--- a/lib/new_relic/agent/instrumentation/grape.rb
+++ b/lib/new_relic/agent/instrumentation/grape.rb
@@ -114,7 +114,15 @@ DependencyDetection.defer do
       ::NewRelic::Agent::Instrumentation::GrapeInstrumentation.send :alias_method, :name_for_transaction, :name_for_transaction_deprecated
     end
 
-    ::Grape::API.class_eval do
+    # Since 1.2.0, the class `Grape::API` no longer refers to an API instance, rather, what used to be `Grape::API` is `Grape::API::Instance`
+    # https://github.com/ruby-grape/grape/blob/c20a73ac1e3f3ba1082005ed61bf69452373ba87/UPGRADING.md#upgrading-to--120
+    grape_api_class = if defined?(Grape::VERSION) && Gem::Version.new(::Grape::VERSION) >= Gem::Version.new("1.2.0")
+                        ::Grape::API::Instance
+                      else
+                        ::Grape::API
+                      end
+
+    grape_api_class.class_eval do
       def call_with_new_relic(env)
         begin
           response = call_without_new_relic(env)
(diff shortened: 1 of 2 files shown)


undees:
Thank you so much for this contribution, @ohbarye! We are evaluating it for possible inclusion in a future Ruby agent version.

undees:
We've incorporated this fix, with credit, into the just-released version 5.7 of the agent. Thanks again! Closing this PR as complete.

ohbarye:
@undees 
I appreciate you!

PR #294, opened 2019-02-05
Fix bunny support for when pop does not return a message

soupmatt:
Currently, if pop a message results in no message (because the queue is 
empty), an error is logged because of a NoMethodError on nil:NilClass. 
This fixes it by skipping recording the segment if no message was 
returned.

I'm not sure if the skip is the correct solution, but it makes it so the log file isn't filled with errors when you do a lot of pops against empty queues.


rachelleahklein:
Thanks for your contribution, @soupmatt! We'll be evaluating it for possible inclusion in a future Ruby agent version.

soupmatt:
I added another commit to handle another NoMethodError we were getting when using Queue.pop.

rachelleahklein:
We've incorporated your fix, with credit, into the just-released version 6.1 of the agent. Once again, thank you! Closing this PR as complete.

PR #297, opened 2019-04-16
Convert value of Socket.gethostname to UTF-8

rafaelpetry:
I was getting a lot of warnings locally that looked like this:
`log writing failed. "\xE2" from ASCII-8BIT to UTF-8`

After much digging, I traced it down to the agent logger, which uses the computer's hostname in the default log formatter. Turns out `Socket.gethostname` does not respect the default character encoding, and I get ASCII-8BIT every time, even though any new string is UTF-8.

```
2.6.2 :004 > Socket.gethostname.encoding
 => #<Encoding:ASCII-8BIT>
2.6.2 :005 > "".encoding
 => #<Encoding:UTF-8>
```

My computer's name has a special character (curved apostrophe in `Rafael-Petry’s-MacBook-Pro`), which meant all writes to `newrelic_agent.log` failed with the warning mentioned earlier. This change forces the encoding to UTF-8, which fixes the problem.

For reference, prior to the change, the test I added failed with the following message:
```
  1) Failure:
NewRelic::Agent::HostnameTest#test_get_returns_socket_hostname_converted_to_utf8 [test/new_relic/agent/hostname_test.rb:26]:
--- expected
+++ actual
@@ -1 +1 @@
-"Elronds’s-Computer"
+"Elronds\xE2\x80\x99s-Computer"
```

### DIFF
--- a/lib/new_relic/agent/hostname.rb
+++ b/lib/new_relic/agent/hostname.rb
@@ -13,7 +13,7 @@ module NewRelic
           dyno_name = "#{matching_prefix}.*" if matching_prefix
           dyno_name
         else
-          Socket.gethostname
+          Socket.gethostname.force_encoding(Encoding::UTF_8)
         end
       end
 
--- a/test/new_relic/agent/hostname_test.rb
+++ b/test/new_relic/agent/hostname_test.rb
@@ -21,6 +21,11 @@ module NewRelic
         assert_equal 'Rivendell', NewRelic::Agent::Hostname.get
       end
 
+      def test_get_returns_socket_hostname_converted_to_utf8
+        Socket.stubs(:gethostname).returns('Elronds’s-Computer'.force_encoding(Encoding::ASCII_8BIT))
+        assert_equal 'Elronds’s-Computer', NewRelic::Agent::Hostname.get
+      end
+
       def test_get_uses_dyno_name_if_dyno_env_set_and_dyno_names_enabled
         with_dyno_name('Imladris', :'heroku.use_dyno_names' => true) do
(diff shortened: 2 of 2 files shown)


rafaelpetry:
The JRuby builds are failing on Travis for reasons unrelated to the code change. Looks like even though in `.travis.yml` we install a specific version of bundler (1.7.3), on JRuby builds bundler 2.0.1 is already installed (maybe in the global gemset), causing `bundle install` to fail.

MRI example (2.6.1 norails):
```
$ bundle --version
Bundler version 1.17.3
```

JRuby example (9.2.6.0 norails):
```
$ bundle --version
Bundler version 2.0.1
$ bundle install
Fetching gem metadata from https://rubygems.org/..........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    bundler (< 2.0) java
```

Unsure what the best solution here is. [The docs](https://docs.travis-ci.com/user/languages/ruby/#bundler-20) suggest adding this command to the config:
`gem uninstall -v '>= 2' -i $(rvm gemdir)@global -ax bundler || true`
Should I add that here or try a separate PR?

justinfoote:
Thank you so much for this contribution, @rafaelpetry! We are evaluating it for possible inclusion in a future Ruby agent version.

justinfoote:
Thank you @rafaelpetry.  We've incorporated this fix, with credit, into our development branch.  It will be released with version 6.3 of the agent.  Thanks again!  
I'm closing this PR as complete.

PR #298, opened 2019-05-31
Fix ActiveRecord 4 monkey patch to align with the original code

kovyrin:
Not sure what happened in https://github.com/newrelic/rpm/commit/04b93b3f72fa16fefe14839b999fffdd3f50b8ee, but the original change introducing the performance improvements in AR instrumentation ended up being broken on AR 4 (you lose AR exception messages when the agent is active).

The original version of this monkeypatched line: https://github.com/rails/rails/blob/4-1-stable/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L380

### DIFF
--- a/lib/new_relic/agent/instrumentation/active_record_notifications.rb
+++ b/lib/new_relic/agent/instrumentation/active_record_notifications.rb
@@ -13,6 +13,8 @@ module NewRelic
   module Agent
     module Instrumentation
       module ActiveRecordNotifications
+        SQL_ACTIVE_RECORD = 'sql.active_record'.freeze
+
         module BaseExtensions41
           def self.included(base)
             base.class_eval do
@@ -24,7 +26,7 @@ module NewRelic
           # https://github.com/rails/rails/blob/4-1-stable/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L371
           def log_with_performance_improvement(sql, name = "SQL", binds = [], statement_name = nil)
             @instrumenter.instrument(
-              "sql.active_record",
+              SQL_ACTIVE_RECORD,
               :sql            => sql,
               :name           => name,
               :connection_id  => object_id,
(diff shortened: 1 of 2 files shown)


justinfoote:
Thank you so much for this contribution, @kovyrin.  

I have to apologize for our lack of sufficient test coverage in this area.  When we merged the rails 4 and 5 instrumentation, we changed this line from `translate_exception_class` to `translate_exception`, because `translate_exception_class` doesn't exist in ActiveRecord 4.0.  This included the backport in version 4.0, but at the expense of correct error handling in versions 4.1 and 4.2.  We broke error handling for ActiveRecord 4.1 and 4.2, and our tests didn't even notify us!

I would love to include this change in our next release, but we must also exclude this backport from ActiveRecord < 4.1, and some tests to verify that we haven't broken exception handling for ActiveRecord in general.

kovyrin:
Let me know if you want me to add a separate monkeypatch for 4.0.

justinfoote:
@kovyrin, that would be great!  Thanks!

kovyrin:
Ok, I think this is ready for a review. CI is green for everything but jruby, but those are broken for reasons unrelated to the PR.

Let me know if this implementation makes sense to you or if you'd like to have two independent versions of the monkeypatch method for different 4.x versions of AR.

justinfoote:
Thanks again!  This is looking really good.  

We generally try to limit version-checking or branching in general in the very hot codepaths, and the log method of the ActiveRecord::Base is a _very_ hot codepath.  But, since this branching only happens in the rescue block, it's likely not a problem.  

I'm going to run this through a bit more testing, then we'll likely merge it into our development branch.  I'll update here when that's happened.

kovyrin:
Good point about the hot path. That's basically why I decided to add the frozen string constant in there to make sure it does not generate garbage. Could absolutely change it to not have the conditional if you think it makes sense to do that. Just let me know.

justinfoote:
@kovyrin We've integrated this PR, with credit, into our development branch.  We will include it in our next agent release.  Thanks again!
I'm closing this PR as complete.

PR #300, opened 2019-06-20
RUBY-2226 Remove all versions of bundler before setting up travis jobs

justinfoote:
Because bundler versions >= 2.0 are problematic for jruby, we must remove them all, then install the correct version.  The existing command to uninstall bundler has been failing on newer travis workers.
running a simple `gem uninstall` without the `rvm @global do...` solves the problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ services:
 
 language: ruby
 
+dist: precise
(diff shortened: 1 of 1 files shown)


kenichi:
i feel your pain

PR #312, opened 2020-01-23
Fixup Content-Length setting in rack/browser_monitoring

ojab:
We have an issue with sidekiq web-interface HTML being cut because of wrong `Content-Length` header. Looks like the test is already here, but it's failing, fix it.

### DIFF
--- a/lib/new_relic/rack/browser_monitoring.rb
+++ b/lib/new_relic/rack/browser_monitoring.rb
@@ -21,6 +21,7 @@ module NewRelic::Rack
 
     CONTENT_TYPE        = 'Content-Type'.freeze
     CONTENT_DISPOSITION = 'Content-Disposition'.freeze
+    CONTENT_LENGTH      = 'Content-Length'.freeze
     ATTACHMENT          = 'attachment'.freeze
     TEXT_HTML           = 'text/html'.freeze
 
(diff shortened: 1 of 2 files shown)


ojab:
Also made tests to run if rack-test if available, since they're not actually running on CI right now.

mwlang:
@ojab Thank you for your PR.  I have pulled your code and starting to go through it and it looks solid on quick glance.  

I can see it's largely concerned with setting the Context Length and I was wondering if you could elaborate on what you meant by the HTML being cut before your fix.

ojab:
Basically it was HTML with `Content-Length` length and since New Relic js is injected to the `<head/>` (in the start of the HTML) — stuff at the end of the file was missing.

mwlang:
@ojab Great find and clear fix to the problem!  Thank you for your contribution.  I have marshalled this through internal testing, which it passed.  If it passes next round of internal review, it should appear in the next release of the Ruby agent.

dentarg:
@mwlang did it pass the next round of review? do you know when you will release the next version of the Ruby agent? 

Thanks!

mwlang:
@dentarg Your fix has been accepted and is merged so it will appear in the next release of the agent.  We are expecting to release later this month.  The root cause was also found with Rack's change in how they're handling `Content-Length` after 2.0.8 release.

PR #322, opened 2020-02-29
Calculate message size in bytes for unicode characters

thaim:
Content-Length requires to set message size in bytes.
There is an issue to return number of characters which isn't equal to message size when message contains unicode characters. Fix to calculate size in bytes.

### DIFF
--- a/lib/new_relic/rack/browser_monitoring.rb
+++ b/lib/new_relic/rack/browser_monitoring.rb
@@ -37,7 +37,7 @@ module NewRelic::Rack
       if (js_to_inject != "") && should_instrument?(env, status, headers)
         response_string = autoinstrument_source(response, headers, js_to_inject)
         if headers.key?(CONTENT_LENGTH)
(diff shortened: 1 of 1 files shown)


rachelleahklein:
Hello, @thaim!

Thanks so much for catching this needed improvement and submitting a pull request. We're currently evaluating your PR and expect to include it in the next released agent version.

dentarg:
I think it would be good if the test are extended to cover this issue. Here's an example: https://github.com/dentarg/rpm/commit/6cebcebe1e0a91f1de9a32c4d31fcc6658bbf161#diff-a9947902f661b107f21c50e767e561ef

rachelleahklein:
This change has been included in version 6.10.0 of the Ruby agent, released today! Thank you again, @thaim!

dentarg:
> I think it would be good if the test are extended to cover this issue

Such a test was added in https://github.com/newrelic/rpm/commit/b1a86f8c561592e20dca4be1dfbd4a4cbb93a745

PR #326, opened 2020-03-12
add distributed tracing to sidekiq instrumentation

andreaseger:
I wanted to test and see how easily distributed tracing support for sidekiq jobs could be added. At the moment it's not doing anything besides storing the trace payload as additional job attribute on the "client" and reading/accepting it in the "server". IMHO this would already be good enough.

Please let me know if or how you'd like to proceed with something like this.

### DIFF
--- a/lib/new_relic/agent/instrumentation/sidekiq.rb
+++ b/lib/new_relic/agent/instrumentation/sidekiq.rb
@@ -65,7 +65,7 @@ DependencyDetection.defer do
         }
       rescue => e
         NewRelic::Agent.logger.error("Failure during deserializing YAML for Sidekiq::Extensions::DelayedClass", e)
-        NewRelic::SidekiqInstrumentation.default_trace_args(msg)
+        NewRelic::SidekiqInstrumentation::Server.default_trace_args(msg)
       end
     end
 
--- a/lib/new_relic/agent/instrumentation/sidekiq.rb
+++ b/lib/new_relic/agent/instrumentation/sidekiq.rb
@@ -26,13 +26,13 @@ DependencyDetection.defer do
           else
             self.class.default_trace_args(msg)
           end
-          trace_info = msg.delete("newrelic")
+          trace_headers = msg.delete(NewRelic::NEWRELIC_KEY)
           
           perform_action_with_newrelic_trace(trace_args) do
             NewRelic::Agent::Transaction.merge_untrusted_agent_attributes(msg['args'], :'job.sidekiq.args',
               NewRelic::Agent::AttributeFilter::DST_NONE)
-              
-            ::NewRelic::Agent::DistributedTracing.accept_distributed_trace_payload(trace_info) if trace_info
+
+            ::NewRelic::Agent::DistributedTracing::accept_distributed_trace_headers(trace_headers, "Other")
             yield
           end
         end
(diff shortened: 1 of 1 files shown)


tannalynn:
Hello @andreaseger,
Thank you for your submission. I noticed this is still a draft, is that still accurate?
As soon as this pr is ready to be reviewed, please let me know and we’ll take a look!
Thanks!

andreaseger:
@tannalynn I created it as draft to see if there is interest in such a PR before doing any more.
This code works and makes distributed tracing work for sidekiq jobs and doesn't break any existing behavior. But I'm not sure if you require tests for something like this. Also I can't say I have time at the moment to dig deeper into the huge codebase here to be even able to even create test.

I marked it now "ready for review".

tannalynn:
Okay sounds good @andreaseger. We will go ahead and review this. We expect to include this with the next release of the agent. We will let you know. Thanks!

mwlang:
@andreaseger This is really good.  All-in-all, the organizational changes are on point to allow both a client and server middleware for the Sidekiq instrumentation and I think distributed tracing in Sidekiq jobs is a great feature enhancement.

Would you be willing to further improve this implementation to use the latest Distributed Tracing API's?  As currently written, the deprecated API's are utilized.

As of 6.9.0 version of the Agent, the new API calls would be:
`insert_distributed_trace_headers` and `accept_distributed_trace_headers` also in the same namespace as the deprecated API calls at 'NewRelic::Agent::DistributedTracing'

lib/new_relic/agent/instrumentation/sidekiq.rb
```diff
@@ -14,32 +14,44 @@
...
+          else
+            self.class.default_trace_args(msg)
+          end
+          trace_info = msg.delete("newrelic")
+          
+          perform_action_with_newrelic_trace(trace_args) do
+            NewRelic::Agent::Transaction.merge_untrusted_agent_attributes(msg['args'], :'job.sidekiq.args',
+              NewRelic::Agent::AttributeFilter::DST_NONE)
+              
+            ::NewRelic::Agent::DistributedTracing.accept_distributed_trace_payload(trace_info) if trace_info
```
mwlang, line 35: This API call is deprecated along with it's companion create_distributed_trace_payload.  Intead, use:
```
::NewRelic::Agent::DistributedTracing::accept_distributed_trace_headers(trace_info) if trace_info
```

lib/new_relic/agent/instrumentation/sidekiq.rb
```diff
@@ -14,32 +14,44 @@
...
-          :class_name => msg['class'],
-          :category => 'OtherTransaction/SidekiqJob'
-        }
+      class Client
+        def call(_worker_class, job, *_)
+          distributed_trace_payload = ::NewRelic::Agent::DistributedTracing.create_distributed_trace_payload
+          distributed_trace_payload = distributed_trace_payload.http_safe if distributed_trace_payload
+          job["newrelic"] = distributed_trace_payload
+          yield
+        end
```
mwlang, line 54: This can be simplified with the new API call like this:
```
      class Client
        def distributed_trace_headers
          headers = {}
          ::NewRelic::Agent::DistributedTracing.insert_distributed_trace_headers headers
          headers
        end

        def call(_worker_class, job, *_)
          if headers = distributed_trace_headers
            job["newrelic"] = headers
          end
          yield
        end 
      end
```

andreaseger:
@mwlang Using the new API now, but I have to say it's less then intuitive. `accept_distributed_trace_headers` by default assume to get something like a rack request where the headers are prefixed with `HTTP_`. This took me quite a while to understand.
I now overwritten this default with `"Other"` which seems to work but I noticed that this could be super easy to break as you pass along the `transport_type` further down but actually always assume header keys with the `HTTP_` prefix. I think this bit should be adapted to prevent this in the future https://github.com/newrelic/rpm/blob/00e60019deab4f407f64998b0145913519e3c274/lib/new_relic/agent/distributed_tracing.rb#L178-L196

tannalynn:
@andreaseger Thank you for updating this! We are reviewing it now and plan to include this in the next release of the agent. We will let you know here when it is released. 

Thank you for the feedback regarding our api. We will look at some ways to make this clearer and more intuitive.

tannalynn:
@andreaseger Thank you for your contribution! We have released the new version of the agent and included these changes.

PR #329, opened 2020-04-03
fix cross_app_payload missing header

nritholtz:
See ticket 396913 for context.


tannalynn:
This issue has been fixed in our most recent version 6.11.0. 
Thank you!

PR #344, opened 2020-06-15
Subtract request buffering from queue time

tlynam:
Follow up from https://github.com/newrelic/rpm/issues/342

- When enabled, this will subtract time spent waiting for the HTTP request body to be received from the queue time.
- Currently only supported for Puma.
- Enabled with config `queue_time_subtract_buffering` set to true.

If we adjust the queue time within `QueueTime`, then it will update queue time, response time, and the apdex calculation.  Please let me know what should be changed.  When we figure out the implementation, then I can update and add tests.  Thank you!

### DIFF
--- a/lib/new_relic/agent/instrumentation/queue_time.rb
+++ b/lib/new_relic/agent/instrumentation/queue_time.rb
@@ -27,7 +27,23 @@ module NewRelic
 
         module_function
 
-        def parse_frontend_timestamp(headers, now=Time.now)
+        def frontend_timestamp(headers, now=Time.now)
+          parsed_timestamp = parse_frontend_timestamp(headers, now)
+
+          # When enabled, this will return an adjusted timestamp
+          # with request buffer time subtracted.  This is the time
+          # spent waiting for the HTTP request body to be received.
+          if parsed_timestamp && NewRelic::Agent.config[:'queue_time_subtract_buffering']
+            # Currently, only Puma is supported
+            request_buffering_in_seconds = headers.fetch('puma.request_body_wait', 0).to_f / 1000
+
+            parsed_timestamp -= request_buffering_in_seconds
+          end
+
+          parsed_timestamp
+        end
+
+        def parse_frontend_timestamp(headers, now)
           earliest = nil
 
           CANDIDATE_HEADERS.each do |header|
--- a/lib/new_relic/agent/instrumentation/action_controller_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/action_controller_subscriber.rb
@@ -91,7 +91,7 @@ module NewRelic
 
         def queue_start(request)
           if request && request.respond_to?(:env)
-            QueueTime.parse_frontend_timestamp(request.env, Time.now)
+            QueueTime.frontend_timestamp(request.env, Time.now)
           end
         end
       end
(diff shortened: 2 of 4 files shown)


tlynam:
Hi @tannalynn, hope you're well.  Please let me know what you think of this approach. Thank you so much!

lib/new_relic/agent/instrumentation/queue_time.rb
```diff
@@ -27,7 +27,23 @@ module QueueTime
...
-        def parse_frontend_timestamp(headers, now=Time.now)
+        def frontend_timestamp(headers, now=Time.now)
+          parsed_timestamp = parse_frontend_timestamp(headers, now)
+
+          # When enabled, this will return an adjusted timestamp
+          # with request buffer time subtracted.  This is the time
+          # spent waiting for the HTTP request body to be received.
+          if NewRelic::Agent.config[:'queue_time_subtract_buffering']
+            # Currently, only Puma is supported
+            request_buffering_in_seconds = headers.fetch('puma.request_body_wait', 0).to_f / 1000
```
mwlang, line 38: How hard to extend this to other rack providers (webrick, thin, unicorn, etc.)?

tlynam: Let me research this and I'll let you know what I find 👍

tlynam: Hi, @mwlang, hope you're well.  I was hoping other web servers produced a similar header like `puma.request_body_wait` but didn't find anything like this for other commonly used web servers.

Nate Berkopec has a great chapter `Webservers and I/O models` in his book `The Complete Guide to Rails Performance` that covers different web server models.  In it, he describes which web servers offer a model that download larger requests before passing them to a worker to be run application code.

I went through each of these servers and tested a 15kB PUT on a simulated slow 3G connection.

### Webrick
Webrick doesn't buffer large requests.  I also didn't see any passed headers that log the buffering time.

### Thin
Thin can buffer large requests but I didn't see any passed headers that log buffering time.

### Unicorn
I didn't see any passed headers that log buffering time.  It looks like it's recommended to use with nginx to buffer requests.  I'm an nginx beginner but wondering if it's possible set a header when a request has finish buffering so we could calculate here what the buffering time is.

### Phusion Passenger
Phusion Passenger also uses nginx.  I didn't see any headers logging the request buffer time.  Also wondering if it would be possible to set a header in nginx when a request has completed buffering so we could then calculate the buffering time here.

tlynam:
Hi @mwlang, hope you're well.  Did you close this by accident?  Having this could make a big difference for Puma users.  Our queue time, response time, and apdex calculations are impacted so much that we can't use them.  This impacts apps that have create/update payloads a couple kB or more.  We're seeing this with 7kB payloads and the data we took shows it also impacts high-bandwidth clients.

I documented more in https://github.com/newrelic/rpm/issues/342

Would it help if I investigate if we can use another header with nginx so other setups could benefit from this?  Other webservers could also expose the buffering time like Puma does.  I could open up issues on those projects.

I think this issue is underreported because I think most don't realize this is happening and could be resolved.

mwlang:
This shouldn't have been closed.  I'll reopen it!

mwlang:
This looks like a system error of some sort.  The issue still appears as "OPEN" to me despite showing Closed above.  I think this might be because we decided to rename 'master' to 'main' and this PR is for merging to `master`

tlynam:
Oh, thank you that makes perfect sense!  I tried to update this PR's target but wasn't able to.  Please let me know if I should create a new PR.

mwlang:
@tlynam Figured it out!  We moved `master` to `main` and dropped the `master` branch altogether.  This triggered closing all PRs targeting `master`  If you would, please reopen or retarget this PR to `main`  The work you have here and the proposal is solid and we definitely want to collaborate further on this effort.

PR #350, opened 2020-07-05
Subtract request buffering from queue time

tlynam:
Follow up from https://github.com/newrelic/rpm/issues/342

- When enabled, this will subtract time spent waiting for the HTTP request body to be received from the queue time.
- Enabled with config `queue_time_subtract_buffering` set to true.
- Currently only supported for Puma.
- Other common Ruby web servers don't expose buffering time.

If we adjust the queue time within `QueueTime`, then it will update queue time, response time, and the apdex calculation.  Please let me know what should be changed.  When we figure out the implementation, then I can update and add tests.  Thank you!

Here's some info on common Ruby web servers:

### Webrick
Webrick doesn't buffer large requests.  I also didn't see any passed headers that log the buffering time.

### Thin
Thin can buffer large requests but I didn't see any passed headers that log buffering time.

### Unicorn
I didn't see any passed headers that log buffering time.  It looks like it's recommended to use with nginx to buffer requests.  I don't know nginx but wondering if it's possible set a header when a request has finish buffering so we could calculate here what the buffering time is.

### Phusion Passenger
Phusion Passenger also uses nginx.  I didn't see any headers logging the request buffer time.  Also wondering if it would be possible to set a header in nginx when a request has completed buffering so we could then calculate the buffering time here.

### DIFF
--- a/lib/new_relic/agent/instrumentation/middleware_tracing.rb
+++ b/lib/new_relic/agent/instrumentation/middleware_tracing.rb
@@ -39,7 +39,7 @@ module NewRelic
         end
 
         def merge_first_middleware_options(opts, env)
-          opts[:apdex_start_time] = QueueTime.parse_frontend_timestamp(env)
+          opts[:apdex_start_time] = QueueTime.frontend_timestamp(env)
           # this case is for the rare occasion that an app is using Puma::Rack
           # without having ::Rack as a dependency
           opts[:request] = ::Rack::Request.new(env) if defined? ::Rack
--- a/lib/new_relic/agent/instrumentation/action_controller_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/action_controller_subscriber.rb
@@ -91,7 +91,7 @@ module NewRelic
 
         def queue_start(request)
           if request && request.respond_to?(:env)
-            QueueTime.parse_frontend_timestamp(request.env, Time.now)
+            QueueTime.frontend_timestamp(request.env, Time.now)
           end
         end
       end
--- a/lib/new_relic/agent/instrumentation/controller_instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/controller_instrumentation.rb
@@ -462,7 +462,7 @@ module NewRelic
         def detect_queue_start_time(request)
           headers = newrelic_request_headers(request)
 
-          QueueTime.parse_frontend_timestamp(headers) if headers
+          QueueTime.frontend_timestamp(headers) if headers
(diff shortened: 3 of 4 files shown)


tlynam:
Hi @mwlang, hope you're well.  Okay here's the new PR.  I'm researching to see if it's possible to calculate the buffering time using nginx.

I see we can buffer requests with `proxy_request_buffering` but don't see a buffering time exposed.  I also don't know if it's possible to set another header with `proxy_set_header` after the buffering is completed.

mwlang:
@tlynam, I should have some more time to look at this next week -- we're hard at work prepping for the next release of the agent at the moment.

lib/new_relic/agent/instrumentation/middleware_tracing.rb
```diff
@@ -39,7 +39,7 @@ def build_transaction_options(env, first_middleware)
         end
 
         def merge_first_middleware_options(opts, env)
-          opts[:apdex_start_time] = QueueTime.parse_frontend_timestamp(env)
+          opts[:apdex_start_time] = QueueTime.frontend_timestamp(env)
```
mwlang, line 42: Changing the value we record for the apdex_start_time will have the undesired effect of changing the `apdex` scores that get computed.  `apdex` needs to retain visibility of the entire time the user is waiting for page updates.  Let's see if we can find a way to subtract the `puma.request_body_wait`

My first thought here is that we should record a metric for the `puma.request_body_wait` OR record the metric of the value computed here, and then a graph can render this data point in the UI.

tlynam: Thanks @mwlang, that makes a lot of sense.  I'm curious, just for clarification, it looks like the documentation refers to the`apdex` using response time as the metric.  https://docs.newrelic.com/docs/apm/new-relic-apm/apdex/apdex-measure-user-satisfaction
I guess my personal definition of response time wouldn't include the time it takes to upload a request body to the server.  Is it just that response time does normally upload time?  Or in a philosophical sense, we should include this time because this large payload for example is coming from our own front end app and we should know what the users are truly experiencing in terms of total time spent for the request?  Perhaps ideally a better name could be request time or something?

I like the idea of recording a metric for `puma.request_body_wait`.  Are you saying we would create a user-friendly name as a metric and populated it with `puma.request_body_wait` for Puma?  If we have a metric, it would be cool if we could just de-select it in the homepage chart if it's present or maybe just simply subtract it when using a query with the New Relic query language.

PR #367, opened 2020-07-23
Add sidekick checklist

rachelleahklein:
# Overview
Adding a checklist for PR reviewers.

### DIFF
--- a/.github/workflows/sidekick_checklist.yml
+++ b/.github/workflows/sidekick_checklist.yml
@@ -1,0 +1,17 @@
+name: PR Sidekick Checklist
+on: pull_request
+jobs:
+  sidekick_checklist:
+    name: 
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checklist
+        uses: unsplash/comment-on-pr@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          msg: "- [ ] Perform code review
+- [ ] Add performance label
+- [ ] Perform appropriate level of performance testing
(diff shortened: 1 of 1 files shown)


rachelleahklein:
After a few edits (sorry for the notifications noise) this has been verified to appear properly on [my fork](https://github.com/rachelleahklein/newrelic-ruby-agent/pull/6#issuecomment-663275091).

.github/workflows/sidekick_checklist.yml
```diff
@@ -0,0 +1,18 @@
...
+on: pull_request
+jobs:
+  sidekick_checklist:
+    name: 
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checklist
+        uses: unsplash/comment-on-pr@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
tannalynn, line 11: secrets aren't available on pull requests from forks, so this would only work on PRs within the repo. We ran into a similar issue with Snyk (we switched to running snyk on a schedule and on PRs to specific branches). I'm not sure that same solution would work for this since we want this to be commented on all the pull requests. Is there a way to do this same thing without secrets?

rachelleahklein: Right you are - seems like [many would like to use GHA for this purpose](https://github.community/t/github-actions-are-severely-limited-on-prs/18179) but it is not supported, at least for now. Some have worked around it, but it may be preferable for us to try to use another tool, such as a bot.

rachelleahklein:
Closing this PR while we explore other options for adding a checklist to PRs.

PR #369, opened 2020-07-28
Decrease allocations for span attributes

tannalynn:

# Overview
Previously, our segments were always creating an attributes object in the initializer. Moving the creation of the attributes object to when it is actually being used decreased allocations by 21% in a basic middleware performance test.
Results:
```
+--------------------------------------------+-----------+-----------+-------+---------------+--------------+--------------+
| name                                       | before    | after     | delta | allocs_before | allocs_after | allocs_delta |
|--------------------------------------------+-----------+-----------+-------+---------------+--------------+--------------|
| RackMiddleware#test_basic_middleware_stack | 446.70 µs | 425.91 µs | -4.7% |           302 |          236 |       -21.9% |
+--------------------------------------------+-----------+-----------+-------+---------------+--------------+--------------+
```

### DIFF
--- a/lib/new_relic/agent/transaction/segment.rb
+++ b/lib/new_relic/agent/transaction/segment.rb
@@ -17,12 +17,15 @@ module NewRelic
 
         def initialize name=nil, unscoped_metrics=nil, start_time=nil
           @unscoped_metrics = unscoped_metrics
-          @attributes = Attributes.new(NewRelic::Agent.instance.attribute_filter)
           super name, start_time
         end
 
+        def attributes
+          @attributes ||= Attributes.new(NewRelic::Agent.instance.attribute_filter)
+        end
+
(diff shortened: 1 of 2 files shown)


mwlang:
This PR is also introducing the stale issues commenter, which isn't presently on the `main` nor `dev` branches.

mwlang:
This PR is also seems to be missing the performance benchmark test that was used to validate this change.

mwlang:
After doing a bit of rebasing between `main` and `dev` and then restoring the performance test's `README.md` this PR is now in tune with `dev` branch so we can safely merge.  However, we're still missing the performance benchmark test was was used to validate this PR's improvements.

rachelleahklein:
Thanks so very much for sorting out the history here, and sorry for what I added to the confusion. 😅 

Once CI passes, this should be good to go - the performance test you're referring to (results captured in the screenshot aboce) above already existed and is not part of this PR.

Issue #338, opened 2020-06-03
NameError: uninitialized constant Rake::VERSION


CamilleDrapier:
Hi,

I have noticed that I'm getting these kind of errors when I startup a typical rails application with rake tasks:

```
ERROR : Error while detecting rake_instrumentation:
ERROR : NameError: uninitialized constant Rake::VERSION
```

The way to reproduce this is:

- Create a new rails application: `rails new blog`
- Add the newrelic_rpm gem to the gem file (and bundle): `gem "newrelic_rpm"` 
- Generate the newrelic configuration: `newrelic install --license_key="YOUR_KEY" "Blog"`
- Add a simple rake/task filter in the `common` section of the newrelic configuration file (`newrelic.yml`):
   ```
     rake:
       tasks: [".*"]
   ```
- Start the rails server: `rails server`

=> The error message should appear in `log/newrelic_agent.log`

I noticed that this problem disappears when adding the `rake` gem to the bundle, but I'm unsure if this is a good way to fix this ? (as rails has a dependency constraint on rake). Another way, of course, would be to configure the newrelic agent to `disable_rake` but I guess that is not the purpose here 😅 

I also noticed that this is happening because rails loads a very small part of the `Rake` lib in [`lib/rails/test_unit/runner.rb#L5`](https://github.com/rails/rails/blob/9734dfa1a1a2d5909652dcbb80947a92ce42c431/railties/lib/rails/test_unit/runner.rb#L5) during its initialization process which result in the [`Rake` module being defined](https://github.com/ruby/rake/blob/master/lib/rake/file_list.rb), thus the [first check passes](https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/instrumentation/rake.rb#L10), but not the `Rake::VERSION` 😅 

I believe this error could be averted by checking if the constant is defined before reading it in the [instrumentation/rake](https://github.com/newrelic/rpm/blob/f42f233c02c3ad85ed4a8b41ea9bf2a664103d26/lib/new_relic/agent/instrumentation/rake.rb#L60) (i.e: prepending `defined?(::Rake::VERSION) &&`), but here again, I'm unsure if this is the right solution as maybe rails shouldn't partially load one of it's dependency during its initialization process and make this kind of problems possible ? And maybe this instrumentation need to be loaded at this point ?... Another solution could be to try to `require` the whole `rake` dependency, but that sounds a bit strange/dangerous/unexpected...

[...]

PR #373, opened 2020-07-31
Check Rake::VERSION is defined before checking for supported versions

rachelleahklein:
# Overview
When running a Rails application with rake tasks, customers could see the following error:

```
ERROR : Error while detecting rake_instrumentation:
ERROR : NameError: uninitialized constant Rake::VERSION
```
This is because a very small piece of Rake is loaded with [the Rails test runner](https://github.com/rails/rails/blob/9734dfa1a1a2d5909652dcbb80947a92ce42c431/railties/lib/rails/test_unit/runner.rb#L5).

On this branch, we're checking to see if `Rake::VERSION` exists before going further with determining whether to install rake instrumentation. This does not seem to have any ill effects on our normal rake instrumentation.

# Related Github Issue
#338 - thanks to @CamilleDrapier for reporting this!

# Testing
Verified in my logs that the error does not appear, and there was no breakage of existing tests when run locally. Let me know, dear reviewer, if you'd like us to add a specific test for this.

Also, I manually tested that I could still see instrumented rake tasks with this fix in place, but I'd appreciate the reviewer double-checking that before merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,23 @@
 # New Relic Ruby Agent Release Notes #
 
+  * **Fix error messages about Rake instrumentation**
+
+    When running a Rails application with rake tasks, customers could see the following error in logs resulting from
+    a small part of rake functionality being loaded with the Rails test runner:
+
+    ```
+    ERROR : Error while detecting rake_instrumentation:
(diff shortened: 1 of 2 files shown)


rachelleahklein:
I'll wait to add a reviewer until we get CI sorted out.

PR #374, opened 2020-08-03
Fixes openssl paths

mwlang:
# Overview
Resolves inconsistency with openssl versions and the gems building native extensions.


.github/workflows/ci.yml
```diff
@@ -9,25 +9,29 @@ jobs:
       fail-fast: false
       matrix:
         ruby-version: [ "2.0.0-p648", "2.1.10", "2.2.10", "2.3.8", "2.4.10", "2.5.8", "2.6.6", "2.7.1", "jruby-9.2.12.0" ]
+        multiverse: [ "agent", "api", "background", "database", "httpclients", "rails", "serialization", "sinatra", "background_2", "infinite_tracing" ]
```
tannalynn, line 12: The infinite tracing tests have too many intermittent failures. I think we should remove the infinite tracing group from here and create a separate workflow to run those (we can set up this workflow later, i don't think it needs to be included in this PR). When we rerun failures, it will rerun the entire workflow and matrix which will be a headache when only infinite tracing intermittent failures occur.

tannalynn: But to be clear, this is something we can set up next. I'll merge this PR and get started on the next part.

mwlang: Sounds good!

PR #377, opened 2020-08-05
Don't modify params from the caller

johnsyweb:

👀 ✔️ 

# Overview

Upgrading newrelic_rpm from 6.11.0.365 to 6.12.0.367 had an unexpected side-effect of breaking the metrics we sent to another tool because these `params` were deleted. We've worked around this by cloning the `params`:

```diff
         new_relic.increment_metric(long_name)
-        new_relic.add_custom_attributes(dimensions)
+        new_relic.add_custom_attributes(dimensions.clone)
```

This PR takes a slightly different approach than https://github.com/newrelic/newrelic-ruby-agent/commit/6348c3a96 and doesn't modify `params`.

# Testing

I looked for some tests associated with 6348c3a96 that I could augment but I couldn't find any. The entire test suite passes locally.


mwlang:
@johnsyweb Thank you for this PR.  We're in the process of bringing our new github actions based continuous integration testing online and retiring Travis CI.  Once we've successfully completed our migration to github actions, we'll be reviewing this PR in-depth and ensuring it runs against all tests. 

On first inspection, this looks like a great change and avoids modifying the passed in hash object as you have described.  I do know we make intentional decisions on implementing interactions with hashes like this for performance reasons, but I don't think this particular API call is on a "hot-path" that would negatively impact performance.  To test that hypothesis out, I plan to set up a performance benchmark test and confirm this is indeed the case as part of accepting and merging this PR.

johnsyweb:
Thanks @mwlang, I appreciate your review.

tannalynn:
Now that our CI is setup, I'm going to close and reopen this PR to trigger our fresh CI

johnsyweb:
Thanks @tannalynn. I merged in the `dev` branch and CI passes :-)

mwlang:
Performance benchmark results show no adverse changes to processing time.  There is additional allocations, but this is expected and it's also unlikely the API is called for > 2% of all spans generated, so deemed an acceptable trade-off to solve the problem this PR solves.

## Baseline vs. PR's implementation
image

It should be noted that the baseline benchmark compared against itself has roughly similar variance:

## Baseline vs. itself
image

PR #383, opened 2020-08-12
Add Retry to running tests

tannalynn:

# Overview
Infinite tracing tests have frequent intermittent failures that cause our CI to be a bit rocky. The main agent also has some less frequent intermittent failures. Retrying the test runs when failing will help weed out the intermittent failures from true failures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,8 +124,12 @@ jobs:
       - name: Bundler install
         run: bundle install 
 
-      - name: rake test
-        run:  bundle exec rake test:env[${{ matrix.rails }}] TESTOPTS="--verbose"
+      - name: Run Unit Tests
+        uses: nick-invision/retry@v1.0.0
+        with:
+          timeout_minutes: 20
+          max_attempts: 2
+          command:  bundle exec rake test:env[${{ matrix.rails }}] TESTOPTS="--verbose"
         env:
           DB_PORT: 3306
           MYSQL_PASSWORD: root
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -192,8 +196,12 @@ jobs:
       - name: Start mysql
         run: sudo systemctl start mysql
 
-      - name: Run unit tests
-        run: bundle exec rake test:multiverse[group=${{ matrix.multiverse }},verbose]
+      - name: Run Multiverse Tests
+        uses: nick-invision/retry@v1.0.0
+        with:
+          timeout_minutes: 20
```
rachelleahklein, line 202: I noticed this caused one of the multiverse test runs to time out. Should probably be bumped up here.

.github/workflows/pr_review_checklist.yml
```diff
@@ -1,5 +1,5 @@
 name: PR Review Checklist
-on: pull_request_target
+on: pull_request
```
rachelleahklein, line 2: Thank you for fixing my earlier error here! Since this seems to be working on PRs from forks now, can we limit this to PR creation rather than [the many default types for the pull_request event](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request)?

```
on:
  pull_request:
    types: [opened]
```

rachelleahklein:
Thanks for making our tests more stable and for fixing the issue I introduced with the PR review checklist. 🎉

Issue #387, opened 2020-08-13
GitHub automation research/implementation


rachelleahklein:
We should explore whether we want to set up additional bots and/or actions to automate some aspects of our GitHub process, e.g. updating stale PRs/issues.

Parent issue: #351

PR #389, opened 2020-08-13
Close stale issues and pull requests

tannalynn:

# Overview
Runs the stale action once a day to check for stale issues and pull requests. They will be marked as stale after 30 days, and then closed 5 days after that if there is no activity. 

These labels will prevent an issue from being marked as stale
- good first issue
- help wanted
- in progress
- in review
- on hold

# Github Issue
https://github.com/newrelic/newrelic-ruby-agent/issues/387

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,21 @@
+name: "Close stale issues"
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This has been marked stale after 30 days with no activity. It will be closed in 5 days if there is no activity.'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This has been marked stale after 30 days with no activity. It will be closed in 5 days if there is no activity.'
+        stale-pr-message: 'Stale pull request message'
+        days-before-stale: 30
+        days-before-close: 5
+        stale-issue-label: 'stale'
+        exempt-issue-labels: 'good first issue,help wanted,in progress,on hold,in review'
```
rachelleahklein, line 18: Thanks for including this thorough list.

Just a note: the items on our [roadmap](https://github.com/orgs/newrelic/projects/12) are currently represented as issues, and we don't want those to be closed as stale, so if we continue to organize things in that way we'll need to add another exemption label.

PR #406, opened 2020-08-22
Ensure preconnect never uses redirect host

rachelleahklein:
# Overview
Stealing shamelessly from @astormnewrelic's [excellent explanation of this issue](https://github.com/newrelic/node-newrelic/issues/462). 😄 

When connecting to New Relic, the Ruby Agent uses the value in `Agent.config[:host]` to post a request to the New Relic preconnect endpoint. This endpoint returns a "redirect host", which (at the time of this writing) often looks like one of the following:

```
collector-00n.newrelic.com
gov-collector-00n.newrelic.com
```
This URL becomes the URL that agents will send data to.

There is a bug in the Ruby Agent related to this functionality. If the agent needs to reconnect to the collector, it will incorrectly use this redirect host to call the preconnect endpoint, when instead it should use the original configured value in `Agent.config[:host]`.

The root cause of this bug is the agent using a single `@collector` for all endpoints, which at the beginning is the configured host and after connect is permanently reset to the redirect host.

This PR stores the original configured host separately and always uses it in the case of preconnect only.

# Testing
I've included a new test for this which checks the logs to make sure we are sending things to the correct host.

This also involved moving `assert_log_contains` into our test helpers.

### DIFF
--- a/lib/new_relic/agent/new_relic_service.rb
+++ b/lib/new_relic/agent/new_relic_service.rb
@@ -38,6 +38,7 @@ module NewRelic
       def initialize(license_key=nil, collector=control.server)
         @license_key = license_key
         @collector = collector
+        @generic_collector = collector
         @request_timeout = Agent.config[:timeout]
         @ssl_cert_store = nil
         @use_bundled_certs = false
@@ -350,7 +351,7 @@ module NewRelic
         if @use_bundled_certs == false
           ::NewRelic::Agent.logger.info("Unable to connect. Falling back to bundled security certificates")
           @use_bundled_certs = true
-          retry 
+          retry
         else
           raise
         end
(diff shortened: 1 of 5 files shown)


lib/new_relic/agent/new_relic_service.rb
```diff
@@ -433,15 +434,18 @@ def invoke_remote(method, payload = [], options = {})
         data, encoding = compress_request_if_needed(data, method)
         size = data.size
 
+        # Preconnect needs to always use the generic collector host, not the redirect host
```
mwlang, line 437: excellent use of comments

rachelleahklein:
Thanks for your feedback on variable names, @mwlang. Definitely want to make sure this is as clear a read as possible. What would you think about `@configured_collector`? Whether the customer has supplied a collector or whether we are using the one in our default config source, this is in contrast with the redirect host that we receive from the preconnect service.

PR #412, opened 2020-09-01
Add workflow for release tasks

rachelleahklein:
Adding a workflow to automate release tasks as part of our migration to GitHub actions. 🎉

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,84 @@
+name: Release
+
+on: 
+  push: 
+    branches:
+      - main
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install rotp
+      run: gem install rotp
+    - name: Configure gem credentials
+      run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install rotp
+      run: gem install rotp
+    - name: Configure gem credentials
+      run: |
+        if [ -f ~/.gem/credentials ]; then
+          echo 'Using credentials in ~/.gem/credentials'
+        else
+          echo -e "---\n:rubygems_api_key: ${{ secrets.RUBYGEMS_API_KEY }} > ~/.gem/credentials
```
tannalynn, line 20: This line seems to be missing the second " for the echo
Should be `echo -e "---\n:rubygems_api_key: ${{ secrets.RUBYGEMS_API_KEY }" > ~/.gem/credentials` I believe

tannalynn: Fixed

tannalynn (approved):

PR #415, opened 2020-09-03
Fix release workflow

rachelleahklein:
A quick fix to our release workflow so we can actually install the `rotp` gem.


rachelleahklein:
Closing to base the fix on `dev`.

PR #420, opened 2020-09-04
Update gem credentials step

rachelleahklein:
A small update to our automated release workflow.


rachelleahklein:
After some discussion with @mwlang, I decided to try using `GEM_HOST_API_KEY` rather than a credentials file when we found the [documentation](https://guides.rubygems.org/command-reference/) for that option.

mwlang (approved):
LGTM.  Getting away from ~/.gem/credentials and using ENV variable was right choice here.

mwlang (approved):

Issue #349, opened 2020-06-30
Ruby 2.7 deprecation warning tracks the method wrongly to new relic line


bishwahang:
I'm not quite sure if this issue belongs here or in Ruby project. 

Ruby tracks the message defined in another library/project to this line on NewRelic gem, https://github.com/newrelic/rpm/blob/main/lib/new_relic/agent/method_tracer.rb#L337, and then throws the deprecation warning.

```ruby
# trigger_profile_confirmed_hooks.rb

class TriggerProfileConfirmedHooks
  Messenger.publish("user.confirmed", {foo: “bar”})
end


# messenger.rb
class Messenger
  def self.publish(message_type, payload = {}, opts = {})
  end
end
```

Warning:

```
app/services/trigger_profile_confirmed_hooks.rb|10 warning| Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call

/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/newrelic_rpm-6.11.0.365/lib/new_relic/agent/method_tracer.rb|337 warning| The called method `publish' is defined here
```

Apparently, it happens only for particular method signature. If I change the method name on my project to:
`def something_else(message_type, payload = {}, opts = {})`,
instead of `def publish(message_type, payload = {}, opts = {})`,
the warning goes away.

May be this will go away, when New Relic gem will handle all the 2.7 warnings. Or may be this is related with Ruby wrongly tracing method call.

tannalynn:
Hello @bishwahang 
Thank you for bringing this instance of this warning to our attention. We are working on eliminating these warnings, and will include this instance as well.
We will let you know when we have released this fix.

tannalynn:
Hello @bishwahang !
While looking into these warnings, I learned that this instance of the warning is a little different from the others.
I was able to replicate it by adding methods that match your signatures and adding add_method_tracer for the method. We had already updated the method tracer to use the new kwargs, so I believe the warning is actually coming from the way the method is being called.
When I change
`Messenger.publish("user.confirmed", {foo: “bar”})`
to 
`Messenger.publish("user.confirmed", **{foo: “bar”})`
The warning no longer appears. So adding the double splat to the locations you call the method should remove these warnings for you.

I found this page very helpful in understanding more of the nuances of the keywords arguments changes, personally. https://blog.saeloun.com/2019/10/07/ruby-2-7-keyword-arguments-redesign.html 

Hopefully that helps!

[...]

Issue #347, opened 2020-06-22
net instrumentation can cause "stack level too deep" with 3rd party gems


MathieuDerelle:
please consider replacing monkey-patching `Net::HTTP` using class reopening and "alias_method_chain" by `prepend`

[more details in the issue I submitted to airbrake](https://github.com/airbrake/airbrake/issues/1104)

mwlang:
@MathieuDerelle Thank you for this report and linking to airbrake issue as well.  This is a great example highlighting the need to move away from method chaining to a prepend approach to instrumenting.  I am placing this issue in our "review bucket" as a reminder to prioritize refactoring all instrumentation hooks that are not already using the prepend approach.

gsar:
also affected by this issue. sadly, the suggested workaround of moving the `newrelic_rpm` line in the `Gemfile` ahead of `airbrake` doesn't seem to fix it here.

MathieuDerelle:
@gsar the order in the `Gemfile` does NOT matter
it is the order of the `require`s that matters

take a look at your `config/application.rb` and look for a `Bundler.require(*Rails.groups)` or `require 'airbrake'` and put a `Bundler.require(:newrelic)` before it

gsar:
@MathieuDerelle thanks for the tip. i tried your suggestion but it doesn't seem to make a difference, i still see the `stack level too deep` error with the stack showing airbrake and newrelic httpclient interceptors.

the issue occurs within a rake task but the main app itself appears to be fine. My `Rakefile` contains just this:

``` ruby
require File.expand_path('config/application', __dir__)

Rails.application.load_tasks
Rails.application.config.autoflush_log = true
Doorkeeper::Rake.load_tasks
```

angelatan2:
This issue is included in the bug smash that we are working on.

PR #422, opened 2020-09-04
WIP version that implements prepend version of method tracer

mwlang:
# Overview
We have a recurring issue with Ruby 2.7 where using add_method_tracer is challenging to get right in all scenarios and argument combinations.  Ruby 2.7 is deprecating automatic keyword arguments and attempting to replicate method signatures causes warnings emitted to the log file.

This work in progress demonstrates using prepend instead of append as the method to latch onto methods the users wish to trace in their apps.  

TODO: 
* Continue fleshing out the solution and adapting test framework to test `prepended` version as fully as the `included` version.
* Extract the `included` version out to its own file and prepare to deprecate it with next major release.
* Clean up implementation of `prepended` version as needed to fully replicate `included`'s functionality.

# Related Github Issue
* See issue #349 

# Testing
exercise your changes.

### DIFF
--- a/lib/new_relic/agent/method_tracer.rb
+++ b/lib/new_relic/agent/method_tracer.rb
@@ -48,12 +48,19 @@ module NewRelic
 
     module MethodTracer
 
-      def self.included clazz
-        clazz.extend ClassMethods
+      def self.included target
+        target.extend ClassMethods
+        NewRelic::Agent.logger.warn \
+          "`include NewRelic::Agent::MethodTracer` is deprecated\n" \
+          "Please use `prepend NewRelic::Agent::MethodTracer` instead"
       end
 
-      def self.extended clazz
-        clazz.extend ClassMethods
+      def self.extended target
+        target.extend ClassMethods
+      end
+
+      def self.prepended target
+        target.extend PrependMethods
       end
 
       # Trace a given block with stats and keep track of the caller.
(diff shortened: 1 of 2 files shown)


lib/new_relic/agent/method_tracer.rb
```diff
@@ -96,6 +103,260 @@ def trace_execution_unscoped(metric_names, options=NewRelic::EMPTY_HASH) #THREAD
...
+            header += options[:code_header].to_s
+            header
+          end
+
+          # returns an eval-able string that contains the traced
+          # method code used if the agent is not creating a scope for
+          # use in scoped metrics.
+          def method_without_push_scope(method_name, metric_name_code, options)
+            define_method _traced_method_name(method_name, metric_name_code) do |*arguments|
+              eval assemble_code_header(method_name, metric_name_code, options)
```
mwlang, line 191: Compare this call to line 207 which doesn't call `assemble_code_header`. Is this an oversight?

lib/new_relic/agent/method_tracer.rb
```diff
@@ -48,12 +48,19 @@ module Agent
...
     module MethodTracer
 
-      def self.included clazz
-        clazz.extend ClassMethods
+      def self.included target
+        target.extend ClassMethods
+        NewRelic::Agent.logger.warn \
+          "`include NewRelic::Agent::MethodTracer` is deprecated\n" \
+          "Please use `prepend NewRelic::Agent::MethodTracer` instead"
       end
```
mwlang, line 56: We may be able to change this to:
```ruby
      def self.included target
        target.prepend PrependMethods
        NewRelic::Agent.logger.warn \
          "`include NewRelic::Agent::MethodTracer` is deprecated\n" \
          "Please use `prepend NewRelic::Agent::MethodTracer` instead"
      end
```
which, if it works, means we can throw away `ClassMethods` altogether.  Users who `include` will still continue to work with a deprecation warning logged.

lib/new_relic/agent/method_tracer.rb
```diff
@@ -96,6 +103,260 @@ def trace_execution_unscoped(metric_names, options=NewRelic::EMPTY_HASH) #THREAD
...
+            end
+          end
+
+          # returns an eval-able string that contains the tracing code
+          # for a fully traced metric including scoping
+          def method_with_push_scope(method_name, metric_name_code, options)
+            define_method _traced_method_name(method_name, metric_name_code) do |*arguments|
+              eval options[:code_header]
+              result = ::NewRelic::Agent::MethodTracerHelpers.trace_execution_scoped(metric_name_code, :metric => options[:metric]) do
+                super(*arguments)
```
mwlang, line 209: Note how we're completely sides-stepping the arguments issue with using `define_method` and calling `super` with original arguments.  We no longer have to concern ourselves with Ruby <= 2.6 vs >= 2.7

lib/new_relic/agent/method_tracer.rb
```diff
@@ -96,6 +103,260 @@ def trace_execution_unscoped(metric_names, options=NewRelic::EMPTY_HASH) #THREAD
...
+        # that are not allowed in the middle of method names
+        def _sanitize_name(name)
+          name.to_s.tr_s('^a-zA-Z0-9', '_')
+        end
+      end
+
+##############################################################################################      
+##############################################################################################      
+##############################################################################################      
+##############################################################################################
```
mwlang, line 358: This marks end of changes and duplicated code of the `ClassMethods` module.

mwlang:
Using the prepend version of method tracing may also solve #424, so tagging the issue here.

gsar:
also might be a resolution for #347?

mwlang:
Abandoning this PR and branch in favor of #499

PR #437, opened 2020-09-10
Update snyk workflow

tannalynn:

# Overview
Updated snyk to run on all PRs to main.


rachelleahklein:
Thanks for making this change. When I was looking at this I also wondered whether we need to use the `setup-ruby` action or whether we can get away with using the default Ruby.

mwlang:
That's a really good point about setup-ruby.  When we first put this together, we were _just_ getting started with github actions and didn't even know Ruby was pre-installed on the default containers.  So yeah, I think we can remove this complexity from the workflow.

tannalynn:
Alright, i removed the ruby setup from the workflow like suggested.

PR #443, opened 2020-09-11
Fix for snyk workflow

tannalynn:

# Overview
Need to setup ruby before bundling because the included ruby doesn't have that


rachelleahklein (approved):
Thanks for reverting this change. I think we're now squarely in "knowing enough to ~be dangerous~ sabotage ourselves" category when it comes to GitHub acitons.

Issue #450, opened 2020-09-18
newrelic_rpm 6.13.0 causes elasticsearch results to fail


sighmon:
Version `6.13.0` seems to be incompatible with elasticsearch-rails `7.1.1`

## Description
When updating from `6.12.0.367` to `6.13.0` the search results from elasticsearch only returned the first ten results, and the pagination failed.

## Expected Behavior
Elasticsearch returns full results, and pagination works.

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results
Unfortunately there weren't any errors. It silently returned the results mentioned above.

## Steps to Reproduce
Unfortunately I don't have time to reproduce this bug in a repository, sorry!

## Your Environment
Heroku
Bonsai Elasticsearch
ruby 2.7.1
rails 5.2.4.4
elasticsearch-rails 7.1.1
newrelic_rpm 6.13.0

## Additional context
Trying to view NewRelic data from Heroku's dashboard is also broken. I can't see any error logs anymore.

tannalynn:
Thank you for bringing this issue to our attention. We will look into this.

mwlang:
@sighmon Thank you for opening this issue.  Hopefully we can get to the bottom of this one, but there's not a lot of information to go on.  

First, I am assuming that reverting back to `6.12.0.367` resolves this issue?  If not, then this suggests another change outside the agent update is causing this issue.  If using 6.12 does resolve current issue, then can you share what instrumentation code (such as `add_method_tracer`, `add_custom_attributes`, etc.) might be wrapping your calls to ES?  

In purely speculative terms, there were two changes that _might_ have an effect on ES results returned in 6.13:

* calling `add_custom_attributes` no longer modifies the params of the caller

  The previous agent's improvements to recording attributes at the span level had an unexpected side-effect of modifying the params passed to the API call as duplicated attributes were deleted in the process. This is now fixed and params passed in are no longer modified.

* `http.url` query parameters spans are now obfuscated

  Previously, the agent was recording the full URL of the external requests, including the query and fragment parts of the URL as part of the attributes on the external request span. This has been fixed so that the URL is obfuscated to filter out potentially sensitive data.

[...]

PR #452, opened 2020-09-23
resolves Issue 451 and possibly 450

### LINKED ISSUE #451: Agent version 6.13.0 removes query parameters from original http request uri object when obfuscating
After upgrading from newrelic_rpm 6.12.0 to 6.13.0 we ran into an issue where our requests seemed to be missing it's query parameters when instrumented by the newrelic agent.  

## Description
In https://github.com/newrelic/newrelic-ruby-agent/pull/404/files#diff-4a15d5cf920433fc7286b4df743a5e6cR17-R25 a change was made to the `filter_uri` method (renamed to obfuscated_uri). The method now calls `parse_and_normalize_url` and the `url.dup`, that was previously present, has been removed. Since the `parse_and_normalize_url` method has some flows in which it does not create a new object it now removes query params (and BA) from the _original_ url object. This breaks the request in many cases.

## Expected Behavior
Obfuscating the request url should not modify the orginal object but always create a new url object instead.

## Steps to Reproduce
Adding a test case to the uri_util_test suite as follows demonstrates the breaking flow:
```
  def test_obfuscate_should_not_modify_input
    original = ::URI.parse("https://foo.com/bar/baz?a=1&b=2#fragment")
    to_obfuscate = original.dup

    NewRelic::Agent::HTTPClients::URIUtil.obfuscated_uri(to_obfuscate).to_s

    assert_equal original, to_obfuscate
  end
```
which results in 
```
  1) Failure:
URIUtilTest#test_obfuscate_should_not_modify_input [test/new_relic/agent/http_clients/uri_util_test.rb:61]:
--- expected
+++ actual
@@ -1 +1 @@
-#<URI::HTTPS https://foo.com/bar/baz?a=1&b=2#fragment>
+#<URI::HTTPS https://foo.com/bar/baz>
```

## Your Environment
newrelic_rpm 6.13.0
ruby 2.6.3
rails 5.2.3

mwlang:

# Overview
A recent change to the Ruby agent to obfuscate URIs sent to external services had the unintended side-effect of removing query parameters from the original URI.  This is fixed to obfuscate while also preserving the original URI

# Related Github Issue
Resolves #451 and potentially #450.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # New Relic Ruby Agent Release Notes #
 
+  ## NEXT VERSION
+
+  * **Bugfix: obfuscating URLs to external services no longer modifying original URI
+
+    A recent change to the Ruby agent to obfuscate URIs sent to external services had the unintended side-effect of removing query parameters
+    from the original URI.  This is fixed to obfuscate while also preserving the original URI
+
+    Thanks to @VictorJimenezKwast for pinpointing and helpful unit test to demonstrate.
+
   ## v6.13.0
 
   * **Bugfix: never use redirect host when accessing preconnect endpoint**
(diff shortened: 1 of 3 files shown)


rachelleahklein:
Thank you for the fix, @mwlang!

For those who are tracking this issue, we plan to release this in a version 6.13.1 as soon as possible.

rachelleahklein (approved):

PR #459, opened 2020-09-29
Update stale PR message

tannalynn:

# Overview
Updates the stale PR message to be more specific


mwlang (commented):
This actually prompts a question:  Do we remove the `stale` label if activity on the PR or Issue occurs after marking it stale?  This seems like a logical thing to do!

rachelleahklein:
> This actually prompts a question: Do we remove the `stale` label if activity on the PR or Issue occurs after marking it stale? This seems like a logical thing to do!

Sounds good to me too. When researching whether this option existed I found it and added it.

mwlang (approved):

PR #468, opened 2020-10-08
Add checks for nil middleware class in use_with_newrelic

rachelleahklein:

# Overview
This PR skips all middleware processing in the case that a `nil` value exists in the middleware chain. Sorry the diff is rendering weird - the only actual change is checking for `nil` here.

# Related Issue
Here's the [internal ticket](https://newrelic.atlassian.net/browse/GTSE-3080) where the bug was reported through support. The customer was seeing `nil` passed in as a middleware (reason unknown on their end), resulting in error messages such as `NoMethodError: undefined method `new' for nil:NilClass`.

This is confirmed to have solved the customer's issue and we should get this fix into our next release.

# Testing
This is a very small change so I have not added a test, but I'm happy to do so if it seems advisable, dear reviewer.

### DIFF
--- a/lib/new_relic/agent/instrumentation/rack.rb
+++ b/lib/new_relic/agent/instrumentation/rack.rb
@@ -99,27 +99,31 @@ module NewRelic
             else
               run_without_newrelic(app, *args, **kwargs)
             end
-          end          
+          end
         end
 
         if RUBY_VERSION < "2.7.0"
           def use_with_newrelic(middleware_class, *args, &blk)
-            if ::NewRelic::Agent::Instrumentation::RackHelpers.middleware_instrumentation_enabled?
-              wrapped_middleware_class = ::NewRelic::Agent::Instrumentation::MiddlewareProxy.for_class(middleware_class)
-              use_without_newrelic(wrapped_middleware_class, *args, &blk)
-            else
-              use_without_newrelic(middleware_class, *args, &blk)
+            unless middleware_class.nil?
+              if ::NewRelic::Agent::Instrumentation::RackHelpers.middleware_instrumentation_enabled?
+                wrapped_middleware_class = ::NewRelic::Agent::Instrumentation::MiddlewareProxy.for_class(middleware_class)
+                use_without_newrelic(wrapped_middleware_class, *args, &blk)
+              else
+                use_without_newrelic(middleware_class, *args, &blk)
+              end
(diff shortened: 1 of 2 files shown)


lib/new_relic/agent/instrumentation/rack.rb
```diff
@@ -99,27 +99,31 @@ def run_with_newrelic(app, *args, **kwargs)
...
         end
 
         if RUBY_VERSION < "2.7.0"
           def use_with_newrelic(middleware_class, *args, &blk)
-            if ::NewRelic::Agent::Instrumentation::RackHelpers.middleware_instrumentation_enabled?
-              wrapped_middleware_class = ::NewRelic::Agent::Instrumentation::MiddlewareProxy.for_class(middleware_class)
-              use_without_newrelic(wrapped_middleware_class, *args, &blk)
-            else
-              use_without_newrelic(middleware_class, *args, &blk)
+            unless middleware_class.nil?
```
mwlang, line 107: use guard clause instead of adding another conditional block around the method's implementation.

lib/new_relic/agent/instrumentation/rack.rb
```diff
@@ -99,27 +99,31 @@ def run_with_newrelic(app, *args, **kwargs)
...
             end
           end
         else
           def use_with_newrelic(middleware_class, *args, **kwargs, &blk)
-            if ::NewRelic::Agent::Instrumentation::RackHelpers.middleware_instrumentation_enabled?
-              wrapped_middleware_class = ::NewRelic::Agent::Instrumentation::MiddlewareProxy.for_class(middleware_class)
-              use_without_newrelic(wrapped_middleware_class, *args, **kwargs, &blk)
-            else
-              use_without_newrelic(middleware_class, *args, **kwargs, &blk)
+            unless middleware_class.nil?
```
mwlang, line 118: same as above.  guard clause here.

lib/new_relic/agent/instrumentation/rack.rb
rachelleahklein: That is a bit too much nesting, isn't it? I will happily make that change for readability.

mwlang (approved):
LGTM

mwlang (approved):
This is now ready to merge.  Only change was to add CHANGELOG entry.

PR #469, opened 2020-10-11
Use connection_db_config if it's available

tomgilligan:
# Overview

This PR is really just to silence this Rails 6.1 deprecation warning

```
DEPRECATION WARNING: The kwarg `spec_name` is deprecated in favor of `name`. `spec_name` will be removed in Rails 6.2 (called from block in <class:EnvironmentReport> at /Users/tomgilligan/.asdf/installs/ruby/2.6.1/lib/ruby/gems/2.6.0/gems/newrelic_rpm-6.13.1/lib/new_relic/environment_report.rb:74)
DEPRECATION WARNING: DatabaseConfig#config will be removed in 6.2.0 in favor of DatabaseConfigurations#configuration_hash which returns a hash with symbol keys (called from block in <class:EnvironmentReport> at /Users/tomgilligan/.asdf/installs/ruby/2.6.1/lib/ruby/gems/2.6.0/gems/newrelic_rpm-6.13.1/lib/new_relic/environment_report.rb:74)
```

# Testing

No tests introduced thus far.  Probably it'd be nice to introduce a Rails 6.1 environment for testing but it's still pre-release so maybe that's a bit premature?

### DIFF
--- a/lib/new_relic/environment_report.rb
+++ b/lib/new_relic/environment_report.rb
@@ -71,7 +71,13 @@ module NewRelic
     report_on('OS'                ) { ::NewRelic::Agent::SystemInfo.ruby_os_identifier     }
     report_on('Database adapter'  ) do
       begin
-        ActiveRecord::Base.configurations.configs_for(env_name: NewRelic::Control.instance.env, spec_name: "primary").config['adapter']
+        if ::ActiveRecord::Base.respond_to?(:connection_db_config)
+          ActiveRecord::Base.configurations.configs_for(env_name: NewRelic::Control.instance.env, name: "primary")
+            .connection_db_config.configuration_hash['adapter']
+        else
+          ActiveRecord::Base.configurations.configs_for(env_name: NewRelic::Control.instance.env, spec_name: "primary").config['adapter']
+        end
+
       rescue NoMethodError
         ActiveRecord::Base.configurations[NewRelic::Control.instance.env]['adapter']
       end


lib/new_relic/environment_report.rb
```diff
@@ -71,7 +71,13 @@ def self.registered_reporters=(logic)
     report_on('OS'                ) { ::NewRelic::Agent::SystemInfo.ruby_os_identifier     }
     report_on('Database adapter'  ) do
       begin
-        ActiveRecord::Base.configurations.configs_for(env_name: NewRelic::Control.instance.env, spec_name: "primary").config['adapter']
+        if ::ActiveRecord::Base.respond_to?(:connection_db_config)
+          ActiveRecord::Base.configurations.configs_for(env_name: NewRelic::Control.instance.env, name: "primary")
```
rachelleahklein, line 75: I tested out your fix with Rails 6.1 and I am actually seeing a NoMethodError arise from this line:

```
NoMethodError: undefined method `connection_db_config' for #<ActiveRecord::DatabaseConfigurations::HashConfig:0x00007fd54e72c6c8>
```

As that error implies, calling `connection_db_config` on `ActiveRecord::Base` directly will work:

```
::ActiveRecord::Base.connection_db_config.configuration_hash[:adapter]
```

But we really want to make sure we are accessing configuration for the agent's current environment, so instead we may want something more like this:

```
ActiveRecord::Base.configurations.configs_for(env_name: NewRelic::Control.instance.env, name: "primary").configuration_hash[:adapter]
```

And since we wouldn't be using `connection_db_config` in this case, it probably makes sense to change the conditional to be on Rails version, unless you can think of a better alternative.

lib/new_relic/environment_report.rb
```diff
@@ -71,7 +71,13 @@ def self.registered_reporters=(logic)
...
+            .connection_db_config.configuration_hash['adapter']
+        else
+          ActiveRecord::Base.configurations.configs_for(env_name: NewRelic::Control.instance.env, spec_name: "primary").config['adapter']
+        end
+
       rescue NoMethodError
```
rachelleahklein, line 81: Incidentally, your PR brought to light another potential improvement: the `NoMethodError` that surfaced in my testing was caught here and raised another deprecation warning:

```
DEPRECATION WARNING: [] is deprecated and will be removed from Rails 6.2 (Use configs_for) (called from rescue in block in <class:EnvironmentReport> at /Users/rklein/newrelic/ruby_agent/lib/new_relic/environment_report.rb:82)
```

If you're interested in addressing this as well, you'd be most welcome. Otherwise, I can add it to our backlog for Rails 6.1 support.

rachelleahklein (commented):
Hi @tomgilligan! Thanks so much for your contribution! You're right that we usually wait until there is a release candidate gem before adding a new version of Rails to our test suites, but we really appreciate PRs like these that help us get a head start on official support.

I've got a few comments on the implementation of this based on my own testing. Please let me know if your tests of this branch are giving you different results.

Again, thank you, and looking forward to future discussion.

tomgilligan:
Hi @rachelleahklein .  Thank you very much for such a considered review. 

> Please let me know if your tests of this branch are giving you different results.

I will re-check this today.

lib/new_relic/environment_report.rb
tomgilligan:
```
NoMethodError: undefined method `connection_db_config' for #<ActiveRecord::DatabaseConfigurations::HashConfig:0x00007fd54e72c6c8>
```

I did not get this error but I will try to recreate it.

> And since we wouldn't be using connection_db_config in this case, it probably makes sense to change the conditional to be on Rails version, unless you can think of a better alternative.

A condition on the Rails version sounds good to me.

lib/new_relic/environment_report.rb
tomgilligan: I will address this also.

mwlang (approved):
LGTM

mwlang:
@tomgilligan Thank you very much for your work on this.  This fix gives us a leg up in supporting the next Rails version.

PR #492, opened 2020-11-02
request headers are now logged

### LINKED ISSUE #472: Http Header Details
<p>We are currently trying to build a solution for mobile distributed tracing and we need the ability to view then header details for the transactions reported within New Relic. Use case, a HTTP request from a mobile application, we are appending a New relic header trace.id to that request. We need to see if the agent is picking that header record up correctly.</p><p></p><p>Can we get the header details exposed in the ruby agent logs?</p>

mwlang:
# Overview
HTTP request headers will be logged when log level is at least debug level.  

Similarly, request headers for exchanges with New Relic servers are now audit logged when audit logging is enabled.

# Related Github Issue
This resolves #472

### DIFF
--- a/lib/new_relic/agent/audit_logger.rb
+++ b/lib/new_relic/agent/audit_logger.rb
@@ -26,6 +26,16 @@ module NewRelic
         !@log.nil?
       end
 
+      def log_request_headers(uri, headers)
+        return unless enabled? && allowed_endpoint?(uri)
+        @log.info("REQUEST HEADERS: #{headers}")
+      rescue StandardError, SystemStackError, SystemCallError => e
+        ::NewRelic::Agent.logger.warn("Failed writing request headers to audit log", e)
+      rescue Exception => e
+        ::NewRelic::Agent.logger.warn("Failed writing request headers to audit log with exception. Re-raising in case of interrupt.", e)
+        raise
+      end
+
       def log_request(uri, data, marshaller)
         return unless enabled? && allowed_endpoint?(uri)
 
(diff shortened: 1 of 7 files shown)


lib/new_relic/agent/audit_logger.rb
```diff
@@ -26,6 +26,16 @@ def setup?
...
       end
 
+      def log_request_headers(uri, headers)
+        return unless enabled? && allowed_endpoint?(uri)
+        @log.info("REQUEST HEADERS: #{headers}")
+      rescue StandardError, SystemStackError, SystemCallError => e
+        ::NewRelic::Agent.logger.warn("Failed writing request headers to audit log", e)
+      rescue Exception => e
+        ::NewRelic::Agent.logger.warn("Failed writing request headers to audit log with exception. Re-raising in case of interrupt.", e)
+        raise
```
tannalynn, line 36: What are the situations that could happen that this section is taking care of?

tannalynn (commented):
Do you think there are any tests we should add, or tests we should update to check for these changes?

lib/new_relic/agent/audit_logger.rb
mwlang: The only thing I can think of that could happen if no header at all `nil` or the headers are malformed in some what that prevents them from casting to `String` when outputting to log file.

It's super critical path, so over abundance of caution here.  Same as method below it in the same file and I just repeated the pattern.

tannalynn (approved):
Looks good to me, Thank you!

PR #494, opened 2020-11-04
restores .erb to the search paths for config files

### LINKED ISSUE #493: The agent is not processing newrelic.yml.erb when present.
## Description

The agent is able to parse `newrelic.yml.erb` but fails to list the `.erb` extension in the paths it searches at start up, so an ERB template file is never attempted at load time.

## Expected Behavior
The agent should look for `newrelic.yml.erb` in all the places it looks for `newrelic.yml` and load and parse the ERB template version of the config file when one is found.

## Additional context
Ensure all findable newrelic.yml files are removed in order to discover the `newrelic.yml.erb` version since `newrelic.yml` will take precedence in the discovery process.

## Use Case

1. The user wants to create the `app_name` from within the app
1. Once set, the `app_name` won’t change.
 
The user wants to construct the name from other environment variables. One solution would be if the newrelic.yml parser recognized ERB style interpolation (i.e. <%= %>).

mwlang:
# Overview
Adds `newrelic.yml.erb` to the list of directories and files searched at start up.

# Related Github Issue
Resolves #493

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
   ## v6.14.0
 
+  * **Bugfix: `newrelic.yml.erb` added to the configuration search path**
+
+    Previously, when a user specifies a `newrelic.yml.erb` and no `newrelic.yml` file, the agent fails to find
+    the `.erb` file because it was not in the list of files searched at startup.  The Ruby agent has long supported this as a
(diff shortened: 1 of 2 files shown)


angelatan2 (approved):
Looks like you are pushing into path two additional paths to search.

PR #496, opened 2020-11-06
Add prepend option for Net::HTTP instrumentation

tannalynn:

# Overview
This change adds a new config option to enable the instrumentation of Net::HTTP to use prepend rather than monkey patching. 

# Related Github Issue
https://github.com/newrelic/newrelic-ruby-agent/issues/347

# Testing
Added a new test suite `net_http_prepend`. I pulled out all of the existing work in net_http test suite into a helper module and included that in both `net_http_prepend` and `net_http`. These needed to be in separate suites because the instrumentation is set up when the agent starts, and cannot switch instrumentation implementations while running.

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -823,6 +823,13 @@ module NewRelic
           :allowed_from_server => false,
           :description => 'If <code>true</code>, uses Module.prepend rather than alias_method for ActiveRecord instrumentation.'
         },
+        :prepend_net_instrumentation => {
+          :default => false,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => 'If <code>true</code>, uses Module.prepend rather than alias_method for Net::HTTP instrumentation.'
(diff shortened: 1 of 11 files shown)


mwlang (changes requested):
Let's make prepend the default so we're accelerating adoption of this technique.

tannalynn:
> Let's make prepend the default so we're accelerating adoption of this technique.

Sounds great, this is now done.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -830,6 +830,13 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'If <code>true</code>, uses Module.prepend rather than alias_method for ActiveRecord instrumentation.'
         },
+        :prepend_net_instrumentation => {
+          :default => false,
```
mwlang: still false here?

mwlang (approved):
All Good.  Friday the 13th, notwithstanding. Let's Roll.

Issue #471, opened 2020-10-12
set-env and add-path in Github Actions is deprecated.


mwlang:
## Description
We have a few places we're using `set-env` in Github Actions.  This command is being deprecated and currently littering our CI workflow with hundreds of warnings:

```
 Check warning on line 1 in .github

@github-actions
github-actions
/ mini-env (rails31, 2.1.10)

.github#L1
The `set-env` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

The `add-path` command is also being deprecated and should be addressed if used.

## Expected Behavior
Github Actions should run without deprecation warnings.

## Additional context

* https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
* https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

Issue #463, opened 2020-10-01
Increase the retry and timeout intervals on the CI workflow


mwlang:
### Feature Description
Github itself seems to have performance issues at times that causes our intermittently failing tests to fail more often than expected during specific times of the day.  Raising the retry counts and timeouts should alleviate the intermittent fail incidents.

### Describe Alternatives
Find and resolve root cause of intermittent fails (note: this could take some significant time!)

### Priority
Must Have

PR #497, opened 2020-11-07
Updates for our github action workflows

tannalynn:

# Overview
- Changed release workflow so it is no longer using the deprecated `set-env`, and is now using [`$GITHUB_ENV`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable). 
- Pulled infinite tracing tests into a separate job in the same CI workflow. This enables us to bump up the retry limits for only infinite tracing, which has the most intermittent failures. This makes it so that we don't have to wait even longer in the case of real failures in other parts of the agent unrelated to infinite tracing. 
- added [`continue-on-error`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) to workflows that run on pull requests, but are not testing. This includes adding the workflows for PR checklist commenting and assigning to the project board. 

# Related Github Issue
https://github.com/newrelic/newrelic-ruby-agent/issues/471 https://github.com/newrelic/newrelic-ruby-agent/issues/463

### DIFF
--- a/.github/workflows/assignproj.yml
+++ b/.github/workflows/assignproj.yml
@@ -11,6 +11,7 @@ env:
 jobs:
   assign_one_project:
     runs-on: ubuntu-latest
+    continue-on-error: true
     name: Assign to One Project
     steps:
     - name: Assign NEW issues and NEW pull requests to project 2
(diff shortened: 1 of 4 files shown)


mwlang (approved):
This is solid work.  Really like the inspiration to breakout the infinite tracing gem into it's own workflow step for retriability (new word I just made up!) purposes.

PR #498, opened 2020-11-11
Don't overwrite existing sidekiq trace info.

liaden:

# Overview
We are using Shoryuken to pull from SQS and push to Sidekiq. This is a small slice of logic that we don't want monitored with with New Relic, but we do want to connect the distributed tracing that is happening from the Ruby application on the other side of SNS. Our Shoryuken worker is putting the key in place and it is being overwritten by sidekiq.

I can have our application delete the New Relic client sidekiq middleware within shoryuken, but I figured it'd be sharing what I ran into here as well.

Maybe `::NewRelic::Agent.config[:'distributed_tracing.enabled']` should be false if `::NewRelic::Agent.config[:agent_enabled]` is false?

# Related Github Issue
N/A

# Testing
exercise your changes.

### DIFF
--- a/lib/new_relic/agent/instrumentation/sidekiq.rb
+++ b/lib/new_relic/agent/instrumentation/sidekiq.rb
@@ -46,7 +46,7 @@ DependencyDetection.defer do
       end
       class Client
         def call(_worker_class, job, *_)
-          job[NewRelic::NEWRELIC_KEY] = distributed_tracing_headers if ::NewRelic::Agent.config[:'distributed_tracing.enabled']
+          job[NewRelic::NEWRELIC_KEY] ||= distributed_tracing_headers if ::NewRelic::Agent.config[:'distributed_tracing.enabled']
           yield
         end
 


mwlang:
@liaden Thank you for sharing the backstory around this change and PR.  I will look at the more, but on the surface, this seems reasonable.

Could you please explain further this comment?
> Maybe `::NewRelic::Agent.config[:'distributed_tracing.enabled']` should be false if `::NewRelic::Agent.config[:agent_enabled]` is false?

What benefit is gained with this suggested change?

mwlang:
@liaden This PR is waiting on further feedback from plus also signing the CLA.  This looks like a useful change we'd be in favor of incorporating.  

Thanks again for the PR and helping us make the Ruby agent better.  Please do let us know if you'd like to move this PR forward, otherwise we will close this PR with no action in a few days if we don't hear back.

liaden:
@mwlang  When I was looking into it, `::NewRelic::Agent.config[:'distributed_tracing.enabled']` was true (truthy) while `::NewRelic::Agent.config[:agent_enabled]` was false. 

I have a Rails application (with newrelic) that enqueues to SNS/SQS. We use shoryuken to pull from SQS and then push it into sidekiq. With shoryuken being a small shim, we see no value in enabling monitoring for it, but maintaining the distributed tracing across the publisher to the other applications sidekiq is pretty beneficial. Given the above settings, my solution prevents overriding the trace headers; however, it might make more sense to not create the trace headers if the agent is disabled. That seems like a bigger change that I don't have context for.

mwlang:
@liaden Thank you for providing more insight into what your use-case.  What you say makes sense in this context.  

After some consideration, I think it's best not to further entangle `agent_enabled` and `distributed_tracing.enabled` settings.  While the idea has merit, it's a bit of added complexity that may not always be the least surprising to users.  We do have precedence for flipping one setting depending on another setting, but these cases have clear mutually exclusive reasons driving those entanglements.

Thank you for signing the CLA and providing this thoughtful PR.  We will merge it as-is and it will appear in the next release of our agent.

Issue #486, opened 2020-10-28
Agent does not work with RackApp on Ruby 2.7


mitnal:
The new relic agent does not work together with [RackApp](http://www.rack-app.com) on ruby 2.7.

## Description

After upgrading to Ruby 2.7 our App did not start anymore in production env. After digging around it seems that one of the ruby 2.7 fixes splats an argument and the RackApp argument is 'converted' to a hash. RackApp then calls `serialize` on the argument expecting it would be a custom RackApp class, but instead it is a hash. This results in `undefined method 'serializer' for #<Hash:0x00007fa43fb8c758> (NoMethodError)` and the app does not start.

Because we have no problems running this combination in Ruby 2.6 I think the problem is the splat here int the Ruby 2.7 path:
https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/middleware_proxy.rb#L30-L40

## Expected Behavior

No `NoMethodError` when using the ruby agent and Rack App in the same app. The agent should not change arguments to rack apps (in this case it should not convert it to a hash).

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag]

Traceback from test app.
```
rails s
=> Booting Puma
=> Rails 6.0.3.4 application starting in production
=> Run `rails server --help` for more startup options
Exiting
Traceback (most recent call last):
	96: from bin/rails:9:in `<main>'
	95: from [...]/.rvm/gems/ruby-2.7.2/gems/activesupport-6.0.3.4/lib/active_support/dependencies.rb:324:in `require'
	94: from [...]/.rvm/gems/ruby-2.7.2/gems/activesupport-6.0.3.4/lib/active_support/dependencies.rb:291:in `load_dependency'
	93: from [...]/.rvm/gems/ruby-2.7.2/gems/activesupport-6.0.3.4/lib/active_support/dependencies.rb:324:in `block in require'
	92: from [...]/.rvm/gems/ruby-2.7.2/gems/bootsnap-1.4.9/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:31:in `require'
	91: from [...]/.rvm/gems/ruby-2.7.2/gems/bootsnap-1.4.9/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:22:in `require_with_bootsnap_lfi'
[...]

PR #499, opened 2020-11-12
Method tracers refactored

### LINKED ISSUE #349: Ruby 2.7 deprecation warning tracks the method wrongly to new relic line

### LINKED ISSUE #423: ruby 2.7 deprecation warning originating in delayed_job_instrumentation.rb
## Description

Upgrading to ruby 2.7.1, we are seeing the following warning lines numerous times when delayed jobs are run:

```
Sept 06 16:48:32 app/worker.1 /app/vendor/bundle/ruby/2.7.0/gems/delayed_job-4.1.8/lib/delayed/tasks.rb:9: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
Sept 06 16:48:32 app/worker.1 /app/vendor/bundle/ruby/2.7.0/gems/newrelic_rpm-6.12.0.367/lib/new_relic/agent/instrumentation/delayed_job_instrumentation.rb:113: warning: The called method `initialize' is defined here
```

## Expected Behavior
There should be no warnings.

### LINKED ISSUE #424: add_method_tracer modifies object arguments
`add_method_tracer` causes object arguments to be cloned.

## Description
When tracing a method that takes an object argument that is supposed to be mutated during the execution of that method, the tracing causes the passed argument to be a clone of the object, meaning that any mutations are lost to outside references to the same object.

## Expected Behavior
No cloning of arguments should happen and the passed object should be included in the call to the method as is.

## Steps to Reproduce
```ruby
class Demonstration

  def call
    (1..10).each_with_object({}) do |i, hash|
      process(i, hash)
    end
  end

  def process(i, hash)
    hash[i] = i * 2
  end

  add_method_tracer :process, 'Demonstration/process'
end

Demonstration.new.call
#=> {}
```

Expected result:

```ruby
Demonstration.new.call
=> {1=>2, 2=>4, 3=>6, 4=>8, 5=>10, 6=>12, 7=>14, 8=>16, 9=>18, 10=>20}
```

## Your Environment
ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-darwin19]

## Additional context
It seems like this is not an issue in Ruby 2.6

mwlang:
# Overview
This PR aims to solve some of the 2.7+ issues encountered with method argument delegation when using `add_method_tracer` to instrument specific methods.

# Related Github Issue
* Resolves #424 by not double-splatting `**kwargs`
* Resolves #423 by removing `(*args, **kwargs)` in the tracing method's definition (we were overzealous as with RackApp #486)
* Resolves #349 by not using `(*args, **kwargs)` and by also adding `ruby2_keywords` to identify the method_tracer wrapper as a delegator of method arguments in a way that's compatible with all the rubies.  

NOTE: `ruby2_keywords` expected to be retired in Ruby 3.2, so must refactor how `add_method_tracer` works in general to continue supporting long-term.


lib/new_relic/agent/method_tracer.rb
```diff
@@ -337,6 +326,7 @@ def _add_method_tracer_now(method_name, metric_name_code, options)
           class_eval traced_method, __FILE__, __LINE__
           alias_method _untraced_method_name(method_name, metric_name_code), method_name
           alias_method method_name, _traced_method_name(method_name, metric_name_code)
+          ruby2_keywords(_traced_method_name(method_name, metric_name_code)) if respond_to?(:ruby2_keywords, true)
```
tannalynn, line 329: What does `ruby2_keywords` do here?

test/new_relic/agent/method_tracer_params_test.rb
```diff
@@ -0,0 +1,178 @@
...
+    assert_expected_results traced_class
+    refute_deprecation_warnings traced_class
+    call_expecting_warning_after_ruby_26 traced_class
+  end
+
+  [ ["untraced_methods", UntracedMethods],
+    ["traced_methods", TracedMethods],
+    ["traced_metric_methods", TracedMetricMethods],
+    ["traced_metric_methods_unscoped", TracedMetricMethodsUnscoped],
+  ].each do |traced_class_name, traced_class, metrics|
```
tannalynn, line 127: what is `metrics` here? I'm not seeing it used below, and it seems like each will only have the first 2 things

mwlang: need to remove `metrics`. no longer using it.

lib/new_relic/agent/method_tracer.rb
mwlang: makes 2.7 behave like Ruby <= 2.6

lib/new_relic/constants.rb
```diff
@@ -28,6 +28,10 @@ module NewRelic
   TRACEPARENT_KEY = "traceparent"
   TRACESTATE_KEY = "tracestate"
 
+  # Right now, old and new Rubies are the same.  Intention is to use "(...)" for
+  # Ruby 2.7+ for argument delegation, but doing so is a breaking API change.
+  ARGS_FOR_RUBY_VERSION = RUBY_VERSION < "2.7.0" ? "*args, &block" : "*args, &block"
+
```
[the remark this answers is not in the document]
mwlang: In certain scenarios, users will still see warnings, but I'm hoping the ruby2_keywords call will help guide them to the right place in their code for further inspection/resolution.

tannalynn (approved):
LGTM, and everything is passing. Thank you!

PR #505, opened 2020-11-20
Force install exit handler flag

mwlang:
# Overview

When an application embeds Sinatra, perhaps as part of another gem such as [splitrb/split] (https://github.com/splitrb/split), Sinatra application will be detected by the Agent and the at_exit hook that is responsible for sending any unsent data to the collector server is not installed.

This is because Sinatra runs its entire app in an at_exit hook and installing another would interfere with that.

For apps that are also managed by a tool like Unicorn Killer, apps are frequently shutdown and the user ends up seeing what looks  like missing data, espcially during high load periods.

The `force_install_exit_handler` configuration option allows a user to override the agent's normal condition to skip the at_exit hook and install it anyway.


mwlang:
This feature may be installed and tested by adding the following to your Gemfile:

```ruby
gem 'newrelic_rpm', 
  :git => 'https://github.com/newrelic/newrelic-ruby-agent.git',  
  :ref => '5c6b03b9acb58b7524a174ca3ab03d6136be20c9'
```

.ruby-version
```diff
@@ -0,0 +1 @@
+2.7.2
```
tannalynn, line 1: Why is this `.ruby-version` file here? I don't think it is supposed to be

mwlang: Accidental check in.  Good catch!

tannalynn (approved):

PR #514, opened 2020-12-03
updated json gem for Ruby 3.0+

### LINKED ISSUE #489: Update String methods used in tests for ruby 3.0 compatibility
The methods `map` and `first` for[ String have been removed in ruby 3.0](https://ruby-doc.org/core-3.0.0.preview1/String.html). The tests using these methods in `newrelic-ruby-agent/test/new_relic/marshalling_test_cases.rb` will need to be updated. These tests are used by multiverse in JsonTest.

mwlang:
# Overview
Resolves #489 - Turns out json gem 2.0.4 isn't compatible with Ruby 3.0, likely because of positional vs. keyword arguments (not confirmed, but error was an "ArgumentError: given 1 where 2 were expected" error.  Testing with 2.3.1 version of the json gem resolves this.

# Related Github Issue
* #489


tannalynn (approved):
Looks great, thank you for taking care of this!

PR #525, opened 2020-12-15
509 handle failed precondition response

### LINKED ISSUE #509: Add FAILED_PRECONDITION for infinite tracing response handler
## Description
A new type of status code can be returned on the grpc connection for infinite tracing, `FAILED_PRECONDITION`. The agent needs to be able to handle this status code. In the case this status code is received, the agent must close the connection and reconnect to the new cell identified by the new resolution of the dns record.

See this section of [the spec](https://source.datanerd.us/agents/agent-specs/blob/master/Infinite-Tracing.md#failed_precondition) for detailed requirements for this.

mwlang:
# Overview
Implements handing FAILED_PRECONDITION responses from the server.

# Related Github Issue
Closes #509

### DIFF
--- a/infinite_tracing/test/support/fake_trace_observer.rb
+++ b/infinite_tracing/test/support/fake_trace_observer.rb
@@ -87,6 +87,14 @@ if NewRelic::Agent::InfiniteTracing::Config.should_load?
       end
     end
 
+    class FailedPreconditionInfiniteTracer < BaseInfiniteTracer
+      def record_span(record_spans)
+        @lock.synchronize { @noticed.signal }
+        msg = "I don't exist!"
+        raise GRPC::BadStatus.new(GRPC::Core::StatusCodes::FAILED_PRECONDITION, msg)
+      end
+    end
+
     class FakeTraceObserverServer
       attr_reader :trace_observer, :worker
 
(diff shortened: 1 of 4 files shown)


infinite_tracing/test/support/fake_trace_observer.rb
```diff
@@ -87,6 +87,20 @@ def record_span(record_spans)
...
+      def initialize
+        super
+        @count = 0
+      end
+
+      def record_span(record_spans)
+        @lock.synchronize { @noticed.signal }
+        msg = "I don't exist!"
+        @count += 1
+        raise GRPC::BadStatus.new(GRPC::Core::StatusCodes::FAILED_PRECONDITION, msg) if @count <= 5
```
tannalynn, line 100: This was the best way i could figure out how to limit how many times this goes through currently. Does this approach seem reasonable for now?

tannalynn (commented):
We should add an entry to the infinite tracing changelog for this, correct?

infinite_tracing/test/support/fake_trace_observer.rb
mwlang: This is perfect!

tannalynn (approved):

Issue #530, opened 2020-12-21
Infinite Recursion of net/http


willcosgrove:
## Description
I think I've tracked back our problem to the same one that was fixed here: #511 
We do not use the airbrake or rack-mini-profiler gems. We do however use both Scout and Honeybadger, either of which I suspect could be monkey patching net/http like airbrake or rack-mini-profiler did.

## Expected Behavior
Net/http requests would be unaffected.

## Steps to Reproduce
Run a rails app with the newrelic_rpm gem and the scout gem or the honeybadger gem.

## Your Environment
I noticed this when I upgraded to 6.14.0. I reverted back to the previous version we were using: 6.11.0.365 and the issue went away.

tannalynn:
Thank you for bringing this to our attention @willcosgrove. We've seen a few gems already that are incompatible with the updated prepend instrumentation due to monkey patching, and it seems there are more and more popping up.

For now, while this behavior is defaulted to true, you can disable this by adding this to your newrelic.yml in version 6.14.0:
```prepend_net_instrumentation: false```

I think the best solution going forward for us, for now, will be to default this option to false, so that our old instrumentation is used by default. This is will prevent unexpected compatibility issues with various gems that have not yet moved away from monkey patching.

Issue #510, opened 2020-11-27
`6.14.0` creates infinite recursion on "net request" with rack-mini-profiler present


PikachuEXE:
^^ Provide a general summary of the issue in the title above. ^^ 

## Description
We use [`fastimage`](https://rubygems.org/gems/fastimage/versions/2.2.0) to check the type of image behind the image URLs input by user before saving them.
Example code would be something like:
```ruby
FastImage.type(
  "https://duckduckgo.com/assets/logo_homepage.alt.v108.svg",
  timeout: 5,
).nil?

```

## Expected Behavior
It should work without infinite recursion (`SystemStackError: stack level too deep`)

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results
<details>

<summary>Error Backtrace</summary>

```
/usr/local/lib/ruby/gems/2.7.0/gems/addressable-2.7.0/lib/addressable/idna/pure.rb:305:in `hash'
/usr/local/lib/ruby/gems/2.7.0/gems/addressable-2.7.0/lib/addressable/idna/pure.rb:305:in `lookup_unicode_composition'
/usr/local/lib/ruby/gems/2.7.0/gems/addressable-2.7.0/lib/addressable/idna/pure.rb:215:in `unicode_compose_pair'
/usr/local/lib/ruby/gems/2.7.0/gems/addressable-2.7.0/lib/addressable/idna/pure.rb:153:in `block in unicode_compose'
/usr/local/lib/ruby/gems/2.7.0/gems/activesupport-6.0.3.4/lib/active_support/core_ext/range/each.rb:9:in `each'
/usr/local/lib/ruby/gems/2.7.0/gems/activesupport-6.0.3.4/lib/active_support/core_ext/range/each.rb:9:in `each'
/usr/local/lib/ruby/gems/2.7.0/gems/addressable-2.7.0/lib/addressable/idna/pure.rb:149:in `unicode_compose'
/usr/local/lib/ruby/gems/2.7.0/gems/addressable-2.7.0/lib/addressable/idna/pure.rb:121:in `unicode_normalize_kc'
/usr/local/lib/ruby/gems/2.7.0/gems/addressable-2.7.0/lib/addressable/uri.rb:566:in `normalize_component'
[...]

PR #531, opened 2020-12-22
Default net_prepend_instrumentation to false

tannalynn:

# Overview
We've had a few issues come through where prepend instrumentation on Net::HTTP is causing compatibility issues with gems that are still using monkey patching on the same objects. For now, we should change the default behavior back to monkey patching to prevent unexpected problems to arise for customers using gems that monkey patch. We should revisit this change when we do our major shift to prepend in our next major release. 

# Related Github Issue
https://github.com/newrelic/newrelic-ruby-agent/issues/530
https://github.com/newrelic/newrelic-ruby-agent/issues/510


mwlang (commented):
Fixed a couple typos in CHANGELOG, but everything else is solid.

mwlang (approved):

PR #532, opened 2020-12-24
Fix database adapter lookup for Rails 6.1

hasghari:
# Overview
The adapter lookup in Rails 6.1 is raising an error which is caught by the rescue block. However, the lookup in the rescue block is producing a deprecation warning that will be removed in Rails 6.2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: Continuous Integration
 
-on: 
-  push: 
+on:
+  push:
     branches:
       - main
       - dev
@@ -10,8 +10,8 @@ on:
 jobs:
   build-ruby:
     runs-on: ubuntu-latest
-    strategy: 
-      matrix: 
+    strategy:
+      matrix:
         ruby-version: [2.0.0-p648, 2.1.10, 2.2.10, 2.3.8, 2.4.10, 2.5.8, 2.6.6, 2.7.1, 3.0.0, jruby-9.2.12.0]
     steps:
       - uses: actions/checkout@v2
@@ -20,14 +20,14 @@ jobs:
         uses: ./.github/actions/build-ruby
         with:
           ruby-version: ${{ matrix.ruby-version }}
-  
+
   mini-env:
     needs: build-ruby
     runs-on: ubuntu-latest
-    strategy: 
+    strategy:
       fail-fast: false
-      matrix: 
-        rails: [norails, rails60, rails52, rails51, rails42, rails41, rails40, rails32, rails31, rails30]
+      matrix:
+        rails: [norails, rails61, rails60, rails52, rails51, rails42, rails41, rails40, rails32, rails31, rails30]
         ruby-version: [2.0.0-p648, 2.1.10, 2.2.10, 2.3.8, 2.4.10, 2.5.8, 2.6.6, 2.7.1, 3.0.0, jruby-9.2.12.0]
         exclude:
           - ruby-version: "3.0.0"
@@ -50,6 +50,8 @@ jobs:
             rails: rails52
           - ruby-version: "3.0.0"
             rails: rails60
+          - ruby-version: "3.0.0"
+            rails: rails61
           - ruby-version: "2.7.1"
             rails: rails30
           - ruby-version: "2.7.1"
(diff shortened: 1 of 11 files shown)


mwlang:
This is interesting and thank you for raising this PR, @hasghari.  

I'm concerned as this is the third time this block of code has been updated and PR'd in the last few months, which means we may be doing something entirely wrong here (not calling your solution here wrong at all -- I just don't know on the surface of things and want to investigate deeper).

Also in reviewing this code, I want to extract this out into its own class/module just like the other environment reports above AR's.  The goal here is to correctly handle ActiveRecord 3.x all the way through 6.2 and beyond w/o warnings, deprecations, or resorting to a rescue block (which in hindsight is a bit hacky on my part) when I initially introduced this for 6.1 support 9 months ago.

lib/new_relic/agent/database_adapter.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      attr_reader :env, :version
+
+      def initialize(env, version)
+        @env = env
+        @version = version
+      end
+
+      def name
+        [Rails61, Rails40, Rails30, Base]
```
hasghari, line 68: The idea here is that these classes define the boundaries where a breaking change was made in the adapter lookup implementation in ActiveRecord. For example, since there was no change in implementation between Rails 4.0 and Rails 6.0, they can all use the implementation provided by `Rails40`

.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     strategy: 
       fail-fast: false
       matrix: 
-        rails: [norails, rails60, rails52, rails51, rails42, rails41, rails40, rails32, rails31, rails30]
+        rails: [norails, rails61, rails60, rails52, rails51, rails42, rails41, rails40, rails32, rails31, rails30]
```
hasghari, line 30: I added a test environment for Rails 6.1 which actually exposed the existing bug.

hasghari (commented):
@mwlang I took a stab at extracting the adapter lookup to its own class. Open to your feedback.

test/environments/rails51/Gemfile
```diff
@@ -11,6 +11,8 @@ gem 'rack-test'
 gem 'sprockets', '3.7.2'
 
 platforms :jruby do
+  gem "activerecord-jdbcmysql-adapter", "~> 51.0"
```
hasghari, line 14: Database connection cannot be established if these gems are missing.

lib/new_relic/agent/database_adapter.rb
```diff
@@ -0,0 +1,33 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+
+module NewRelic
+  module Agent
+    class DatabaseAdapter
+      VERSIONS = {
```
hasghari, line 8: The idea here is that this hash defines the boundaries where a breaking change was made in the adapter lookup implementation in ActiveRecord. For example, since there was no change in implementation between Rails 4.0 and Rails 6.0, they can all use the implementation provided by `'4.0.0'`

mwlang:
@hasghari This is a great PR and perfectly in line with what I had in mind with extracting to separate class.  Even better adding the Rails 6.1 suite to exercise the code properly in automated testing.  Well done!

Can you rebase on the current `dev` branch?  I just looked at a few of the failures and at least two fails are resolved fairly recently with their fixes merged to the `dev` branch.

hasghari:
> @hasghari This is a great PR and perfectly in line with what I had in mind with extracting to separate class. Even better adding the Rails 6.1 suite to exercise the code properly in automated testing. Well done!
> 
> Can you rebase on the current `dev` branch? I just looked at a few of the failures and at least two fails are resolved fairly recently with their fixes merged to the `dev` branch.

@mwlang My branch is already rebased against the latest in dev. With respect to the `mini-env (norails, jruby-9.2.12.0)` failure, I can attest to that being intermittent. I recall seeing that error once when I was testing locally but can't reproduce it consistently.

mwlang:
Thanks for confirming the PR is up to date with `dev`. I subsequently found that there's a consistently failing test in the `dev` branch.  #535 that's blocking this PR's tests from completing successfully.  Once I get this new Issue merged, will circle back to this PR.

mwlang:
@hasghari I was trying to see if I can do a rebase of dev to your fork/PR here, but ended up with a separate branch (first time I ever tried such with Github!).  Reading deeper, it looks like I would need to ask for you to grant me commit access to your repo.  Would you be willing to do a quick rebase on current `dev` and we carry this PR forward.  All Ruby 3.0 failures should be resolved and merged to `dev` at this point.  The [following branch](https://github.com/newrelic/newrelic-ruby-agent/tree/hasghari-fix-db-adapter-lookup) may or may not be useful to you, but does resolve the conflicts on the edits to the ci.yml for reference.

hasghari:
@mwlang No problem. I've rebased my branch against `dev`

mwlang (approved):
LGTM

PR #534, opened 2021-01-05
Update CI to run on ruby 3.0.0

tannalynn:

# Overview
Removed ruby 3.0.0-preview2 and replaced it with just 3.0.0 for our CI.


mwlang:
Need to resolve #536 as part of this PR

mwlang (commented):
I went through all these changes and everything seems solid here.  One comment to make about the changes to multiverse where an Envfile's conditionals end up resulting in no Gemfile.X outputs for a given Ruby version -- that's only partially implemented at this point, but is a good start.  Over time, as conditions arise locally where we can continue to incrementally improve this, let's do so.  For now, exclusion clauses in the workflow files are guarding against invalid scenarios running, so it's not problematic to keep these incremental improvements to multiverse.

mwlang (approved):
@tannalynn and I both reviewed and feel the changes in this PR are solid.  Approving and merging.

PR #539, opened 2021-01-07
Ruby 3 has removed :global_method_state from VM stats

hasghari:
This change fixes the 2 test suites that are currently failing in `dev` for `3.0.0-preview2`.

https://github.com/ruby/ruby/commit/278450de803c59137fff69c4db03cd40c5b5d08a


mwlang (approved):
Nice catch!  Ruby did indeed drop or renamed these parameters.  
Changing target branch to `update_to_ruby3` since this is active right now and focused on Ruby 3 specific fixes.

PR #549, opened 2021-01-27
Remove reference to non-existing create_distributed_trace_headers method

koleksiuk:
# Overview
We did implement custom version of distributed tracing in Sidekiq when it was not available back in time. In the logs, after upgrading newrelic_rpm gem, I've noticed following:

```
[2021-01-26 13:17:16 +0000 host.local (44425)] WARN : The method create_distributed_trace_payload is deprecated.
[2021-01-26 13:17:16 +0000 host.local (44425)] WARN : Please use create_distributed_trace_headers instead.
```

Problem is, `create_distributed_trace_headers` never existed. What we should use is `insert_distributed_trace_headers`.

# Testing
It's not required, no code changes.

### DIFF
--- a/lib/new_relic/agent/distributed_tracing.rb
+++ b/lib/new_relic/agent/distributed_tracing.rb
@@ -36,10 +36,10 @@ module NewRelic
       #
       # @api public
       #
-      # @deprecated See {#create_distributed_trace_headers} instead.
+      # @deprecated See {#insert_distributed_trace_headers} instead.
       #
       def create_distributed_trace_payload
-        Deprecator.deprecate :create_distributed_trace_payload, :create_distributed_trace_headers
+        Deprecator.deprecate :create_distributed_trace_payload, :insert_distributed_trace_headers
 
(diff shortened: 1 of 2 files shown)


tannalynn (approved):
Hello @koleksiuk.
Thank you for bringing this to our attention! That's a good eye, you're correct that that method does not actually exist. I suspect this was a mixup due to the how similarly named these methods all are.

As part of our current efforts for our next release, these methods are being removed completely. [Here is a link ](https://github.com/newrelic/newrelic-ruby-agent/pull/544/files#diff-bd1447fbf499088524b9570c6f1ffa6c5f3c4ca215de9cd528a02c3a043fd24f) to the changes made here in our current efforts.

However, I think it's a great idea for us to go ahead and merge this still to correct this mistake before we actually remove these methods. In the case this question gets raised in the future from users on these older versions, we'll be able to see this change easily and notice that error. 
Thank you for taking the time to correct this!

PR #565, opened 2021-02-05
Prepend instrumentation

tannalynn:
# Overview

The agent instruments many libraries using the method chaining (a.k.a. method aliasing) approach to monkey-patching.  This is a form of monkey-patching was deprecated with Ruby's introduction of `:prepend` in Ruby version 2.0.0 as a means to insert additional functionality into a module's method lookup mechanism ahead of the library's original method implementation.  

Prepend strategy will be the default choice in 7.0+.  However, the prepend strategy may be reverted back to method chaining by adding the following (sidekiq example) configuration flag:

This PR collects all auto-instrumentation prepend implementation efforts and should be merged when all related work is completed and merged to this PR's branch.  Each individual instrumentation issue should have its own PR against this branch.

```YAML
instrumentation:
  sidekiq: chain
```
> NOTE: The use case for this is for when an unknown gem is found to be conflicting.  The user is able to revert to method-chaining to deal with the conflict until the agent can be updated to auto-detect and handle the conflict. 

To disable instrumentation altogether:

```YAML
instrumentation:
  sidekiq: disable
```

In some cases, we may know specific gems conflict with prepend.  To facilitate, we offer an `auto` config option (which is the default) which will automatically degrade to `chain` in such cases.

The default setting in most cases is thus:

```YAML
instrumentation:
  sidekiq: auto
```

It is possible to force using `prepend` strategy by specifying it in the config like so:
```YAML
instrumentation:
  sidekiq: prepend
```
> NOTE: The use case for this is for when a newer version of the conflicting gem is released and it's known to no longer conflict with `prepend` strategy.

### Additional Context

# Related Github Issue


mwlang (commented):
Review Checklist:

* [ ] DRY implementations of Chain and Prepend
* [ ] new instrumentation configuration attribute: (i.e. `:'instrumentation.foo``)
* [ ] all replaced instrumentation configuration attributes have `deprecated: true` added.
* [ ] all replaced instrumentation configuration descriptions refer to new instrumentation attribute.
* [ ] any `:public => :true` have been corrected with `:public => true`
* [ ] `:disable_foo` is not referenced in dependency detection nor anywhere else (the logic to disable is already wired!)
* [ ] file naming is consistent! (i.e. instrumentation, chain, prepend, helpers)

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1365,89 +1567,109 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :description => "Controls auto-instrumentation of Excon at start up.  May be one of [enabled|disabled]."
         },
         :disable_httpclient => {
           :default      => false,
           :public       => true,
           :type         => Boolean,
           :dynamic_name => true,
           :allowed_from_server => false,
-          :description  => 'If <code>true</code>, disables instrumentation for the httpclient gem.'
+          :description  => deprecated_description(:'instrumentation.httpclient', 'If <code>true</code>, disables instrumentation for the httpclient gem.')
```
tannalynn, line 1607: missing deprecated tag

lib/new_relic/agent/instrumentation/net_http/chain.rb
```diff
@@ -0,0 +1,45 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+
+module NewRelic::Agent::Instrumentation
+  module NetHTTP
+    module Chain
+      def self.instrument!
+        Net::HTTP.class_eval do
```
tannalynn, line 9: DRY this one up

mwlang (approved):
Everything looks good!

PR #574, opened 2021-02-17
Support current Rails edge

casperisfine:
Rails recently bumped its version number, which breaks the NewRelic agent.

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -101,7 +101,8 @@ module NewRelic
               when 4..6
                 :rails_notifications
               else
-                ::NewRelic::Agent.logger.error "Detected unsupported Rails version #{Rails::VERSION::STRING}"
+                ::NewRelic::Agent.logger.warn "Detected untested Rails version #{Rails::VERSION::STRING}"
+                :rails_notifications
               end
             when defined?(::Sinatra) && defined?(::Sinatra::Base) then :sinatra
             when defined?(::NewRelic::IA) then :external


mwlang:
@casperisfine, Thank you for your PR.  It's hard to believe Rails is about to turn the 7.0 corner.

Before we can merge this, there are a couple of more things to check into:

1. Do you have any insight into what API changes are coming for Rails edge?  The most important of those is the Rails [subscriber/notification API](https://guides.rubyonrails.org/active_support_instrumentation.html) and any signature changes in that plumbing.  If this is not something documented somewhere in the Rails community presently, then I'd definitely want to inject a logged message somewhere noting that Rails 7 is experimental so all users are aware not to just expect Rails edge to "just work" even though my suspicion is that it likely will, given Rails' maturity these days.

2. All tests are passing, but Rails Edge wasn't covered, so this is a false sense of security at this stage.  Would you have the bandwidth to add the latest `rails70` equivalent to the [test/environments folder](https://github.com/newrelic/newrelic-ruby-agent/tree/dev/test/environments).  If you can, then we can move this PR along much faster.  If not, no worries, but please do make sure maintainers have privilege to add to your fork so we can prepare this PR for merging.

casperisfine:
> It's hard to believe Rails is about to turn the 7.0 corner.

Well, it's still months down the road. It's just that we're running rails-edge on a few apps, and they recently bumped the version number.

> Do you have any insight into what API changes are coming for Rails edge?

I'm afraid I can't predict the future. So far latest new_relic works worth Rails edge.

> All tests are passing, but Rails Edge wasn't covered, so this is a false sense of security at this stage.

I'll take a look. But really my goal is just to allow it to run, edge being a moving target it's better to allow people to test rather than to hard break and prevent them from discovering and reporting errors to you.

If anything I'd suggest to simply check for `>= 4` rather than `4..7`.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -101,7 +101,8 @@ def self.framework
               when 4..6
                 :rails_notifications
               else
-                ::NewRelic::Agent.logger.error "Detected unsupported Rails version #{Rails::VERSION::STRING}"
+                ::NewRelic::Agent.logger.warning "Detected untested Rails version #{Rails::VERSION::STRING}"
+                :rails_notifications
```
casperisfine, line 105: So in the end I think this is the best way to go. Simply log a warning if Rails is newer than either of the tested versions.

mwlang:
I didn't think about it earlier, but we don't want failing Rails 7 tests to block any PRs for now.  Your suggested change is a good compromise for now and then we can officially support Rails 7 when they start putting out the release candidates.

lib/new_relic/agent/configuration/default_source.rb
mwlang: I agree.

mwlang (approved):
A simple change that warns users for upcoming major Rails releases rather than not setting at all.  Allows early adopters of Rails Edge to use and test the Agent in their setting.

mwlang:
@casperisfine, excellent work on this PR.  Could you please sign the CLA?  This is the only blocker to accepting this PR and merging.

casperisfine:
@mwlang done. It was confused by my double user thingy. I had to sign it as @byroot as well.

mwlang (approved):
The CI fails in this PR are unrelated to the changes made.  All tests are passing where concerned and this PR does allow Rails edge to be instrumented.

PR #575, opened 2021-02-19
Skip constructing Time for transactions

viraptor:

# Overview
The `time` parameter was never used, but each call constructed the current time
to fill out the default value.


viraptor:
I see the same `infinite_tracing` tests failing on `dev` branch, so they don't seem specific to this PR.

mwlang:
@viraptor -- thank you for this PR -- Looks like `google-protobuf` released 3.15.0 with a change that breaks the infinite tracing tests.  We'll open a separate ticket for that. 

Meanwhile, very nice catch on finding this unused parameter.  I would be in complete favor of removing the parameter altogether and removing it where given in the test suite.  

```
lib/new_relic/agent/sql_sampler.rb:
  67        # This is called when we are done with the transaction.
  68:       def on_finishing_transaction(state, name, time=Time.now)
  69          return unless enabled?

test/new_relic/agent/pipe_channel_manager_test.rb:
  223        sampler.notice_sql("SELECT * FROM table", "ActiveRecord/Widgets/find", nil, 100, state)
  224:       sampler.on_finishing_transaction(state, 'noodles', Time.now)
  225      end

test/new_relic/agent/sql_sampler_test.rb:
  440  
  441:   def test_on_finishing_transaction_with_busted_transaction_state_does_not_crash
  442      state = NewRelic::Agent::Tracer.state
  443:     @sampler.on_finishing_transaction(state, "whatever", Time.now)
  444    end

test/new_relic/agent/transaction_sampler_test.rb:
  445        end
  446:       @sampler.on_finishing_transaction(@state, @txn, Time.now.to_f)
  447      end

  460        @sampler.notice_pop_frame(state, "a")
  461:       @sampler.on_finishing_transaction(state, @txn, (stop || Time.now.to_f))
  462      end

```

viraptor:
I split it in two - removing the use in rest of the code, and removing the param, just in case you want to keep the API.

mwlang (approved):
Looks good to me and we're now saving on not instantiating an unused Time.now call.

The CI fails are not due to this particular change, so approving and merging.

PR #578, opened 2021-02-19
Prepend grape

mwlang:
Implements Prepend method of auto-instrumenting the Grape library.

* Reduced number of versions tested from 26 environments to 6.
* DRY implementation between chain and prepend patching.


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -15,6 +15,35 @@ def self.value_of(key)
...
+      #    if :disable_net_http is true, then returned value is "disabled"
+      #    if :prepend_net_instrumentation is false, then returned value is "chain"
+      #    otherwise, "auto" is returned.
+      #
+      # Intent is: 
+      #     - if user sets disable_xxx property, then don't instrument
+      #     - if user set prepend to `false` then we use method_alias chaining
+      #     - auto, when returned means, try to use prepend unless conflicting gems discovered
+      #
+      def self.instrumentation_value_of(disable_key, prepend_key=nil)
```
tannalynn, line 30: hmmm I'm confused about this diff, because the target branch, prepend_instrumentation, has this change on it already.

lib/new_relic/agent/instrumentation/grape.rb
```diff
@@ -2,90 +2,23 @@
...
 DependencyDetection.defer do
   # Why not just :grape? newrelic-grape used that name already, and while we're
   # not shipping yet, overloading the name interferes with the plugin.
   named :grape_instrumentation
-
+  configure_with :grape
+  
   depends_on do
     ::NewRelic::Agent.config[:disable_grape] == false
   end
```
tannalynn, line 17: We should remove this check on the old config here

tannalynn (approved):
After discussing this, looks good. The failure that is showing is unrelated to this PR(due to rails issues with mimemagic), so I'll be merging this even with that failure.

PR #579, opened 2021-02-19
Remove unused functions in test

viraptor:

# Overview
While looking at https://github.com/newrelic/newrelic-ruby-agent/pull/575 I noticed the two helper functions are never used anymore.


mwlang (approved):
I really love it when we can remove unused code, especially code like this that adds a lot of mental load just to figure out what it is, let alone if it's being utilized.  Thank you for your contribution.  I reviewed our code base and these methods are indeed entirely unused.

PR #605, opened 2021-02-26
Deprecate legacy excon instrumentation

### LINKED ISSUE #582: Implement prepend instrumentation for Excon
As part of implementing prepend auto-instrumentation for all libraries instrumented, introduce a prepend version of auto-instrumenting for the [Excon gem](https://rubygems.org/gems/excon)
### Feature Description
The agent instruments many libraries using the method chaining (a.k.a. method aliasing) approach to monkey-patching.  This is a form of monkey-patching was deprecated with Ruby's introduction of `:prepend` as a means to insert additional functionality into a module's method lookup mechanism ahead of the library's original method implementation.  
Prepend strategy will be the default choice in 7.0+.  However, the prepend strategy may be reverted back to method chaining by adding the following configuration flag:
```YAML
instrumentation:
  excon: chain
```
All configuration options are [described in-depth here](https://github.com/newrelic/newrelic-ruby-agent/pull/565):
### Additional Context

tannalynn:

# Overview
We discovered prepend was not needed for excon instrumentation. Very old excon versions require method chaining, but we are removing these old versions from support. This PR deprecates the legacy method chaining instrumentation for these older unsupported versions.

# Related Github Issue
closes #582


mwlang (changes requested):
Is not conforming to our new configuration setup.  Still references the :disable_excon for example.  We need to honor 
```yaml
instrumentation:
  excon: disabled
```
To disable.  Since we're auto-instrumenting via adding to excon's middleware stack, I suggest, "auto" is equivalently the same to "enabled". If "chain" or "prepend" are encountered, we effectively treat this as "enabled" and produced to install middleware accordingly.  (i.e. there's no difference in what happens between these options).  We may want to log a debug message saying as much.)

tannalynn:
> Is not conforming to our new configuration setup. Still references the :disable_excon for example. We need to honor
> 
> ```yaml
> instrumentation:
>   excon: disabled
> ```
> 
> To disable. Since we're auto-instrumenting via adding to excon's middleware stack, I suggest, "auto" is equivalently the same to "enabled". If "chain" or "prepend" are encountered, we effectively treat this as "enabled" and produced to install middleware accordingly. (i.e. there's no difference in what happens between these options). We may want to log a debug message saying as much.)

Alright, I added a check for the allowed_by_config? method for excon, which should check if it's enabled. I didn't make any other changes since we don't need to check if it's auto/chain/prepend since the behavior is the same for all of those anyways, I believe.

mwlang (changes requested):
There are not changes to the default_source to add `:'instrumentation.excon'` so it can be enabled/disabled in the new standard way.  Here, `prepend`, `chain` will stand to mean `enabled` and doesn't actually toggle between a prepend or chain method of instrumenting.  My suggestion (and we can discuss this more as a team): just list `[auto|enabled|disabled]` as choices in the description.

lib/new_relic/agent/instrumentation/excon.rb
```diff
@@ -65,7 +65,8 @@ def install_middleware_excon_instrumentation
   end
 
   def install_legacy_excon_instrumentation
-    ::NewRelic::Agent.logger.info 'Installing legacy Excon instrumentation'
+    Deprecator.deprecate :install_legacy_excon_instrumentation, :install_middleware_excon_instrumentation, "8.0.0"
```
mwlang, line 68: I'm not sure how this is working!  I just discovered the same on the prepend_mongo branch required the full namespace: `NewRelic::Agent::Deprecator.deprecate(...)`

mwlang (changes requested):

mwlang (approved):
This looks good.  A couple of failing jobs in CI completely unrelated to Excon, so moving forward with merging.

PR #608, opened 2021-03-03
install multiarch-support when downgrading MySQL client libs

mwlang:
Resolves #607

PR #609, opened 2021-03-03
updating npm dependencies; fixing older mysqlclient installs

mwlang:
Resolves #608 
* Updates all nodeJS packages on the GHA scripts
* Adds multiarch Debian package that older MySQL clients depend on.


mwlang:
We're opting to merge this directly in the main branch since old (and thus broken!) versions of the CI workflow scripts are running and these are known to be broken.

tannalynn (approved):
Thanks for these fixes for our CI!

PR #613, opened 2021-03-09
Implement prepend instrumentation for redis

tannalynn:

# Overview
Updated Redis instrumentation to have both prepend and chain options hooked up and defaults to prepend.

# Related Github Issue
#591


lib/new_relic/agent/instrumentation/redis.rb
```diff
@@ -5,43 +5,17 @@
...
-
-  depends_on do
-    defined? ::Redis::VERSION
-  end
-
+  configure_with :redis
+  
   depends_on do
-    NewRelic::Agent.config[:disable_redis] == false
+    defined?(::Redis) && defined?(::Redis::VERSION)
```
tannalynn, line 18: Shouldn't we also still be checking for the config option being disabled for redis?

test/helpers/config_scanning.rb
```diff
@@ -9,7 +9,7 @@ module ConfigScanning
       # NewRelic::Agent.config into classes as long as we call the variable config
       AGENT_CONFIG_PATTERN       = /config\[:['"]?([a-z\._]+)['"]?\s*\]/
       DEFAULT_VALUE_OF_PATTERN   = /:default\s*=>\s*value_of\(:['"]?([a-z\._]+)['"]?\)\s*/
-      DEFAULT_INST_VALUE_OF_PATTERN   = /:default\s*=>\s*instrumentation_value_of\(:['"]?([a-z\._]+)['"]?,?\s?:?['"]?([a-z\._]+)['"]?\)/
+      DEFAULT_INST_VALUE_OF_PATTERN   = /:default\s*=>\s*instrumentation_value_of\(:['"]?([a-z._]+)['"]?\)|\(:['"]?([a-z._]+)['"]?\s*,\s*:['"]?([a-z._]+)['"]?\)/
```
tannalynn, line 12: What is this change doing?

mwlang: It's picking up both parameters correctly.  We can have:
```ruby
instrumentation_value_of(:attribute1)
```
or 
```ruby
instrumentation_value_of(:attribute1, :attribute2)
```
Previously, we were not only getting just one of the attributes, it was actually splitting a single value into two values, like `:attribute` and `:1` which was the bug that led to me realizing I coded that Regex incorrectly to begin with.

lib/new_relic/agent/instrumentation/redis.rb
mwlang: It's checked before the execute block is called.  In dependency_detection.rb...

```ruby
    def allowed_by_config?
      !(disabled_configured? || deprecated_disabled_configured?)
    end

    def check_dependencies
      return false unless allowed_by_config? && dependencies

      dependencies.all? do |dep|
        begin
          dep.call
        rescue => err
          NewRelic::Agent.logger.error( "Error while detecting #{self.name}:", err )
          false
        end
      end
    end
```

tannalynn: OH! I thought we were supposed to be _replacing_ things like `NewRelic::Agent.config[:disable_redis] == false` in the depends_on block with allowed_by_config?, I didn't realize the new way was hooked up to a different place.

test/helpers/config_scanning.rb
tannalynn: Okay cool, that makes sense. thank you for explaining that!

lib/new_relic/agent/instrumentation/redis.rb
mwlang: We're definitely supposed to reference the deprecated `:disable_foo` when we introduce the new `:instrumentation.foo` setting by setting the new attribute's default to `instrumentation_value_of(:disable_foo, ...)`. But otherwise, yes, I wired it into the Dependency Detection so that all instrumentation followed the same rules for enabling, disabling, and toggling prepend and method-chaining.

PR #621, opened 2021-03-17
Dry prepend httpclient

mwlang:
# Overview
Provides a DRY implementation shared between CHAIN and PREPEND implementations


tannalynn (approved):
Looks good to me!

PR #624, opened 2021-03-18
Fix required ruby version for infinite tracing

tannalynn:

# Overview
The gemspec for infinite tracing had the required minimun ruby version as 2.3, however the correct minimum ruby version is 2.5. Our CI has already been reflecting that we support 2.5+, so this change is just to our gemspec to more accurately reflect the ruby versions infinite tracing is expected to run on.


mwlang (approved):
Good catch!

mwlang:
We have internally confirmed that gRPC is not working correctly with Ruby 2.4.9, so syncing the gemspec with known minimum working version is solid.

PR #625, opened 2021-03-19
Update ci.yml

angelatan2:
Move nightly from 1 am to 2 am PDT (5 AM EST).


angelatan2:
I've merged this change of time to 1 hour later.  Both devs are in after-hours now.  Putting more weight into the ability to run over the weekend at the new time and seeing the results.

PR #636, opened 2021-03-31
prepend implemented for memcache

### LINKED ISSUE #585: Implement prepend instrumentation for memcached
As part of implementing prepend auto-instrumentation for all libraries instrumented, introduce a prepend version of auto-instrumenting for the [Memcached gem](https://rubygems.org/gems/memcached)
### Feature Description
The agent instruments many libraries using the method chaining (a.k.a. method aliasing) approach to monkey-patching.  This is a form of monkey-patching was deprecated with Ruby's introduction of `:prepend` as a means to insert additional functionality into a module's method lookup mechanism ahead of the library's original method implementation.  
Prepend strategy will be the default choice in 7.0+.  However, the prepend strategy may be reverted back to method chaining by adding the following configuration flag:
#### For `dalli` gem
```YAML
instrumentation:
  memchache: chain
```
or
#### for `memcache-client` gem
```YAML
instrumentation:
  memchache-client: chain
```
or
#### for `memcached` gem
```YAML
instrumentation:
  memchached: chain
```
All configuration options are [described in-depth here](https://github.com/newrelic/newrelic-ruby-agent/pull/565):
### Additional Context

mwlang:
Implements prepend and chain auto-instrumentation for `memcached`, `memcache-client`, and 'dalli` MemCache gems.

* Moves all dependency detection blocks to the main instrumentation file.
* Moves tracing logic/wrappers to `NewRelic::Instrumentation::Memcache::Tracer`
* Helper class to share code needed in preparing auto-instrumenting
* newrelic.yml introduced to control prepend vs chain in tests

Closes #585


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -986,22 +1010,25 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
         :disable_dalli => {
           :default => value_of(:disable_memcache_instrumentation),
           :public => true,
           :type => Boolean,
+          :deprecated => true,
           :allowed_from_server => false,
-          :description => 'If <code>true</code>, disables instrumentation for the dalli gem.'
+          :description => deprecated_description(:'instrumentation.memcache', 'If <code>true</code>, disables instrumentation for the dalli gem.')
         },
         :disable_dalli_cas_client => {
```
tannalynn, line 1033: Should we be updating this config value as well?

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -986,22 +1010,25 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :description => deprecated_description(:'instrumentation.memcache-client', 'If <code>true</code>, disables instrumentation for the memcache-client gem.')
         },
         :disable_dalli => {
           :default => value_of(:disable_memcache_instrumentation),
           :public => true,
           :type => Boolean,
+          :deprecated => true,
           :allowed_from_server => false,
-          :description => 'If <code>true</code>, disables instrumentation for the dalli gem.'
+          :description => deprecated_description(:'instrumentation.memcache', 'If <code>true</code>, disables instrumentation for the dalli gem.')
```
tannalynn, line 1031: How come this one is set up to just use instrumentation.memcache instead of a new instrumentation.dalli?

mwlang: Design decision.  Nothing was using memcache anyway, so I figured since Dalli is overwhelmingly used, just make memcache drive Dalli.

lib/new_relic/agent/configuration/default_source.rb
mwlang: sorry, I didn't quite understand the question since that description is updated...or is it that I forgot the deprecated => true bit and, if so, then yes, we need that.

tannalynn: ooh yeah I meant the `disable_dalli_cas_client` since it didn't have any changes on it

lib/new_relic/agent/configuration/default_source.rb
tannalynn: Makes sense to me! Just wanted to make sure we weren't missing something

tannalynn (approved):
LGTM!

PR #645, opened 2021-04-08
Fix passing keyword arguments in ruby 3.0

knapo:
# Overview
Passing keyword arguments to `middleware_class` in MiddlewareProxy is broken in ruby 3.0. It's fine, when middleware doesn't require `keyword arguments. Bu when it doesn't they are passed as an hash argument.

Here is an example issue from with honeycomb integration:

Screenshot 2021-04-08 at 11 22 57

# Testing
exercise your changes.

### DIFF
--- a/lib/new_relic/agent/instrumentation/middleware_proxy.rb
+++ b/lib/new_relic/agent/instrumentation/middleware_proxy.rb
@@ -32,7 +32,7 @@ module NewRelic
             MiddlewareProxy.wrap(middleware_instance)
           end
 
-          # ruby2_keywords(:new) if respond_to?(:ruby2_keywords, true)
+          ruby2_keywords(:new) if respond_to?(:ruby2_keywords, true)
         end
 
         def self.is_sinatra_app?(target)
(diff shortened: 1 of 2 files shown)


mwlang:
@knapo -- Thank you for this PR!  Unfortunately, this particular approach breaks our support for earlier Rubies, which includes Ruby 2.2 through 3.0 as well as jRuby.  I believe the right way to solve this is to keep the original `def new(*args, &blk)` signature but add a call to ruby2_keywords like this:

```Ruby
def new(*args, &blk)
  middleware_instance = @middleware_class.new(*args, &blk)
  MiddlewareProxy.wrap(middleware_instance)
end
ruby2_keywords(:new) if respond_to?(:ruby2_keywords, true)
```

Here's a more [complete extraction](https://gist.github.com/mwlang/49bf3b55d4d7eefb30d2a929726075b0) of the delegation problem we're solving here.  Lastly, to complete the PR, the middleware_proxy_test.rb needs some scenarios to cover Rack app examples that are defined using keyword arguments.  If you're up for seeing this through, we'll definitely support your effort as needed.  If not, we'll open an issue to track this and address in an upcoming release cycle.

knapo:
@mwlang Yes, I realized that seeing the build failing and wasn't aware about so many ruby versions support. Thanks for the explanation.

knapo:
I added a test covering such scenario, which got passed having `ruby2_keywords` thingie you recommended.

tannalynn (approved):
Looks good to me, thank you for making this change, @knapo!

Issue #561, opened 2021-02-04
missing postgres stats after upgrade to 6.15.0


gsar:
## Description

The Databases area is no longer showing any stats for postgres queries since upgrading from 6.14.0 to 6.15.0. We also upgraded to rails 6.1.1 from rails 6.0.3.4 at the same time.  Everything went silent after the Feb 3rd deploy that introduced both upgrades.

![image](https://user-images.githubusercontent.com/6981812/106953842-bbd97000-66e7-11eb-8c6d-331bb1f4a903.png)

I also noticed that all database queries look different, the table names are missing, screenshot comparison of a typical query below:

![image](https://user-images.githubusercontent.com/6981812/106954143-3904e500-66e8-11eb-9b2a-83b7658a7ab7.png)

vs

![image](https://user-images.githubusercontent.com/6981812/106954267-6356a280-66e8-11eb-955a-86dc5b90624d.png)

The changelog for 6.15.0 doesn't mention any activerecord configuration changes, hope I'm not missing a configuration step.

## Expected Behavior

Postgres stats should be shown.

## Your Environment

ruby 2.7.2
rails 6.1.1
newrelic_rpm 6.15.0

tannalynn:
Hello, @gsar 
Thank you for bringing this to our attention.
Looking at this, I suspect this may be related to the rails upgrade to 6.1. 
We added rails 6.1 to our test suite and everything is working as expected there, however it's possible that some things maybe changed in 6.1 that are not covered by our current test cases. We will need to spend some time digging deeper on 6.1 and experiment with it to confirm our test cases are accurately reflecting how activerecord 6.1 is working. 

I do have a couple questions about how you guys are using 6.1, to help guide us in our investigations. 
Are you setting `legacy_connection_handling` to false?
Are you utilizing multiple connections?
Any other changes or new features of rails that you guys have started using?
Thank would be very helpful for us to track this down.

Apologies for the delayed response, thank you for your patience.

motionless:
We have the same observation/problem after upgrading to Rails 6.1

Environment:
Rails version:	6.1.3
NewRelic gem: 6.1.5
Ruby version:	2.7.1

legacy_connection_handling is not set.

We use multiple connections trough - ActiveRecord::Base.connected_to for a few specific queries. The behavior is consistent throughout the whole application.

The queries are showing up as `ActiveRecord` instead of `Postgres`.
This wouldn't be a big deal in the long run, even though it is inconvenient at the moment.

[...]

PR #665, opened 2021-05-07
Use `connection` from sql.active_record payload over `connection_id`

amhuntsman:
As of Rails 6.1, ActiveRecord instrumentation does not provide `connection_id` in the payload for "sql.active_record". The agent currently expects this key to be there; in 6.1, this results in no connection config being fetched, and the default vendor falls back to "ActiveRecord" rather than the appropriate database name - other side effects may also be occurring.

The nature of this fix should ensure no behavior changes for the agent in existing Rails installations. `connection` will be used if present; otherwise it will fall back to `connection_id` (the previous behavior).

Also updated the unit test `test_active_record_config_for_event` to cover both pre- and post-Rails 6.1. This test was not working as intended and has been branched into two new tests.

Fixes https://github.com/newrelic/newrelic-ruby-agent/issues/561

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,15 @@
 
   ## v7.1.0
 
-
   * **Update known conflicts with use of Module#Prepend**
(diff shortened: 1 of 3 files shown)


tannalynn (approved):
Looks good, we talked about it in pairing also, and this appears to be a straight forward change that will keep previous rails versions working the same way as before, while allowing it to also work in rails 6.1

tannalynn (changes requested):
Oh actually, I almost forgot, but if you could add a changelog entry for this under 7.1.0, that would be great.

angelatan2 (approved):
Reviewed.  Changelog looks good.

Issue #401, opened 2020-08-13
Update yajl-ruby instrumentation


rachelleahklein:
The Ruby agent is known to support up to version 1.3.x of the yajl-ruby gem. We would love your help in making sure our instrumentation works with the latest version.

For guidance on how to proceed, see our [Guide to Updating Instrumentation](https://github.com/newrelic/newrelic-ruby-agent/wiki/Guide-to-Updating-Instrumentation), and please let us know if you have any questions not covered there.

creaturenex:
Hello,

This is the first time I am contributing to open source. I wanted to see if I could update the envfile to test the new version of "yajl-ruby" v 1.4.1. When following the directions [Guide to Updating Instrumentation](https://github.com/newrelic/newrelic-ruby-agent/wiki/Guide-to-Updating-Instrumentation) the first step says you can test your changes using `rake:test multiverse[name_of_suite]`. This command is not working on my end I wasn't sure if this was a typo. I can get `rake test:multiverse` to work and it run all the test suites. When I try to run `rake test:multiverse[yajl] or rake:test multiverse[yajl]` it get the following error `zsh: no matches found: test:multiverse[yajl]`.

I wanted to know if there is something on my end that I am doing incorrectly. I am testing on a linux machine, elementary os, unbuntu based, using ohmyzsh, with a hyper terminal and using RVM. Let me know if you need additional information.

amhuntsman:
@creaturenex Hello, and thank you for your contribution!

If you're using zsh, you'll need to escape the square brackets in that command:

    rake test:multiverse\[yajl\]

creaturenex:
@amhuntsman Thanks I did not know I had to do that! lesson learned.
I hate to bother but now I am getting, I ran `rake test:multiverse\[yajl\]` & `bundle exec rake test:multiverse\[yajl\]`, RVM is current in Ruby 3.0.0.
```ruby
SKIPPED "/home/oscar/Projects/Github_repos/newrelic-ruby-agent/test/multiverse/suites/yajl"
: Yajl not supported for JRuby or 2.4
```
I see the Envfile in  `newrelic-ruby-agent/test/multiverse/suites/yajl` has the statement. I assumed maybe incorrectly that the test would load the required gems. 
```ruby
suite_condition("Yajl not supported for JRuby or 2.4") do
  RUBY_PLATFORM != 'java' && RUBY_VERSION < '2.4.0'
end
```

amhuntsman:
[...]

PR #669, opened 2021-05-11
#401

creaturenex:

# Overview
I was able to get a passing test using `rake test:multiverse[yajl]` after adding an if statement and adding the `#{ruby3_gem_webrick}` as ruby 3.0. no longer has webrick-gem bundled.

Also the suite_condition would prevent the test from running so that was also wrapped in an if statement. Please review if any issues arise. 

I read the yajl_test.rb file and saw the comment that `require 'yajl/json_gem'` overrides your JSON marshalling. Reading the [yajl documentation](https://rdoc.info/github/brianmario/yajl-ruby#json-gem-compatibility-api) it looks like `require 'yajl/json_gem'` just replaces  `require 'json'`. I was able to get a passing test by using `require 'json'` in the yajl_test file instead of `require 'yajl/json_gem'` on line 8 and by using `require: ['yajl', 'json']` in the Envfile on line 10. Those changes were reverted back to `'yajl/json_gem'` and continued to work.

The test currently pass as I have it but I did have some follow up questions like are the following code required?

- `json_gems = ['~> 2.3.0', '~> 2.4.0', '~> 2.5.0']`
  - Do I need to specify json version? I see require 'yajl/json_gem'
    This example is from suite/json/Envfile.
  - Not sure if my comment above corrects the issue
- `instrumentation_methods :chain, :prepend`
  - Most suite that where updated recently have this.
- `rack_test_version = RUBY_VERSION < "2.2.0" ? "< 0.8.0" : ">= 0.8.0"`
  - Some suite that where updated recently have this.

# Related Github Issue
[Issue #401](https://github.com/newrelic/newrelic-ruby-agent/issues/401)

# Testing
See comments but this change was tested locally with rake test:multiverse[yajl]
rake test: multiverse was ran initially but didn't know how to parse the data into something meaningful for me to understand.

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/test/multiverse/suites/yajl/Envfile
+++ b/test/multiverse/suites/yajl/Envfile
@@ -1,65 +1,31 @@
-
-# the suite_condition method prevents the test from running
-# but the test runs if I change it to... or comment it out
-# if RUBY_VERSION < '2.4.0'
-#   code below
-# end
-
-suite_condition("Yajl not supported for JRuby or 2.4") do
-  RUBY_PLATFORM != 'java' && RUBY_VERSION < '2.4.0'
+if RUBY_PLATFORM != 'java' && RUBY_VERSION < '2.4.0'
+  suite_condition("Yajl not supported for JRuby or 2.4") do
+    RUBY_PLATFORM != 'java' && RUBY_VERSION < '2.4.0'
+  end
 end
 
-=begin
-Do I need to specify json version I see yajl/json_gem
-This is from suite/json/Envfile
-json_gems = ['~> 2.3.0', '~> 2.4.0', '~> 2.5.0']
-
-Do I need to add this method?
-Most suite that where updated have this.
-instrumentation_methods :chain, :prepend
-
-Do I need to add this,
-some suite that where updated have this.
-rack_test_version = RUBY_VERSION < "2.2.0" ? "< 0.8.0" : ">= 0.8.0"
-
-I believe each test, tests a different a version of yaml, but the same ruby version in my case 3.0.0
-
-my main question is this meant to just run and compare to what a user has installed?
-say average joes install the newrelic agent, they have Ruby 2.4.0
-when rake is run this test should fail and then not install? or does it fail and prompt the use to update?
-
-if they have ruby 2.4 or higher this test should pass..no issues during install
-
-=end
-
-
-# OMR edit, this test passes, added #{ruby3_gem_webrick} because Ruby 3.0 no longer bundles webrick
-if RUBY_VERSION >= '2.4'
+if RUBY_VERSION >= '2.4.0'
   gemfile <<-RB
-  gem 'rack'
-  gem 'yajl-ruby', '~> 1.4.0', require: ['yajl', 'yajl/json_gem']
-  #{ruby3_gem_webrick}
+    gem 'rack'
+    gem 'yajl-ruby', '~> 1.4.0', require: ['yajl', 'yajl/json_gem']
+    #{ruby3_gem_webrick}
+  RB
+else
+  gemfile <<-RB
+    gem 'rack'
+    gem 'yajl-ruby', '~> 1.3.0', require: ['yajl', 'yajl/json_gem']
+    #per rubygems.org 1.2.1 requires >= ruby v1.8.6
+  RB
+
+  gemfile <<-RB
+    gem 'rack'
+    gem 'yajl-ruby', '~> 1.2.1' #per rubygems.org 1.2.1 requires >= ruby v1.8.6
+  RBi didnt
+
+  gemfile <<-RB
+    gem 'rack'
+    gem 'yajl-ruby', '~> 1.1.0' #per rubygems.org 1.1.0 requires >= ruby v1.8.6
   RB
 end
-
-=begin
-this are the old tests, they all fail, if I comment out suite_condition method
-
-gemfile <<-RB
-  gem 'rack'
-  gem 'yajl-ruby', '~> 1.3.0', require: ['yajl', 'yajl/json_gem']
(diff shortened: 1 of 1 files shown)


tannalynn (changes requested):
@creaturenex thank you for getting this PR started! We love seeing community involvement!

I noticed that this change is failing on our jruby suite, because it is unable to install yajl-ruby on the jruby platform. I believe that this is due to the if added surrounding the suite_condition statement. We should remove that if around the suite_condition, and since we don't want these running on jruby, we should leave that part of the suite_condition requirement. It looks like you have added some versions that DO work on ruby >= 2.4, so I think we can safely remove that part from the suite condition, and then just have the if/else below that control which versions on those ruby versions, like you've added. 

As for instrumentation methods chain/prepend, that is not needed here, it's related to some prepend work we did in our 7.0.0 release, but this gem was not effected by those changes.

For rack, this suite doesn't appear to need any additional checks on it. We recently removed support for ruby versions 2.0 and 2.1, so those checks aren't really doing anything anymore anyways, it's just left over from when we were testing on those ruby versions.

As for json, I think it's fine to leave that as is, since there is a separate suite that tests the different json versions. This suite is just responsible for yajl.

Hopefully that answers your questions! Let me know if you have any other questions or need help getting jruby fixed.

creaturenex:
Hello Tanna,

First of all, nice to meet you! I agree that my if statement on the
suite_condition is causing the issue with jruby. I guess my issue is
running the tests on my machine. I need some time to see if I have set
something up wrong on my end. When I submitted my commit, the yajl test
would only pass on my machine if the if statement on the suite_condition
was present. I can correct my commit and resubmit with the if statement
from the suite_condition removed if you think that will solve the issue.
it's just at the moment I do not have a way of confirming that it will
pass/fail on my machine and I rather not submit the commit if I cannot
confirm it will pass the test.

Thank you for answering my other questions.

Oscar

tannalynn:
@creaturenex hmm that's odd that it would be failing for you. What was the error that happens?

creaturenex:
[image: Screenshot from 2021-05-17 16-35-57.png]

tannalynn:
@creaturenex it looks like you did an email reply to github and attached an image, however the image isn't working here on github.

creaturenex:
I removed the if statement from the suite_condition from the Envfile.
From the directory executed `rake test:multiverse\[yajl\]
newrelic-ruby-agent/tests/multiverse/suites/yajl

looks like the suite condition is triggered.
SKIPPED
"/home/oscar/Projects/Github_repos/newrelic-ruby-agent/test/multiverse/suites/yajl":
Yajl not supported for JRuby or 2.4
There were no test failures


currently using RVM ruby 3.0.0
system ruby 2.5.1

tannalynn:
@creaturenex It's skipping because of the ruby version you are on. Currently, it only runs on ruby version <2.4, so I'm assuming you are running on a more recent version. As previously mentioned, I see that you added versions that should work on ruby 2.4+, correct? If that is the case, you should try removing the part of the condition that is checking for ruby <2.4, and leave only the jruby part. Something like this:
```
  suite_condition("Yajl not supported for JRuby") do
    RUBY_PLATFORM != 'java'
  end
```
Let me know how that works out.

creaturenex:
That worked!! awesome totally makes sense the < 2.4 was needed because yajl
gem was not updated. Once we updated yajl the latter half of && is no
longer necessary. How does the agent know which version of yajl-ruby to use
in the else statement?
ie 1.3.0, 1.2.1 or 1.1.0? which would only apply in ruby version less than
2.4?

I have pushed the changes. Do you know how the checks are done on github
check <https://github.com/newrelic/newrelic-ruby-agent/pull/669/checks>? is
that something I can do or is that on new relic ends?

tannalynn:
@creaturenex That's great to hear that worked! The changes look great now! I'll go ahead and merge the changes, thank you so much for working on this!

So to answer your question about how does it know which version to use, it's definitely a bit complicated, but basically, the envfile is used only in our multiverse testing. Multiverse allows the agent to be tested in multiple different environment with various gems and versions of gems. Every time you see something like
```
  gemfile <<-RB
    ...
  RB
```
what this does is actually create another gemfile for that section. When there are multiple sections like that, it will create multiple gemfiles and will start the agent with those gems, one after another, and run the entire suite with every gemfile. 
If you are running the multiverse suites on ruby 2.3, then the entire suite will run 3 times, one using each of those versions in that else.
The if statement there allows us to choose what to run based on the ruby version that is currently running. It's a bit complicated, and is specific to the ruby agent. This is where the code that controls multiverse lives https://github.com/newrelic/newrelic-ruby-agent/tree/dev/test/multiverse/lib/multiverse, if you are curious. 


The checks done on github run with github actions, and are configured in the repo. This is the file that controls the CI running like that https://github.com/newrelic/newrelic-ruby-agent/blob/dev/.github/workflows/ci.yml
It's a yml for github actions, but you can see the commands run and services needed in there. You can run those on your own too, but this runs ALL the tests, and there would likely be some setup needed to run some of the other multiverse suites using databases or other services that would need to be installed. Personally, I don't run every single thing locally when working on the agent, I usually run the unit tests (bundle exec rake test), and whatever multiverse suite is related to the changes I'm making (in this case, it would be the yajl one). The other tests I just let github run for me when I start a PR since running everything locally every time would take a very long time.

Hopefully that helps explain how both of those things work! let me know if you have any other questions!

tannalynn (approved):
Looks great! Thanks for working on this, and making the changes!

creaturenex:
Awesome! That was great and I really appreciate you taking the time to answer all my questions. I learned Ruby and this was the best place to apply what I have learned. I am so excited to see this pull request accepted!

PR #672, opened 2021-05-12
Bugfix: ensure FORK_PER_JOB Resque option is correctly supported in agent

amhuntsman:
# Overview
The FORK_PER_JOB environment flag used by Resque is only partially supported in the Ruby agent. Some areas of agent code still expect the default forking behavior of Resque in determining how/when instrumentation data is harvested and sent to NR. This PR should make the agent functional with FORK_PER_JOB=false set.

# Testing
With an existing Resque app, update `Gemfile` to use this branch:

    gem 'newrelic_rpm', git: 'https://github.com/newrelic/newrelic-ruby-agent', branch: 'fork_per_job'

Start a Resque worker programatically or on the command line:

    FORK_PER_JOB=false QUEUE=* bundle exec rake resque:work

Verify the `newrelic_agent.log` shows data being harvested and sent in the single worker process, and this data is displayed in NR1.

# Known Issues
Setting `FORK_PER_JOB=false` and `BACKGROUND=1` in a Resque worker is not yet supported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
     Support for Resque's FORK_PER_JOB flag within the Ruby agent was incomplete and nonfunctional. The agent should now behave
     correctly when running in a non-forking Resque worker process.
 
+
   ## v7.0.0
 
   * **Ruby Agent 6.x to 7.x Migration Guide Available**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,20 @@
 
   ## v7.1.0
 
-
   * **Update known conflicts with use of Module#Prepend**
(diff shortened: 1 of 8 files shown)


lib/new_relic/agent/instrumentation/resque_helper.rb
```diff
@@ -0,0 +1,17 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+
+module NewRelic
```
tannalynn, line 5: I think it'd make sense to move this helper into the `lib/new_relic/agent/instrumentation/resque/` directory, and name it just helper. I know not all the helpers have been moved into directories like that, but it's a pattern we tried to follow starting when we did the prepend work.

tannalynn (approved):
Looks great! Thanks for making those changes! 🎉

PR #673, opened 2021-05-12
Add support for CSP nonces when injecting the browser agent

### LINKED ISSUE #332: Support for CSP nonces
Currently the injected JS instrumentation uses a simple `<script>...</script>` block.
This is not playing well with services trying to use nonce-based CSP support.

Specifically, with a rails app, I'd do:
```
<%= javascript_tag nonce: true do -%> .... <% end -%>
```
to get a script tag with the current nonce added to it.

I see two potential solutions:
- either the configuration could receive some callback to receive the current nonce, if there is one, and construct the tag itself, or
- there could be some per-framework plugin which tries to do everything in the native way

I'm happy to submit some code for this. Do you have a preferred approach or some existing plans to deal with this?

Ref. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src

tannalynn:

# Overview
This PR adds support for CSP nonces when injecting the browser agent for our API method `browser_timing_header`. This change was implemented in a way that is in line with the other language agents, based on the updated agent specs for the Browser Monitoring API. 
In order to utilize this change, you must disable auto instrumentation for the browser agent, and use the API method `browser_timing_header` to inject the script manually.
Example:
```
<html>
  <head>
    <%= ::NewRelic::Agent.browser_timing_header("nonce") %>
    ...
  </head>
  <body>
```

# Related Github Issue
closes #332 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
   ## v7.1.0
 
+  * **Add support for CSP nonces when using our API to insert the browser agent**
+    We now support passing in a nonce to our API method `browser_timing_header` to allow the browser agent to run on applications using CSP nonces. This allows users to inject the browser agent themselves and use the nonce required for the script to run. In order to utilize this new feature, you must disable auto instrumentation for the browser agent, and use the API method browser_timing_header to pass the nonce in and inject the script manually.
+    
   * **Update known conflicts with use of Module#Prepend**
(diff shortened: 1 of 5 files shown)


tannalynn:
oh I forgot to add an entry to the changelog, I'll need to do that before we merge this

amhuntsman (approved):

PR #681, opened 2021-05-18
Workaround for 403 response when using the staging collector

### LINKED ISSUE #657: Agent is unable to communicate with collector until a restart
## Description

The collector responds with a 403 Forbidden error after having been connected for a while. After this error starts, it continues to receive the 403 forbidden error until the application is restarted, causing all data during that time to be lost. 

So far, we've only seen this issue on 6.13+, downgrading to 6.11 seems to resolve the issue. 

Potentially related to changes made in https://github.com/newrelic/newrelic-ruby-agent/pull/406, but more investigation is needed.

We have mostly seen this issue happening for internal teams that are reporting to the staging collector, so perhaps specifying the host in the newrelic.yml is related to the issue.

tannalynn:
# Overview
This PR is a workaround for the issue happening with the staging collector. The issue this PR addresses only impacts internal New Relic teams that are using the agent to report to the staging collector.
This makes a change to the logic added in PR https://github.com/newrelic/newrelic-ruby-agent/pull/406 when making a preconnect request. The changes present in that pull request work correctly on production, but currently the staging collector is having issues when using the original configured collector. This change will allow the agent to use the redirect host for all preconnect requests rather than the configured collector. 
This change is intended to be a temporary workaround until the issue is resolved with the staging collector. 

# Related Github Issue
closes #657 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/new_relic_service.rb
+++ b/lib/new_relic/agent/new_relic_service.rb
@@ -426,7 +426,16 @@ module NewRelic
         size = data.size
 
         # Preconnect needs to always use the configured collector host, not the redirect host
-        endpoint_specific_collector = (method == :preconnect) ? @configured_collector : @collector
+        # endpoint_specific_collector = (method == :preconnect) ? @configured_collector : @collector
+
+        # This is a temporary workaround due to errors occurring on the staging collector. The prod collector does not have the same issue.
+        # The staging collector does not respond correctly when using the configured collector host for preconnect, so must use the redirect host
+        # Once this issue is resolved on the staging collector, use the original line that is commented out above.
+        endpoint_specific_collector = if method == :preconnect && (@configured_collector && @configured_collector.name != 'staging-collector.newrelic.com')
+                                        @configured_collector
+                                      else
+                                        @collector
+                                      end
 
         uri = remote_method_uri(method)
         full_uri = "#{endpoint_specific_collector}#{uri}"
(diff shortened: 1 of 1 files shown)


lib/new_relic/agent/new_relic_service.rb
```diff
@@ -426,7 +426,16 @@ def invoke_remote(method, payload = [], options = {})
...
 
         # Preconnect needs to always use the configured collector host, not the redirect host
-        endpoint_specific_collector = (method == :preconnect) ? @configured_collector : @collector
+        # endpoint_specific_collector = (method == :preconnect) ? @configured_collector : @collector
+
+        # This is a temporary workaround due to errors occurring on the staging collector. The prod collector does not have the same issue.
+        # The staging collector does not respond correctly when using the configured collector host for preconnect, so must use the redirect host
+        # Once this issue is resolved on the staging collector, use the original line that is commented out above.
+        endpoint_specific_collector = if method == :preconnect && (@configured_collector && @configured_collector.name != 'staging-collector.newrelic.com')
+                                        @configured_collector
```
angelatan2, line 435: Is there a way to log when this condition is entered for staging?  The concern here is that this if condition will mask the problem and we will not come back to it to remove the if condition.

Could we also create another issue to check on logs after this is release. In essence to visit in x months whether this condition that we a adding is still necessary.

angelatan2 (commented):
Please see my comment.

lib/new_relic/agent/new_relic_service.rb
tannalynn: I think adding log messages here to which part of the if is being used is a great idea, I'll add that.

amhuntsman (approved):
+1 with logging updates.

Issue #670, opened 2021-05-11
DT with Server Side Config support for Ruby Agent


angelatan2:
### Is your feature request related to a problem? Please describe.
The Ruby agent has defaults set in the default.rb for DT in the file.  We need to change the default for SSC to true now that the UI is available for DT.

### Feature Description
Turn on SSC for DT to default to true.  

### Additional context
Please do the following:

- [ ] End to end testing of DT with SSC
- [ ] Review the cross agent spec
- [ ] Communicate with GTS the support of this feature and how will it change for them so that they are ready to field any questions that may come along.

### Priority
[Must Have]

PR #683, opened 2021-05-19
allow distributed_tracing.enabled to be set from server

amhuntsman:
https://github.com/newrelic/newrelic-ruby-agent/issues/670

Enabled SSC for distributed tracing in the Ruby agent. Tested in a sample application, verified that enabling/disabling DT in NR1 sends the signal to force restart the agent, and the configs are correctly updated. *Note: tested in prod as staging reconnect issue is still being investigated.*

Cross-app spec for DT was reviewed. The only change here is to allow `distributed_tracing.enabled` from the server; existing config combinations should continue working as intended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
   ## v7.1.0
 
+  * **Enable server-side configuration of distributed tracing**
+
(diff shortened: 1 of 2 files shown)


tannalynn (approved):
Looks great! Thanks for working on this and testing it out.

Issue #652, opened 2021-04-23
Add latest Pumas to Rack instrumentation test matrix


mwlang:
We're only testing up to Puma 3.11.x in our Rack instrumentation tests.

Add 4.x, 5.x to the matrix.  5.2.2 is latest Puma version at the time of this writing.

ChaelCodes:
Does this just involve adding [new Puma versions here](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/test/multiverse/suites/rack/Envfile)? Or is it more complicated?

PR #690, opened 2021-05-21
update rack and puma versions in multiverse

amhuntsman:
https://github.com/newrelic/newrelic-ruby-agent/issues/652

Update rack multiverse Envfile for Rack versions 2.0 and 2.2, Puma versions 4 and 5. Added a version check so Rack 2.2 on runs in Ruby 2.3+.

I tried to follow the intent of existing Puma 3.11 tests which run against both Rack 1.6 and Rack 2.0. Only Rack 2.0 and up have Ruby version requirements, and certain tests only run if Puma is present, so I included all of these combinations for Puma 3 and up.

# Reviewer Checklist
- [x] Perform code review
- [ ] Add performance label


tannalynn (approved):
Looks great! Thanks for getting this updated.

PR #694, opened 2021-06-03
Fix Rack instrumentation for "kwarg middlewares" in Ruby 3.0

walro:
# Overview

Prior to this change, in Ruby 3.0, using a middleware with keyword arguments would fail with an `ArgumentError`.

This fix is pretty much the same as https://github.com/rack/rack/pull/1505

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/instrumentation/rack/chain.rb
+++ b/lib/new_relic/agent/instrumentation/rack/chain.rb
@@ -31,7 +31,8 @@ module NewRelic::Agent::Instrumentation
             def use_with_newrelic(middleware_class, *args, &block)
               use_with_tracing(middleware_class) { |wrapped_class| use_without_newrelic(wrapped_class, *args, &block) }
             end
-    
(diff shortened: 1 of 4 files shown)


walro:
The GHA failures seems to be due to this branch lacking the fix from #696

tannalynn:
@walro yes, it does look like that is the case, and it is preventing our tests from running on those older ruby versions. If you merge (or rebase) the new changes in dev into your branch, this should allow the tests to run, so we can confirm that the tests pass on all ruby versions the agent supports.

walro:
> f you merge (or rebase) the new changes in dev into your branch, this should allow the tests to run, so we can confirm that the tests pass on all ruby versions the agent supports.

Rebased and pushed!

tannalynn (approved):
This looks great, thank you for pulling the recent updates into your branch to get the tests passing. Also thank you so much for adding the test coverage for this situation! It's great to get these type of oversights in our test coverage fixed.

Issue #573, opened 2021-02-16
Rails 5 + Rack + Rack Cors + Puma produces "can't add a new key into hash during iteration" Errors


mwlang:
## Description
Error: can't add a new key into hash during iteration
/root/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rack-2.2.3/lib/rack/request.rb:79:in `set_header'

We're fighting with this persistent and long-standing error again on production with New Relic and Rack when deploying using Puma with multiple threads in Rails. It appears that the env hash isn't atomic and Rack (or rack-cors) attempts to add a key while New Relic iterates over it internally leading to a somewhat rare but breaking error. It usually takes 20-30 minutes, and the frequency increases based on number of threads, especially when not using workers. This happens on any route at random intervals.

There are some related topics I’ve found online but no public solutions, and the only solution I’ve found is either to not use New Relic at all or to set the thread count to 1 and compensate for that concurrency slack by increasing the number of workers that Puma uses. I’ve spoken to an engineer on another project using a similar setup but they don’t seem to be affected by this issue, possibly due to the fact they’re using Rails 4 and outdated gems, or that they wrap their rack-cors implementation in Grape middleware.

There seem to be [other tickets](https://discuss.newrelic.com/t/ruby-runtime-error-cant-add-a-new-key-into-hash-during-iteration/77571/10) out there referring to this issue as well, if any relevant communication can be relayed from those, that'd be appreciated. 

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results
Provide any other relevant log data.
TIP:  Scrub logs and diagnostic information for sensitive information 

## Steps to Reproduce

This can be reproduced by deploying a Rails app with Puma, that has New Relic set up, and using Rack and / or rack-cors. Disable workers and increase number of threads to 16+, the higher the number of threads the more frequently this error will occur.

## Your Environment

Our gem versions are:
rails 5.2.4.4
rack 2.2.3
rack-cors 1.1.1
newrelic_rpm 6.15.0
puma 5.2.1

## Additional context
* [New Relic Explorers - Ruby runtime error: “can’t add a new key into hash during iteration”](https://discuss.newrelic.com/t/ruby-runtime-error-cant-add-a-new-key-into-hash-during-iteration/77571)
[...]

PR #702, opened 2021-06-15
Kh.fix hash thread conflict

wyhaines:
# Overview
Issue #573 is being caused by a simple threading conflict with the ENV hash. This small fix eliminates the issue.

# Related Github Issue
#573 

# Testing
The basic testing procedure is outlined in the original issue. It's not something that has a spec. Rather, it just depends on a multithreaded environment to trigger the problem, so run your Puma instance with a lot of threads.

# Reviewer Checklist
- [x] Perform code review
- [x] Add performance label
- [x] Add version label prior to acceptance


amhuntsman:
Thank you @wyhaines!

PR #708, opened 2021-06-24
Ensure blank SSC settings don't overwrite local expected errors

amhuntsman:
# Overview
One more Expected Errors change that didn't make it into #704. This commit ensures that if SSC sends an empty array to the agent, this will not overwrite any locally-set expected/ignored errors.

Also fixes Error Filter to accept integers as status codes (as these will be sent by SSC as integers).

# Reviewer Checklist
- [x] Perform code review
- [x] Add performance label
- [x] Add version label prior to acceptance

### DIFF
--- a/lib/new_relic/agent/error_collector.rb
+++ b/lib/new_relic/agent/error_collector.rb
@@ -87,6 +87,14 @@ module NewRelic
         @error_filter.expected?(ex, status_code)
       end
 
+      def load_error_filters
+        @error_filter.load_all
+      end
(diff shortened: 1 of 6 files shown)


amhuntsman:
I finally found the source of most of the test failures. Since I changed the way error filters are initialized, I had to make sure that the setup/teardown was properly loading and resetting the filters for each test.

tannalynn (approved):
looks good! Thanks!

Issue #720, opened 2021-07-07
Validate options for add_method_tracer earlier in call


tannalynn:
Refactor validating options to occur earlier with the API call (current implementation raises exception that appears to go unhandled, change to logging error instead).  Ensure invalid options aren’t preventing agent from starting and also that exceptions raised here aren’t stopping customer’s app from running properly.

angelatan2:
7/26/2021 This story will be reviewed once the method tracer PR is merged.  We need more information on the exception itself and will monitor after the merge.

Issue #719, opened 2021-07-07
Use procs for code_header and code_footer in add_method_tracer


tannalynn:
Currently, the agent allows a string to be passed as an optional parameter for the `code_header` and the `code_footer`. We are moving away from using string interpolation in the `add_method_tracer`, so we should update this to accept a proc for these optional params.

Issue #718, opened 2021-07-07
Implement callback option for dynamic naming of metric name in add_method_tracer


tannalynn:
Currently the agent allows a string to be passed as the metric_name when using `add_method_tracer`.  We should update this to use Procs to dynamically set the metric_name, and no longer interpolate the metric_name string. Passing a static string for the metric_name should still be allowed.

Issue #717, opened 2021-07-07
Update method tracer to use Module#prepend


tannalynn:
Currently our method tracer uses an eval'ed string with method chaining, however we want to update this to no longer use string eval, and to use Module#prepend instead of method chaining. The args should use `(...)` on ruby 2.7+.
Also this breaking change may break some our auto instrumented libraries, so updating those uses of add_method_tracer is important as well.
For more details please see https://github.com/newrelic/newrelic-ruby-agent/issues/502

PR #728, opened 2021-07-19
Refactor method tracers to use Module#prepend, no longer class_eval strings

amhuntsman:
This PR incorporates the four proposed changes from https://github.com/newrelic/newrelic-ruby-agent/issues/502.
* Implement using prepend instead of eval and call super to invoke the original method that is wrapped.
* Change code_header and code_footer to accept Procs instead of Strings that is Ruby code. This will allow us to also avoid having to use eval
* Allow metric names to be String or Proc and no longer an embedded, interpolatable String.
* Handle method parameter delegation in a way that is future-proof and works from Ruby 2.0 through Ruby 3.2+ and does not cause compiler warnings that attribute syntax issues to the Ruby agent. That is, any such syntax errors are correctly surfacing at the call site or point of original method's definition rather than as side-effect of Ruby agent wrapping the original method.

These are wide-reaching concerns that were difficult to separate from each other in the same PR, hence their being grouped together here.

## Switch to Module#prepend
`add_method_tracer` now sets a class instance variable that holds an anonymous Module, which is prepended to the calling class. A method is created with the same name as the traced method, that wraps the original method in tracing logic. This works with tracing class methods as well.

One point of confusion here is that the Ruby agent runs the following on initialization:

    Module.send :include, NewRelic::Agent::MethodTracer::ClassMethods
    Module.send :include, NewRelic::Agent::MethodTracer

This gives all classes and modules access to `add_method_tracer`, though in our documentation and custom instrumentation (and many tests) we still explicitly `include NewRelic::Agent::MethodTracer`. Changing this is beyond the scope of this milestone, but we should revisit whether the above step is necessary.

## Change `:code_header` and `:code_footer` to accept Procs, no longer `class_eval` strings
Fairly straightforward change: anything that was a string is now a Proc, which is evaluated by `instance_exec` in the traced method.

## Allow metric names to be String or Proc; String no longer interpolated
Similar to the above. One big catch here: the arity of this Proc is important. It should have the same arity of the method being traced (or accept any arguments, i.e. `-> (*) { #code }`. This was done in order to preserve a test case like this:

    add_method_tracer :foo, '#{args[0]}.#{args[1]}'

Previously `args` was provided by the tracer method body built by string. After this change, the provided Proc must accept at least the arguments passed to it, or an `ArgumentError` will be raised. The above will be rewritten as:

    add_method_tracer :foo, -> (*args) { "#{args[0]}.#{args[1]}" }

## Future-proof method delegation (Ruby 3 kwargs)
This is handled as suggested in this article: https://eregon.me/blog/2021/02/13/correct-delegation-in-ruby-2-27-3.html

That is, the delegate method uses the `(*args, &block)` parameter format, and the `ruby2_keywords` feature in Ruby 2.7+.

We've had some discussion on what to do when `ruby2_keywords` is removed from Ruby. I recommend not worrying about it until that actually happens, so we can see what the Ruby community recommends. The good news is that, with this refactoring, there's only one line in method tracing that we'll need to change.

# Breaking changes
* Strings provided to `:code_header` and `:code_footer` will be ignored; only Procs will be processed. Strings passed as metric names will not be interpolated at instance level.
* Calling `add_method_tracer` twice for the same method will not overwrite the first (as of this PR; we can probably change this behavior)
* `add_method_tracer`-generated methods are public regardless of the original method's visibility - a natural consequence of the `Module#prepend` paradigm
* `remove_method_tracer` now only takes a single argument (the method name)

# Related Issues
https://github.com/newrelic/newrelic-ruby-agent/issues/717
https://github.com/newrelic/newrelic-ruby-agent/issues/718
https://github.com/newrelic/newrelic-ruby-agent/issues/719
https://github.com/newrelic/newrelic-ruby-agent/issues/720

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


lib/new_relic/agent/method_tracer.rb
```diff
@@ -103,126 +103,72 @@ module ClassMethods
...
           # assist with debugging, so that errors occur at class
           # construction time rather than instrumentation run time
-          def validate_options(method_name, options)
+          def _nr_validate_method_tracer_options(method_name, options)
             unless options.is_a?(Hash)
               raise TypeError.new("Error adding method tracer to #{method_name}: provided options must be a Hash")
             end
-            check_for_illegal_keys!(method_name, options)
+
+            unrecognized_keys = options.keys - ALLOWED_KEYS
```
tannalynn, line 117: Curious about why you removed the method call and replaced it with the contents of the method?

lib/new_relic/agent/method_tracer.rb
```diff
@@ -293,64 +239,63 @@ def derived_class_name
...
         private
 
-        def _add_method_tracer_now(method_name, metric_name_code, options)
+        def _nr_add_method_tracer_now(method_name, metric_name, options)
           NewRelic::Agent.record_api_supportability_metric(:add_method_tracer)
 
           return unless newrelic_method_exists?(method_name)
-          metric_name_code ||= default_metric_name_code(method_name)
-          return if traced_method_exists?(method_name, metric_name_code)
+          metric_name = _nr_default_metric_name(method_name) if metric_name.to_s.strip.empty?
```
tannalynn, line 268: How come you changed it from using `||=` to `if metric_name.to_s.strip.empty?`? 
I think ||= might be more performant, although since this is for setting up the method tracer, it only happening on startup, so performance on that scale isn't as big of a deal compared to things that happen during every transaction or span. was curious though.

test/new_relic/agent/method_tracer_params_test.rb
```diff
@@ -1,178 +0,0 @@
-# encoding: utf-8
```
tannalynn, line 1: Why was this file removed? I think this is still the type of thing we'll want to be testing, especially because of kwarg reasons?

test/new_relic/agent/method_tracer_test.rb
```diff
@@ -302,12 +291,12 @@ def test_trace_method_with_block
   def test_trace_module_method
     NewRelic::Agent.add_method_tracer :module_method_to_be_traced, '#{args[0]}'
```
tannalynn, line 292: There is an interpolated string used as the metric name for this test as well, does this need to be updated to use a proc?

test/new_relic/agent/method_visibility_test.rb
```diff
@@ -67,19 +61,11 @@ def setup
...
-    define_method "test_should_preserve_visibility_of_#{visibility}_traced_method" do
-      assert @instance.send("#{visibility}_methods").map{|s|s.to_sym}.include?(:"#{visibility}_method!"), "Method #{visibility}_method should be #{visibility}"
-    end
-
     define_method "test_should_preserve_visibility_of_#{visibility}_traced_transaction" do
       assert @instance.send("#{visibility}_methods").map{|s|s.to_sym}.include?(:"#{visibility}_transaction!"), "Transcation #{visibility}_transaction should be #{visibility}"
     end
   end
 
-  def test_tracing_non_public_methods_doesnt_add_public_methods
```
tannalynn, line 79: I remember you mentioning in pairing that prepend makes things public, is that correct? I'm guessing that's why this test was removed. I'm curious what kind of impacts having it be public due to prepend would have for customers?

lib/new_relic/agent/method_tracer.rb
amhuntsman: Those methods were only being used by `validate_options` before, and each of those methods were 3-4 lines long. I'm not a fan of this pattern because debugging means tracking 2-3 separate methods in the same file, and in this case the error messages make it fairly clear what the code is doing.

lib/new_relic/agent/method_tracer.rb
amhuntsman: I was thinking of the odd case where someone passes `metric_name` as a string of just whitespace. I didn't specifically test if that would even make it through to the UI, so I just tried avoiding it altogether.

test/new_relic/agent/method_tracer_params_test.rb
amhuntsman: I have mixed feelings about this.

For one thing, this file is testing Ruby functionality and not our actual agent code. For another, it's asserting deprecation warnings which don't fire because we're using `ruby2_keywords` as intended.

I think it's a good idea to test different parameter formats to verify the delegation works, but not necessarily at this level of granularity. We can add this to our existing method tracer tests and let the CI test against different Ruby versions.

test/new_relic/agent/method_visibility_test.rb
amhuntsman: I'm going to revisit this and add these tests back in. Whatever I was trying last week wasn't working and I decided to shelve it temporarily.

I misspoke, basically, since I tend to forget the fundamental of private methods when mixins get involved. Private methods aren't totally isolated to the class they're defined in; they just can't be called with an explicit receiver. This, for example, works:

    module M
      private
      def fruit
        "pine#{super}"
      end
    end

    class A
      prepend M

      def eat_something
        puts "Mmm, #{fruit}."
      end

      private
      def fruit
        'apple'
      end
    end

    A.new.eat_something  # "Mmm, pineapple."

test/new_relic/agent/method_tracer_test.rb
amhuntsman: My eyes glazed over this particular test because it's not doing anything interesting. It's just testing that tracing a module method works without error, which we're already basically doing below in `test_add_class_method_tracer`. I think I'll rewrite this test to work more like that one.

amhuntsman:
Making a TODO list while I'm thinking of it:

- update changelog
- add parameter checks to method tracer tests (spread over multiple tests or have one test that uses a variety of formats)
- replace method visibility code and tests
- allow for setting multiple metrics/options on the same method tracer

The last item came about when I realized that's why the ActiveMerchant tests are broken. In our instrumentation we have:

    add_method_tracer operation, -> (*) { "ActiveMerchant/gateway/#{gateway_name}/#{operation}" }
    add_method_tracer operation, -> (*) { "ActiveMerchant/gateway/#{gateway_name}" }, :push_scope => false
    add_method_tracer operation, -> (*) { "ActiveMerchant/operation/#{operation}" }, :push_scope => false

which currently only records the first tracer and ignores the next two. This functionality is not really documented, and looking at our current code I'm not even sure how it's working right now. At any rate, this shouldn't be too hard to implement under the new model.

test/new_relic/agent/method_tracer_params_test.rb
tannalynn: i think that makes sense. My main concern related to this is making sure we are forwarding the arguments correctly, no matter if someone is using old style args, kwargs, etc, to make sure the arguments are exactly as the user expects. If we add some tests elsewhere that can check for that, I think that sounds good to me.

lib/new_relic/agent/method_tracer.rb
tannalynn: Yeah, the performance impacts of something that only happens on agent start up isn't too concerning, I just wanted to ask about it. I think that reasoning makes sense.

test/new_relic/agent/method_tracer_params_test.rb
```diff
@@ -146,12 +146,6 @@ def assert_common_tracing_behavior traced_class
       refute_deprecation_warning { instance.wildcard_args(:foo, bar: "foobar") }
       refute_deprecation_warning { instance.wildcard_args(:foo, {bar: "foobar"}) }
  
-      if RUBY_VERSION >= "2.7.0" && RUBY_VERSION < "3.0.0"
```
tannalynn, line 149: How come these tests were removed? you may have mentioned it in pairing at one point, but I don't remember.

amhuntsman: These very specific tests were the ones that were failing in a seemingly random order - the first test run would not produce the deprecation warning, but subsequent tests would. Since this seems to be specific to the testing environment, I removed them for the time being.

In general I think testing for the deprecation warnings is fine, but the important tests are the ones that ensure the arguments passed to the tracer method are the same as those received by the original method, and those tests are currently passing.

tannalynn (commented):
@amhuntsman oh, can you add a changelog entry for this?

Issue #551, opened 2021-01-28
Deprecate CAT


azeazasloper:
Change the default for `cross_application_tracer.enabled` configuration flag to `false` and add deprecation warnings.

The [public docs](https://docs.newrelic.com/docs/agents/ruby-agent/features/cross-application-tracing-ruby) will also need to be updated.

We will completely remove support for this feature in a future major version release.

kaylareopelle:
The work to deprecate CAT in the ruby-agent's code has been merged into `dev`. The remaining work involves updating the docs-website repo with deprecation messages where CAT is referenced. There are a handful of areas where CAT is mentioned  that I would like to test with DT to make sure it is an applicable replacement.

kaylareopelle:
Docs WIP Branch: https://github.com/kaylareopelle/docs-website/tree/ruby-agent/deprecate-cat 

Steps before close: 
* Can parallel requests using the `Hydra` mechanism in Typhoeus receive entire tracing support through DT? 
* Can requests made via the `Curl::Multi` Api have full tracing support with distributed tracing? 
* Is `NewRelic::Agent::Hooks` still a viable middleware for inclusion in Rack-based applications? (If not, which version of the gem did it lose support?)
* Can DT be used on pure Rack applications?

kaylareopelle:
Testing complete (notes above). Documentation PR ready for ruby-agent team review: https://github.com/newrelic/docs-website/pull/3583

PR #753, opened 2021-08-06
Update CAT deprecation message

kaylareopelle:

# Overview
Update the location of the CAT deprecation message and the accompanying text.

The previous location printed a message to the logs every time an external request is made. By moving the message to `CrossAppMonitor`, it will only be printed when CAT is initialized. 

# Related Github Issue
#551 

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/test/new_relic/agent/monitors/cross_app_monitor_test.rb
+++ b/test/new_relic/agent/monitors/cross_app_monitor_test.rb
@@ -163,7 +163,7 @@ module NewRelic::Agent
         'OtherTransactionTotalTime',
         'OtherTransactionTotalTime/transaction',
         'Supportability/API/record_metric',
-        'Supportability/Deprecated/cross_application_tracer.enabled configuration'
+        'Supportability/Deprecated/cross_application_tracer'
         ])
     end
 
--- a/test/new_relic/agent/monitors/cross_app_monitor_test.rb
+++ b/test/new_relic/agent/monitors/cross_app_monitor_test.rb
@@ -157,8 +157,14 @@ module NewRelic::Agent
     def test_doesnt_write_metric_if_id_blank
       when_request_runs(for_id(''))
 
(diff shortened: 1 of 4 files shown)


amhuntsman (approved):

test/new_relic/agent/monitors/cross_app_monitor_test.rb
```diff
@@ -157,8 +157,14 @@ def test_writes_metric
...
 
-      assert_metrics_recorded_exclusive(['transaction', 'Supportability/API/drop_buffered_data',
-        'OtherTransactionTotalTime', 'OtherTransactionTotalTime/transaction'])
+      assert_metrics_recorded_exclusive([
+        'transaction',
+        'Supportability/API/drop_buffered_data',
+        'OtherTransactionTotalTime',
+        'OtherTransactionTotalTime/transaction',
+        'Supportability/API/record_metric',
+        'Supportability/Deprecated/cross_application_tracer.enabled configuration'
```
tannalynn, line 166: I'm not sure if we are allowed to have a space or period in a metric like this. I don't think I've seen any metrics that have either of those characters. When we get back we should look into that and what the limitations are on the server. Since it's coming from the call to the deprecator, that's where we'd need to update what we're calling the metric.

tannalynn (changes requested):

test/new_relic/agent/monitors/cross_app_monitor_test.rb
kaylareopelle: @tannalynn - how would we go about looking into the limitations on the server for the metrics?

kaylareopelle: The agent specs seem to prefer underscores only as separators for the last metric section. Nice catch!

amhuntsman (approved):

tannalynn (approved):
TY!

Issue #713, opened 2021-07-06
Failure to load a configuration YAML file containing integer values


cbeer:
## Description

NewRelic breaks when it loads a configuration containing integer values:

```
NoMethodError: undefined method `empty?' for 404:Integer
/opt/app/mise/mise/shared/bundle/ruby/3.0.0/gems/newrelic_rpm-7.2.0/lib/new_relic/agent/error_filter.rb:32:in `load_from_config'
/opt/app/mise/mise/shared/bundle/ruby/3.0.0/gems/newrelic_rpm-7.2.0/lib/new_relic/agent/error_collector.rb:32:in `block (2 levels) in initialize'
/opt/app/mise/mise/shared/bundle/ruby/3.0.0/gems/newrelic_rpm-7.2.0/lib/new_relic/agent/configuration/manager.rb:157:in `register_callback'
/opt/app/mise/mise/shared/bundle/ruby/3.0.0/gems/newrelic_rpm-7.2.0/lib/new_relic/agent/error_collector.rb:31:in `block in initialize'
/opt/app/mise/mise/shared/bundle/ruby/3.0.0/gems/newrelic_rpm-7.2.0/lib/new_relic/agent/error_collector.rb:30:in `each'
/opt/app/mise/mise/shared/bundle/ruby/3.0.0/gems/newrelic_rpm-7.2.0/lib/new_relic/agent/error_collector.rb:30:in `initialize'
/opt/app/mise/mise/shared/bundle/ruby/3.0.0/gems/newrelic_rpm-7.2.0/lib/new_relic/agent/agent.rb:63:in `new'
/opt/app/mise/mise/shared/bundle/ruby/3.0.0/gems/newrelic_rpm-7.2.0/lib/new_relic/agent/agent.rb:63:in `initialize'
/opt/app/mise/mise/shared/bundle/ruby/3.0.0/gems/newrelic_rpm-7.2.0/lib/new_relic/agent/agent.rb:106:in `new'
/opt/app/mise/mise/shared/bundle/ruby/3.0.0/gems/newrelic_rpm-7.2.0/lib/new_relic/agent/agent.rb:106:in `instance'
/opt/app/mise/mise/shared/bundle/ruby/3.0.0/gems/newrelic_rpm-7.2.0/lib/new_relic/control/instance_methods.rb:70:in `init_plugin'
[...]

PR #777, opened 2021-09-13
Accept Integers for ignored status codes in YAML config

elaguerta:
# Overview
- Fixes bug where agent errors on ignore_status_codes with Integer values, identified in https://github.com/newrelic/newrelic-ruby-agent/issues/713
- Adds a test to confirm that the fix works

# Related Github Issue
https://github.com/newrelic/newrelic-ruby-agent/issues/713

# Testing
- Additional config added to existing test to verify that Integer status codes can pass

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/error_filter.rb
+++ b/lib/new_relic/agent/error_filter.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 # This file is distributed under New Relic's license terms.
 # See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+require 'pry'
 
 module NewRelic
   module Agent
@@ -29,7 +30,9 @@ module NewRelic
       def load_from_config(setting, value = nil)
         errors = nil
         new_value = value || fetch_agent_config(setting.to_sym)
-        return if new_value.nil? || new_value.empty?
+        #edit 33 to account for integer (DONE)
+        
+        return if new_value.nil? || (new_value.instance_of?(String) && new_value.empty?) 
 
         case setting.to_sym
         when :ignore_errors, :ignore_classes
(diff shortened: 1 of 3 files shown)


lib/new_relic/agent/error_filter.rb
```diff
@@ -144,8 +145,16 @@ def log_filter(setting, errors)
         end
       end
 
+      #edit this to account for integers
```
kaylareopelle, line 148: I think this might be a leftover from the WIP! Do you think this comment is still needed?

lib/new_relic/agent/error_filter.rb
```diff
@@ -144,8 +145,16 @@ def log_filter(setting, errors)
...
       def parse_status_codes(codes)
-        code_list = codes.is_a?(String) ? codes.split(',') : codes
+        code_list = case codes
+          when String
+            codes.split(',')
+          when Integer
+            [codes]
+          else
+            codes
+          end
```
kaylareopelle, line 157: Ahhh, I really like the look of this solution. I remember our struggle during pairing with this yesterday. This solution is very easy to read. Nice work!

lib/new_relic/agent/error_filter.rb
```diff
@@ -29,7 +29,8 @@ def load_all
       def load_from_config(setting, value = nil)
         errors = nil
         new_value = value || fetch_agent_config(setting.to_sym)
-        return if new_value.nil? || new_value.empty?
+
+        return if new_value.nil? || (new_value.respond_to?(:empty?) && new_value.empty?)
```
kaylareopelle, line 33: Great pivot from yesterday! `respond_to?` is such a handy method. This is a common idiom in Ruby.

test/new_relic/agent/error_filter_test.rb
```diff
@@ -88,6 +88,13 @@ def test_skip_invalid_status_codes
           assert @error_filter.ignore?(TestExceptionA.new, 401)
           assert @error_filter.ignore?(TestExceptionA.new, 500)
         end
+
```
kaylareopelle, line 91: When writing tests, I lean toward writing one assertion per example. I understand this file isn't currently following these practices, but would you be open to refactoring the added test into its own method? This helps use the test suite as documentation (each definition provides a bit more context about what the file is trying to accomplish) and helps with debugging.

That could look something like: 
```ruby
def test_ignore_integer_status_codes
  # ... code written for your test
end 
```

kaylareopelle (commented):
Awesome work, @elaguerta and @brammerl! Two small notes before wrapping this up. Let me know if you have any questions!

lib/new_relic/agent/error_filter.rb
elaguerta: Nope! Deleting it.

test/new_relic/agent/error_filter_test.rb
elaguerta: Sounds good, I also prefer one assertion per example. Thanks for suggesting a test name!

kaylareopelle (approved):
LGTM! Thanks for your hard work on this! I'm going to add a label to this ticket so that we merge it in after the 8.0 major release is complete.

PR #778, opened 2021-09-15
Fix multiverse "rest" group when running multiple groups

tannalynn:

Adds a fix to the way multiverse checks the "rest" group (every suite not in any other group) when passing in multiple groups at the same time. previously it would skip the rest group check, but this has been fixed.
This PR also includes a fix for a failing bunny test that was previously not being run in the CI due to this multiverse bug.

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


kaylareopelle (approved):
LGTM! Thanks for catching and fixing this. I double checked the test groups and everything is there. ✨

PR #798, opened 2021-09-28
Logger instrumentation and perf suite

jasonrclark:
# Overview 🪓 
This pull request introduces basic instrumentation for logging line counts and sizes, with faceting by severity. It hooks into the Ruby standard library `Logger` class.

This new instrumentation is controlled by the configuration flags `instrumentation.logger` or the older style `disable_logger_instrumentation`.

I've confirmed that the point that we're instrumenting is available with the same signature all the way back to Ruby 2.2.0, so nothing special needed to support it on all our currently allowed Rubies.

There did end up being one subtle bit here because the agent uses `Logger` ourselves via `AgentLogger`. This meant that if our metric recording caused us to write to the `AgentLogger` again (which does happen in some paths), we could end up in a recursion. This surfaced immediately in one of the unit tests. To address this, we flag when a given instance of `Logger` is already in the midst of a call, so if we end up writing to that logger again it won't instrument the further calls.

# Testing 🔍 
Because the `Logger` is available in the standard library, I could exercise the functionality entirely through unit tests. The tests check that:

* The underlying device still sees messages with the contents we expect (i.e. we don't interfere with normal operation)
* We record the metrics we expect with all the various permutations
* Flagging to skip further instrumentation is obeyed.

~I haven't run the entire Multiverse locally to see whether any of the metric changes show up in unexpected places. Will keep an eye on the PR testing and if anything shows up like that I'll get it fixed.~ Multiverse is running 🟢  now. In some cases I added to the lists of expected metrics, in a few others I took advantage of the `ignore_filter` where it was already in use for `Supportability` metrics to do the same for `Logging/*`. Could arguably just disable that instrumentation on the whole multiverse suite instead if we preferred to take that out of consideration entirely 🤔 

# Performance 💨 
As logging is a common action, we do want to pay attention to the impact of this. I've updated to get the performance suite running on the latest couple of Rubies (was just flagged to not) and added a test case to the Logging performance suite that doesn't use the decorator.

I observed that while there is a relatively sizable relative increase in the call time, we were still in the 10000's of iterations per second with the new path clocking in for me around 20 microseconds (not millis). I will be doing some further testing to confirm these findings in a real app with some significant logging, but all signs point to this being just another method instrumentation like lots of other frequently called things we do in the agent.

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1000,7 +1000,7 @@ module NewRelic
           :description  => 'Controls auto-instrumentation of dalli gem for Memcache at start up.  May be one of [auto|prepend|chain|disabled].'
         },
         :'instrumentation.logger' => {
-          :default => instrumentation_value_of(:disable_logger_instrumentation),
+          :default => "auto",
           :public => true,
           :type => String,
           :dynamic_name => true,
@@ -1747,15 +1747,6 @@ module NewRelic
           :type         => Boolean,
           :allowed_from_server => false,
           :description  => 'Internal name for controlling Rails 3+ middleware instrumentation'
-        },
-        :disable_logger_instrumentation => {
-          :default      => false,
-          :public       => true,
-          :type         => Boolean,
-          :deprecated   => true,
-          :dynamic_name => true,
-          :allowed_from_server => false,
-          :description  => deprecated_description(:'instrumentation.logger', 'If `true`, disables instrumentation for the Ruby standard library Logger.')
         },
         :'heroku.use_dyno_names' => {
           :default      => true,
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -999,6 +999,14 @@ module NewRelic
           :allowed_from_server => false,
           :description  => 'Controls auto-instrumentation of dalli gem for Memcache at start up.  May be one of [auto|prepend|chain|disabled].'
         },
+        :'instrumentation.logger' => {
+          :default => instrumentation_value_of(:disable_logger_instrumentation),
+          :public => true,
+          :type => String,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => 'Controls auto-instrumentation of Ruby standard library Logger at start up.  May be one of [auto|prepend|chain|disabled].'
+        },
         :disable_data_mapper => {
           :default => false,
           :public => true,
@@ -1739,6 +1747,15 @@ module NewRelic
           :type         => Boolean,
           :allowed_from_server => false,
           :description  => 'Internal name for controlling Rails 3+ middleware instrumentation'
+        },
+        :disable_logger_instrumentation => {
+          :default      => false,
+          :public       => true,
+          :type         => Boolean,
+          :deprecated   => true,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description  => deprecated_description(:'instrumentation.logger', 'If `true`, disables instrumentation for the Ruby standard library Logger.')
         },
         :'heroku.use_dyno_names' => {
           :default      => true,
--- a/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
@@ -37,8 +37,8 @@ module NewRelic
           return formatted_message if skip_instrumenting?
 
           begin
-            # It's critical we don't instrumention further logging from our
-            # metric recording in the agent itself or we'll stack overflow!!
+            # It's critical we don't instrument logging from metric recording
+            # methods within NewRelic::Agent, or we'll stack overflow!!
             mark_skip_instrumenting
 
             NewRelic::Agent.increment_metric(LINES)
--- a/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
@@ -10,25 +10,34 @@ module NewRelic
           defined?(@skip_instrumenting) && @skip_instrumenting
         end
 
+        def mark_skip_instrumenting
+          @skip_instrumenting = true
+        end
+
+        def clear_skip_instrumenting
+          @skip_instrumenting = false
+        end
+
         def format_message_with_tracing(severity, datetime, progname, msg)
-          return yield if skip_instrumenting?
+          formatted_message = yield
+          return formatted_message if skip_instrumenting?
 
           begin
             # It's critical we don't instrumention further logging from our
             # metric recording or we'll stack overflow!!
-            @skip_instrumenting = true
+            mark_skip_instrumenting
 
             sev = severity || UNKNOWN
             NewRelic::Agent.increment_metric("Logging/lines")
             NewRelic::Agent.increment_metric("Logging/lines/#{sev}")
 
-            size = msg.nil? ? 0 : msg.length
+            size = formatted_message.nil? ? 0 : formatted_message.length
             NewRelic::Agent.record_metric("Logging/size", size)
             NewRelic::Agent.record_metric("Logging/size/#{sev}", size)
 
-            yield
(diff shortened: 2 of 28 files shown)


lib/new_relic/agent/instrumentation/logger/instrumentation.rb
```diff
@@ -0,0 +1,59 @@
...
+
+          begin
+            # It's critical we don't instrumention further logging from our
+            # metric recording in the agent itself or we'll stack overflow!!
+            mark_skip_instrumenting
+
+            NewRelic::Agent.increment_metric(LINES)
+            NewRelic::Agent.increment_metric(line_metric_name_by_severity(severity))
+
+            size = formatted_message.nil? ? 0 : formatted_message.length
```
jasonrclark, line 47: bytesize?

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -999,6 +999,14 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description  => 'Controls auto-instrumentation of dalli gem for Memcache at start up.  May be one of [auto|prepend|chain|disabled].'
         },
+        :'instrumentation.logger' => {
+          :default => instrumentation_value_of(:disable_logger_instrumentation),
```
jasonrclark, line 1003: "auto"

test/multiverse/suites/mongo/mongo2_instrumentation_test.rb
```diff
@@ -101,7 +101,6 @@ def test_records_metrics_for_insert_one
             metrics = build_test_metrics(:insert, true)
             expected = metrics_with_attributes(metrics)
 
-            assert_metrics_recorded(expected)
```
jasonrclark, line 104: wut?

tannalynn (commented):
Oh another question I have is, do we want to mention this in the changelog? If we do, is there specific language that we should be using? If there is something specific you/someone has in mind, would you be willing to add that to the changelog?

test/performance/lib/performance/instrumentation/gc_stats.rb
```diff
@@ -5,7 +5,7 @@
 module Performance
   module Instrumentation
     class MRIGCStats < Instrumentor
-      platforms :mri_20, :mri_21, :mri_22, :mri_23, :mri_24, :mri_25, :mri_26
+      platforms :mri_22, :mri_23, :mri_24, :mri_25, :mri_26, :mri_27, :mri_30
```
kaylareopelle, line 8: THANK YOU VERY MUCH FOR FINDING AND FIXING THIS! 🚀

kaylareopelle (commented):
Thanks for this work, Jason! I appreciate the walkthrough presentation this morning for context. 

I happened to see your earlier WIP code snippet. In this version, I noticed that statement that would set "ANY" level logs to read as "Unknown" was removed. Why did you choose to pull this out from the PR?

lib/new_relic/agent/configuration/default_source.rb
jasonrclark: 🆒  📚

jasonrclark:
@kaylareopelle 

> I happened to see your earlier WIP code snippet. In this version, I noticed that statement that would set "ANY" level logs to read as "Unknown" was removed. Why did you choose to pull this out from the PR?

When I examined the logger.rb code more closely, I found that 1) it was already guarding the severity to set, so it was unnecessary and 2) my first draft in this PR had actually failed to defined the UNKNOWN constant so it was getting it from _somewhere_. Given point 1) it seemed better to just remove it, but I'm open to restoring it if you'd feel more comfortable.

lib/new_relic/agent/configuration/default_source.rb
jasonrclark: Added in https://github.com/newrelic/newrelic-ruby-agent/pull/798/commits/2a9cb148da9365e9aa08331338d52deff570ba65

lib/new_relic/agent/instrumentation/logger/instrumentation.rb
```diff
@@ -0,0 +1,59 @@
...
+        end
+
+
+        def format_message_with_tracing(severity, datetime, progname, msg)
+          formatted_message = yield
+          return formatted_message if skip_instrumenting?
+
+          begin
+            # It's critical we don't instrumention further logging from our
+            # metric recording in the agent itself or we'll stack overflow!!
```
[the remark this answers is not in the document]
jasonrclark: :+1: I think your version is a little bit clearer and more explicit about where the overflow would happen so taking it.

jasonrclark:
@tannalynn Updated the CHANGELOG with notes on the addition -- seems like since it's automatic and in public, we shouldn't be hiding the fact we added it.

kaylareopelle:
> @kaylareopelle
> 
> > I happened to see your earlier WIP code snippet. In this version, I noticed that statement that would set "ANY" level logs to read as "Unknown" was removed. Why did you choose to pull this out from the PR?
> 
> When I examined the logger.rb code more closely, I found that 1) it was already guarding the severity to set, so it was unnecessary and 2) my first draft in this PR had actually failed to defined the UNKNOWN constant so it was getting it from _somewhere_. Given point 1) it seemed better to just remove it, but I'm open to restoring it if you'd feel more comfortable.

Thanks for clarifying, @jasonrclark! No need to restore it.

kaylareopelle (approved):
Thanks for those adjustments! This looks good to me.

jasonrclark:
Realized that I hadn't checked this with chaining and it was busted until the latest commit. Is there a way that we typically test chained vs prepend on instrumentation? 🤔  Best I could do was adding an ENV variable to set the mode on `Logger` instrumentation (ala `NEW_RELIC_INSTRUMENTATION_LOGGER=chain bundle exec rake`)

kaylareopelle:
@jasonrclark  - Good catch! As far as testing goes, I'm only aware of testing chain vs. prepend in Multiverse right now. The Envfiles include `instrumentation_methods :chain, :prepend` to run the tests with each method. @tannalynn -  Do we have a similar method available for unit tests?

jasonrclark:
@kaylareopelle Ah cool, hadn't see that before. Open to adding multiverse testing if we need to translate the unit tests over to get that coverage. Curious whether that could come in a follow-up or would be needed in this PR before it can land.

tannalynn:
> @jasonrclark - Good catch! As far as testing goes, I'm only aware of testing chain vs. prepend in Multiverse right now. The Envfiles include `instrumentation_methods :chain, :prepend` to run the tests with each method. @tannalynn - Do we have a similar method available for unit tests?

Hmm, for unit tests, we do have a with_config block we can wrap around tests, although I haven't tried using that with something that is just used on agent startup like instrumentation type, so I'm not 100% sure it would work as intended with that, but if it does, that would allow us to skip having to add a multiverse suite for this. I can mess with it a little today and see how it works with this config option

tannalynn:
> > @jasonrclark - Good catch! As far as testing goes, I'm only aware of testing chain vs. prepend in Multiverse right now. The Envfiles include `instrumentation_methods :chain, :prepend` to run the tests with each method. @tannalynn - Do we have a similar method available for unit tests?
> 
> Hmm, for unit tests, we do have a with_config block we can wrap around tests, although I haven't tried using that with something that is just used on agent startup like instrumentation type, so I'm not 100% sure it would work as intended with that, but if it does, that would allow us to skip having to add a multiverse suite for this. I can mess with it a little today and see how it works with this config option

My suspicion about with_config not working for config options that are used at startup like that was correct, it does not work. So we will need to do multiverse testing for this in order to make sure chain and prepend are running in the CI. I went ahead and moved the logger tests into its own multiverse suite and set it up to test both prepend and chain, so we should be good there now 👍

tannalynn (approved):

kaylareopelle (approved):
Thanks for moving the suite, @tannalynn! 🚀

jasonrclark:
Awesome, thanks for digging into that and tidying it up @tannalynn ✨ !

PR #799, opened 2021-09-29
[BUGFIX] add_method_tracer errors on BasicObject methods

### LINKED ISSUE #789: Jbuilder method tracer exception in v8.0.0
## Description
After a gem upgrade from v7.2.0 to v8.0.0, an exception is raised by the `Jbuilder#target!` method tracer.

## Expected Behavior
To continue working as it did before the upgrade.

```
[ActionView::Template::Error] undefined method `trace_execution_scoped' for nil:NilClass | Stack trace:
	/Users/tonci/.rvm/gems/ruby-2.6.7/gems/newrelic_rpm-8.0.0/lib/new_relic/agent/method_tracer.rb:326:in `block (2 levels) in _nr_define_traced_method'
	/Users/tonci/.rvm/gems/ruby-2.6.7/gems/jbuilder-2.11.2/lib/jbuilder/jbuilder_template.rb:81:in `target!'
	/Users/tonci/Documents/server/app/views/api/v1/authentication/login.json.jbuilder:8:in `_app_views_api_v__authentication_login_json_jbuilder___2930598681177930957_70166338224360'
	... 
```

## Steps to Reproduce
We add a method tracer to measure actual JSON serialization performance of `Jbuilder`:

```ruby
class Jbuilder

  add_method_tracer :target!, 'View/Jbuilder/target!'
end
```

This code is executed as a Rails initializer in `config/initializers/newrelic.rb`.

Afterwards, just call the `target!` method:

```ruby
Jbuilder.new.target!
```

## Your Environment
Ruby v2.6.7, Rails v5.2.6, both macOS and Linux affected.

## Additional context
Nothing right now.

amhuntsman:
Fixes #789 

When using `add_method_tracer`, the created tracer method called `self.class.trace_execution_[un]scoped`; however, `Kernel#class` is not available on BasicObjects. The tracer method is updated here to call `::NewRelic::Agent::MethodTracer.trace_execution_scoped` as these are effectively static methods. A new test case has been added for adding method tracers to BasicObjects.

# Other thoughts
Note that the above method call relies on the fact that `MethodTracer` is included in `Module` at startup. The historic reasons for this are unclear, and it's difficult to trace *how* `trace_execution_scoped` is available as a class method on `MethodTracer`. Making this more explicit in the code would require more refactoring than is in scope here.

If/when we do change `MethodTracer` not to be included in everything via `Module`, this will have to be changed again. Luckily we already have test coverage for this.

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/method_tracer.rb
+++ b/lib/new_relic/agent/method_tracer.rb
@@ -320,6 +320,11 @@ module NewRelic
 
               instance_exec(&code_header) if code_header.kind_of?(Proc)
 
+              # NOTE: Calling ::NewRelic::Agent::MethodTracer.trace_execution_scoped and
+              # .trace_execution_unscoped below relies on the fact that MethodTracer is included
+              # in Module on agent startup. If/when this changes, these methods should be
+              # explicitly namespaced and extended in MethodTracer.
+
               # If tracing multiple metrics on this method, nest one unscoped trace inside the scoped trace.
               begin
                 if scoped_metric_eval
--- a/lib/new_relic/agent/method_tracer.rb
+++ b/lib/new_relic/agent/method_tracer.rb
@@ -323,17 +323,17 @@ module NewRelic
               # If tracing multiple metrics on this method, nest one unscoped trace inside the scoped trace.
               begin
                 if scoped_metric_eval
-                  self.class.trace_execution_scoped(scoped_metric_eval, metric: record_metrics, internal: true) do
+                  ::NewRelic::Agent::MethodTracer.trace_execution_scoped(scoped_metric_eval, metric: record_metrics, internal: true) do
                     if unscoped_metrics_eval.empty?
                       super(*args, &block)
                     else
-                      self.class.trace_execution_unscoped(unscoped_metrics_eval, internal: true) do
+                      ::NewRelic::Agent::MethodTracer.trace_execution_unscoped(unscoped_metrics_eval, internal: true) do
                         super(*args, &block)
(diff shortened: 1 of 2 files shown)


lib/new_relic/agent/method_tracer.rb
```diff
@@ -323,17 +323,17 @@ def _nr_define_traced_method(method_name, scoped_metric: nil, unscoped_metrics:
               # If tracing multiple metrics on this method, nest one unscoped trace inside the scoped trace.
               begin
                 if scoped_metric_eval
-                  self.class.trace_execution_scoped(scoped_metric_eval, metric: record_metrics, internal: true) do
+                  ::NewRelic::Agent::MethodTracer.trace_execution_scoped(scoped_metric_eval, metric: record_metrics, internal: true) do
```
kaylareopelle, line 326: Could you add a comment above this line that summarizes the information from the "Other Thoughts" section on the PR comment? I think that would be helpful in case MethodTracer is removed from the Module class on startup if this PR is lost to time.

kaylareopelle (dismissed):
Great work, Aaron. Thanks for the informative PR comment. One small suggestion for posterity.

lib/new_relic/agent/method_tracer.rb
kaylareopelle: Looks great, thanks!

kaylareopelle (approved):
LGTM!

PR #820, opened 2021-10-26
Use Dalli::Protocol::Binary rather than Dalli::Server

kylesnowschwartz:
# Overview
Since Dalli 3.0.2 was released, usage of Dalli::Server has been
deprecated. see: https://github.com/petergoldstein/dalli/pull/760

This commit removes those references and replaces them with the updated
API through Dalli::Protocol::Binary

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/test/multiverse/suites/memcache/Envfile
+++ b/test/multiverse/suites/memcache/Envfile
@@ -1,4 +1,4 @@
-%w(2.7.6 2.7.8 2.7.11).each do |version|
+%w(2.7.6 2.7.8 2.7.11 3.0.2).each do |version|
   gemfile <<-RB
     gem 'dalli', '~> #{version}'
   RB
--- a/test/multiverse/suites/memcache/Envfile
+++ b/test/multiverse/suites/memcache/Envfile
@@ -1,6 +1,12 @@
-%w(2.7.6 2.7.8 2.7.11 3.0.2).each do |version|
+%w(2.7.6 2.7.8 2.7.11).each do |version|
   gemfile <<-RB
     gem 'dalli', '~> #{version}'
+  RB
+end
+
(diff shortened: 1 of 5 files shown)


kylesnowschwartz:
I've updated this PR with what I think is a reasonable approach based on your suggestions @tannalynn 

I was not able to run the multiverse tests locally, unfortunately. Each test case came back with the simulated error defined in `dalli_test.rb`. I'm not sure what configuration I need to run the multiverse tests locally,  so I must wait to see what CI says regarding these test cases.

test/multiverse/suites/memcache/Envfile
```diff
@@ -1,4 +1,4 @@
-%w(2.7.6 2.7.8 2.7.11).each do |version|
+%w(2.7.6 2.7.8 2.7.11 3.0.2).each do |version|
   gemfile <<-RB
     gem 'dalli', '~> #{version}'
   RB
```
tannalynn, line 4:
```suggestion
%w(2.7.6 2.7.8 2.7.11).each do |version|
  gemfile <<-RB
    gem 'dalli', '~> #{version}'
  RB
end

if RUBY_VERSION >= '2.5.0'
  gemfile <<-RB
    gem 'dalli', '~> 3.0.2'
  RB
```
I didn't realize this previously, but it looks like dalli 3.0.2 only supports ruby 2.5+, which is what the current CI failures are related to. This change should make to so 3.0.2 is only running on 2.5+ and get everything passing!

tannalynn (approved):
Looks great, thank you so much for making those changes to get this compatible on all supported versions!

Issue #833, opened 2021-11-03
[TILT-1] Add file structure for instrumentation


angelatan2:
Add the file structure for the following components: 
* default_source configuration options for auto, chain, prepend, disabled
* chain instrumentation
* prepend instrumentation
* multiverse test suite
* configuration documentation in `newrelic.yml` 

### Acceptance Criteria
- [x] When the gem is installed, the instrumentation auto-loads with the prepend setting
- [x] When the gem isn't installed, the instrumentation does not load
- [x] When the gem is installed and the prepend configuration setting is passed, prepend instrumentation is used
- [x] When the gem is installed and the chain configuration setting is passed, chain instrumentation is used  
- [x] When looking at the instrumentation setting for tilt is available on `newrelic.yml` with a clear description of its behavior
- [x] When the CI runs, a multiverse test suite for tilt is included on supported Ruby versions

PR #840, opened 2021-11-04
[TILT-1] Create structure for Tilt instrumentation

kaylareopelle:
# Overview
Adds files for Tilt instrumentation, configuration, and multiverse testing.

# Related Github Issue
Closes: #833 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1007,6 +1007,14 @@ module NewRelic
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of Ruby standard library Logger at start up.  May be one of [auto|prepend|chain|disabled].'
         },
+        :'instrumentation.tilt' => {
+          :default => "auto",
+          :public => true,
+          :type => String,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => 'Controls auto-instrumentation of Tilt at start up.  May be one of [auto|prepend|chain|disabled].'
+        },
         :disable_data_mapper => {
           :default => false,
           :public => true,
--- a/lib/new_relic/agent/instrumentation/tilt.rb
+++ b/lib/new_relic/agent/instrumentation/tilt.rb
@@ -1,0 +1,25 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+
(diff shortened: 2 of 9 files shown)


tannalynn:
> @tannalynn - I was trying to create a test for the code added as part of this PR, something like:
> 
> ```ruby
> def test_tilt_instrumentation_installation_logged
>   log = with_array_logger { app }  
>   assert_log_contains(log, /INFO.*Installing Tilt instrumentation/)
>   assert_log_contains(log, /INFO.*Installing New Relic supported Tilt instrumentation using #{Regexp.quote($instrumentation_method).capitalize}/)
> end
> ```
> 
> Where `app` was a config.ru file added to the Tilt multiverse suite folder. Even though I could see the messages added to the generated `newrelic.log` file, they weren't included in the `log` variable.
> 
> I'm also on the fence as to whether these are useful tests, since we are not currently testing instrumentation installation in our other multiverse suites.
> 
> Do you believe we need a test to verify config loading/dependency detection works with the instrumentation? Can you think of another kind of test that might be helpful for this PR?
> 
> Manual QA was performed with a Tilt playground app, and everything is loading as expected.

Hmm those are some great questions. I don't think we really need to test the installation like that, since i believe dependency detection has its own tests. And then also once we add real tests, those will all just fail if it doesn't get installed at all. The only issues we've had with installation related to tests are like with padrino, where we don't know how to control the order things being loaded which can affect if instrumentation is installed successfully, but that is something we haven't solved yet, and also might not be an issue here. 
I can't really think of any other types of tests that would really make sense here, since this is kind of just setting up a skeleton for us to work in. So I'm good with leaving the tests empty as it is currently. 
One thing I am curious about, github is showing me that the branch is `experiment/disable_rails_2_instrumentation`, that seems weird though? Not a huge deal, but I was wondering if that is what it actually is, or if i'm having some weird github UI bug happening (haven't seen this particular ui bug, but i have seen several over the last year or so).

tannalynn (approved):
Looks good to me! Thanks for getting this set up!

kaylareopelle:
> One thing I am curious about, github is showing me that the branch is experiment/disable_rails_2_instrumentation, that seems weird though? Not a huge deal, but I was wondering if that is what it actually is, or if i'm having some weird github UI bug happening (haven't seen this particular ui bug, but i have seen several over the last year or so).

@tannalynn - excellent catch! I applied these changes to an old branch. Creating a new PR with a branch that has the most up-to-date content...

PR #847, opened 2021-11-11
[TILT-2] Tilt::Template#render instrumentation

### LINKED ISSUE #835: [TILT-2] Add instrumentation for Tilt::Template#render
**This must be completed after [TILT-1](#833)** 

### Resources
* [Tilt::Template#render code](https://github.com/rtomayko/tilt/blob/master/lib/tilt/template.rb#L102-L112)
* [Playground application with Tilt](https://source.datanerd.us/ruby-agent/ruby_agent_playground/tree/master/tilt) (internal)

### Description
* New metrics should follow [the style used by ActionView](https://github.com/newrelic/newrelic-ruby-agent/blob/74259ff6451a62a379176fa4cb4b96d0d94953b0/lib/new_relic/agent/instrumentation/action_view_subscriber.rb#L48): `View/template_path/render`
  * Consider whether we should also include a metric for the class that was used to render the template, ex: `Tilt::HamlTemplate`
* This should become a trace/span within the transaction that calls #render 
* Multiverse tests should be added for the new instrumentation
* Consider whether any performance tests could be added for this instrumentation

### Potential Gotchas 
* #render is a method name used by many view rendering tools
* Similar data flows through ActionView, we need to be careful about duplication if tilt is loaded within a Rails context

### Acceptance Criteria
- [ ] When an error is raised by Tilt, it is recorded by the agent as a TransactionError
- [ ] When an error is raised by New Relic code, it does not escape the agent and stop the user's application from executing
- [ ] When a Tilt template renders, a metric is recorded
- [ ] When a Tilt template renders, it is saved as a trace/span
- [ ] When a Tilt template renders, it does not create an overwhelming amount of noise within the UI

### PR Content
 - [ ] Have you attached a screenshot where the data is reporting in the UI?
 - [x] Do you have a blurb in the PR about why the instrumented method(s) were chosen? Consider how this instrumentation might help our customers solve problems.
 - [x] Does this instrumentation create any new metrics? What are they?
 - [x] Does this instrumentation create any new events? What are they?

### LINKED ISSUE #836: [TILT-3] Add instrumentation for Tilt::Template#prepare
**This must be completed after [TILT-1](#833)** 

### Resources
* [Tilt::Template#prepare code](https://github.com/rtomayko/tilt/blob/master/lib/tilt/template.rb#L151-L158)
* [Playground application with Tilt](https://source.datanerd.us/ruby-agent/ruby_agent_playground/tree/master/tilt) (internal)

### Description
* This should become a trace/span within the transaction that calls `#prepare`  
* Multiverse tests should be added for the new instrumentation
* Consider whether any performance tests could be added for this instrumentation

### Potential Gotchas 
* Similar data flows through ActionView, we need to be careful about duplication if tilt is loaded within a Rails context
* We need to be careful about instrumenting versions below 0.8. Before this version, the `#prepare` method was called `#compile`.

### Acceptance Criteria
- [ ] When an error is raised by Tilt, it is recorded by the agent as a TransactionError
- [ ] When an error is raised by New Relic code, it does not escape the agent and stop the user's application from executing
- [ ] When a Tilt template calls #prepare, it is saved as a trace/span
- [ ] When a Tilt template calls #prepare, it does not create an overwhelming amount of noise within the UI

### PR Content
 - [ ] Have you attached a screenshot where the data is reporting in the UI?
 - [ ] Do you have a blurb in the PR about why the instrumented method(s) were chosen? Consider how this instrumentation might help our customers solve problems.
 - [ ] Does this instrumentation create any new metrics? What are they?
 - [ ] Does this instrumentation create any new events? What are they?

kaylareopelle:
# Overview
This instrumentation creates new metrics with the style of: `View/#{class_name}/#{file_path}/Rendering`
An example metric would look like `View/Tilt::HamlTemplate/blog/layout.html/Rendering` 

Traces and spans related to view rendering within Tilt should be nested within the action that causes the template to render.

If a template is rendered within the block of another template rendering call, it will appear as nested within the outer template. 

### Example
```ruby 
require 'tilt'
require 'haml'

def index
  Tilt.new('layout.haml').render {
    Tilt.new('body.haml')
  }
end
```

Should result in a nested span that looks like this: 
Screen Shot 2021-11-11 at 4 52 41 PM

## Branch Name/Background 
This project began with the intention of instrumenting `Tilt::Template#prepare`, to discover this is a protected method. I decided to pivot to instrument `#initialize` instead, for the time being. 

With initial testing in the UI, I was uncertain whether anything of additional value will be captured through `#initialize` that cannot be captured using `#render`.

After talking with @tannalynn, we decided to opt-out of instrumenting `#initialize` and instrument `#render` instead so that it is on-par with our ActionView instrumentation. 

# Related Github Issue
Closes #836 
Closes #835

# Testing
Multiverse tests added to verify functionality for prepend, chain, disabled recording of metrics 
Tested with Rails calling a Haml view, and did not see a duplicated metric for Tilt, just the existing ActionView metric.

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/instrumentation/tilt.rb
+++ b/lib/new_relic/agent/instrumentation/tilt.rb
@@ -9,7 +9,8 @@ require_relative 'tilt/prepend'
 DependencyDetection.defer do
   named :tilt
 
-  depends_on { defined?(::Tilt) }
+  # prior to 0.8.0, the prepare method was known as compile
+  depends_on { defined?(::Tilt) && ::Tilt::VERSION >= '0.8.0' }
 
   executes do
     ::NewRelic::Agent.logger.info  "Installing Tilt instrumentation"
--- a/lib/new_relic/agent/instrumentation/tilt.rb
+++ b/lib/new_relic/agent/instrumentation/tilt.rb
@@ -12,14 +12,14 @@ DependencyDetection.defer do
   depends_on { defined?(::Tilt) }
 
   executes do
-    ::NewRelic::Agent.logger.info  "Installing Tilt instrumentation"
+    ::NewRelic::Agent.logger.info("Installing Tilt instrumentation")
   end
 
   executes do
     if use_prepend?
       prepend_instrument ::Tilt::Template, NewRelic::Agent::Instrumentation::Tilt::Prepend
     else
-      chain_instrument NewRelic::Agent::Instrumentation::Tilt
+      chain_instrument NewRelic::Agent::Instrumentation::Tilt::Chain
     end
   end
 end
--- a/lib/new_relic/agent/instrumentation/tilt.rb
+++ b/lib/new_relic/agent/instrumentation/tilt.rb
@@ -12,7 +12,7 @@ DependencyDetection.defer do
   depends_on { defined?(::Tilt) }
 
   executes do
-    ::NewRelic::Agent.logger.info("Installing Tilt instrumentation")
+    ::NewRelic::Agent.logger.info('Installing Tilt instrumentation')
   end
 
   executes do
--- a/lib/new_relic/agent/instrumentation/tilt.rb
+++ b/lib/new_relic/agent/instrumentation/tilt.rb
@@ -9,8 +9,7 @@ require_relative 'tilt/prepend'
 DependencyDetection.defer do
   named :tilt
 
-  # prior to 0.8.0, the prepare method was known as compile
-  depends_on { defined?(::Tilt) && ::Tilt::VERSION >= '0.8.0' }
+  depends_on { defined?(::Tilt) }
 
   executes do
     ::NewRelic::Agent.logger.info  "Installing Tilt instrumentation"
--- a/lib/new_relic/agent/instrumentation/tilt/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/tilt/instrumentation.rb
@@ -11,10 +11,19 @@ module NewRelic
           "View/#{klass}/#{file}/Rendering"
         end
 
+        # Sinatra uses #caller_locations for the file name in Tilt (unlike Rails/Rack)
+        # So here we are only grabbing the file name and name of directory it is in
+        def create_filename_for_metric(file)
+          return file unless defined?(::Sinatra) && defined?(::Sinatra::Base)
+          file.split('/')[-2..-1].join('/')
+        rescue NoMethodError
+          file
+        end
+
         def render_with_tracing(*args, &block)
           begin
             finishable = Tracer.start_segment(
-              name: metric_name(self.class, self.file)
+              name: metric_name(self.class, create_filename_for_metric(self.file))
             )
             begin
               yield
--- a/lib/new_relic/agent/instrumentation/tilt/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/tilt/instrumentation.rb
@@ -6,7 +6,120 @@ module NewRelic
   module Agent
     module Instrumentation
       module Tilt
+
+        def prepare_with_tracing
+          # Trace stuff
+          # get the template name?
+          # require 'pry'; binding.pry
+          finishable = Tracer.start_segment(name: 'Tilt/prepare')
+          yield
+          finishable.finish
+        rescue => e
+          log_notification_error(e, '', 'start')
+        end
       end
     end
   end
 end
+
+=begin
+        def start(name, id, payload) #THREAD_LOCAL_ACCESS
+          parent = segment_stack[id].last
+          metric_name = format_metric_name(name, payload, parent)
+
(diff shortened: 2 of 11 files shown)


lib/new_relic/agent/instrumentation/tilt.rb
```diff
@@ -12,14 +12,14 @@
   depends_on { defined?(::Tilt) }
 
   executes do
-    ::NewRelic::Agent.logger.info  "Installing Tilt instrumentation"
+    ::NewRelic::Agent.logger.info("Installing Tilt instrumentation")
```
kaylareopelle, line 15: Could make this single quotes since it's not interpolated

lib/new_relic/agent/instrumentation/tilt/instrumentation.rb
```diff
@@ -6,6 +6,35 @@ module NewRelic
...
+          return file unless defined?(::Sinatra) && defined?(::Sinatra::Base)
+          file.split('/')[-2..-1].join('/')
+        rescue NoMethodError
+          file
+        end
+
+        def render_with_tracing(*args, &block)
+          begin
+            finishable = Tracer.start_segment(
+              name: metric_name(self.class, create_filename_for_metric(self.file))
```
kaylareopelle, line 26: I wonder if we actually need `self` here to make it work?

test/multiverse/suites/tilt/tilt_instrumentation_test.rb
```diff
@@ -6,13 +6,184 @@ class TiltInstrumentationTest < Minitest::Test
   include MultiverseHelpers
```
kaylareopelle, line 6: I wonder if we're actually using any of these in this PR... we might be able to pull this out.

test/multiverse/suites/tilt/tilt_instrumentation_test.rb
```diff
@@ -6,13 +6,184 @@ class TiltInstrumentationTest < Minitest::Test
...
+
+    assert_equal(haml_render_metric('layout.haml'),
+                 template_node.metric_name)
+    assert_equal(haml_render_metric,
+                 partial_node.metric_name)
+  end
+
+  ### File name parsing tests ###
+  def call_create_filename_for_metric(filename)
+    Class.new.extend(NewRelic::Agent::Instrumentation::Tilt).create_filename_for_metric(filename)
```
kaylareopelle, line 123: I'd like to change up the spacing on this line to make it shorter

kaylareopelle (commented):
It's really just nit-picking things that stood out to me for fixing. I'll hold off on making changes until your review is complete, @tannalynn !

test/multiverse/suites/tilt/tilt_instrumentation_test.rb
tannalynn: The multiverse helpers? hmmm. I'm not very familiar with that part of multiverse, but looking at it it seems to have some things that are used indirectly. I _think_ we do need to keep that since I believe it's needed for multiverse to do all of the set up it needs to for the agent and fake collector? We could experiment with that though.

lib/new_relic/agent/instrumentation/tilt/instrumentation.rb
tannalynn: I'm not sure if it's necessary. Personally, I like having self.whatever for things we're accessing like that, it seems a little more immediately obvious to me what it's using. If you feel differently we could talk about it though!

tannalynn:
Oh also, we should the changelog entry for this! Almost forgot about that.

lib/new_relic/agent/instrumentation/tilt/instrumentation.rb
kaylareopelle: Nice! I don't have strong preferences toward using self or eliminating it. I was taught to use it for more clarity, but I've had people who prefer eliminating it when it isn't needed. Happy to keep it!

test/multiverse/suites/tilt/tilt_instrumentation_test.rb
kaylareopelle: I'll remove them for this CI run and if trouble brews, I'll stick them back in.

kaylareopelle: CI passed! Marking this as resolved.

tannalynn: Huh, that's surprising! I wonder how those things are happening then? Multiverse has some mystery to it lol. thanks for checking that out!

tannalynn (approved):
Looks great! Thanks for making those changes.

tannalynn (approved):

PR #868, opened 2021-11-30
Use pessimistic versioning for Rails test Gemfiles

kaylareopelle:

# Overview
The Gemfiles/Envfiles used to test Rails will sometimes use a hard-coded version of the gem. This prevents patched changes from being tested, potentially causing unseen regressions in our support for Rails. 

This PR would change the references to Rails gems to use [pessimistic versioning](https://guides.rubygems.org/patterns/) that preserves the major and minor versions of Rails but leaves patches open for automatic upgrades.

**QUESTION**
* Should we add a block to each Rails testing suite to run tests against the major version, so that the latest major version is always grabbed and we automatically test the newest minor version release?

# Related Github Issue

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        ruby-version: [2.2.10, 2.3.8, 2.4.10, 2.5.9, 2.6.8, 2.7.4, 3.0.2, jruby-9.2.19.0]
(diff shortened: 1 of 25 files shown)


tannalynn:
> **QUESTION**
> 
> * Should we add a block to each Rails testing suite to run tests against the major version, so that the latest major version is always grabbed and we automatically test the newest minor version release?

I think this is a good idea to do. If we have it automatically grabbing the latest we'll find out right away if there is an issue for a new minor version rather than waiting until we add it ourselves.

tannalynn (approved):

PR #882, opened 2021-12-17
Make generate_guid faster

jdelStrother:
# Overview

[generate_guid](https://github.com/newrelic/newrelic-ruby-agent/blob/e8f9651ca179c9d07d4b2ee0207c0bd1a8febcaa/lib/new_relic/agent/guid_generator.rb#L18) can be called many times during a single Rails request - in our app, it's called 70 times just in middleware, and over 1000 in some heavy Rails actions. This might shave a millisecond or two off those requests.

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'benchmark-ips', require: 'benchmark/ips'
end

module GuidGenerator
  HEX_DIGITS = (0..15).map{|i| i.to_s(16)}

  module_function

  # generate a random 64 bit uuid
  def generate_guid length = 16
    guid = ''
    length.times do |a|
      guid << HEX_DIGITS[rand(16)]
    end
    guid
  end

  def generate_guid2 length = 16
    guid = rand(16**length).to_s(16)
    guid.length < length ? guid.rjust(length, "0") : guid
  end
end



Benchmark.ips do |x|
  x.report("old version") { GuidGenerator.generate_guid }
  x.report("new version") { GuidGenerator.generate_guid2 }
  x.compare!
end
```

```
Warming up --------------------------------------
         old version    50.760k i/100ms
         new version   278.462k i/100ms
Calculating -------------------------------------
         old version    507.072k (± 0.3%) i/s -      2.538M in   5.005250s
         new version      2.785M (± 0.2%) i/s -     14.202M in   5.099937s

Comparison:
         new version:  2784661.4 i/s
         old version:   507071.9 i/s - 5.49x  (± 0.00) slower
```

# Reviewer Checklist
- [x] Perform code review
- [x] Add performance label


tannalynn:
Thank you for making this pr to contribute to the ruby agent. I see that the target of this PR is the `main` branch, but that branch is only used for releasing new versions. Could you update this pr to target the `dev` branch please? Thank you!

jdelStrother:
OK, that's updated.

tannalynn:
We've merged in a branch to dev that resolves the errors related to rails 7, so if you could merge dev back into your branch to get those tests passing again, that would be great! Thank you.

tannalynn (approved):
Awesome! This change looks good to go. We're currently already in the process of releasing our next version, so this will have to go into the next release after this one. I'll tag this as pending merge, so that we know this one is ready to merge once we're ready. Thanks for making this PR, that performance benchmark shows a big difference!

fallwith:
Hi @jdelStrother, this improvement has now made it out to RubyGems in the v8.4.0 release. Thanks again for your contribution!

Issue #788, opened 2021-09-22
Support JRuby 9.3.0.0


kaylareopelle:
### Feature Description
JRuby [released version 9.3.0.0 on September 22](https://www.jruby.org/2021/09/22/jruby-9-3-0-0.html 
)! 🎉 

Let's update the agent to support this version and make adjustments to the documentation.

### Additional context
Ruby-Build did not have an image for jruby-9.3.0.0 on September 22. This is blocked until the image is available.

### Priority
Must Have

kaylareopelle:
Ruby-build added support for JRuby 9.3.0.0! https://github.com/rbenv/ruby-build/issues/1796 

There are failures on the CI to address to resolve this issue.

PR #784, opened 2021-09-22
Update ci to use jruby 9.3.0.0

kaylareopelle:
_Before contributing, please read our [contributing guidelines](https://github.com/newrelic/newrelic-ruby-agent/blob/main/CONTRIBUTING.md) and [code of conduct](https://github.com/newrelic/.github/blob/master/CODE_OF_CONDUCT.md)._

# Overview
JRuby 9.3.0.0 was released today (September 22). This is a test to observe how our CI reacts.

# Related Github Issue
Closes: https://github.com/newrelic/newrelic-ruby-agent/issues/788 

# Testing
The agent includes a suite of unit and functional tests which should be used to
verify your changes don't break existing functionality. These tests will run with 
Github Actions when a pull request is made. More details on running the tests locally can be found 
[here for our unit tests](https://github.com/newrelic/newrelic-ruby-agent/blob/main/test/README.md), 
and [here for our functional tests](https://github.com/newrelic/newrelic-ruby-agent/blob/main/test/multiverse/README.md).
For most contributions it is strongly recommended to add additional tests which
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label
- [ ] Perform appropriate level of performance testing
- [ ] Confirm all checks passed
- [ ] Add version label prior to acceptance

PR #919, opened 2022-01-25
jruby 9.3 fixes

ahorek:
# Overview
this PR fixes compatibility with jruby 9.3+

usually, programs don't need to access ```JRuby.runtime``` directly, so it's no longer available automatically since jruby 9.3

for special cases like accessing object space, we need to load the runtime explicitly. This change is backward compatible.

# Related Github Issue
https://github.com/newrelic/newrelic-ruby-agent/issues/788 https://github.com/newrelic/newrelic-ruby-agent/pull/784

# Testing
the current CI on this project is more complex than I originally thought, many tested libraries are very outdated, rails 3 doesn't work well on jruby 9.3 anymore and I think there's one new bug in jruby's autoload... so CI switch should be done separately.

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/language_support.rb
+++ b/lib/new_relic/language_support.rb
@@ -22,7 +22,8 @@ module NewRelic
     end
 
     def object_space_usable?
-      if defined?(::JRuby) && JRuby.respond_to?(:runtime)
+      if jruby?
+        require 'jruby'
         JRuby.runtime.is_object_space_enabled
       else
         defined?(::ObjectSpace)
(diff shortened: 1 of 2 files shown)


kaylareopelle:
Thank you for submitting the fix in this PR and the others for JRuby support! Our team will review the PRs today and get back to you with any suggestions.

kaylareopelle (approved):
Thanks for explaining the reasoning behind this change in the description of the PR!

Issue #902, opened 2022-01-15
Update ActiveRecord instrumentation for Ruby 3+ compatibility


kaylareopelle:
### Feature Description
Our multiverse tests for ActiveRecord do not run on Ruby 3+. Let's address the test failures for this version and make code changes as needed so that our instrumentation remains up to date.

### Priority
Really Want

kaylareopelle:
Look into updating the Envfile to use Ruby 3.0-approved ActiveRecord versions to resolve the failures.

kaylareopelle:
Functionality delivered in PR #923

PR #923, opened 2022-01-25
Run active record tests on ruby 3.0+

tannalynn:
# Overview
This PR adds a new multiverse suite `active_record_pg` that runs the active record tests with postgres instead of mysql like the `active_record` suite uses. Mysql2 gem is difficult to get working on ruby 3.0+, so we moved to Postgres for all ruby 3.0+ active record tests. 

# Changes 
* Run jruby tests on Rails 5.0, 5.2, 6.0, 6.1 
* Add Postgres service to the multiverse CI actions
* Add activerecord-jdbcmysql-adapter and activerecord-jdbcsqlite-adapter to active record suite
* Refine Postgresql tests to support Rails 7.0+

# Related Github Issue
#902

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -161,6 +161,8 @@ jobs:
         env:
           DB_PORT: 3306
           MYSQL_PASSWORD: root
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_USERNAME: postgres
 
       - name: Annotate errors
         if: ${{ failure() }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,6 +158,8 @@ jobs:
         env:
           DB_PORT: 3306
           MYSQL_PASSWORD: root
+          POSTGRES_USERNAME: postgres
+          POSTGRES_PASSWORD: password
 
       - name: Annotate errors
         if: ${{ failure() }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,6 +93,9 @@ jobs:
     services:
       postgres:
         image: postgres:latest
+        env:
+          POSTGRES_USERNAME: postgres
+          POSTGRES_PASSWORD: password
         ports:
           - 5432:5432
         options: >-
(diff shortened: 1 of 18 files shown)


kaylareopelle:
**Note:** An earlier suggestion included reducing the duplication of files between the `active_record` and `active_record_pg` suites. Unfortunately, sharing a single test file raised errors with finding the database primary keys that I wasn't able to resolve. I'm happy to revisit this as a team if we'd like to continue efforts to reduce duplication.

.github/workflows/ci.yml
```diff
@@ -91,6 +91,18 @@ jobs:
     needs: build-ruby
     runs-on: ubuntu-18.04
     services:
+      postgres:
+        image: postgres:latest
+        env:
```
fallwith, line 96: If we dropped these credentials ("postgres"/"postgres") and went with the defaults ("postgres"/"password") instead, then I believe you could just eliminate all related env vars from the config (both from the service config and the app configs), assuming they are what gets used when database.yml ends up with blank values.

test/multiverse/lib/multiverse/runner.rb
```diff
@@ -107,6 +107,8 @@ def execute_suites(filter, opts)
     end
 
     if RUBY_VERSION >= '3.0'
+      # this suite uses mysql2 which has issues on ruby 3.0+
+      # a new suite, active_record_pg, has been added to run active record tests on ruby 3.0+
```
fallwith, line 111: 👍

test/multiverse/suites/active_record/Envfile
```diff
@@ -90,10 +90,12 @@ end
 # Only 3.2.22 fixes the above mentioned issue, so it's the only one 
 # surviving in our coverage of Rails 3.x series here.
 #
-if RUBY_VERSION <= '2.2.0'
+if RUBY_VERSION < '2.3'
+  mysql_vsn = '0.3.20'
```
fallwith, line 94: I'm not sure if 0.3.21 gives us anything we can leverage, but it's the version I've been testing with (latest stable 0.3.x)

fallwith (approved):
👍

fallwith (approved):

PR #942, opened 2022-02-10
Bugfix: Puma "can't add a new key into hash during iteration"

### LINKED ISSUE #933: The old error "can't add a new key into hash during iteration"
The same [issue](https://github.com/newrelic/newrelic-ruby-agent/issues/573) happens and wasn't resolved yet.

## Description
When an exception is triggered and the puma runs in multiple threads and workers it throws `can't add a new key into hash during iteration` error. Yes, same `request.rb:79:in set_header'` as it was in [this](https://github.com/newrelic/newrelic-ruby-agent/issues/573) issue
```
Error during failsafe response: can't add a new key into hash during iteration
  .rbenv/versions/2.6.8/lib/ruby/gems/2.6.0/gems/rack-2.2.3/lib/rack/request.rb:79:in `set_header'
  .rbenv/versions/2.6.8/lib/ruby/gems/2.6.0/gems/actionpack-5.2.6/lib/action_dispatch/middleware/show_exceptions.rb:48:in `render_exception'
  .rbenv/versions/2.6.8/lib/ruby/gems/2.6.0/gems/actionpack-5.2.6/lib/action_dispatch/middleware/show_exceptions.rb:36:in `rescue in call'
  .rbenv/versions/2.6.8/lib/ruby/gems/2.6.0/gems/actionpack-5.2.6/lib/action_dispatch/middleware/show_exceptions.rb:31:in `call'
  .rbenv/versions/2.6.8/lib/ruby/gems/2.6.0/gems/newrelic_rpm-8.4.0/lib/new_relic/agent/instrumentation/middleware_tracing.rb:100:in `call'
  .rbenv/versions/2.6.8/lib/ruby/gems/2.6.0/gems/railties-5.2.6/lib/rails/rack/logger.rb:38:in `call_app'
  .rbenv/versions/2.6.8/lib/ruby/gems/2.6.0/gems/railties-5.2.6/lib/rails/rack/logger.rb:26:in `block in call'
  .rbenv/versions/2.6.8/lib/ruby/gems/2.6.0/gems/activesupport-5.2.6/lib/active_support/tagged_logging.rb:71:in `block in tagged'
  .rbenv/versions/2.6.8/lib/ruby/gems/2.6.0/gems/activesupport-5.2.6/lib/active_support/tagged_logging.rb:28:in `tagged'
  .rbenv/versions/2.6.8/lib/ruby/gems/2.6.0/gems/activesupport-5.2.6/lib/active_support/tagged_logging.rb:71:in `tagged'
```

## Expected Behavior
The error shouldn't happen

## Steps to Reproduce
I created [a brand new Rails app](https://github.com/sasharevzin/newrelic_app) (added master.key as well) and can 100% time reproduce the issue.
1. Start with a production environment like `RAILS_ENV=production bundle exec rails s`
2. Monitor logs by running `tail -f log/production.log`
3. Go to http://localhost:3000
4. You'll get the page `we are sorry...`
5. Now refresh the page multiple times by hitting `COMMAND + R` in the Chrome browser

## Your Environment
It's a brand new rails app with only `newrelic_rpm` gem on it
```
ruby '2.6.8'
rails', '~> 5.2.4'
newrelic_rpm (8.4.0)
```

tannalynn:

# Overview
It was brought to our attention that this error continues to occur despite previous attempts to prevent this error. I believe I have identified the root cause and prevented this from occurring in the future. I was no longer able to see the error happening using the reproduction provided. 

During my investigation, I found similar issues in a couple other gems:
https://github.com/rack/rack/issues/589
https://github.com/getsentry/sentry-ruby/issues/1183
https://github.com/airbrake/airbrake/issues/233

# Related Github Issue
closes #933 (original #573)

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/instrumentation/middleware_tracing.rb
+++ b/lib/new_relic/agent/instrumentation/middleware_tracing.rb
@@ -106,7 +106,7 @@ module NewRelic
 
             result
           rescue Exception => e
-            # NewRelic::Agent.notice_error(e)
+            NewRelic::Agent.notice_error(e)
             raise e
           ensure
             finishable.finish if finishable
--- a/lib/new_relic/agent/instrumentation/middleware_tracing.rb
+++ b/lib/new_relic/agent/instrumentation/middleware_tracing.rb
@@ -45,7 +45,7 @@ module NewRelic
           opts[:apdex_start_time] = QueueTime.parse_frontend_timestamp(env)
           # this case is for the rare occasion that an app is using Puma::Rack
           # without having ::Rack as a dependency
-          opts[:request] = ::Rack::Request.new(env) if defined? ::Rack
+          opts[:request] = ::Rack::Request.new(env.dup) if defined? ::Rack
           opts
         end
 
@@ -106,7 +106,7 @@ module NewRelic
 
             result
           rescue Exception => e
-            NewRelic::Agent.notice_error(e)
+            # NewRelic::Agent.notice_error(e)
             raise e
           ensure
             finishable.finish if finishable
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
   ## v8.5.0
 
+  * **Bugfix: Rails 5 + Puma errors in rack "can't add a new key into hash during iteration"**
+  
+    When using rails 5 with puma, the agent would intermittently cause rack to raise a `RuntimeError: can't add a new key into hash during iteration`. We have identified the source of the error in our instrumentation and corrected the behavior so it no longer interferes with rack. Thanks to @sasharevzin for bringing attention to this error and providing a reproduction of the issue for us to investigate.
+
   * **Eliminated warnings for redefined constants in ParamaterFiltering**
 
     Fixed the ParameterFiltering constant definitions so that they are not redefined on multiple reloads of the module. Thank you to @TonyArra for bringing this issue to our attention.


fallwith (approved):
👍

kaylareopelle (approved):
🌟

lib/new_relic/agent/instrumentation/middleware_tracing.rb
```diff
@@ -45,7 +45,7 @@ def merge_first_middleware_options(opts, env)
           opts[:apdex_start_time] = QueueTime.parse_frontend_timestamp(env)
           # this case is for the rare occasion that an app is using Puma::Rack
           # without having ::Rack as a dependency
```
kaylareopelle, line 47: Out of curiosity, did this pan out as being true in your research? (Were all the sample apps ones that used Puma::Rack, without having ::Rack as a dependency?) I'm wondering how true this comment is today.

tannalynn: Hmm, i'm not sure. honestly i didn't really look into using Puma::Rack without ::Rack much. It looks like something that might just be an edge case? I believe Rack has to be actually be defined for this error to happen though, otherwise that line would just be skipped entirely, avoiding the error.

kaylareopelle: Thanks for clarifying!

n-ikhil:
@tannalynn  I am facing similar issue, with stacktrace pointing to aws-sdk-core gem (we also use newrelic). Can you help me understand how this fixes the bug, so that i can debug my issue ?
Is the env object being iterated in caller, and current execution is in separate thread ?

n-ikhil:
@tannalynn  I am facing similar issue, with stacktrace pointing to aws-sdk-core gem (we also use newrelic). Can you help me understand how this fixes the bug, so that i can debug my issue ?
Is the env object being iterated in caller, and current execution is in separate thread ?

tannalynn:
> @tannalynn I am facing similar issue, with stacktrace pointing to aws-sdk-core gem (we also use newrelic). Can you help me understand how this fixes the bug, so that i can debug my issue ?
> Is the env object being iterated in caller, and current execution is in separate thread ?

So for this bug in this particular context, what was happening was that the agent was grabbing the request headers from rack so that we could report the data as part of the transaction. Later on, we are iterating over this value as part of how we record this data. Most of the time this was fine, but sometimes during our iterating another thread in the application would attempt to insert a new value into the header, which would cause the error `can't add a new key into hash during iteration`. 

Since rack inserting a header was something that should be fine, we knew that the thread in the agent doing the iteration was what we needed to make changes to. It was very tricky to track down because the stack trace given with the error only pointed to the location in rack that was inserting a header, so we had no idea where the other part of the issue, the iteration, was happening. 

Since I knew the object that the issue was with was request headers from rack, I searched through the agent code for the places where we interact with the request headers to find where we were shallow copying that object. Once the correct location was found (it did take a while though lol), I was able to simply add `.dup` so we could create a deep copy that we could iterate on without any fear of causing issues in rack.

The agent doesn't have any instrumentation specific to the aws-sdk-core gem, so I'm not sure how the agent would be part of the equation there. Though it is still possible if the agent and the aws-sdk-core gem are both grabbing references to the same object from somewhere else. 

Hopefully that helps! Please let me know if there is anything else I can do to help.

n-ikhil:
@tannalynn  Thank you for your time and detailed explanation ! I will try to figure out the cause of error in our case and possibly share it.

PR #952, opened 2022-02-11
We have several configuration options for "analytic events", which are called transaction events. The term analytic events is an older term. We've updated the language for these configuration options to reference transaction events instead, and deprecated the older analytic event configs.  Note: While I did update all references to the deprecated configuration options in the main agent code, I did not rename other references to analytic events in the internal agent code (such as method/variable names).

tannalynn:
# Overview

We have several configuration options for "analytic events", which are called transaction events. The term analytic events is an older term. We've updated the language for these configuration options to reference transaction events instead, and deprecated the older analytic event configs.

Note: While I did update all references to the deprecated configuration options in the main agent code, I did not rename other references to analytic events in the internal agent code (such as method/variable names).

# Related Github Issue

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -652,8 +652,17 @@ module NewRelic
           :public => false,
           :type => Integer,
           :dynamic_name => true,
+          :deprecated => true,
           :allowed_from_server => true,
-          :description => 'Number of seconds betwixt connections to the New Relic analytic event collection services.'
+          :description => deprecated_description(:'event_report_period.transaction_event_data', 'Number of seconds betwixt connections to the New Relic transaction event collection services.')
+        },
+        :'event_report_period.transaction_event_data' => { 
+          :default => value_of(:'event_report_period.analytic_event_data'),
+          :public => false,
+          :type => Integer,
+          :dynamic_name => true,
+          :allowed_from_server => true,
+          :description => 'Number of seconds betwixt connections to the New Relic transaction event collection services.'
         },
         :'event_report_period.custom_event_data' => {
           :default => 60,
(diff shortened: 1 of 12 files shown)


kaylareopelle (approved):
This looks good to me, Tanna. The only code-based change that comes to mind is to add a positive test case to verify config callbacks will work with procs in the future.

It may also be worth adding a sentence to the deprecation issue we discussed on the call earlier today with a line item to replace references to analytics events with transaction events inside the code to prevent future confusion about what analytics events are.

Thanks for your innovation! 🧠

fallwith (approved):

tannalynn:
> This looks good to me, Tanna. The only code-based change that comes to mind is to add a positive test case to verify config callbacks will work with procs in the future.

Added this test! Thanks!

> It may also be worth adding a sentence to the deprecation issue we discussed on the call earlier today with a line item to replace references to analytics events with transaction events inside the code to prevent future confusion about what analytics events are.

I added the new issue for removing deprecated configs. I'm not sure we can remove all of the references to analytic events in the internal agent code because that's still what the endpoint is called and other things related to the server.

kaylareopelle:
>I added the new issue for removing deprecated configs. I'm not sure we can remove all of the references to analytic events in the internal agent code because that's still what the endpoint is called and other things related to the server.

@tannalynn  - I didn't realize analytic events was still referenced in the overall structure of the system! No need to remove it then. :)

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
   ## v8.5.0
 
+  * **Added updated configuration options for transaction events and deprecate previous configs**
+    This release deprecates and replaces the following configuration options
```
kaylareopelle, line 6:
```suggestion
    This release deprecates and replaces the following configuration options:
```

kaylareopelle (approved):
🌟

CHANGELOG.md
kaylareopelle: That's ok! No need to go back and change it now. We usually review the Changelog for little things like that before a release anyway. 😄

PR #954, opened 2022-02-11
NewRelic::VERSION optimisations

fallwith:
when defining `NewRelic::VERSION::STRING`, don't bother loading a file
that won't be there, rescuing, compacting an array, splatting an arg
list, joining an array, etc.  Just build a string.

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


tannalynn (approved):
LGTM! Thanks for getting this updated since we stopped using build numbers a while ago! 🎉

PR #956, opened 2022-02-14
Add rubocop step to run first in ci.yml

tannalynn:

# Overview
Runs rubocop first before continuing with all the test suites

# Related Github Issue

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/setup-ruby@v1
         with:
           ruby-version: '2.7'
-      - run: gem install rubocop
+      - run: bundle
       - run: rubocop
 
   build-ruby:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,19 @@ on:
     - cron:  '0 9 * * *'
 
 jobs:
+  run-rubocop:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '2.7'
+      - run: gem install rubocop
+      - run: rubocop
+
   build-ruby:
+    needs: run-rubocop
     runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


kaylareopelle (approved):
Nice work! Thank you for getting this set up!

.github/workflows/ci.yml
```diff
@@ -10,7 +10,19 @@ on:
...
 
 jobs:
+  run-rubocop:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '2.7'
```
fallwith, line 20: @tannalynn What do you think about having this match the `TargetRubyVersion` defined in `.rubocop.yml`? It's currently set to 2.5. Perhaps keep this at 2.7 and bump up the `.rubocop.yml` one?

tannalynn: Alright, I updated the .rubocop.yml to 2.7! thank you!

fallwith (approved):
👍

.github/workflows/ci.yml
kaylareopelle: I really like the ideas of these matching. Nice recommendation, James! Is this something we could set through a constant shared between the two files to reduce the toil of updating the version?

fallwith: RuboCop will send `.rubocop.yml` through an ERB parse before handing the content off to YAML, so there we could do something dynamic, such as:

```yaml
  TargetRubyVersion: <%= ENV.fetch('RUBY_VERSION', '2.7') %>
```

On the `ci.yml` side while trying to share an approach with RuboCop, we face these issues:

* The `ci.yml` parser won't have access to the source code locally
* The `ci.yml` content will be ran through a JavaScript parse instead of a Ruby one

I suppose we could take the approach I did with `test/helpers/ruby_rails_mappings.rb`, which uses `ci.yml` as a single source of truth. We could define the Ruby version once in `ci.yml` and then have RuboCop parse it out of there.

kaylareopelle: I think I'd rather move forward with merging this in, but keep this improvement in mind for the future. Thanks for walking through some of the options on how this might be achieved.

kaylareopelle (approved):

PR #960, opened 2022-02-15
Update version to 8.5.0

tannalynn:
Update version number and changelog for upcoming release.


fallwith (approved):
👍

kaylareopelle (commented):
Thanks for kicking off the release process! I have a few questions about the CHANGELOG.

1. Should we add an entry to the CHANGELOG about creating tests for postgres support within Active Record and/or Ruby 3? 

2. Could the CHANGELOG be reorganized so that anything we're considering a bugfix starts with "Bugfix" and comes after any new features? 

3. Could we add some entries to represent the work ahorek did on JRuby? 

4. Could we call out the changes that were made during innovation week to deprecate `analytics_events` settings? I think we have in our workflow document that deprecations should appear first on the changelog. I'm not sure why I don't see those PRs on our Ruby Engineering board in the "Done" column. We might want to double-check GitHub to see which PRs have been merged since our last release to make sure nothing else is missing.

Thanks!

kaylareopelle (approved):
Thanks for expanding the CHANGELOG, @fallwith!

PR #963, opened 2022-02-17
Docker updates

fallwith:
With the decision made to continue to use standard GitHub Actions images
for CI, we are now free to leverage official Ruby Docker images for
local development. The following Docker related changes have been made:

* Use official `ruby` images
* Allow both `Dockerfile` and `docker-compose.yml` to accept a Ruby
  version value
* Update `DOCKER.md`
    * With information about specifying the desired Ruby version
    * With a new quick start guide at the top
    * Updated `brew` usage syntax
    * Recommend the simpler 'rpm' as a tag value
    * Suggest the use of `--rm` with `docker run`
* Standardize on MySQL 5.7
    * Used by our GitHub Actions CI
    * Some multiverse tests (older versions of frameworks?) seem to
      prefer it
* Update the newly added `active_record_pg` multiverse suite for Docker
  compatibility
* Update the `ruby_rails_mappings.rb` helper to avoid `Hash#dig`,
  which isn't¬ available on older Rubies (introduced in v2.3).
* Remove the Docker specific `ruby_installer.rb` script (GitHub Actions
  uses JS)

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


DOCKER.md
```diff
@@ -40,27 +54,35 @@ Docker Compose and the project `docker-compose.yml` file will be needed to
...
+```shell
+docker build --build-arg ruby_version=2.7 .
+```
+
+**Legend:**
+* `build -t <TAG>` applies a tag to the image during building.
 * `.` indicates "here" and tells Docker that the `Dockerfile` file can be found
   in the current directory
-* `run -it` tells Docker to run interactively (`i`) and with a pseudo TTY (`t`)
+* `run --rm` tells Docker to remove the container after the tests complete.
```
kaylareopelle, line 79: Could you add an entry to the legend for `--build-arg` for completeness? I think that's the only switch that's missing

DOCKER.md
```diff
@@ -72,6 +94,15 @@ containers with Docker Compose:
 $ docker-compose up
 ```
 
+By default, `docker-compose` will use the default Ruby version specified in the
+`Dockerfile` file. To override this version with a custom desired version, set
+the `RUBY_VERSION` environment variable before calling `docker-compose`,
+like so:
+
+```shell
+RUBY_VERSION=3.0 docker-compose up
```
kaylareopelle, line 103: I notice that the plain ol' Docker command uses lowercase `ruby_version`, but the Docker Compose command here uses uppercase `RUBY_VERSION`. Is there a reason behind the different cases?

DOCKER.md
```diff
@@ -117,6 +148,13 @@ ran after the `docker-compose up` process has been stopped. All relevant
 containers and images can then be optionally discarded using the `docker` CLI
 commands described in the previous paragraph.
 
+Use `docker ps -a` to show a list of all containers. Pass a container id to
+`docker rm` (ex: `docker rm 5c15ee2f1c4f`) to remove it.
+
+Use `docker images` to show a list of all images. Typically, you'll want to
+keep these images if you plan on running Docker with them again in the future.
+If you are done with them, you can pass an image id to `docker rmi` to remove
+an image (ex: `docker rmi 4253856b2570`).
```
kaylareopelle, line 157: As an aside, I'm so thrilled with this documentation. This is something I wish I had as an early career developer interested in making contributions to open source projects, but still learning the basics with Docker. Thanks for taking the time to break this down!

Dockerfile
```diff
@@ -1,105 +1,10 @@
...
-    && gem update bundler \
-    && bundle install \
-    && gem install bundler:1.17.3
-
-# TODO: re-enable for multi-ruby testing
-# This script will install all CI supported Rubies (can take an hour to complete)
-# RUN ruby test/script/ruby_installer.rb
-
-CMD [ "bundle", "exec", "rake" ]
+CMD ["bundle", "exec", "rake"]
```
kaylareopelle, line 10: the simplicity here is BEAUTIFUL. 🤩 gotta love that ruby image.

docker-compose.yml
```diff
@@ -1,91 +1,85 @@
...
     restart: always
     volumes:
       - /data/db
-    ports:
-      - "27017:27017"
+    expose:
+      - "27017"
   postgres:
-    image: postgres:14.1
+    image: postgres:14.2
```
kaylareopelle, line 36: What are your thoughts about pulling directly from `postgres:latest` instead?

docker-compose.yml
```diff
@@ -1,91 +1,85 @@
...
-      - "6379:6379"
+    expose:
+      - "6379"
   app:
-    build: .
-    image: newrelic/newrelic-ruby-agent-tester
+    build:
+      context: .
+      args:
+        - ruby_version=${RUBY_VERSION}
```
kaylareopelle, line 56: ahh, I think this answers my `ruby_version` vs `RUBY_VERSION` question from earlier... but I don't want to assume, so I'll leave the earlier comment.

docker-compose.yml
```diff
@@ -1,91 +1,85 @@
...
+    image: rpm
```
kaylareopelle, line 57: I like this simplicity here. I wonder if we should go straight for the gem name in that case? `newrelic_rpm`? I'm already getting it confused with RPM site.

kaylareopelle (approved):
Thanks for closing the loop on this issue. These changes look great. I have a few questions in the code, but they're quite small and this would be fine to merge without addressing them.

DOCKER.md
fallwith: Great idea. Implemented with [ca0c5e3](https://github.com/newrelic/newrelic-ruby-agent/pull/963/commits/ca0c5e337b2b032cf1dfdbe84a924fe23a89ba35)

DOCKER.md
fallwith: From what I understand (an understanding that may very well be incomplete or outdated), the `docker` CLI command can be given build arguments via `--build-arg`, but the `docker-compose` command can't be given something similar but can process environment variables. This is why when we use `docker` directly, we can use `ruby_version` but when building the app image via `docker-compose` we need to specify the `RUBY_VERSION` env var which in turn gets used by `docker-compose` as a build argument.

We can update this if we end up coming across a related best practice or if something gets added to a new Docker Compose syntax version.

DOCKER.md
fallwith: My pleasure! I was delighted to see this!

Dockerfile
fallwith: Agreed!

docker-compose.yml
fallwith: My take: we should only commit the specific tags that we've personally tested with and acknowledged as working. I'd be more inclined to switch to "latest" if the latest version was actively being brought down and tested by an automated process (CI for PRs or whatnot) and we could at least have some proactive heads up when a breaking change is introduced.

I'm very happy to consider benefits to going with "latest" though, let me know.

As far as this bump here from 14.1 -> 14.2, it wasn't actually necessary in the end. The issue I was seeing turned out to be due to the `database.yml` changes needed in this PR. But after doing so much testing on 14.2 I went ahead and left it.

docker-compose.yml
fallwith: Yep, `docker-compose` will parse out env vars in `docker-compose.yml` but from what I can tell can't be passed key/val pairs for this type of usage via CLI arguments. So here an env var is passed to Docker Compose and in turn being used to pass a build arg to Docker.

docker-compose.yml
fallwith: Good call. We want to disambiguate from RedHat too! Addressed with [ca0c5e3](https://github.com/newrelic/newrelic-ruby-agent/pull/963/commits/ca0c5e337b2b032cf1dfdbe84a924fe23a89ba35)

DOCKER.md
kaylareopelle: Thanks for explaining this! 😄

docker-compose.yml
kaylareopelle: I completely agree. The main benefits I considered were avoiding minor bump image updates since it seemed like the latest version was regularly being adopted. I agree that we should be testing these things before the versions change, especially because people outside our organization may be using this tool and I'd hate to have it break on them.

docker-compose.yml
kaylareopelle: Thanks!

PR #967, opened 2022-02-17
remove the echo statements from the label workflow

fallwith:
Remove the echo statements from the label workflow as per [GitHub's
recommendation](https://securitylab.github.com/advisories/GHSL-2020-183-checkstyle-checkstyle-workflow/)


angelatan2 (approved):
Looks good to me.

PR #970, opened 2022-02-18
AWS: include a token with IMDS metadata requests

### LINKED ISSUE #552: AWS metadata request not compatible with v2
The metadata requests fail on AWS if instance metadata v1 is disabled on the host.

## Description
When my app boots up the New Relic gem fires off requests to all known cloud provider metadata endpoints. The request to AWS uses the deprecated v1 request method. It is a best practice to not only use the v2 service but to disable access to v1 on hosts.

See:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/utilization/aws.rb

## Expected Behavior
I expect New Relic to use the v2 metadata service.

## Steps to Reproduce
Disable instance metadata v1 on your host and then start an app using any version of the New Relic gem.

## Your Environment
My hosts are in AWS and have instance metadata v1 disabled

## Additional context
For now I'm disabling requests to the metadata service completely to avoid the errors.

fallwith:
When fetching instance metadata from the AWS IMDS endpoint, first
generate an API token and then include that token in the metadata
request's headers.

With this change:

* The fetching of metadata is more secure
* Users can now configure their instances to require tokens for API
calls without breaking the agent's metadata fetching functionality

resolves #552

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,19 @@ on:
     - cron:  '0 9 * * *'
 
 jobs:
+  run-rubocop:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '2.7'
+      - run: bundle
+      - run: rubocop
+
   build-ruby:
+    needs: run-rubocop
     runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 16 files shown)


kaylareopelle (dismissed):
Thanks for digging into this!

kaylareopelle (approved):
One more try! 🤞

PR #980, opened 2022-02-22
C debug

fallwith:
CI debug


fallwith:
Debugging complete. Closing.

Lessons learned:

* AWS and Azure use the same "IMDS" acronym for metadata and they both use the same IP address for querying IMDS
* When unit tests are ran via GitHub Actions, we need to account for the possibility that the Azure IMDS host will be reachable and that calls intended for AWS IMDS are made against it, they will fail. These failures need to be anticipated (debug instead of error).

PR #987, opened 2022-02-24
performance test enhancements

fallwith:
enhancements for the performance tests:

* `redis` is a required gem, so add it to `Gemfile` and alphabetize the
  list
* Make the baseline output file location configurable via an environment
  variable while still defaulting to the user's home directory
* Add a `--markdown` / `-M` flag to generate the comparison table in
  Markdown format. Usage: `script/runner -C --markdown`

### DIFF
--- a/test/performance/Gemfile
+++ b/test/performance/Gemfile
@@ -1,9 +1,10 @@
 source 'https://rubygems.org'
 
 gem 'newrelic_rpm', :path => '../../..'
-gem 'rack'
-gem 'mocha'
 
-gem 'stackprof'
+gem 'mocha'
 gem 'pry'
 gem 'pry-nav'
+gem 'rack'
+gem 'redis'
+gem 'stackprof'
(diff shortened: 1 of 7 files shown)


fallwith:
I had a fourth enhancement I wanted to add, which is to have the runner output some progress info during the (rather lengthy) test run to give some hint as to its progress. Unfortunately given how the outer `script/runner` (non-inline) backticks other `script/runner` processes (inline) and relies on STDOUT, the progress info I was testing with was being misinterpreted as test suite output. I then tried using STDERR but had issues there as well.

kaylareopelle (approved):
Excellent enhancement! Thanks for cleaning this up, James!

kaylareopelle:
> I had a fourth enhancement I wanted to add, which is to have the runner output some progress info during the (rather lengthy) test run to give some hint as to its progress. Unfortunately given how the outer `script/runner` (non-inline) backticks other `script/runner` processes (inline) and relies on STDOUT, the progress info I was testing with was being misinterpreted as test suite output. I then tried using STDERR but had issues there as well.

Perhaps this could be made into an issue for a future technical debt milestone?

PR #993, opened 2022-02-24
v8.5.0 CHANGELOG reordering

fallwith:
* reorder v8.5.0 changelog to list general, then bugs, then commit
* added missing period


kaylareopelle (approved):
Thanks for updating the Changelog! 📝

PR #994, opened 2022-02-25
config docs: tweak HTML output formatting

fallwith:
Update the `rake newrelic:config:docs[html]` target:

* Eliminate all trailing whitespace
* Separate multiple sentences with a single space, not two

These two formatting issues are responsible for ~60% of the diff noise
we encounter when updating the docs site repo.


kaylareopelle (approved):
This looks great, James. Thanks for putting in this fix. I spent a few minutes looking at the diff generated with these changes and came up with #995 to get the rake task a little closer to parity.

PR #997, opened 2022-02-28
Logging: Apply attribute value limit on messages

kaylareopelle:
# Overview
Adds a private method, `#truncate_message` to shorten messages to the maximum allowed bytes, 32 * 1024 (1 kibibyte). This method is called when an event is created if the message's bytesize exceeds the limit.

# Related Github Issue
Closes #997

# Testing
Adds unit tests for truncating the message if over the bytes and leaving the message as-is if equal to or less than the required bytes.

### DIFF
--- a/lib/new_relic/agent/log_event_aggregator.rb
+++ b/lib/new_relic/agent/log_event_aggregator.rb
@@ -22,7 +22,7 @@ module NewRelic
       METRICS_SUPPORTABILITY_FORMAT = "Supportability/Logging/Metrics/Ruby/%s".freeze
       FORWARDING_SUPPORTABILITY_FORMAT = "Supportability/Logging/Forwarding/Ruby/%s".freeze
       DECORATING_SUPPORTABILITY_FORMAT = "Supportability/Logging/LocalDecorating/Ruby/%s".freeze
-      MAX_BYTES = 32768 # 32 * 1024 = 1 kibibyte
+      MAX_BYTES = 32768 # 32 * 1024 bytes (32 kibibytes)
 
       named :LogEventAggregator
       buffer_class PrioritySampledBuffer
@@ -91,7 +91,7 @@ module NewRelic
       end
 
       def create_event priority, formatted_message, severity
-        formatted_message = truncate_message(formatted_message) if formatted_message.bytesize > MAX_BYTES
+        formatted_message = truncate_message(formatted_message)
 
         event = LinkingMetadata.append_trace_linking_metadata({
           LEVEL_KEY => severity,
@@ -208,7 +208,7 @@ module NewRelic
       end
 
       def truncate_message(message)
-        return if message.bytesize <= MAX_BYTES
+        return message if message.bytesize <= MAX_BYTES
         message.byteslice(0...MAX_BYTES)
       end
     end
--- a/lib/new_relic/agent/log_event_aggregator.rb
+++ b/lib/new_relic/agent/log_event_aggregator.rb
@@ -22,6 +22,7 @@ module NewRelic
       METRICS_SUPPORTABILITY_FORMAT = "Supportability/Logging/Metrics/Ruby/%s".freeze
       FORWARDING_SUPPORTABILITY_FORMAT = "Supportability/Logging/Forwarding/Ruby/%s".freeze
       DECORATING_SUPPORTABILITY_FORMAT = "Supportability/Logging/LocalDecorating/Ruby/%s".freeze
+      MAX_BYTES = 32768 # 32 * 1024 = 1 kibibyte
 
       named :LogEventAggregator
       buffer_class PrioritySampledBuffer
@@ -90,6 +91,8 @@ module NewRelic
       end
 
       def create_event priority, formatted_message, severity
+        formatted_message = truncate_message(formatted_message) if formatted_message.bytesize > MAX_BYTES
+
         event = LinkingMetadata.append_trace_linking_metadata({
           LEVEL_KEY => severity,
           MESSAGE_KEY => formatted_message,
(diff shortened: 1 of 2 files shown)


kaylareopelle:
Planning to fix the branching issues after lunch. Ideally, there's only one commit for this branch, [ef6f84](https://github.com/newrelic/newrelic-ruby-agent/pull/997/commits/ef6f848706ca40e09c42677830d941b3703a5948)

kaylareopelle:
Blocked by original logging PR being merged into dev

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -22,6 +22,7 @@ class LogEventAggregator < EventAggregator
       METRICS_SUPPORTABILITY_FORMAT = "Supportability/Logging/Metrics/Ruby/%s".freeze
       FORWARDING_SUPPORTABILITY_FORMAT = "Supportability/Logging/Forwarding/Ruby/%s".freeze
       DECORATING_SUPPORTABILITY_FORMAT = "Supportability/Logging/LocalDecorating/Ruby/%s".freeze
+      MAX_BYTES = 32768 # 32 * 1024 = 1 kibibyte
```
fallwith, line 25: The way this comment is written might confuse some.

This part is helpful:

```ruby
32768 # 32 * 1024
```

This part is helpful:

```ruby
1024 = 1 kibibyte
```

But when both of these parts are combined, it looks like we're saying that 32,768 bytes equals 1 kibibyte (whereas it's really 32 kibibytes).

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -90,6 +91,8 @@ def record_batch txn, logs
       end
 
       def create_event priority, formatted_message, severity
+        formatted_message = truncate_message(formatted_message) if formatted_message.bytesize > MAX_BYTES
```
fallwith, line 94: I think we should drop the conditional aspect (`if formatted_message.bytesize > MAX_BYTES`) here. The new `truncate_message` method performs a very similar check. As things are currently, we are always performing one check and sometimes two. If we get rid of this conditional here, we'll only ever be performing one check.

We could also have `truncate_message` be destructive and then just have this one line be `truncate_message(formatted_message)`.

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -203,6 +206,11 @@ def record_supportability_metrics total_count, captured_count, dropped_count
         NewRelic::Agent.increment_metric(SEEN_METRIC, total_count)
         NewRelic::Agent.increment_metric(SENT_METRIC, captured_count)
       end
+
+      def truncate_message(message)
+        return if message.bytesize <= MAX_BYTES
+        message.byteslice(0...MAX_BYTES)
```
fallwith, line 212: (not a trick question - I don't know the answer) What does `#byteslice` give us here over `#truncate` (`message.truncate(MAX_BYTES)`)?

test/new_relic/agent/log_event_aggregator_test.rb
```diff
@@ -265,5 +265,27 @@ def test_basic_conversion_to_melt_format
       assert_equal 1, size
       assert_equal expected, payload
     end
+
+    def test_create_event_truncates_message_when_exceeding_max_bytes
+      right_size_message = String.new("a" * LogEventAggregator::MAX_BYTES)
+      message = right_size_message + 'b'
+      event = @aggregator.create_event(1, message, 'INFO')
+
+      assert_equal(right_size_message, event[1]["message"])
```
fallwith, line 274: It's great that you're not only checking the size but the content as well!

fallwith (changes requested):

lib/new_relic/agent/log_event_aggregator.rb
kaylareopelle: That's my bad, I forgot that the limit was 32 kb. Would you like me to change the text to show 32 kibibytes instead of 1 kibibyte?

lib/new_relic/agent/log_event_aggregator.rb
kaylareopelle: Great question! Truncate is a Rails-created string extension. Byteslice is pure Ruby.

lib/new_relic/agent/log_event_aggregator.rb
kaylareopelle: Thanks for noticing this! I meant to get rid of one of these based on a performance test before I pushed this up. I haven't run said test yet. I will and remove whichever one is slower. 

If they're about the same, do you have a preference for one over the other?

lib/new_relic/agent/log_event_aggregator.rb
fallwith: Yeah, I think that would read better. Something like this:

```ruby
MAX_BYTES = 32768 # 32 * 1024 bytes (32 kibibytes)
```

lib/new_relic/agent/log_event_aggregator.rb
fallwith: I think `truncate_message` should be the one to do the check.

lib/new_relic/agent/log_event_aggregator.rb
fallwith: Ah! Thank you. Ok, well then we should definitely use a pure Ruby approach. But now that I know that `#truncate` is an ActiveSupport-ism, I took a look at `#truncate`'s source and found that it does `string[0, MAX]` instead of our `0...MAX` approach. Now I want test the two approaches! But I won't let that hold up approval.

kaylareopelle: Nice! I hesitate to use that approach because it relies on character length rather than bytesize. Though the two are often the same, the true limit is on the number of bytes, so I'd rather lean on specificity unless there's a big performance difference.

https://ruby-doc.org/core-3.1.1/String.html#method-i-length

fallwith (approved):

Issue #988, opened 2022-02-24
Add log decoration support


kaylareopelle:
_Duplicate of [LOGGING-7303: Local Log Decoration (Internal)](https://newrelic.atlassian.net/browse/LOGGING-7303)_

Specifications described in [Agent Logs-in-Context Decoration from Engineering Brief: Logging in Agents (Internal)](https://docs.google.com/document/d/1HG9RD0NAYiLtR8ZFfH0obLbVZiVmQxOt3k8xBz-RZaQ/edit#heading=h.xgjthpm746nl)

Consider **Option A: Direct Injection** as the selected strategy.  This involves appending the following onto the log line itself: `NR-LINKING | {entity.guid} | {hostname} | {trace.id} | {span.id} |` if the setting is enabled.

Adding the config option to enable local log decoration should be completed as part of this ticket if it is not already complete with #983 

Log decoration should be allowed in high security mode and on by default.

As this is getting completed after #978, adding functionality to disable log decoration when `application_logging.enabled` is false is also within the scope of this issue.

kaylareopelle:
The values needed to provide the log decorating data are accessible through the LinkingMetadata class added here: https://github.com/jasonrclark/newrelic-ruby-agent/commit/66cbdd38d203cbc21d5344ecca193c446c8b55ae

kaylareopelle:
Closed by #1101

PR #1007, opened 2022-03-03
Logging: Create overall application logging enabled flag

### LINKED ISSUE #978: Create overall application_logging.enabled flag
_Duplicate of [LOGGING-7293: Overall application_logging.enabled flag (Internal)](https://newrelic.atlassian.net/browse/LOGGING-7293)_

Add a configuration option to  `lib/new_relic/agent/configuration/default_source.rb` named `application_logging.enabled` toggles the enablement of all new features implemented by the logging initiative. 

This behavior is similar to that of `transation_tracer.enabled`. It should turn off logging metrics, log forwarding and log decorating when set to false. It should allow logging metrics, forwarding, and decorating to be enabled based on their settings when set to true. Supportability metrics related to logging should still be sent regardless.

See [Engineering Brief: Logging in Agents - Configuration section](https://docs.google.com/document/d/1HG9RD0NAYiLtR8ZFfH0obLbVZiVmQxOt3k8xBz-RZaQ/edit#heading=h.n9gfg79u8nzd) for implementation details.

_Note: Log decorating has not been enabled yet. The acceptance criteria related to the overall flag toggling log decorating will be moved to that ticket, #988_

kaylareopelle:
# Overview

Adds `application_logging.enabled` config option to toggle logging features
* The default value of `instrumentation.logger` is dependent on the default of `application_logging.enabled`
* Within the LogEventAggregator, the feature is `@enabled?` when both `instrumentation.logger` and `application_logging.enabled` have truthy values.
* When `application_logging.enabled` is false, all logging-related config options are disabled:
  * `application_logging.forwarding.enabled`
  * `application_logging.metrics.enabled`
  * The work to set `application_logging.local_decorating.enabled` with the overall toggle will be completed in #988
* When `application_logging.enabled` is false, supportability metrics will still be recorded
* When `application_logging.enabled` is true, all logging-related config options to use their own values to determine their enabled status

Adds supportability metric related to the overall toggle `Supportability/Logging/Ruby/Logger/{enabled | disabled}`. The spec states that the fourth argument should be the logging framework. The Ruby language has a built-in class called `Logger` the agent instruments that will always hold the framework value.

# Related Github Issue
Closes #978 

# Testing
Added unit tests to confirm the toggle works in the cases of:
* metric recording
* log event recording/harvest (forwarding)
* high security mode
* overall enabled/disabled functionality
* `instrumentation.logger` and `application_logging.enabled` relationship

# Q & A raised during development
* This setting is intended to relate to `instrumentation.logger`, added in v8.1.0 with logging metrics. Our instrumentation config options provide four strategies: `auto|prepend|chain|disabled`. The `auto`, `prepend`, and `chain` options would all map to a truthy value. The `disabled` option maps to a falsey value. This setting is not currently connected to `application_logging.enabled`, but likely should be. _Answer: There was a precedent with a method called #instrumentation_value_of that returned a string that matched the related setting's default value. This wasn't quite right for these purposes, so I created a new method #instrumentation_boolean_value_of_

* The EventAggregator class has a method, `enabled?` that should take the config option(s) assigned to `enabled_key` and return a boolean based on their setting. This is not working in the LogEventAggregator#record method after changing `enabled_key` from `application_logging.forwarding.enabled` to `application_logging.enabled`. _Answer: Add `NewRelic::Agent.config.notify_service_source_added` after the config changes so that the accurate value is returned from `enabled?`_

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -38,6 +38,12 @@ module NewRelic
         end
       end
 
+      def self.instrumentation_boolean_value_of(key)
+        Proc.new do
+          NewRelic::Agent.config[key] ? 'auto' : 'disabled'
+        end
+      end
+
       # Marks the config option as deprecated in the documentation once generated.
       # Does not appear in logs.
       def self.deprecated_description new_setting, description
@@ -1004,7 +1010,7 @@ module NewRelic
           :description => 'Controls auto-instrumentation of dalli gem for Memcache at start up.  May be one of [auto|prepend|chain|disabled].'
         },
         :'instrumentation.logger' => {
-          :default => "auto",
+          :default => instrumentation_boolean_value_of(:'application_logging.enabled'),
           :public => true,
           :type => String,
           :dynamic_name => true,
@@ -1865,7 +1871,7 @@ module NewRelic
           :dynamic_name => true
         },
         :'application_logging.enabled' => {
-          :default => true,
+          :default => false,
           :public => true,
           :type => Boolean,
           :allowed_from_server => false,
@@ -1890,7 +1896,7 @@ module NewRelic
           :default => true,
           :public => true,
           :type => Boolean,
-          :allowed_from_server => false,
+          :allowed_from_server => true,
           :description => 'If `true`, the agent captures metrics related to logging for your application.'
         },
         :disable_grape_instrumentation => {
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -38,7 +38,7 @@ module NewRelic
         end
       end
 
-      def self.instrumentation_boolean_value_of(key)
+      def self.instrumentation_value_from_boolean(key)
         Proc.new do
           NewRelic::Agent.config[key] ? 'auto' : 'disabled'
         end
@@ -1018,7 +1018,7 @@ module NewRelic
           :description => 'Controls auto-instrumentation of dalli gem for Memcache at start up.  May be one of [auto|prepend|chain|disabled].'
         },
         :'instrumentation.logger' => {
-          :default => instrumentation_boolean_value_of(:'application_logging.enabled'),
+          :default => instrumentation_value_from_boolean(:'application_logging.enabled'),
           :public => true,
           :type => String,
           :dynamic_name => true,
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1871,7 +1871,7 @@ module NewRelic
           :dynamic_name => true
         },
         :'application_logging.enabled' => {
-          :default => false,
+          :default => true,
           :public => true,
           :type => Boolean,
           :allowed_from_server => false,
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1864,6 +1864,13 @@ module NewRelic
           :description => 'Defines the maximum number of span events reported from a single harvest. Any Integer between 1 and 10000 is valid.',
           :dynamic_name => true
         },
+        :'application_logging.enabled' => {
+          :default => true,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => 'If `true`, enables log decoration and the collection of log events and metrics.'
+        },
         :'application_logging.forwarding.enabled' => {
           :default => false,
           :public => true,
--- a/lib/new_relic/agent/instrumentation/logger.rb
+++ b/lib/new_relic/agent/instrumentation/logger.rb
@@ -24,7 +24,5 @@ DependencyDetection.defer do
     else
       chain_instrument NewRelic::Agent::Instrumentation::Logger
     end
-
-    ::NewRelic::Agent.increment_metric("Supportability/Logging/Ruby/Logger/enabled")
   end
 end
--- a/lib/new_relic/agent/instrumentation/logger.rb
+++ b/lib/new_relic/agent/instrumentation/logger.rb
@@ -9,7 +9,10 @@ require_relative 'logger/prepend'
 DependencyDetection.defer do
   named :logger
 
-  depends_on { defined?(::Logger) }
+  depends_on do
+    defined?(::Logger) &&
+      NewRelic::Agent.config[:'application_logging.enabled']
+  end
 
   executes do
     ::NewRelic::Agent.logger.info "Installing Logger instrumentation"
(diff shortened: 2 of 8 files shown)


lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -213,6 +230,10 @@ def truncate_message(message)
         return message if message.bytesize <= MAX_BYTES
         message.byteslice(0...MAX_BYTES)
       end
+
+      def instrumentation_logger_enabled?
+        NewRelic::Agent.config[:'instrumentation.logger'] != 'disabled'
```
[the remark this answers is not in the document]
kaylareopelle: I like that idea! I'll move things around 😄

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -50,9 +52,13 @@ def capacity
       end
 
       def record(formatted_message, severity)
-        @counter_lock.synchronize do
-          @seen += 1
-          @seen_by_severity[severity] += 1
+        return unless enabled?
+
+        if NewRelic::Agent.config[METRICS_ENABLED_KEY]
```
jasonrclark, line 57: Similar to how we memoize high security settings to a local bool in `register_for_done_configuring`, we could consider doing that as well for settings like this that are in the hot path (and other config checking below).

It's a bit ugly, but might be worth measuring whether it makes a difference or not in the microbenchmarks 🤔

kaylareopelle: In a microbenchmarking test on hash access vs. instance variable memoization, it looks like there's a difference that falls within the margin of error for these options. Something similar is reflected when comparing these approaches in the log instrumentation performance tests. 

Since there isn't a difference, do you have a preference, @jasonrclark?

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -153,6 +163,7 @@ def register_for_done_configuring(events)
         events.subscribe(:server_source_configuration_added) do
           @high_security = NewRelic::Agent.config[:high_security]
```
kaylareopelle, line 164: @jasonrclark  - I noticed this seems to be the same value as in the `#initialize` method. Do you remember why the instance variable was set twice?

lib/new_relic/agent/log_event_aggregator.rb
jasonrclark: Mild preference to just using the `config[]` if the data doesn't clearly support removing it. Thanks!

lib/new_relic/agent/log_event_aggregator.rb
jasonrclark: My intent there was that I didn't know if the server response could send a different value, so it seemed safest to grab the value after all the configs have settled (which is basically what this event represents).

kaylareopelle: Nice, thanks for clarifying!

lib/new_relic/agent/instrumentation/logger.rb
```diff
@@ -22,6 +25,6 @@
       chain_instrument NewRelic::Agent::Instrumentation::Logger
     end
 
-    ::NewRelic::Agent.increment_metric("Supportability/Logging/enabled/Ruby/Logger")
+    ::NewRelic::Agent.increment_metric("Supportability/Logging/Ruby/Logger/enabled")
```
kaylareopelle, line 28: This now seems like a duplicate of the metric recorded on `lib/new_relic/agent/log_event_aggregator.rb:166`. I like this location, but can't seem to find a path through the `DependencyDetection` class to increment the metric as `disabled` if `depends_on` fails.

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -174,7 +189,9 @@ def after_harvest metadata
       # To avoid paying the cost of metric recording on every line, we hold
       # these until harvest before recording them
       def record_customer_metrics
+        return unless enabled?
         return unless NewRelic::Agent.config[:'application_logging.metrics.enabled']
```
tannalynn, line 193: Metrics can be disabled separately? So someone could have logs enabled and sending them up but have the metrics for it disabled? That seems pretty different from most of what the agent usually does, so I'm just curious

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -38,6 +38,12 @@ def self.instrumentation_value_of(disable_key, prepend_key = nil)
         end
       end
 
+      def self.instrumentation_boolean_value_of(key)
```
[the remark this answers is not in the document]
kaylareopelle: These are great points! I was struggling with the name and like your suggestion, `instrumentation_value_from_boolean` a lot. I was trying to make the `instrumentation_value_of` method work for this, but was hitting a bit of a wall since it seems like it sets the opposite value. Perhaps we could pair on it?

lib/new_relic/agent/instrumentation/logger.rb
kaylareopelle: I like that idea! This metric seems more appropriate in my opinion in this instrumentation class than the aggregator. I was trying to add something like that to the depends_on block with an `unless`, but it felt a bit messy to me: 

```ruby
  unless depends_on { defined?(::Logger) && NewRelic::Agent.config[:'application_logging.enabled'] }
    ::NewRelic::Agent.increment_metric("Supportability/Logging/Ruby/Logger/enabled")
  end
```
What are your thoughts about that approach? Would you prefer the `not_installed` block? I'm open to either.

lib/new_relic/agent/log_event_aggregator.rb
kaylareopelle: Yes. It seems odd to me, but that is one of the core logging features. Supportability metrics will still be sent up. The metrics that won't be sent are `Logging/lines` and `Logging/lines/{severity}`. I believe this was added because some people may want to fine-tune their performance and adding incrementing metrics like this have more of a performance impact in other language agents.

fallwith: (discussed via mobbing)

lib/new_relic/agent/configuration/default_source.rb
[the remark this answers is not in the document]
kaylareopelle: Decided during mobbing to use `instrumentation_value_from_boolean` and leave the two methods separated.

lib/new_relic/agent/instrumentation/logger.rb
kaylareopelle: Decided during mobbing to move the #increment_metric call within .depends_on and return false. If this comes up again, we'll consider adding a `not_installed` method to DependencyDetection

kaylareopelle: After testing this code out in practice, we'll never get the disabled metric because if `application_logging.enabled` is false, `instrumentation.logging` is disabled, so the code within `lib/new_relic/agent/instrumentation/logger.rb` will not run. It only runs when the instrumentation is enabled. 

It seems like keeping the metric recording within `NewRelic::Agent::LogEventAggregator` is the safest place after all.

kaylareopelle: Due to the above, I'm going to remove the `::NewRelic::Agent.increment_metric` call within this file.

fallwith (approved):
👍

tannalynn (approved):

PR #1008, opened 2022-03-03
Logging: Implement data usage supportability metrics

### LINKED ISSUE #976: Implement Data Usage Supportability Metrics Agent Spec
The newly approved [Data Usage Supportability Metrics Agent Specs (Internal)](https://source.datanerd.us/agents/agent-specs/blob/main/Data-Usage-Supportability-Metrics.md) needs to be implemented into the Ruby Agent. 

The basics for this spec are already implemented in the codebase with the name `Supportability/invoke_remote_size` metrics defined in `NewRelic::Agent::NewRelicService#record_size_supportability_metrics`, however, the metric names need to be updated to match the new format: 
`Supportability/{language}/Collector/Output/Bytes`, `Supportability/{language}/Collector/{method}/Output/Bytes`

kaylareopelle:

# Overview
Updates existing metrics to the new format outlined in the [Data Usage Supportability Metrics Agent Spec (Internal)](https://source.datanerd.us/agents/agent-specs/blob/main/Data-Usage-Supportability-Metrics.md):
* `Supportability/invoke_remote_size` => `Supportability/{language}/Collector/Output/Bytes`
* `Supportability/invoke_remote_size/{method}` => `Supportability/{language}/Collector/{method}/Output/Bytes`

# Related Github Issue
Closes #976 

# Testing
Existing unit tests updated to use new format

### DIFF
--- a/lib/new_relic/agent/new_relic_service.rb
+++ b/lib/new_relic/agent/new_relic_service.rb
@@ -461,7 +461,7 @@ module NewRelic
       def handle_serialization_error(method, e)
         NewRelic::Agent.increment_metric("Supportability/serialization_failure")
         NewRelic::Agent.increment_metric("Supportability/serialization_failure/#{method}")
-        msg = "Failed to serialize #{method} data using #{@marshaller.class.to_s}: #{e.inspect}"
+        msg = "Failed to serialize #{method} data using #{@marshaller.class}: #{e.inspect}"
         error = SerializationError.new(msg)
         error.set_backtrace(e.backtrace)
         raise error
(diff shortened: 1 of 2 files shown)


lib/new_relic/agent/new_relic_service.rb
```diff
@@ -489,8 +489,8 @@ def record_timing_supportability_metrics(method, start_ts, serialize_finish_ts,
       # of items as arguments.
       def record_size_supportability_metrics(method, size_bytes, item_count)
         metrics = [
-          "Supportability/invoke_remote_size",
-          "Supportability/invoke_remote_size/#{method.to_s}"
+          "Supportability/Ruby/Collector/Output/Bytes",
+          "Supportability/Ruby/Collector/#{method.to_s}/Output/Bytes"
```
fallwith, line 493: It looks like it's left over from the previous code, but there's no need for the `#to_s` here. The use of `#{}` will automatically see `#to_s` called on the object being interpolated, so our extra `#to_s` call is redundant. I'm not sure if Ruby actually ends up invoking `#to_s` twice, but it's definitely ok to remove ours and let `#{}` do its thing.

fallwith (approved):
Approved as-is, as it is an improvement over what exists now. Recommend the removal of the redundant `#to_s` though.

lib/new_relic/agent/new_relic_service.rb
kaylareopelle: Thanks for pointing this out, @fallwith! I took this as an opportunity to remove the other redundant #to_s calls in this file.

PR #1009, opened 2022-03-03
Actually set proper things for faster harvest on logging data

jasonrclark:
# Overview
Start setting the `event_harvest_config.harvest_limits.log_event_data` key in the connect payload properly so we can opt into the fast harvest when it's available.

I'd completely misunderstood before looking at the code how those values are sent so this is needed on the agent-side.

# Testing
At the moment apart from all the automated tests passing this shouldn't show any real difference. Once connect is fixed up to handle the new value, we should see it coming back in the response and we can then wire up the agent's use of that setting.

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1864,6 +1864,21 @@ module NewRelic
           :description => 'Defines the maximum number of span events reported from a single harvest. Any Integer between 1 and 10000 is valid.',
           :dynamic_name => true
         },
+        :'application_logging.forwarding.enabled' => {
+          :default => true,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => true,
+          :description => 'If `true`, the agent captures log records emitted by your application.'
+        },
+        :'application_logging.forwarding.max_samples_stored' => {
+          :default => 2000,
+          :public => true,
+          :type => Integer,
+          :allowed_from_server => true,
+          :description => 'Specify a maximum number of log records to buffer in memory at a time.',
+          :dynamic_name => true
+        },
         :disable_grape_instrumentation => {
           :default => false,
           :public => false,
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -354,21 +354,21 @@ module NewRelic
           :public => false,
           :type => String,
           :allowed_from_server => true,
-          :description => 'The [Entity GUID](/attribute-dictionary/span/entityguid) for the entity running this agent.'
+          :description => 'The [Entity GUID](/attribute-dictionary/span/entityguid) for the entity running your agent.'
         },
         :monitor_mode => {
           :default => value_of(:enabled),
           :public => true,
           :type => Boolean,
           :allowed_from_server => false,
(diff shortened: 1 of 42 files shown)


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1875,7 +1875,7 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :default => 10000,
           :public => true,
           :type => Integer,
-          :allowed_from_server => false,
+          :allowed_from_server => true,
```
jasonrclark, line 1878: This turns out to be necessary for how we translate back from the `harvest_limits` in the connect response to the configs. You can see the other faster harvest config values (look for `max_samples_stored` ) also default to `:allowed_from_service => true`

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1875,7 +1875,7 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :default => 10000,
```
kaylareopelle, line 1875: Would this PR be a good time to add `event_harvest_config.harvest_limits.log_event_data`? Or should we create the setting and switch them up later on?

test/new_relic/agent/configuration/event_harvest_config_test.rb
```diff
@@ -33,7 +35,8 @@ def test_to_config_hash
           'harvest_limits' => {
             'analytic_event_data' => 833,
             'custom_event_data' => 83,
-            'error_event_data' => 8
+            'error_event_data' => 8,
+            'log_event_data' => 833
```
kaylareopelle, line 39: Is this the number of spans per faster event harvest (5 seconds)? 833 * 12 = 9996?

jasonrclark: Yeah, that's per 5 second harvest. 10,000/12 = 833.3333 is where that's coming from.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1875,7 +1875,7 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
           :public => true,
           :type => Integer,
-          :allowed_from_server => false,
+          :allowed_from_server => true,
           :description => 'Defines the maximum number of log records to buffer in memory at a time.',
           :dynamic_name => true
         },
```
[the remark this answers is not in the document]
jasonrclark: If you look in the rest of the `default_source.rb`, you'll see that the other faster harvest settings don't reference `event_harvest_config.harvest_limits` at all --  we use the feature-focused style of named config keys for everything. I think logging should follow along on that.

How this flows is that the connect response has a `event_harvest_config.harvest_limits` section in it -- but _not_ part of the typical config settings. The agent then reads the `event_harvest_config` part of the connect response, [converts it to our expected keys](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/configuration/event_harvest_config.rb) and then [merges that into the `ServerSource`](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/configuration/server_source.rb#L87-L96).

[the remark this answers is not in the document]
kaylareopelle: Got it! Thank you for explaining that! I was wondering why there weren't any other config values that included `event_harvest_config.harvest_limits`.

PR #1011, opened 2022-03-05
Logging: Add log decoration

### LINKED ISSUE #988: Add log decoration support

kaylareopelle:
# Overview
Add local log decoration with linking metadata when `application_logging.local_decoration.enabled` is true.

# Related Github Issue
Closes #988 

# Testing
* Adds unit tests for new class, LocalLogDecorator
* Adds integration tests to Logger instrumentation
* Adds performance tests for log decorating

### DIFF
--- a/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
@@ -47,7 +47,7 @@ module NewRelic
               formatted_message = LocalLogDecorator.decorate(formatted_message)
             end
 
-            return formatted_message
+            formatted_message
           ensure
             clear_skip_instrumenting
           end
--- a/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
@@ -44,6 +44,7 @@ module NewRelic
 
             unless ::NewRelic::Agent.agent.nil?
               ::NewRelic::Agent.agent.log_event_aggregator.record(formatted_message, severity)
+              formatted_message = LocalLogDecorator.decorate(formatted_message)
             end
 
             return formatted_message
--- a/lib/new_relic/agent/local_log_decorator.rb
+++ b/lib/new_relic/agent/local_log_decorator.rb
@@ -1,0 +1,29 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    #
+    # This module contains helper methods related to decorating log messages
+    module LocalLogDecorator
+      extend self
+
+      def decorate(message)
+        return message unless decorating_enabled?
+        metadata = NewRelic::Agent.linking_metadata
+        formatted_metadata = " NR-LINKING|#{metadata["entity.guid"]}|#{metadata["hostname"]}|#{metadata["trace.id"]}|#{metadata["span.id"]}|"
+        message.partition("\n").insert(1, formatted_metadata).join
+      end
+
+      private
+
+      def decorating_enabled?
+        NewRelic::Agent.config[:'application_logging.enabled'] &&
+          NewRelic::Agent::Instrumentation::Logger.enabled? &&
+          NewRelic::Agent.config[:'application_logging.local_decorating.enabled']
+      end
+    end
+  end
+end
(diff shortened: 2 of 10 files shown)


lib/new_relic/agent/local_log_decorator.rb
```diff
@@ -0,0 +1,28 @@
...
+      def decorate(message)
+        return message unless decorating_enabled?
+        metadata = NewRelic::Agent.linking_metadata
+        "#{message} NR-LINKING|#{metadata["entity.guid"]}|#{metadata["hostname"]}|#{metadata["trace.id"]}|#{metadata["span.id"]}|"
+      end
+
+      private
+
+      def decorating_enabled?
+        # NewRelic::Agent.config[:'application_logging.enabled'] &&
```
kaylareopelle, line 22: Commenting out until PR #1007 is merged

lib/new_relic/agent/instrumentation/logger/instrumentation.rb
```diff
@@ -39,6 +39,7 @@ def format_message_with_tracing(severity, datetime, progname, msg)
             mark_skip_instrumenting
 
             unless ::NewRelic::Agent.agent.nil?
+              formatted_message = LocalLogDecorator.decorate(formatted_message)
```
kaylareopelle, line 42: I'm a little hesitant about putting the instrumentation here. I keep going back and forth with the `LogEventAggregator`. Though it's a rare case, if someone has forwarding and decoration enabled, and the message is longer than 32 kb, the decoration would be truncated. If it's in the LogEventAggregator, a lot of extra code is run to add decoration that may not be necessary. 

One idea could be to add some complexity to the truncation, where it finds the size of the log decoration, truncates the message for decoration_size - max bytes and re-applies the decoration with the truncated message. This seems like it might be too much overhead for each logging line.

Thoughts?

lib/new_relic/agent/local_log_decorator.rb
```diff
@@ -0,0 +1,28 @@
...
+      def decorate(message)
+        return message unless decorating_enabled?
+        metadata = NewRelic::Agent.linking_metadata
+        "#{message} NR-LINKING|#{metadata["entity.guid"]}|#{metadata["hostname"]}|#{metadata["trace.id"]}|#{metadata["span.id"]}|"
+      end
+
+      private
+
+      def decorating_enabled?
+        # :'application_logging.enabled' &&
```
kaylareopelle, line 22: PR #1007 has not been merged yet, so the setting doesn't exist. I plan to add this back once that PR is merged, unless anyone has objections!

lib/new_relic/agent/instrumentation/logger/instrumentation.rb
jasonrclark: So my perspective here -- I'd say that we should be doing it in the instrumentation vs the aggregator... but I think we should do it _after_ we've done record.

If we're sending from the agent, we don't need the linking in the message as it'll be tacked on with the collector post. We also don't want that our truncating of what we're sending up to NR ends up truncating their local output, as they may care about that being complete.

fallwith (approved):
👍

kaylareopelle:
Blocked by #1007

tannalynn (approved):
Looks good! 👍

lib/new_relic/agent/instrumentation/logger/instrumentation.rb
```diff
@@ -44,9 +44,10 @@ def format_message_with_tracing(severity, datetime, progname, msg)
 
             unless ::NewRelic::Agent.agent.nil?
               ::NewRelic::Agent.agent.log_event_aggregator.record(formatted_message, severity)
+              formatted_message = LocalLogDecorator.decorate(formatted_message)
             end
 
-            return formatted_message
+            formatted_message
```
fallwith, line 50: 😎

PR #1015, opened 2022-03-09
improve usage of the 'hostname' executable, others

### LINKED ISSUE #697: Fix Hostname.get_fqdn on AWS Lambda
Hi! My name is Ken Collins and I work on Ruby/Rails integration with AWS Lambda. More details here. https://lamby.custominktech.com. I've done posts before on how to [use APM with Rails and Lambda](https://dev.to/aws-heroes/using-new-relic-apm-with-rails-on-aws-lambda-51gi) and today I noticed a new issue when shutting down an instance when working on a simple task runner project. There is no `hostname` on AWS Lambda.

```
** [NewRelic][2021-06-04 19:24:15 -0400 169.254.216.237 (27)] INFO : Starting Agent shutdown
/var/runtime/lib/runtime.rb: No such file or directory - hostname
```

My assumption was that `Hostname.get_fqdn` is the reason for this. If I use `NEW_RELIC_AGENT_ENABLED=false` the error disappears. The `get_fqdn` method is written in such a that I should see an `"Unable to determine fqdn #{e}"` message but I do not. I attempted to override the `get_fqdn` method but still saw the "No such file or directory - hostname" message.

I've gone thru the code here, but can not find where the error happens. It does not cause a major issue but thought I'd ask is there anywhere any ideas you may have on where this message is coming from?

fallwith:
* Do not shell out directly with `%x()` to any executable. Instead, use
the new Open3 based `NewRelic::Helper.run_command` method.
* The new `NewRelic::Helper.run_command` should prevent any unwanted
error messages pertaining to binary execution from appearing in any
output or log unless the log level is set to 'debug'.
* When executing an external binary, make sure it exists first
* If the 'hostname' binary does not exist, have
`NewRelic::Agent::Hostname.get_fqdn` fall back to `get`, which uses
`Socket`.
* When using 'uname' for JRuby to determine the platform, use 'unknown'
if the 'uname' call fails
* Introduce new `NewRelic::CommandExcutableNotFoundError` and
`NewRelic::CommandRunFailedError` error classes to help with logic flow.
* Bring in 'minitest-stub-const' as a development dependency for
stubbing constants

resolves #697

Thank you to @metaskills for submitting #697 and pointing out that
'hostname' will not be available to AWS Lambda Ruby functions.

Thank you to @brcarp for suggesting the use of Open3 as an improvement
over `%x()` to handle output.

Thank you to both @metaskills and @brcarp for their continued input and
maintainer collaboration on issue #697


tannalynn (approved):
🎉 Thanks for working on this issue!

lib/new_relic/helper.rb
```diff
@@ -40,5 +44,41 @@ def instance_methods_include?(klass, method_name)
...
+      output.chomp
+    end
+
+    # TODO: Open3 defers the actual excecution of a binary to Process.spawn,
+    #       which will raise an Errno::ENOENT exception for a file that
+    #       cannot be found. We might want to take the time to evaluate
+    #       relying on that Process.spawn behavior instead of checking for
+    #       existence ourselves. We'd need to see what it does, how efficient
+    #       it is, if it differs in functionality between Ruby versions and
+    #       operating systems, etc.
```
kaylareopelle, line 74: We talked a bit about todos recently, but not quite what to do with them! Should we make issues? Leave them for analysis at the beginning of a bugsmash?

kaylareopelle (approved):
🍵

lib/new_relic/helper.rb
fallwith: Issues would be good. The issue could reference the class and method where the TODO lives. It would be nice to have the TODO text reference the issue as well so that they're linked bidirectionally but that requires a PR. Maybe we can do some bulk issue creation and then produce a single PR with all of the updated TODO comments.

kaylareopelle: I like it! I can make a ticket for bulk TODO ticket making 😉

PR #1016, opened 2022-03-09
agent_helper: fix for 'assert_operator' usage

fallwith:
TestUnit's `assert_operator` takes arguments in the order that makes
sense when read left to right: `[6, :>=, 5]`, and we had the first and
third values swapped.

fixes buggy syntax introduced by #1013


tannalynn (approved):
Thanks for getting this updated!

PR #1019, opened 2022-03-10
Logging: Mark ActiveSupport::Logger.broadcast arguments as @skip_instrumenting

### LINKED ISSUE #1005: Don't forward duplicate logs when using Rails development defaults
~~Testing the logging initiative code within a Rails development environment revealed that when the environment variable `RAILS_LOG_TO_STDOUT` is true, which is the default in the development environment, identical logs are output to STDOUT and development.log. This causes both logs to be reported to the New Relic logging tool, unnecessarily duplicating data.~~

Testing the logging initiative code with Rails in a development environment revealed all logs were reported twice.

# Acceptance Criteria 
~~When `RAILS_LOG_TO_STDOUT` is true, then logs should be sent to the server once~~

When someone has the agent running on Rails development mode, logs should be reported once.

This should be completed before the GA release.

kaylareopelle:
# Overview
The Rails server command has an option, `--log_to_stdout`, that defaults true in development and when the server is not daemonized. The output gets doubled by a method called `ActiveSupport::Logger.broadcast`, that sends all logs to both `development.log` and `STDOUT`. This caused our logging instrumentation to forward emitted logs twice.

This PR prepends/aliases `ActiveSupport::Logger.broadcast` to mark any logger passed to the method with `@skip_instrumenting`. This instance variable is currently used by our Logging instrumentation to skip the NewRelic loggers `newrelic_agent.log` and `newrelic_audit.log`.

# Related Github Issue
Closes #1005

# Testing
Added integration tests with ActiveSupport::Logger

### DIFF
--- a/lib/new_relic/agent/instrumentation/active_support_logger.rb
+++ b/lib/new_relic/agent/instrumentation/active_support_logger.rb
@@ -1,0 +1,26 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+
+require_relative 'active_support_logger/instrumentation'
+require_relative 'active_support_logger/chain'
+require_relative 'active_support_logger/prepend'
+
+DependencyDetection.defer do
+  named :'active_support.logger'
+
+  depends_on { defined?(::ActiveSupport::Logger) }
+
+  executes do
+    ::NewRelic::Agent.logger.info 'Installing ActiveSupport::Logger instrumentation'
+  end
+
+  executes do
+    if use_prepend?
+      # the only method currently instrumented is a class method
+      prepend_instrument ::ActiveSupport::Logger.singleton_class, NewRelic::Agent::Instrumentation::ActiveSupportLogger::Prepend
+    else
+      chain_instrument NewRelic::Agent::Instrumentation::ActiveSupportLogger::Chain
+    end
+  end
+end
--- a/lib/new_relic/agent/instrumentation/active_support_logger.rb
+++ b/lib/new_relic/agent/instrumentation/active_support_logger.rb
@@ -13,9 +13,7 @@ DependencyDetection.defer do
 
   executes do
     ::NewRelic::Agent.logger.info 'Installing ActiveSupport::Logger instrumentation'
-  end
 
-  executes do
     if use_prepend?
       # the only method currently instrumented is a class method
       prepend_instrument ::ActiveSupport::Logger.singleton_class, NewRelic::Agent::Instrumentation::ActiveSupportLogger::Prepend
--- a/lib/new_relic/agent/instrumentation/active_support_logger.rb
+++ b/lib/new_relic/agent/instrumentation/active_support_logger.rb
@@ -7,7 +7,7 @@ require_relative 'active_support_logger/chain'
 require_relative 'active_support_logger/prepend'
 
 DependencyDetection.defer do
-  named :'active_support.logger'
+  named :active_support_logger
 
   depends_on { defined?(::ActiveSupport::Logger) }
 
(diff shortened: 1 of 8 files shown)


jasonrclark:
This looks reasonable to me overall. In this use-case I'm assuming that the broadcast logger that we're overriding _only_ gets used for the broadcast (i.e. it's not used in other spots for logging that would actually be unique). As long as that's true for the base-case here in Rails I think the approach is workable.

lib/new_relic/agent/instrumentation/active_support_logger.rb
```diff
@@ -0,0 +1,26 @@
...
+require_relative 'active_support_logger/chain'
+require_relative 'active_support_logger/prepend'
+
+DependencyDetection.defer do
+  named :active_support_logger
+
+  depends_on { defined?(::ActiveSupport::Logger) }
+
+  executes do
+    ::NewRelic::Agent.logger.info 'Installing ActiveSupport::Logger instrumentation'
```
fallwith, line 15: From what I understand, these `execute` blocks ultimate result in an array of separate executable blocks that are fired later. I could see how it provides value to keep separate independent chunks of logic in different blocks. In this case though with a single logger statement describing what the subsequent `executes` block is doing, I think it would make more sense to use a single `executes` block and have the logger statement live with the code it's describing.

test/multiverse/suites/rails/active_support_logger_test.rb
```diff
@@ -0,0 +1,41 @@
...
+      @aggregator.reset!
+    end
+
+    def test_broadcasted_logger_marked_skip_instrumenting
+      assert @broadcasted_logger.instance_variable_get(:@skip_instrumenting)
+      assert_nil @logger.instance_variable_get(:@skip_instrumenting)
+    end
+
+    def test_logs_not_forwarded_by_broadcasted_logger
+      message = 'Can you hear me, Major Tom?'
```
fallwith, line 29: 👨‍🎤 love it!

I love the (unofficial) sequel and its cover too:

https://www.youtube.com/watch?v=wO0A0XcWy88

https://www.youtube.com/watch?v=KyFe5m2MU9Q

fallwith (approved):
Approved as-is. Left an optional suggestion to use a single `executes` block in the new `active_support_logger.rb` file.

lib/new_relic/agent/instrumentation/active_support_logger.rb
tannalynn: Yeah, I think that seems reasonable. I looked through a few files and found that we do it both ways in different places, so I think either one would be consistent with other code.

jasonrclark (approved):

lib/new_relic/agent/instrumentation/logger.rb
```diff
@@ -16,9 +16,7 @@
 
   executes do
     ::NewRelic::Agent.logger.info "Installing Logger instrumentation"
-  end
 
-  executes do
```
tannalynn, line 21: I saw you combined these 2 execute blocks like we were discussing above, but not on the other file active_support_logger.  Did you want to do it in both places?

newrelic.yml
```diff
@@ -325,6 +325,10 @@ common: &default_settings
 # May be one of [auto|prepend|chain|disabled].
 # instrumentation.logger: auto
 
+# Controls auto-instrumentation of ActiveSupport::Logger at start up.
+# May be one of [auto|prepend|chain|disabled].
+# instrumentation.active_support.logger: auto
```
tannalynn, line 330:
```suggestion
# instrumentation.active_support_logger: auto
```
I saw that this spot still had the .logger and we had changed it to _logger

lib/new_relic/agent/instrumentation/logger.rb
kaylareopelle: Great point! I forgot about that file. I'll make the fix :)

tannalynn (approved):

PR #1024, opened 2022-03-11
Take down the ROADMAP for now

kaylareopelle:
# Overview
The roadmap content is no longer relevant. While we update the roadmap with the latest content, let's point readers to our [Ruby Engineering Board](https://github.com/orgs/newrelic/projects/17).


angelatan2 (approved):
Looks good. Thank you for your attention to details.

PR #1028, opened 2022-03-16
leverage minitest/pride

fallwith:
leverage `minitest/pride` for colorful output

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -15,7 +15,7 @@ require 'rubygems'
 require 'rake'
 
 require 'minitest/autorun'
-require 'minitest/pride'
+require 'minitest/pride' unless ENV['CI']
 require 'minitest/stub_const'
 require 'mocha/setup'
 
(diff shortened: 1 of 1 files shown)


kaylareopelle:
@fallwith - Looking at the CI runs, it seems like the minitest dot test progress output is blank in the GitHub Actions view. Do you see something similar?

fallwith:
> @fallwith - Looking at the CI runs, it seems like the minitest dot test progress output is blank in the GitHub Actions view. Do you see something similar?

@kaylareopelle Oh I see. The dots (plaintext periods) are replaced entirely by terminal escape sequences. I would have thought that there we be a sequence to enable a color, a plaintext dot, and then a sequence to switch to the next color or disable color. Instead, there are no plaintext dots and only escape sequences. These show up just fine in the raw logs (`[31m*[0m[32m*[0m[33m*[0m[34m*[0m[35m*[0m[36m*[0m`) but when we use the HTML terminal output view, they are replaced with blank spaces and it makes it look like there are huge multiline gaps in the output (the line numbers jump by 50 from the start of the escape sequences to the end of them).

I'll see what can be done about disabling color under CI.

kaylareopelle (approved):
🌈

PR #1032, opened 2022-03-23
Logging: Add entity name to log events and local log decoration

### LINKED ISSUE #1031: Add entity.name to local log decoration and log events
The attribute `entity.name` was removed from the log event payload and linking metadata earlier in the release. The EAP raised concerns about it missing, so we're adding it back to the events and metadata.

Tasks:
* The attribute `entity.name` should be added to the linking metadata included in the LocalLogDecorator.
* The attribute ENTITY_NAME_KEY should not be deleted in LogEventAggregator#payload_to_melt_format

kaylareopelle:

# Overview
* Append `entity.name` to the linking metadata used by the `LocalLogDecorator`
* Remove the `#delete` call on the constant holding `entity.name` information

# Related Github Issue
Closes #1031

# Testing
Existing tests updated to include the new attribute

### DIFF
--- a/lib/new_relic/agent/local_log_decorator.rb
+++ b/lib/new_relic/agent/local_log_decorator.rb
@@ -1,10 +1,10 @@
 # encoding: utf-8
 # This file is distributed under New Relic's license terms.
 # See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
 
 module NewRelic
   module Agent
-    #
     # This module contains helper methods related to decorating log messages
     module LocalLogDecorator
       extend self
--- a/lib/new_relic/agent/local_log_decorator.rb
+++ b/lib/new_relic/agent/local_log_decorator.rb
@@ -13,7 +13,9 @@ module NewRelic
         return message unless decorating_enabled?
 
         metadata = NewRelic::Agent.linking_metadata
-        formatted_metadata = " NR-LINKING|#{metadata[ENTITY_GUID_KEY]}|#{metadata[HOSTNAME_KEY]}|#{metadata[TRACE_ID_KEY]}|#{metadata[SPAN_ID_KEY]}|#{escape_entity_name(metadata[ENTITY_NAME_KEY])}|"
+        formatted_metadata = " NR-LINKING|#{metadata[ENTITY_GUID_KEY]}|#{metadata[HOSTNAME_KEY]}|" \
+                             "#{metadata[TRACE_ID_KEY]}|#{metadata[SPAN_ID_KEY]}|" \
+                             "#{escape_entity_name(metadata[ENTITY_NAME_KEY])}|"
 
         message.partition("\n").insert(1, formatted_metadata).join
       end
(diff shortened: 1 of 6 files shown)


lib/new_relic/agent/local_log_decorator.rb
```diff
@@ -1,7 +1,6 @@
 # encoding: utf-8
 # This file is distributed under New Relic's license terms.
 # See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
-# frozen_string_literal: true
```
fallwith, line 4: If there's only one or two strings that need to not be frozen, I'd prefer to leave the magic comment and initialize the strings as unfrozen (use `my_string = +""` instead of `my_string = ""`).

kaylareopelle: Thanks for pointing this out, James. I deleted it for testing and forgot to return it after deeming the shovel process to be less performant in this context.

lib/new_relic/agent/local_log_decorator.rb
```diff
@@ -5,15 +5,16 @@
...
     # This module contains helper methods related to decorating log messages
     module LocalLogDecorator
       extend self
 
       def decorate(message)
         return message unless decorating_enabled?
+
         metadata = NewRelic::Agent.linking_metadata
-        formatted_metadata = " NR-LINKING|#{metadata["entity.guid"]}|#{metadata["hostname"]}|#{metadata["trace.id"]}|#{metadata["span.id"]}|"
+        formatted_metadata = " NR-LINKING|#{metadata[ENTITY_GUID_KEY]}|#{metadata[HOSTNAME_KEY]}|#{metadata[TRACE_ID_KEY]}|#{metadata[SPAN_ID_KEY]}|#{escape_entity_name(metadata[ENTITY_NAME_KEY])}|"
```
fallwith, line 16: Disregard if there's a performance reason for this format, but can we do anything about this line's length? Here are two suggestions for keeping things under 120 columns at least:

```ruby
formatted_metadata = " NR-LINKING|#{metadata[ENTITY_GUID_KEY]}|#{metadata[HOSTNAME_KEY]}|" \
                     "#{metadata[TRACE_ID_KEY]}|#{metadata[SPAN_ID_KEY]}|" \
                     "#{escape_entity_name(metadata[ENTITY_NAME_KEY])}|"

formatted_metadata = [' NR-LINKING',
                      metadata[ENTITY_GUID_KEY],
                      metadata[HOSTNAME_KEY],
                      metadata[TRACE_ID_KEY],
                      metadata[SPAN_ID_KEY],
                      escape_entity_name(metadata[ENTITY_NAME_KEY])].join('|')
```

fallwith (approved):
Functionally great as-is and approved - suggested a possible formatting improvement for line length.

lib/new_relic/agent/local_log_decorator.rb
kaylareopelle: Thanks, James! I tried the multi-line interpolation earlier, but left the `\` character inside the string, which made it appear differently. Thanks for sharing the correct syntax! It is slightly faster to have everything on one line, but it's in the range of 10-20 μs, so I think we're safe.

PR #1036, opened 2022-03-24
Curb instrumentation method_with_tracing fix

### LINKED ISSUE #1033: Typo in Curb instrumentation causes ArgumentError
## Description
When instrumented by NewRelic, following code:
```ruby
Curl::Easy.new.method(:to_s)
```
raises `` `method_with_tracing': wrong number of arguments (given 0, expected 1) (ArgumentError)``:

## Expected Behavior
An object like `#<Method: Curl::Easy(Kernel)#to_s()>` should be returned

## Steps to Reproduce
Paste following code into `newrelic_test.rb` and invoke `ruby newrelic_test.rb`

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'curb', '= 1.0.0'
  gem 'newrelic_rpm', '= 8.5.0'
end

Curl::Easy.new.method(:to_s)
```

## Your Environment
`ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-darwin19]`

## Additional context
I believe the problem can be solved by changing [this line](https://github.com/newrelic/newrelic-ruby-agent/blob/8e453822c38b7fef829818883e81ae07ecc4a498/lib/new_relic/agent/instrumentation/curb/prepend.rb#L37) to
```ruby
method_with_tracing(verb) { super }
```

fallwith:
To address an issue apparently dating back to when the Curb
instrumentation was first split into separate "prepend" and "chain"
approaches, be sure to pass a verb argument to `method_with_tracing`,
which requires it.

Thank you to @knarewski for bringing this issue to our attention,
for providing a means of reproducing an error, and for providing a fix.

resolves #1033

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
   ## v8.6.0
 
+  * **Bugfix: Curb - satify method_with_tracing's verb argument requirement**
+
+    When Curb instrumentation is used (either via prepend or chain), be sure to always pass the verb argument over to `method_with_tracing` which requires it. Thank you to @knarewski for bringing this issue to our attention, for providing a means of reproducing an error, and for providing a fix. That fix has been replicated by the agent team with permission. See [Issue 1033](https://github.com/newrelic/newrelic-ruby-agent/issues/1033) for more details.
+
   * **Improve the usage of the 'hostname' executable and other executables**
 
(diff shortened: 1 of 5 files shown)


tannalynn (approved):
Thanks for adding a test to exercise this code! Great to see a gap in our testing is covered now!

kaylareopelle (approved):
I'm comfortable with this fix being merged into dev along with the logging work in the next release! Thanks for asking and fixing this, @fallwith!

Issue #1037, opened 2022-03-24
Add license file


cmboling:
### Is your feature request related to a problem? Please describe.

I noticed in the README there's a section describing the license, which is Apache 2.0. Please add a license file, so developers can do a quick lookup on a license file rather than the readme. 😸 

### Priority

Must Have

cmboling:
Nvm!!!

fallwith:
Hi @cmboling, thanks very much for looking through the README and raising this issue.

Even though you ended up finding our `LICENSE` file in the root of the repo and closing this issue, your suggestion was still helpful because it prompted us to look at our other `newrelic` namespaced repos and find some minor inconsistencies between our license related documentation that we're going to discuss. This is much appreciated, thank you!

PR #1040, opened 2022-03-25
License documentation updates

fallwith:
* Remove the "Versions 6.12.0 and above" note from `LICENSE` so that
  that file will exactly match the Apache License, version 2.0 text.
* Update `README.md` to reference the fact that other licenses may be in
  play for versions older than 6.12.
* Update `README.md` to directly reference `LICENSE` instead of the
  Apache web site.
* Update `README.md` to explain that third party software _may_ be used
* Update `THIRD_PARTY_NOTICES.md` to explain that third party software
  _may_ be used, but that none is currently used. Call attention to the
  fact that the actual development of the agent absolutely depends on
  third party software even though that software is not itself
  distributed with the agent.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -175,10 +175,6 @@
 
    END OF TERMS AND CONDITIONS
 
-   Versions 6.12.0 and above for this project are licensed under 
-   Apache 2.0.  For prior versions of this project, please see the 
-   LICENSE file in the root directory of that version for more information.
-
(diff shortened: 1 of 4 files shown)


kaylareopelle (approved):
Looks great, James! Thanks for working on this.

fallwith:
These changes were inspired by feedback from [cmboling](https://github.com/cmboling) with #1037. Thank you!

PR #1041, opened 2022-03-25
log forwarding: don't forward STDOUT/STDERR dupes

### LINKED ISSUE #1039: Mark multiple loggers from the same source for skip instrumenting
Goal: Prevent log events with duplicate severities, messages, and timestamps for the same entity from being sent up to New Relic twice

Working solution:
Maintain a hash with Logger instance object ids as keys and log device references as values to answer these questions:

- Are you one of multiple loggers? (does more than 1 key exist in the hash?)
- Is your log device set to write to STDOUT/STDERR?

If the logger answers yes to both, then it is marked for skipping. If it has already been marked for skipping, the logger will not get asked the questions again.

fallwith:
Prevent standard stream (STDOUT/STDERR) based Logger instances from
having their output forwarded to New Relic if other (file or string
based) Logger instances are known to the aggregator.

For Rails 3/4 especially, this prevents duplicate log entries from
reaching New Relic.

resolves #1039

### DIFF
--- a/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
@@ -7,6 +7,7 @@ module NewRelic
     module Instrumentation
       module Logger
         def skip_instrumenting?
+          ::NewRelic::Agent.agent.log_event_aggregator.register_logger(self) unless defined?(@newrelic_registered)
           defined?(@skip_instrumenting) ? @skip_instrumenting : false
         end
 
@@ -35,7 +36,6 @@ module NewRelic
 
         def format_message_with_tracing(severity, datetime, progname, msg)
           formatted_message = yield
-          ::NewRelic::Agent.agent.log_event_aggregator.register_logger(self) unless skip_instrumenting?
           return formatted_message if skip_instrumenting?
 
           begin
--- a/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
@@ -7,7 +7,7 @@ module NewRelic
     module Instrumentation
       module Logger
         def skip_instrumenting?
-          defined?(@skip_instrumenting) && @skip_instrumenting
+          defined?(@skip_instrumenting) ? @skip_instrumenting : false
         end
 
         # We support setting this on loggers which might not have
@@ -35,6 +35,7 @@ module NewRelic
 
         def format_message_with_tracing(severity, datetime, progname, msg)
           formatted_message = yield
+          ::NewRelic::Agent.agent.log_event_aggregator.register_logger(self) unless skip_instrumenting?
           return formatted_message if skip_instrumenting?
 
           begin
--- a/lib/new_relic/agent/log_event_aggregator.rb
+++ b/lib/new_relic/agent/log_event_aggregator.rb
@@ -157,8 +157,7 @@ module NewRelic
       end
 
       def register_logger(logger)
-        return if @loggers.key?(logger)
-
+        logger.instance_variable_set(:@newrelic_registered, true)
         logdev = logger.instance_variable_get(:@logdev)
         dev = logdev.instance_variable_get(:@dev) if logdev
         @loggers[logger] = dev
@@ -240,21 +239,14 @@ module NewRelic
         message.byteslice(0...MAX_BYTES)
       end
 
-      # don't forward logs from a logger writing to a standard streams unless
-      # it happens to be the only logger in use
+      # don't forward logs from a logger writing to a standard stream
+      # (stdout/stderr) unless it happens to be the only logger in use
       def mark_standard_streams_for_skipping
         return unless @loggers.keys.size > 1
 
-        to_prune = []
         @loggers.each do |logger, dev|
-          if dev === $stdout || dev === $stderr
-            logger.mark_skip_instrumenting
-            to_prune << logger
-          else
-            logger.clear_skip_instrumenting
-          end
+          logger.mark_skip_instrumenting if dev.respond_to?(:tty?) && dev.tty?
         end
-        to_prune.each { |logger| @loggers.delete(logger) }
       end
     end
   end
--- a/lib/new_relic/agent/log_event_aggregator.rb
+++ b/lib/new_relic/agent/log_event_aggregator.rb
@@ -44,6 +44,7 @@ module NewRelic
         @seen_by_severity = Hash.new(0)
         @high_security = NewRelic::Agent.config[:high_security]
         @instrumentation_logger_enabled = NewRelic::Agent::Instrumentation::Logger.enabled?
+        @loggers = {}
         register_for_done_configuring(events)
       end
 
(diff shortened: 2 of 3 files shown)


lib/new_relic/agent/instrumentation/logger/instrumentation.rb
```diff
@@ -35,6 +35,7 @@ def self.enabled?
 
         def format_message_with_tracing(severity, datetime, progname, msg)
           formatted_message = yield
+          ::NewRelic::Agent.agent.log_event_aggregator.register_logger(self) unless skip_instrumenting?
```
jasonrclark, line 38: If this was moved below `return formatted_message if skip_instrumenting?` then it wouldn't need the conditional check at all right?

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -228,6 +239,23 @@ def truncate_message(message)
...
+      def mark_standard_streams_for_skipping
+        return unless @loggers.keys.size > 1
+
+        to_prune = []
+        @loggers.each do |logger, dev|
+          if dev === $stdout || dev === $stderr
+            logger.mark_skip_instrumenting
+            to_prune << logger
+          else
+            logger.clear_skip_instrumenting
```
jasonrclark, line 254: Not sure that it's right for us to clear if they aren't STDOUT/STDERR. At the very least, this raises the possibility of multi-threaded issues if the logger was in the midst of another call since we set the flags during our methods. Maybe just marking the skips is the right thing to do here?

lib/new_relic/agent/instrumentation/logger/instrumentation.rb
fallwith: Thanks, @jasonrclark!

I was misusing the existing skipped/cleared instance variables to see if a logger was registered yet or not. This caused the confusing duplicated conditional check here and the problematic interference with existing behavior that you noted in your other comment.

With [f32f947](https://github.com/newrelic/newrelic-ruby-agent/pull/1041/commits/f32f947a34b1c04d80c56bf029c494e21a21a632), I have reworked things so that:

a) We have a new @newrelic_registered instance variable to track if a logger has been registered
b) Non stdout/stderr loggers are no longer cleared for instrumentation, they're simply not marked for skipping.

lib/new_relic/agent/log_event_aggregator.rb
fallwith: Great call. With [f32f947](https://github.com/newrelic/newrelic-ruby-agent/pull/1041/commits/f32f947a34b1c04d80c56bf029c494e21a21a632), non stdout/stderr loggers are no longer cleared for instrumenting.

lib/new_relic/agent/instrumentation/logger/instrumentation.rb
jasonrclark: Totally makes sense and changes look good to me 👍

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -228,6 +238,16 @@ def truncate_message(message)
...
+
+      # don't forward logs from a logger writing to a standard stream
+      # (stdout/stderr) unless it happens to be the only logger in use
+      def mark_standard_streams_for_skipping
+        return unless @loggers.keys.size > 1
+
+        @loggers.each do |logger, dev|
+          logger.mark_skip_instrumenting if dev.class == IO
+        end
+      end
```
kaylareopelle, line 250: I think I might not be fully understanding this solution. I'm concerned about how things will be have if, for example we have the following sources/output locations: 

| source | output location |
| ------ | ---------------- |
| rails | production.log |
| serveza | serveza.log |
| mail | STDOUT |

Would the mail logs be marked with `@skip_instrumenting`? This gems outputs only to STDOUT.

Maybe it's the construction of the hash that I can't quite see. I think I'm hoping for a data type more like this: 

`{:rails => ['production.log'], :serveza => ['serveza.log'], :mail => ['STDOUT']}` so that as long as each source only has one item in the array, it will be kept. If the array is longer than one, the logger is marked `@skip_instrumenting`. Is that how this works?

kaylareopelle (commented):
I think I'm still a little confused overall. Perhaps we can sync up later to discuss? I've added comments in an attempt to convey my confusion.

lib/new_relic/agent/instrumentation/logger/instrumentation.rb
```diff
@@ -7,7 +7,8 @@ module Agent
     module Instrumentation
       module Logger
         def skip_instrumenting?
-          defined?(@skip_instrumenting) && @skip_instrumenting
+          ::NewRelic::Agent.agent.log_event_aggregator.register_logger(self) unless defined?(@newrelic_registered)
+          defined?(@skip_instrumenting) ? @skip_instrumenting : false
```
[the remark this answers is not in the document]
fallwith: I wanted this question mark suffixed method to return a boolean (whereas both the old and proposed new approaches would return `nil`).

There's precedence in stlib for going with `nil` as a response, so you're in good company if you prefer it that way even though it's not my own preference. Will rework.

lib/new_relic/agent/log_event_aggregator.rb
fallwith: The hash key is a reference to an instance of `::Logger`. The hash value is a reference to that logger's underlying output handle (the `::Logger` instance has a `@logdev` instance var, and that `@logdev` object has a `@log` instance var, and it's the `@dev` object's reference we're storing in the hash.

So it's just a 2 dimensional hash with references for both the keys and values (and no arrays anywhere).

The logic is...

* When an unregistered `::Logger` instance writes for the first time, register it.
* When registering, check the registration hash. If this logger is the only logger, short circuit.
* When registering, if multiple loggers exist in the hash, then mark all ones that write to `IO` for skipping.

NOTE that `klass == IO` will match $stdout, $stderr, STDOUT.dup, etc., but not an instance of `File` (which itself would pass an `#is_a?(IO)` check).

So we end up with a hash that tells us of all the `::Logger` objects we've seen (skipped or otherwise) and the device they're writing to.

I'm using this approach to take advantage of the fact that we effectively have a singleton class to track the hash with and we can avoid setting class vars on the `::Logger` class itself.

lib/new_relic/agent/instrumentation/logger/instrumentation.rb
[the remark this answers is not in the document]
kaylareopelle: I understand wanting a `?` method to return false. I think this is fine to leave as-is in that case. Thanks for clarifying. We should talk about this as a team to decide a convention and enforce it using rubocop.

lib/new_relic/agent/log_event_aggregator.rb
kaylareopelle: Ok, thanks for clarifying. I think I want to run a quick test with the Mail library on a playground app to confirm those logs make it through this setup before approving. If that works, then I'll feel more at ease.

kaylareopelle (approved):
Ok! This seems to be working fine in the Rails context, which is the main context our customers are working in. Apologies for my extended hesitation! Thanks for your patience!

Issue #810, opened 2021-10-14
Update Ruby-Agent-Configuration Docs for instrumentation settings


kaylareopelle:
The rake task used to generate new configuration documentation uses `(Dynamic)` when the default value is generated using a method call. This is unclear for the new instrumentation settings that have the options of `[auto|prepend|chain|disabled]`. `"auto"` is the default setting.

This PR by Jason Clark shows the default value in a clearer way than our other settings that begin with `instrumentation-*`. 
https://github.com/newrelic/docs-website/pull/4310/files 

This change could be performed manually or through revising the config options rake task. As the file is getting quite large and hand edits are commonly required, it may be time to refactor or retire the rake task to avoid future issues.

**docs-website file**: https://github.com/newrelic/docs-website/blob/develop/src/content/docs/agents/ruby-agent/configuration/ruby-agent-configuration.mdx 

### Priority
Nice to Have

kaylareopelle:
I tagged this with `version:9.0.0` because I hope to have the `Dynamic` value removed from public documentation before our next major release, when we can expect more folks perusing our documentation.

kaylareopelle:
This was closed by #1051

PR #1051, opened 2022-03-30
Update script to generate config html docs

tannalynn:

# Overview
I updated a bunch of wordings and other things that were changed on the docs site so that we are matching that.
I also updated the config rake task to allow grouping for config options with more than 2 parts (ex: `application_logging.forwarding.enabled`). 
I also updated the script to use `documentation_default` from the default_source configs if that field is present. 

# Related Github Issue
 #810

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -329,7 +329,7 @@ module NewRelic
           :public => true,
           :type => String,
           :allowed_from_server => false,
-          :description => 'Your New Relic [license key](/docs/accounts-partnerships/accounts/account-setup/license-key).'
+          :description => 'Your New Relic [license key](/docs/apis/intro-apis/new-relic-api-keys/#ingest-license-key).'
         },
         :agent_enabled => {
           :default => DefaultSource.agent_enabled,
@@ -367,7 +367,7 @@ module NewRelic
           :public => true,
           :type => Boolean,
           :allowed_from_server => false,
-          :description => 'When `true`, the agent transmits data about your application to the New Relic [collector](/docs/using-new-relic/welcome-new-relic/get-started/glossary/#collector).'
+          :description => 'When `true`, the agent transmits data about your app to the New Relic [collector](/docs/using-new-relic/welcome-new-relic/get-started/glossary/#collector).'
         },
         :test_mode => {
           :default => false,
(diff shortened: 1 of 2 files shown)


lib/tasks/config.rake
```diff
@@ -102,15 +107,24 @@ namespace :newrelic do
...
         "#{spec[:default].inspect}"
       end
     end
 
+    # currently the only 'instrumentation.XXXX' configs that
+    # do not use a default of auto are mongo and excon
+    def default_value_auto?(key)
+      return false if key.match(/mongo/) || key.match(/excon/)
+      key.match(/^instrumentation/)
+    end
```
tannalynn, line 124: I don't really love the way I'm doing this right now. It seems like it could get accidentally broken pretty easily. 
I'm thinking what might be a better overall solution for this situation is to add another field on the default values in `default_source.rb` to list what we want the docs to use as the default value if it's present.
Something like: 
```
        :'instrumentation.net_http' => {
          :default => instrumentation_value_of(:disable_net_http, :prepend_net_instrumentation),
          :documentation_default => "auto",
          :public => true,
          :type => String,
          :dynamic_name => true,
          :description => "Controls auto-instrumentation of Net::HTTP at start up.  May be one of [auto|prepend|chain|disabled]."
        },
```

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -815,7 +816,33 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          "\n" \
+          '          enabled: true'\
+          "\n" \
+          '        distributed_tracing:'\
+          "\n" \
+          '          enabled: false'\
+          "\n" \
+          '      ```'\
+          "\n" \
+          '    </Callout>'
```
fallwith, line 845: This multiline text might be a good candidate for a Heredoc, so that you can skip all of the `\n`, `'`, and `\` characters.

```ruby
          description => <<~DESCRIPTION
          If `true`, disables agent middleware for Sinatra. This middleware is responsible for advanced feature support such as [cross application tracing](/docs/apm/transactions/cross-application-traces/cross-application-tracing), [page load timing](/docs/browser/new-relic-browser/getting-started/new-relic-browser), and [error collection](/docs/apm/applications-menu/events/view-apm-error-analytics).


              <Callout variant="important">

                Cross application tracing is deprecated in favor of [distributed tracing](https://docs.newrelic.com/docs/apm/distributed-tracing/getting-started/introduction-distributed-tracing). Distributed tracing is on by default for Ruby agent versions 8.0.0 and above. Middlewares are not required to support distributed tracing.


                To continue using cross application tracing, update the following options in your `newrelic.yml` configuration file:


                ```

                # newrelic.yml


                  cross_application_tracer:

                    enabled: true

                  distributed_tracing:

                    enabled: false

                ```

              </Callout>
          DESCRIPTION
```

lib/tasks/config.rake
kaylareopelle: I think that's a great idea! I say go for it if you feel motivated or add the snippet to the issue (https://github.com/newrelic/newrelic-ruby-agent/issues/810) if you'd rather wait!

kaylareopelle: Whether it's changed now or later, what do you think about adding tests for these methods?

fallwith: You could use a negative look ahead to combine the 3 separate regex matches into a single one:

```ruby
def default_value_auto?(key)
  key.match?(/^instrumentation(?!.*(?:mongo|excon))/
end
```

tannalynn: Updated to not check the config key value instead, and use `documentation_default` field if it exists.

lib/new_relic/agent/configuration/default_source.rb
tannalynn: Huh, so I updated all of the longer descriptions to use heredoc like that, but it appears to not be supported by ruby 2.2. I remember you mentioning that there were a couple different versions of heredoc that came out at different times, so I'll look and see what works on 2.2

lib/tasks/config.rake
```diff
@@ -111,6 +118,7 @@ namespace :newrelic do
     end
 
     def format_env_var(key)
+      return "None" if NON_ENV_CONFIGS.include? key
```
kaylareopelle, line 121: AWESOME ADDITION! 🐐

kaylareopelle (approved):
Thank you for updating the config task! These changes look great! 🦄

PR #1061, opened 2022-04-07
Separate PR/daily CI coverage, integrate Ruby 3.2.0-preview1 and Rails Edge

### LINKED ISSUE #1098: Ruby 3.2 has trouble building mysql2 native extensions
When the continuous integration system uses Ruby 3.1+ to perform Rails 6 and Rails 7 environment tests, it uses `mysql2` v0.5.3 (latest, November 2019) in conjunction with the default MySQL client libraries provided by the GitHub Actions Ubuntu 20.04 image.

With Ruby 3.1, this works fine. With Ruby 3.2, the `mysql2` gem's native extensions fail to build.

The problem can be reproduced locally with `gem install mysql2`. The build issues can be ignored via `--with-cflags=-Wno-error=implicit-function-declaration`, but this results in a binary that ultimately crashes when activated.

Curiously, the multiverse tests that involve `mysql2` are passing, and it is only the "env" tests for Rails 6 and 7 that are failing. We'll have to check to see if the tests are doing something specially to get Ruby 3+ working, or if they are skipping tests.

Other options include upgrading the Ubuntu MySQL client libraries and switching the Rails tests over to Posgres.

### LINKED ISSUE #1157: CI: Test fewer Rubies by default, still test all of them periodically
One clear way to simplify and lessen the cost of our CI process is to have it do less.

For the most part, we really only have 3 concerns when it comes to testing Ruby itself:

- Does our oldest supported version handle all of our syntax and supported third party instrumented libraries?
- Does the current version of Ruby work?
- Does JRuby work too?

As such, we should be able to narrow our normal field of testing to just these 3 Rubies:

- CRuby 2.2
- CRuby (latest)
- JRuby (latest)

We additionally want to keep an eye on these other concerns, only not as frequently as on every PR:

- Is there any Ruby version specific quirk with the agent or any of its instrumented libraries? Every supported Ruby needs tested.
- Will the upcoming next Ruby version work when it gets released? The latest preview release needs tested.

We ought to address the first list of concerns with our PR level frequency tests and the second list of concerns with our daily tests.

fallwith:
use separate PR and cron CI workflows
for the PR workflow:

* only test the oldest supported Ruby, the newest production Ruby, and
  the newest production JRuby
* only test the latest minor version for each Rails version

for the Cron workflow:

* do everything we used to do (all Rubies, all Rails versions)
* also test Ruby 3.2.0-preview1
* also test Rails Edge for all Rubies previously testing Rails 7

resolves #1157
resolves #1158
resolves #1098


kaylareopelle (approved):
I appreciate your proactivity!

fallwith:
With the split to have separate PR and cron workflows, the PR workflow has gone from 60 jobs to 26!

test/environments/railsedge/config/application.rb
```diff
@@ -0,0 +1,20 @@
...
+require "rails/all"
+
+# Require the gems listed in Gemfile, including any gems
+# you've limited to :test, :development, or :production.
+Bundler.require(*Rails.groups)
+
+module RpmTestApp
+  class Application < Rails::Application
+    # Initialize configuration defaults for originally generated Rails version.
+    config.load_defaults 7.0 # TODO: keep this number up to date (and keep this TODO)
```
kaylareopelle, line 16: Great addition!

kaylareopelle (approved):
This looks good to me! I'm excited to see how faster CI runs change our development workflow! 🚀

PR #1062, opened 2022-04-07
update File.exists to File.exist

fallwith:
File.exists has been deprecated for quite awhile and will be dropped in
favor of File.exist entirely with Ruby 3.2


tannalynn (approved):
Awesome! Thanks for getting this updated throughout the agent 🎉

PR #1063, opened 2022-04-08
Revert overriding logic to original

kaylareopelle:
# Overview
`set_overriding_transaction_name` ended up with the same logic as `set_default_transaction_name`, preventing most transactions from getting their names overridden and all API calls with a category of `:task` from being overridden.

This PR goes back to the original logic of accepting the passed-in name and category unless the name has already been frozen.

# Related Github Issue
[GTSE-14114 - Internal](https://newrelic.atlassian.net/jira/software/c/projects/GTSE/boards/1186?modal=detail&selectedIssue=GTSE-14114)

# Testing
Updated test to check for category, but this doesn't feel robust enough at the moment. I'm considering adding tests that look at `set_transaction_name` calls by nesting depth and the specific `:task` category. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


lib/new_relic/agent/transaction.rb
```diff
@@ -337,10 +337,8 @@ def set_default_transaction_name(name, category)
 
       def set_overriding_transaction_name(name, category)
         return log_frozen_name(name) if name_frozen?
-        if influences_transaction_name?(category)
```
fallwith, line 340: I think the removal of this check will probably work great. From what I can tell though, by removing this we will no longer have _any_ validation whatsoever on the user supplied category value. I'm thinking we should still perform some validation like so:

* Category is present
* Category is a symbol
* Category is found within an allow list
* Category is not on the "web" category list

lib/new_relic/agent/transaction.rb
```diff
@@ -337,10 +337,8 @@ def set_default_transaction_name(name, category)
 
       def set_overriding_transaction_name(name, category)
```
fallwith, line 338: Maybe we could define a new shared `set_transaction_name` which accepts an `override` boolean parameter, defaulting to `false`. Then both `set_default_transaction_name` and `set_overriding_transaction_name` can call that shared method with the latter setting `override` to `true`.

kaylareopelle:
This might need to be a major version release change to address the failing tests.

kaylareopelle:
We can come back to this one later if the issue resurfaces

PR #1066, opened 2022-04-08
tests: don't install 'memcached' for Ruby 3.2

fallwith:
for memcached tests, don't bother to install the 'memcached' gem for
Ruby 3.2. we can likely discard this gem entirely in future, but for now
we'll scope the removal to Ruby 3.2 to avoid gem build issues.


kaylareopelle (approved):
LGTM! Thanks for proactively addressing this!

PR #1067, opened 2022-04-08
tests: update Mongo tests for Ruby 3.2

fallwith:
for Mongo multiverse tests:

* drop coverage of 'mongo' gem <= v1.10. this allows us to stop
  bothering with compiling 'bson_ext' which doesn't work with Ruby 3.2
  and is slow for others
* stop testing the older (<= v2.9.0) 'mongo' gem versions with Ruby
  version 2.5 and above
* specify the latest 'bson' gem version

### DIFF
--- a/test/multiverse/suites/mongo/Envfile
+++ b/test/multiverse/suites/mongo/Envfile
@@ -31,8 +31,11 @@ def mongo_versions
   versions += MONGO_VERSIONS_RUBY26 if ruby_gte?('2.6.0')
   return versions if RUBY_PLATFORM.eql?('java')
 
-  versions += MONGO_VERSIONS_RUBY25 if ruby_gte?('2.5.0')
-  versions += MONGO_VERSIONS_OLD_RUBIES unless ruby_gte?('2.5.0')
+  if ruby_gte?('2.5.0')
+    versions += MONGO_VERSIONS_RUBY25
+  else
+    versions += MONGO_VERSIONS_OLDER_RUBIES
+  end
   versions
 end
 
--- a/test/multiverse/suites/mongo/Envfile
+++ b/test/multiverse/suites/mongo/Envfile
@@ -19,9 +19,11 @@ MONGO_VERSIONS_OLDER_RUBIES = ['2.9.0',
                                '2.4.0',
                                '2.1.0',
                                '2.0.0',
-                               ['1.9.2', 2.6],
-                               ['1.8.6', 2.6],
-                               '1.7.1']
+                               '1.9.2',
+                               '1.8.6']
+
+# The highest Mongo server version to test 1.x gems against
+MAX_SERVER_VERSION_FOR_1X_GEMS = 2.6
 
 # * JRuby uses Mongo gem versions for v2.6.0+ Ruby only
 # * Newer CRubies use newer Mongo gem versions and older ones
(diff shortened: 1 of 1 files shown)


test/multiverse/suites/mongo/Envfile
```diff
@@ -13,19 +13,15 @@ MONGO_VERSIONS_RUBY25 = ['2.15.1',
                          '2.12.0']
 
 # Mongo versions to test with all CRubies
```
kaylareopelle, line 15:
```suggestion
# Mongo versions to test with CRubies below 2.5.0
```

test/multiverse/suites/mongo/Envfile
```diff
@@ -13,19 +13,15 @@ MONGO_VERSIONS_RUBY25 = ['2.15.1',
...
-# Include bson_ext for older Mongo gem versions <= 1.10.0
-BSON_EXT_NEEDED_BEFORE_VERSION = Gem::Version::new("1.10.0")
+MONGO_VERSIONS_OLDER_RUBIES = ['2.9.0',
+                               '2.7.0',
+                               '2.5.0',
+                               '2.4.0',
+                               '2.1.0',
+                               '2.0.0',
+                               ['1.9.2', 2.6],
+                               ['1.8.6', 2.6],
```
kaylareopelle, line 23: What does 2.6 mean here?

test/multiverse/suites/mongo/Envfile
```diff
@@ -35,8 +31,11 @@ def mongo_versions
   versions += MONGO_VERSIONS_RUBY26 if ruby_gte?('2.6.0')
   return versions if RUBY_PLATFORM.eql?('java')
 
-  versions += MONGO_VERSIONS_RUBY25 if ruby_gte?('2.5.0')
-  versions += MONGO_VERSIONS_ALL_RUBIES
+  if ruby_gte?('2.5.0')
```
kaylareopelle, line 34: What does `gte` mean? I'm not familiar with that abbreviation.

kaylareopelle (approved):
This can be merged as-is if you'd like, however, I have a few questions/suggestions on the PR. Thanks for taking this on!

test/multiverse/suites/mongo/Envfile
fallwith: `gte` = "greater than or equal to"

In many places we simply compare strings mathematically and benefit from Ruby's minor version never hitting double digits (`'3.10.0'` would be considered smaller than `'3.2.0'`). The `gte` helper method used `Gem::Version` comparisons instead.

test/multiverse/suites/mongo/Envfile
fallwith: Wow, I'd forgotten about this one from the last time I was in this file. It represents the "last supported server version" for the given gem version, and will prevent the testing of the gem if the server version is too new. But weirdly we don't apply that same constraint to the oldest gem version, `'1.7.1'`. I'll see if I can clean this up while we're in here. Depending on how old these oldest 1.x versions are, I might just drop them entirely as part of our gem EOL efforts.

fallwith: Addressed with [ecdc7b8](https://github.com/newrelic/newrelic-ruby-agent/pull/1067/commits/ecdc7b89b1a30684e6a98d54b87f2c267904ad56)

tannalynn (approved):
Looks great! Thanks for getting this updated

PR #1068, opened 2022-04-08
pipe service: address Ruby 3.2 warning

fallwith:
Pipe service - address the Ruby 3.2 `Struct#initialize` warning

> warning: Passing only keyword arguments to Struct#initialize will behave differently from Ruby 3.2. Please use a Hash literal like .new({k: v}) instead of .new(k: v).


kaylareopelle (approved):
👍

PR #1069, opened 2022-04-08
method level telemetry

fallwith:
MLT: associate source code information with each monitored method

* file path
* line number within the defining .rb file
* module namespace
* method name

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1993,6 +1993,13 @@ A map of error classes to a list of messages. When an error of one of the classe
           :allowed_from_server => false,
           :description => 'If `true`, the agent decorates logs with metadata to link to entities, hosts, traces, and spans.'
         },
+        :'code_level_metrics.enabled' => {
+          :default => true,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => true,
+          :description => 'If `true`, the agent will report source code level metrics for monitored methods.'
+        },
         :'instrumentation.active_support_logger' => {
           :default => instrumentation_value_from_boolean(:'application_logging.enabled'),
           :documentation_default => 'auto',
--- a/lib/new_relic/agent/instrumentation/action_controller_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/action_controller_subscriber.rb
@@ -16,6 +16,7 @@ module NewRelic
           controller_class = controller_class(payload)
 
           if state.is_execution_traced? && !should_ignore(payload, controller_class)
+            # TODO: MLT - possible Rails entrypoint for mlt
             finishable = start_transaction_or_segment(payload, request, controller_class)
             push_segment(id, finishable)
           else
--- a/lib/new_relic/agent/instrumentation/controller_instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/controller_instrumentation.rb
@@ -165,10 +165,6 @@ module NewRelic
           def add_transaction_tracer(method, options = {})
             NewRelic::Agent.record_api_supportability_metric(:add_transaction_tracer)
 
-            # The metric path:
-            options[:name] ||= method.to_s
-
-            argument_list = generate_argument_list(options)
             traced_method, punctuation = parse_punctuation(method)
             with_method_name, without_method_name = build_method_names(traced_method, punctuation)
 
@@ -176,6 +172,12 @@ module NewRelic
               ::NewRelic::Agent.logger.warn("Transaction tracer already in place for class = #{self.name}, method = #{method.to_s}, skipping")
               return
             end
+
+            # The metric path:
+            options[:name] ||= method.to_s
+
+            code_info = NewRelic::Agent::MethodTracerHelpers.code_information(self, method)
+            argument_list = generate_argument_list(options.merge(code_info))
 
             class_eval <<-EOC
               def #{with_method_name}(*args, &block)
@@ -441,6 +443,9 @@ module NewRelic
           txn_options[:apdex_start_time] = queue_start_time
           txn_options[:ignore_apdex] = ignore_apdex?
           txn_options[:ignore_enduser] = ignore_enduser?
+          NewRelic::Agent::MethodTracerHelpers::SOURCE_CODE_INFORMATION_PARAMETERS.each do |parameter|
+            txn_options[parameter] = trace_options[parameter]
+          end
           txn_options
         end
 
--- a/lib/new_relic/agent/method_tracer_helpers.rb
+++ b/lib/new_relic/agent/method_tracer_helpers.rb
@@ -6,6 +6,7 @@ module NewRelic
   module Agent
     module MethodTracerHelpers
       MAX_ALLOWED_METRIC_DURATION = 1_000_000_000 # roughly 31 years
+      SOURCE_CODE_INFORMATION_PARAMETERS = %i[filepath lineno function namespace]
 
       extend self
 
@@ -26,11 +27,57 @@ module NewRelic
           segment.record_metrics = false
         end
 
+        unless !options.key?(:code_information) || options[:code_information].nil? || options[:code_information].empty?
+          segment.code_information = options[:code_information]
+        end
+
         begin
           Tracer.capture_segment_error(segment) { yield }
         ensure
           segment.finish if segment
         end
+      end
+
+      def class_info(object)
+        name = object.name
+
+        return [name, true] if name
+        return [Regexp.last_match(1), false] if object.to_s =~ /^#<Class:(.*)>$/
+
+        ::NewRelic::Agent.logger.error("Unable to determine name for what was expected to be a singleton class, #{object}")
+        [nil, nil]
+      end
+
+      def class_or_instance(name, is_instance_method)
+        klass = Object.const_get(name)
+        return klass unless is_instance_method
+
+        klass.new
+      end
+
+      def location(name, method_name, is_instance_method)
+        begin
+          class_or_instance(name, is_instance_method).method(method_name).source_location
+        rescue => e
+          ::NewRelic::Agent.logger.error("Unable to determine source code info for class '#{name}', " \
+                                          "method '#{method_name}' - #{e.class}: #{e.message}")
+          []
+        end
+      end
+
+      def code_information(object, method_name)
+        return NewRelic::EMTPY_HASH unless NewRelic::Agent.config[:'code_level_metrics.enabled']
+
+        klass_name, is_instance_method = class_info(object)
+        return NewRelic::EMPTY_HASH unless klass_name
+
+        file_info = location(klass_name, method_name, is_instance_method)
+        namespace = Regexp.last_match(1) if klass_name =~ /(.*)::[^:]+/
+
+        {filepath: file_info.first,
+         lineno: file_info.last,
+         function: method_name,
+         namespace: namespace}
       end
     end
   end
--- a/lib/new_relic/agent.rb
+++ b/lib/new_relic/agent.rb
@@ -143,6 +143,9 @@ module NewRelic
     #
     def add_or_defer_method_tracer(receiver, method_name, metric_name, options)
       @tracer_lock.synchronize do
+        # TODO: MLT - need to pass this info downstream
+        code_info = NewRelic::Agent::MethodTracerHelpers.code_information(receiver, method_name)
+
         if @agent
           receiver.send(:_nr_add_method_tracer_now, method_name, metric_name, options)
         else
--- a/lib/new_relic/agent/instrumentation/notifications_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/notifications_subscriber.rb
@@ -16,6 +16,8 @@ module NewRelic
         end
 
         def self.find_all_subscribers
+          # TODO: MLT - need to make sure we have MLT coverage for these
+
           # TODO: need to talk to Rails core about an API for this,
           # rather than digging through Listener ivars
           instance_variable_names = [:@subscribers, :@string_subscribers, :@other_subscribers]
--- a/lib/new_relic/agent/method_tracer.rb
+++ b/lib/new_relic/agent/method_tracer.rb
@@ -84,6 +84,8 @@ module NewRelic
       # @api public
       #
       def trace_execution_unscoped(metric_names, options = NewRelic::EMPTY_HASH) # THREAD_LOCAL_ACCESS
+        # TODO: MLT - need to accept options[:code_information] and pass it forward down to segment creation
+
         NewRelic::Agent.record_api_supportability_metric :trace_execution_unscoped unless options[:internal]
         return yield unless NewRelic::Agent.tl_is_execution_traced?
         t0 = Process.clock_gettime(Process::CLOCK_MONOTONIC)
@@ -100,7 +102,7 @@ module NewRelic
       module ClassMethods
         # contains methods refactored out of the #add_method_tracer method
         module AddMethodTracer
-          ALLOWED_KEYS = [:metric, :push_scope, :code_header, :code_footer].freeze
+          ALLOWED_KEYS = [:metric, :push_scope, :code_header, :code_information, :code_footer].freeze
 
           DEFAULT_SETTINGS = {:push_scope => true, :metric => true, :code_header => "", :code_footer => ""}.freeze
 
@@ -274,7 +276,8 @@ module NewRelic
 
           _nr_define_traced_method(method_name, scoped_metric: scoped_metric, unscoped_metrics: unscoped_metrics,
             code_header: options[:code_header], code_footer: options[:code_footer],
-            record_metrics: options[:metric], visibility: visibility)
+            record_metrics: options[:metric], visibility: visibility,
+            code_information: options[:code_information])
 
           prepend(_nr_traced_method_module)
 
@@ -298,7 +301,8 @@ module NewRelic
 
         def _nr_define_traced_method(method_name, scoped_metric: nil, unscoped_metrics: [],
           code_header: nil, code_footer: nil, record_metrics: true,
-          visibility: :public)
+          visibility: :public, code_information: {})
+
           _nr_traced_method_module.module_eval do
             define_method(method_name) do |*args, &block|
               return super(*args, &block) unless NewRelic::Agent.tl_is_execution_traced?
@@ -325,17 +329,24 @@ module NewRelic
               # If tracing multiple metrics on this method, nest one unscoped trace inside the scoped trace.
               begin
                 if scoped_metric_eval
-                  ::NewRelic::Agent::MethodTracer.trace_execution_scoped(scoped_metric_eval, metric: record_metrics, internal: true) do
+                  ::NewRelic::Agent::MethodTracer.trace_execution_scoped(scoped_metric_eval,
+                    metric: record_metrics,
+                    internal: true,
+                    code_information: code_information) do
                     if unscoped_metrics_eval.empty?
                       super(*args, &block)
                     else
-                      ::NewRelic::Agent::MethodTracer.trace_execution_unscoped(unscoped_metrics_eval, internal: true) do
+                      ::NewRelic::Agent::MethodTracer.trace_execution_unscoped(unscoped_metrics_eval,
+                        internal: true,
+                        code_information: code_information) do
                         super(*args, &block)
                       end
                     end
                   end
                 elsif !unscoped_metrics_eval.empty?
-                  ::NewRelic::Agent::MethodTracer.trace_execution_unscoped(unscoped_metrics_eval, internal: true) do
+                  ::NewRelic::Agent::MethodTracer.trace_execution_unscoped(unscoped_metrics_eval,
+                    internal: true,
+                    code_information: code_information) do
                     super(*args, &block)
                   end
                 end
--- a/lib/new_relic/agent/transaction.rb
+++ b/lib/new_relic/agent/transaction.rb
@@ -124,7 +124,7 @@ module NewRelic
         txn = Transaction.new(category, options)
         state.reset(txn)
         txn.state = state
-        txn.start
+        txn.start(options)
         txn
       end
 
@@ -394,7 +394,7 @@ module NewRelic
         @frozen_name ? true : false
       end
 
-      def start
+      def start(options = {})
         return if !state.is_execution_traced?
 
         sql_sampler.on_start_transaction(state, request_path)
@@ -403,7 +403,7 @@ module NewRelic
 
         ignore! if user_defined_rules_ignore?
 
-        create_initial_segment
+        create_initial_segment(options)
         Segment.merge_untrusted_agent_attributes \
           @filtered_params,
           :'request.parameters',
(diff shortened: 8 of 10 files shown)


lib/new_relic/agent/instrumentation/action_controller_subscriber.rb
```diff
@@ -16,6 +16,7 @@ def start(name, id, payload) # THREAD_LOCAL_ACCESS
           controller_class = controller_class(payload)
 
           if state.is_execution_traced? && !should_ignore(payload, controller_class)
+            # TODO: MLT - possible Rails entrypoint for mlt
```
kaylareopelle, line 19: This could be a good entry point specifically for ActionController. We will likely need to add something like this to each of the `lib/new_relic/agent/*_subscriber.rb` files to get full Rails support.

This also reminds me of some [pre-spike work](https://docs.google.com/document/d/19NkLomuuCjvOudnqu2FeeWNeTqEhRqNxI-6oOzj09j4/edit?usp=sharing) (Internal) we discussed in the fall for ActionPack. We were thinking about rolling our ActionController instrumentation up into ActionPack as a whole to use the actionpack gem as the top-level reference.

lib/new_relic/agent/method_tracer_helpers.rb
```diff
@@ -26,12 +27,58 @@ def trace_execution_scoped(metric_names, options = NewRelic::EMPTY_HASH) # THREA
...
         end
       end
+
+      def class_info(object)
+        name = object.name
+
+        return [name, true] if name
+        return [Regexp.last_match(1), false] if object.to_s =~ /^#<Class:(.*)>$/
+
+        ::NewRelic::Agent.logger.error("Unable to determine name for what was expected to be a singleton class, #{object}")
```
kaylareopelle, line 47: I wonder if this might be a nice area to add a Supportability metric to capture data on how often this occurs?

lib/new_relic/agent/instrumentation/action_controller_subscriber.rb
fallwith: I took a pass at satisfying the TODO here with #1085 by updating `NewRelic::Agent::Instrumentation::ActionControllerSubscriber#start_transaction_or_segment`. By the time that method is called, we have access to the controller class and method names that we need.

We should be able to do something similar for the other subscriber classes.

lib/new_relic/agent/method_tracer_helpers.rb
fallwith: This is a terrific idea! This code has been updated again with #1085, but once it settles down I'd like to layer in support for that metric for sure.

fallwith: @kaylareopelle can you let me know if these lines accomplish what you wanted to see happen?

[method_tracer_helpers.rb#75](https://github.com/newrelic/newrelic-ruby-agent/pull/1069/files#diff-7e20a306b4d12a8e387870b566cf31215725ae5abad2e26d592b15521da78bb4R75)

[method_tracer_helpers.rb#84](https://github.com/newrelic/newrelic-ruby-agent/pull/1069/files#diff-7e20a306b4d12a8e387870b566cf31215725ae5abad2e26d592b15521da78bb4R84)

kaylareopelle: Yes! They look great. My one recommendation would be to add `Ruby` in the metric name so that we can see these values in Angler (in case any other agents have the same metric). The language was the second-to-last value in all the Supportability metrics we created for Logging.

lib/new_relic/agent/instrumentation/controller_instrumentation.rb
```diff
@@ -165,10 +165,6 @@ def newrelic_read_attr(attr_name) # :nodoc:
           def add_transaction_tracer(method, options = {})
             NewRelic::Agent.record_api_supportability_metric(:add_transaction_tracer)
```
fallwith, line 167: We were doing work to set these variables before doing a conditional check that could potentially return early. These variables are not involved in the check, so it is a waste to set them before the check has completed.

lib/new_relic/agent/instrumentation/action_controller_subscriber.rb
```diff
@@ -47,27 +47,31 @@ def finish(name, id, payload) # THREAD_LOCAL_ACCESS
         end
 
         def start_transaction_or_segment(payload, request, controller_class)
+          code_info = NewRelic::Agent::MethodTracerHelpers.code_information(controller_class, payload[:action])
+
+          options = {
```
fallwith, line 52: None of these original hash key/values have been modified, I just moved the anonymous hash up here into a new `options` variable to make the new `#merge` below look/read better.

lib/new_relic/agent/span_event_primitive.rb
```diff
@@ -165,18 +165,18 @@ def custom_attributes segment
...
       end
 
-      def merge_and_freeze_attributes agent_attributes, error_attributes
-        return agent_attributes.freeze unless error_attributes
-        return error_attributes.freeze if agent_attributes.equal?(NewRelic::EMPTY_HASH)
-        agent_attributes.merge!(error_attributes).freeze
+      def merge_and_freeze_attributes *hashes
+        hashes.each_with_object({}) do |hash, merged|
+          merged.merge!(hash) if hash
+        end.freeze
```
tannalynn, line 171: I'm curious how this behaves when there is only one hash passed in in `hashes`? I'm thinking that maybe if there is just one that it might create extra allocations vs having an early return like we did before? I'm not sure how it works though.

lib/new_relic/agent/span_event_primitive.rb
```diff
@@ -165,18 +165,18 @@ def custom_attributes segment
...
       end
 
       def agent_attributes segment
         agent_attributes = segment.attributes
           .agent_attributes_for(NewRelic::Agent::AttributeFilter::DST_SPAN_EVENTS)
         error_attributes = error_attributes(segment)
-        return NewRelic::EMPTY_HASH unless agent_attributes || error_attributes
-        merge_and_freeze_attributes(agent_attributes, error_attributes)
+        code_attributes = segment.code_attributes
+        merge_and_freeze_attributes(agent_attributes, error_attributes, code_attributes)
```
tannalynn, line 179: How come the `return NewRelic::EMPTY_HASH` statement was removed?

test/new_relic/agent/instrumentation/rails/action_controller_subscriber.rb
```diff
@@ -316,4 +322,17 @@ def test_records_span_level_error
...
+
+  def test_records_code_level_metrics
+    @subscriber.start('process_action.action_controller', :id, @entry_payload)
+    txn = NewRelic::Agent::Transaction.tl_current
+    @subscriber.finish('process_action.action_controller', :id, @entry_payload)
+    attributes = txn.segments.first.code_attributes
+
+    assert_equal __FILE__, attributes['code.filepath']
+    assert_equal 'index', attributes['code.function']
+    assert_equal 18, attributes['code.lineno']
```
tannalynn, line 334: This seems like it could break easily if someone adds something above where it's currently defined (same with the tests in the other file). Is it possible for us to get the expected line number in a more flexible way?

test/new_relic/agent/method_tracer_helpers_test.rb
```diff
@@ -0,0 +1,73 @@
...
+
+module ::The
+  class Example
+    def self.class_method; end
+    def instance_method; end
+  end
+end
+
+class NewRelic::Agent::MethodTracerHelpersTest < Minitest::Test
+  # TODO: trace_execution_scoped should have test coverage
```
tannalynn, line 15: weird that there isn't any tests for it yet, although definitely not unheard of. I wonder if there are tests tucked into other files/suites that are testing?

lib/new_relic/agent/span_event_primitive.rb
fallwith: The early `return` statement was removed because I believe it to be unreachable. It is guarded by an `unless`, and I believe that that `unless` will always be satisfied. The `segment.attributes` call will always return a hash. Even an empty hash will pass a truthiness test and satisfy the `unless`.

This change isn't directly related to MLT, but I noticed the dead code while I was in this method so I removed it.

lib/new_relic/agent/span_event_primitive.rb
fallwith: This is a good callout. I had a version that retained the early return functionality, but it became really complicated with the number of possible combinations after a third hash was tossed into the mix. Still, this method is only called in exactly one place, and we should strive to maintain the current level of object allocation. I'll put a TODO here and make sure this is addressed before it gets merged.

test/new_relic/agent/instrumentation/rails/action_controller_subscriber.rb
fallwith: The easiest dynamic way to get at the correct value would be to leverage `#source_location`, just like the method being tested does. I'll put build together that does that.

fallwith: Addressed with 5cfe616 and 105f061

lib/new_relic/agent/span_event_primitive.rb
fallwith: Addressed with b86f528

lib/new_relic/agent/method_tracer_helpers.rb
fallwith: Updated the metric name and dropped the success condition reporting with 5cfe616

lib/new_relic/agent/instrumentation/action_controller_subscriber.rb
fallwith: As discussed, we don't currently hang method tracers on any other Rails methods via subscriber messages being parsed, so we won't start doing so at this time just for the sake of MLT. We do, however, explicitly look for ActiveJob and set method tracers there, and those tracers will automatically have code metrics now.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1993,6 +1993,13 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'If `true`, the agent decorates logs with metadata to link to entities, hosts, traces, and spans.'
         },
+        :'code_level_metrics.enabled' => {
+          :default => true,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => true,
+          :description => 'If `true`, the agent will report source code level metrics for monitored methods.'
```
kaylareopelle, line 2001: I have a small concern about the word "monitored". Monitor mode for the agent and browser monitoring somewhat indicate all instrumented code. When I read this the first time, I expected code level metrics to be reported for every method the agent touches. 

What are your thoughts about using "traced" so that this leans more in the scope of `add_method_tracer`? 

My concern is about over-promising the scope of code level metrics in our documentation.

lib/new_relic/agent/instrumentation/action_controller_subscriber.rb
kaylareopelle: If you were interested in taking it one step further, it might be nice to move the options hash generation into a separate method.

lib/new_relic/agent/method_tracer_helpers.rb
```diff
@@ -6,6 +6,8 @@ module NewRelic
   module Agent
     module MethodTracerHelpers
       MAX_ALLOWED_METRIC_DURATION = 1_000_000_000 # roughly 31 years
+      SOURCE_CODE_INFORMATION_PARAMETERS = %i[filepath lineno function namespace]
```
kaylareopelle, line 9: I wonder if we should add a comment about the use of the word "function"? I itch to change it to "method", but know it's there because of the specs.

lib/new_relic/agent/method_tracer_helpers.rb
```diff
@@ -26,12 +28,83 @@ def trace_execution_scoped(metric_names, options = NewRelic::EMPTY_HASH) # THREA
...
+                                        "method '#{method_name}' - #{e.class}: #{e.message}")
+        ::NewRelic::Agent.increment_metric(CODE_INFORMATION_FAILURE_METRIC, 1)
+        ::NewRelic::EMPTY_HASH
+      end
+
+      private
+
+      # The string representation of a singleton class looks like
+      # '#<Class:MyModule::MyClass>'. Return the 'MyModule::MyClass' part of
+      # that string
```
kaylareopelle, line 68: This comment is so helpful. I love getting regex decoding/intentions in the code.

lib/new_relic/agent/method_tracer_helpers.rb
```diff
@@ -26,12 +28,83 @@ def trace_execution_scoped(metric_names, options = NewRelic::EMPTY_HASH) # THREA
...
+        name = Regexp.last_match(1) if object.to_s =~ /^#<Class:(.*)>$/
+        return name if name
+
+        raise "Unable to glean a class name from string '#{object}'" unless name
+      end
+
+      # get at the underlying class from the singleton class
+      #
+      # note: even with the regex hit from klass_name(), `Object.const_get`
+      # is more performant than iterating through `ObjectSpace`
```
kaylareopelle, line 79: Nice!

lib/new_relic/agent/method_tracer_helpers.rb
```diff
@@ -26,12 +28,83 @@ def trace_execution_scoped(metric_names, options = NewRelic::EMPTY_HASH) # THREA
...
+      #
+      # traced class methods:
+      #     * object is a singleton class, `#<Class::MyClass>`
+      #     * get at the underlying non-singleton class
+      #
+      # traced instance methods and Rails controller methods:
+      #     * object is a class, `MyClass`
+      #
+      # anonymous class based methods (`c = Class.new { def method; end; }`:
+      #    * `#name` returns `nil`, so use '(Anonymous)' instead
```
kaylareopelle, line 96: It looks like this asterisk is one space outdented from the others.

lib/new_relic/agent/method_tracer_helpers.rb
```diff
@@ -26,12 +28,83 @@ def trace_execution_scoped(metric_names, options = NewRelic::EMPTY_HASH) # THREA
...
+      def namespace_and_location(object, method_name)
+        klass = object.singleton_class? ? klassify_singleton(object) : object
+        name = klass.name || '(Anonymous)'
+        method = if klass.instance_methods.include?(method_name)
+          klass.instance_method(method_name)
+        else
+          klass.method(method_name)
+        end
+        [name, method.source_location]
+      end
```
kaylareopelle, line 107: Nice refactor!

lib/new_relic/agent/method_tracer_helpers.rb
kaylareopelle: Alternately, it could be called method except for the final key? https://github.com/newrelic/newrelic-ruby-agent/pull/1069/files#diff-1bf0a4e80d440cbe0b83bfe60e1b97e75696a89369dffaacbc442cd16ff0a6a1R134

test/new_relic/agent/transaction/abstract_segment_test.rb
```diff
@@ -210,6 +210,55 @@ def test_root_segment_gets_transaction_name_attribute
...
+          end
+        end
+
+        def test_code_level_metrics_attributes_are_empty_if_the_metrics_are_empty
+          with_segment do |segment|
+            assert_equal(segment.code_attributes, {})
+          end
+        end
+
+        def test_code_level_metrics_are_all_or_nothing
```
kaylareopelle, line 256: Interesting! I'm glad this test is here. I don't think I absorbed that all metrics needed to be present.

kaylareopelle (commented):
This is looking great, James!

lib/new_relic/agent/instrumentation/action_controller_subscriber.rb
fallwith: Taken one (or two) steps further with 6ff053b

lib/new_relic/agent/configuration/default_source.rb
fallwith: Changed to "traced" with 6ff053b

lib/new_relic/agent/method_tracer_helpers.rb
fallwith: I really wanted to retain the 4 variable names from the spec, but I liked the commenting idea. Added comments with 6ff053b.

lib/new_relic/agent/method_tracer_helpers.rb
fallwith: Thank you! Fixed up with 6ff053b.

test/new_relic/agent/instrumentation/rails/action_controller_subscriber.rb
```diff
@@ -14,6 +14,13 @@ def action_name
       'test'
     end
 
+    # these need to exist for code level metrics to find them
```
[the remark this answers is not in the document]
fallwith: Somehow the tests were previously able to make references to these phantom methods, most of which I believe to not be defined at all. They needed to actually exist on a line in order for CLM to note the line number.

test/new_relic/agent/method_tracer_test.rb
```diff
@@ -52,23 +52,25 @@ def log(msg)
   end
 end
 
-class MyClass
-  def self.class_method
-  end
+with_config(:'code_level_metrics.enabled' => true) do
```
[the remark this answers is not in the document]
fallwith: Originally the tests were written with CLM enabled by default. Now we have flipped that default for the initial release, and some interesting things like this needed to be in the tests to have CLM enabled  before the ActiveSupport::Notifications system discovers the methods.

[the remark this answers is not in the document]
fallwith: Actually this is a non controller class. Let me double check that the wrapper is needed here. I added it everywhere once I had controller tests working.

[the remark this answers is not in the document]
fallwith: Ah, I remember. The feature needs to be enabled when the `add_method_tracer` calls take place. Those methods fire when Ruby reads in the code, not when the tests run. So the wrapper is needed.

test/new_relic/agent/instrumentation/rails/action_controller_subscriber.rb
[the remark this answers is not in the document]
tannalynn: Huh, interesting, that's not what I expected. Thanks for explaining!

test/new_relic/agent/method_tracer_test.rb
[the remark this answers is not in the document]
tannalynn: Ahhh I see, that makes sense! TY!

tannalynn (approved):

lib/new_relic/agent/instrumentation/active_job.rb
```diff
@@ -58,14 +58,21 @@ def self.perform(job, block)
         end
 
         def self.run_in_trace(job, block, event)
-          trace_execution_scoped("MessageBroker/#{adapter}/Queue/#{event}/Named/#{job.queue_name}") do
+          trace_execution_scoped("MessageBroker/#{adapter}/Queue/#{event}/Named/#{job.queue_name}",
```
kaylareopelle, line 61: I'm puzzled by this metric's usage of the MessageBroker format. According to the [agent specs (internal)](https://source.datanerd.us/agents/agent-specs/blob/0c52d16310bcef40cea84873408389f3d1597a98/APIs/messaging.md), ActiveJob does not seem like a MessageBroker.

I think we should look into this more. I've made a technical debt ticket to triage.
https://github.com/newrelic/newrelic-ruby-agent/issues/1151

lib/new_relic/agent/method_tracer_helpers.rb
```diff
@@ -5,6 +5,13 @@
 module NewRelic
   module Agent
     module MethodTracerHelpers
+      # These are code level metrics (CLM) attributes. For Ruby, they map like so:
+      #   filepath: full path to an .rb file on disk
+      #   lineno: the line number a Ruby method is defined on within a given .rb file
+      #   function: the name of the Ruby method
+      #   namespace: the Ruby class' namespace as a string, ex: 'MyModule::MyClass'
```
kaylareopelle, line 12: Great comment!

test/multiverse/suites/rails/Envfile
```diff
@@ -40,7 +40,8 @@ end
 
 if RUBY_VERSION < '3.0.0'
   gemfile <<-RB
-    gem 'rails', '~> 5.2.0'
+    # Rails 5.2.7.1 uses a '~>>' style heredoc that breaks Ruby 2.2 compatibility
+    gem 'rails', '~> 5.2.0'#{", '<= 5.2.7'" if RUBY_VERSION < '2.3.0'}
```
kaylareopelle, line 44: Is this still true? If your PR has been merged in, could you remove this?

kaylareopelle: Same goes for: test/multiverse/suites/rails_prepend/Envfile

kaylareopelle (approved):
Great work, James! A few thoughts, but nothing that blocks this being merged in.

kaylareopelle:
> Great work, James! A few thoughts, but nothing that blocks this being merged in.

Though, it looks like there's a merge conflict, so please address that first! :)

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2012,6 +2012,13 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'If `true`, the agent decorates logs with metadata to link to entities, hosts, traces, and spans.'
         },
+        :'code_level_metrics.enabled' => {
+          :default => false,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => true,
+          :description => 'If `true`, the agent will report source code level metrics for traced methods.'
```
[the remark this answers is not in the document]
fallwith: Yeah, ideally you'd be met with optional documentation experiences coming and going via the repo, the agent docs, the CodeStream docs, marketing materials, blog posts, etc., and they'd all be coordinated and consistent.

I'll surface this internally to make sure someone owns the overall objective.

lib/new_relic/agent/instrumentation/active_job.rb
fallwith: Thank you for filing that one. I could see how someone may have looked at a Redis backed queueing system for ActiveJob and saw a resemblance to an existing MQ type system and stretched "message broker" to cover the use case. Both systems feature a regulation of distributed clients operating over multiple queues. But there's certainly not (in the default case) any actual message broking going on.

test/multiverse/suites/rails/Envfile
fallwith: Ah! I've been scanning for a 5.2.7.2 release that never came, but there is a 5.2.8. I'll give it a try!

test/new_relic/agent/method_tracer_helpers_test.rb
fallwith: Prior to the CLM functionality being added to it, this module only did one thing - it defined a `trace_execution_scoped` method.

What's fascinating is that the only consumer of this module at the time was `MethodTracer`, which also defined its own `trace_execution_scoped` method. So the lone consumer of the method has a method with the same name (which presumably _does_ have test coverage) which defers everything over to the method in the helpers module. Perhaps there was an original intent to share the module more widely.

kaylareopelle (approved):
Great work, James!

Issue #1057, opened 2022-03-31
Create a dedicated new section for Asynchronous Custom Instrumentation


kaylareopelle:
Create a new section within [Ruby Custom Instrumentation](https://docs.newrelic.com/docs/apm/agents/ruby-agent/api-guides/ruby-custom-instrumentation) for "Multithreading" / "Concurrency" / "Asynchronous Transactions" / "Threads" / etc. 

This section should clearly lay out the fact that we have 2 options: all threads, or just those we're explicitly told about, with source/configuration code block examples for each.

Bonus: Call out on our documentation more prominently that we have detailed info on RubyDoc

kaylareopelle:
This ticket may not be necessary if our solution to #1056 is able to integrate thread tracking within transactions as part of the agent's normal behavior.

kaylareopelle:
@angelatan2  - After @tannalynn  and I chatted, it seems like we'll just need to do a simple re-write of the current release notes for this feature within an existing document. This changes the estimate, in our opinion, to a 1.

kaylareopelle:
Merged into https://github.com/newrelic/docs-website/pull/7233

PR #1080, opened 2022-04-11
Improved async support and Thread instrumentation

### LINKED ISSUE #1056: Span Parent ID not set properly using Async Transaction workaround
We have a functional workaround for allowing threads to be traced as unique operations within a transaction, however, the parents are not correctly assigned in all cases and sometimes concurrent threads appear as children of each other. This occurs in both the Transaction and Distributed Tracing views. 

Also, in other cases there is a warning raised with this approach that states the parent ID for the span is not being set properly.

This ticket is to investigate the cause and create a solution.

### LINKED ISSUE #1060: Intermittently missing spans created inside threads
While working on another async related ticket (#1056), I discovered a bug in how the agent handles asynchronous spans in a transaction. It appears that intermittently a span happening in another thread from the starting thread of the transaction will get dropped and will not be reported by the agent. 

It appears that in instances of this missing span, the harvest begins while the missing span is still working (making http requests for our example app), and then it appears to be dropped completely. It's not clear yet how (or if) the harvest is causing the span to go missing. Maybe something to do with how we exclude the agent http calls from being recorded on transactions is related?
This behavior appears to be already present in the agent for some time. 

Example transaction in dt: https://staging.onenr.io/0YBR684K2QO

Can reproduce using the ruby_agent_playground/multi-threading application. 
snippet from test app:
```
    threads = []
    threads << Application.perform_with_instrumentation(name: 'Custom/Thread1') do
      long_request
    end
    threads << Application.perform_with_instrumentation(name: 'Custom/Thread2') do
      post_request
    end
    threads.each(&:join)
```
```
  def self.perform_with_instrumentation(name:)
    instrumentation = Thread.current[:newrelic_tracer_state]
    Thread.new do
      Thread.current[:newrelic_tracer_state] = instrumentation
      segment = NewRelic::Agent::Tracer.start_segment(name: name)
      yield
      segment.finish
    end
  end
```

tannalynn:
# Overview
Previously, the agent was not able to record events and metrics from inside Threads created during a running transaction unless they manually inserted New Relic into the thread. This PR adds support for automatically instrumenting threads and bugfixes for issues in the agent that made it unable to handle asynchronous spans.

- Fix parent span ids for multithreaded transactions, this prevents incorrect nesting of spans by keeping track of all current_segments for every thread id
- Added NewRelic::TracedThread class - allows the agent to instrument only threads you create using this class
- Thread instrumentation 
  - allows the agent to keep track of the thread id of the parent thread spawning child threads
  - allows the agent to automatically insert instrumentation to trace the events occurring inside of threads (disabled by default)
- added a lock in the Stats class (where we keep track of a metric), in order to avoid race conditions when modifying the values of the metric. Also had to override marshal dump and load methods for this class to prevent the new lock instance variable from being included when serializing the class. 

# Related Github Issue
closes #1056 
closes #1060

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


test/new_relic/agent/transaction/external_request_segment_test.rb
```diff
@@ -868,7 +868,7 @@ def test_non_sampled_segment_does_not_record_span_event
         assert_empty last_span_events
       end
 
-      def test_non_sampled_segment_does_not_record_span_event
+      def test_ignored_segment_does_not_record_span_event
```
tannalynn, line 871: this change is unrelated, it's just because I noticed a warning that this test method was being defined twice, so I wanted to get that fixed as well.

test/new_relic/agent/stats_engine_test.rb
```diff
@@ -183,10 +183,14 @@ def test_record_scoped_and_unscoped_metrics_is_thread_safe
     nthreads = 25
     iterations = 100
 
-    nthreads.times do |tid|
-      threads << Thread.new do
-        iterations.times do
-          in_transaction('txn') do
+    # TODO: why are no scoped metrics being recorded?
```
tannalynn, line 216: This is the test that fails if `instrumentation.thread.auto_instrument` is enabled

lib/new_relic/traced_thread.rb
```diff
@@ -0,0 +1,27 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
```
tannalynn, line 5: Need to add documentation for public API on how to use

lib/new_relic/traced_thread.rb
```diff
@@ -0,0 +1,27 @@
...
+module NewRelic
+  class TracedThread < Thread
+    def initialize(*args, &block)
+      # we should add metrics
```
tannalynn, line 9: we should add metrics still

test/new_relic/agent/stats_engine_test.rb
```diff
@@ -178,22 +178,58 @@ def test_record_scoped_and_unscoped_metrics_records_multiple_unscoped_metrics
     ])
   end
 
-  def test_record_scoped_and_unscoped_metrics_is_thread_safe
+  def test_record_scoped_and_unscoped_metrics_is_thread_safe_without_traced_threads
```
tannalynn, line 181: So i updated the failing test use disable `instrumentation.thread.tracing`, and then added a second version of this test below that has it enabled and moves the transaction outside. that one still fails, but I wanted to have both situations represented.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -901,6 +901,21 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+        :'instrumentation.thread' => {
+          :default => 'auto',
+          :public => true,
+          :type => String,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => "Controls auto-instrumentation of the Thread class at start up to enable accurate parent tracking for spans in applications manually inserting new relic tracing into threads. This does not enable the agent to automatically trace all threads created (see `instrumentation.thread.tracing`). May be one of [auto|prepend|chain|disabled]."
+        },
+        :'instrumentation.thread.tracing' => {
+          :default => false,
```
tannalynn, line 913: I believe turning this on by default would be a breaking change (so needs a major release) for any customers who are creating multiple concurrent transactions using threads. This used to work because new relic could not see transactions across threads, but with this change a transaction will encompass all threads running in the process. We even have a test that confirmed this behavior works this way when thread tracing is off.

lib/new_relic/agent/agent.rb
```diff
@@ -9,6 +9,7 @@
 require 'zlib'
 require 'stringio'
 require 'new_relic/constants'
+require 'new_relic/traced_thread'
```
fallwith, line 12: From looking at the existing "constants" and "coerce" it looks like we were already out of alphabetical order here. Is there any sort order that we follow for the requires here?

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -901,6 +901,21 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => "Controls auto-instrumentation of resque at start up.  May be one of [auto|prepend|chain|disabled]."
         },
+        :'instrumentation.thread' => {
+          :default => 'auto',
+          :public => true,
+          :type => String,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => "Controls auto-instrumentation of the Thread class at start up to enable accurate parent tracking for spans in applications manually inserting new relic tracing into threads. This does not enable the agent to automatically trace all threads created (see `instrumentation.thread.tracing`). May be one of [auto|prepend|chain|disabled]."
```
fallwith, line 910: All other instances of "New Relic" here seem to have the first letters of each word capitalized.

lib/new_relic/agent/instrumentation/thread.rb
```diff
@@ -0,0 +1,26 @@
...
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+
+require_relative 'thread/chain'
+require_relative 'thread/prepend'
+
+DependencyDetection.defer do
+  named :thread
+
+  depends_on do
+    defined?(Thread)
```
fallwith, line 12: I think I'd yank out `depends_on` entirely (or make it a noop like `depends_on; true; end` and also smoosh the two `executes` blocks together.

lib/new_relic/agent/instrumentation/thread/chain.rb
```diff
@@ -0,0 +1,24 @@
...
+  module Thread
+    module Chain
+      def self.instrument!
+        ::Thread.class_eval do
+          include NewRelic::Agent::Instrumentation::Thread
+
+          alias_method :initialize_without_new_relic, :initialize
+
+          def initialize(*args, &block)
+            traced_block = add_thread_tracing(*args, block)
```
fallwith, line 17: I was surprised to see `NewRelic::Agent::Instrumentation::Thread#add_thread_tracing` take `block` as a second argument instead of `&block` and then I see that it's because of the way it's being passed here. Would it be possible to use `add_thread_tracing(*args, &block)` here (prefixing `block` with `&`) and then update the method definition and block usage code for `add_thread_tracing` accordingly?

I think there's likely only a slight chance that the block isn't guaranteed to be present, and you're already guarding against `nil` by checking that `#call` is available, but I wonder if using `&block` would match other block passing patterns already in use.

lib/new_relic/traced_thread.rb
```diff
@@ -0,0 +1,27 @@
...
+      traced_block = create_traced_block(*args, block)
+      super(*args, &traced_block)
+    end
+
+    def create_traced_block(*args, block)
+      return block if NewRelic::Agent.config[:'instrumentation.thread.tracing'] # if this is on, don't double trace
+
+      instrumentation = ::Thread.current[:newrelic_tracer_state]
+      Proc.new do |*args|
+        ::Thread.current[:newrelic_tracer_state] = instrumentation
```
fallwith, line 19: I'd love a demo / overview of this bit when you get a chance. I'm not understanding how the tracer state gets set originally in order for it to be passed as `instrumentation` to the Proc. From looking at the tests, it seems like they explicitly set the state beforehand. Is doing so required?

lib/new_relic/agent/agent.rb
kaylareopelle: I'm not aware of a sort order, but we could create a new standard?

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: I'm surprised to see this use a boolean value instead of the `auto|prepend|chain|disabled` options. Why did you decide to use a boolean?

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -901,6 +901,21 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => "Controls auto-instrumentation of resque at start up.  May be one of [auto|prepend|chain|disabled]."
         },
+        :'instrumentation.thread' => {
```
kaylareopelle, line 904: @tannalynn, I vaguely remember you creating a new attribute for the DefaultSource that would set the default value in the config docs to be the string value. Is that in this branch? Should it be used here?

lib/new_relic/agent/instrumentation/thread.rb
kaylareopelle: Would it make sense to use `depends_on` to toggle the instrumentation for the Thread-related configuration, like I think it's done in `lib/new_relic/agent/instrumentation/active_record_notifications.rb`? I wonder if it might be more performant than using a conditional? (Though it may also require some refactoring...)

lib/new_relic/agent/transaction.rb
```diff
@@ -261,6 +261,30 @@ def initialize(category, options)
...
+      def current_segment
+        # if this thread id has a current segment use that
+        current_thread_id = Thread.current.object_id
+        return current_segment_by_thread[current_thread_id] if current_segment_by_thread[current_thread_id]
+
+        # if this thread id does not have a current segment, this is where we'll need thread parent
+        return current_segment_by_thread[parent_thread_id] if current_segment_by_thread[parent_thread_id]
+
+        # if both of those fail, fall back to starting thread current segment
+        return current_segment_by_thread[@starting_thread_id]
```
kaylareopelle, line 277: This is slightly reading as a `case` statement to me. Any idea if there's a performance impact for using that instead? (or if it really is applicable here?)

lib/new_relic/agent/transaction.rb
```diff
@@ -719,9 +743,9 @@ def record_exceptions
       # Do not call this.  Invoke the class method instead.
       def notice_error(error, options = {}) # :nodoc:
         # Only the last error is kept
-        if @current_segment
-          @current_segment.notice_error error, expected: options[:expected]
-          options[:span_id] = @current_segment.guid
+        if current_segment
+          current_segment.notice_error error, expected: options[:expected]
```
kaylareopelle, line 747: Would you be open to adding parens here?

lib/new_relic/traced_thread.rb
```diff
@@ -0,0 +1,27 @@
...
+        segment = NewRelic::Agent::Tracer.start_segment(name: "Thread#{::Thread.current.object_id}")
```
kaylareopelle, line 20: possible future feature idea - allow customers to set the name of the thread

kaylareopelle:
One other thought! Does testing this with `concurrent-ruby` feel like its within the scope of this work?

test/new_relic/agent/stats_engine_test.rb
kaylareopelle: First, I love that you added the specific configuration settings to each of these tests! 🎉 
Next, I wonder if there is a part of the StatsHash or Metric recording mechanisms that isn't getting a lock on it, but should be, causing the mismatched #'s?

lib/new_relic/agent/configuration/default_source.rb
tannalynn: Since the default value here is already string, it's not really needed since that is mostly to help default values that are procs that can't be evaluated when we generate the docs

lib/new_relic/agent/configuration/default_source.rb
tannalynn: Since the way we install the instrumentation (which uses the prepend/chain stuff) is on the other config (`instrumentation.thread`), this is just controlling whether or not we insert ourselves into the thread when it's created.

lib/new_relic/agent/instrumentation/thread.rb
tannalynn: @kaylareopelle the way the dependency detection works actually automatically checks the config for us (for anything using instrumentation.something), that way we don't need to add a depends on block to check it. I think that file you mentioned has the configs there because the configs are not the same format because it has been around longer than the `instrumentation.something` configs. That file also has 2 depends on blocks, the other one is checking to see if things are defined that need to be.
But also yeah, there is no reason Thread wouldn't be defined right? since it's part of ruby itself. We probably don't need the depends on block to check that then.

lib/new_relic/agent/instrumentation/thread/chain.rb
tannalynn: Sure, I can totally change it to be that way instead!
The only reason i didn't do that originally is because since i'm making a new block to wrap around it, I'm not treating it how we typically deal with blocks in the agent, so I just had it be a normal arg instead.

lib/new_relic/agent/transaction.rb
tannalynn: hmmm I'm not sure how we could use a case here really. Since case statements are usually checking a single variable against multiple potential values, but in this situation we are checking multiple hash entries to see if they just exist before using them. I'm not sure how that would work, but i'd be open to refactoring this area. It's kind of an awkward method, so when i tried refactoring it before i didn't end up with anything good, so i just put it back. I'd be happy to pair on it if you'd like.

lib/new_relic/agent/transaction.rb
tannalynn: for sure, I didn't even notice the lack of them since i was eagle eye searching for `@current_segment` part of things lol.

lib/new_relic/traced_thread.rb
tannalynn: Sure, i'd be happy to go over this stuff! How this works was mostly taken from ancient lore uncovered on google at one point about how to make threads work in the agent (aka some super old github gist from like 8 or 10 years ago irrc, i'll see if i can find it again)

As for where it gets set originally, there is the method `state_for` in `lib/new_relic/agent/transaction.rb` that will create a new state and set it if it doesn't exist yet.

lib/new_relic/agent/transaction.rb
kaylareopelle: I'm not sure how a case statement would exactly work here either, but I'm also happy to pair if you want to test it out later. I consider this a low priority compared to other things, so let's table it unless we have some extra time!

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: Riiiiight. Thanks for clarifying!

lib/new_relic/traced_thread.rb
tannalynn: Oh not quite as old as I thought it was https://gist.github.com/mikeantonelli/0d7563db4ad28d7ee10d365b295ce3f4

lib/new_relic/traced_thread.rb
tannalynn: hmm interesting idea. right now all args passed in are passed straight to the thread, but maybe we could peek at them to just see if something specific is present for the name.
also, yeah i wasn't sure what to really call the segment, i'm definitely open to talking about it and seeing what might be a better name also.

test/new_relic/agent/stats_engine_test.rb
tannalynn: Actually, by moving the in_transaction block outside of the threads (since only one transaction can be happening at once, and now we can see inside threads) that solved that specific issue. Currently all the numbers match up consistently, however there are no scoped metrics being recorded, which is very odd. I haven't dug back into that issue yet, but that is where i left it previously. I suspect the inconsistent numbers we were seeing was due to the agent being unable to track more than one transaction at a time now that the threads are getting the agents state.

lib/new_relic/agent/agent.rb
tannalynn: Yeah it didn't look like anything was trying to be in alphabetical order at all, I just put this one up here instead of lower in the list since this one was just a `new_relic/something` rather than a lot of the other ones that are `new_relic/agent/something`.

tannalynn:
> One other thought! Does testing this with `concurrent-ruby` feel like its within the scope of this work?

hmm, I think it would be interesting to test that out, but i guess also that is a totally separate gem. Since we're not adding support for concurrent ruby at this time I don't think we should add any tests about it to multiverse, but it might be good to try it out in a playground app and see how it reacts to concurrent ruby.

lib/new_relic/traced_thread.rb
```diff
@@ -0,0 +1,38 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  #
+  # This class creates a thread that contains instrumentation to allow the agent to see spans created inside of this thread.
```
kaylareopelle, line 8: What about something like this?
```suggestion
  # This class allows the state of a transaction to be passed to a Thread so that nested segments can be created from the operations performed within the Thread's block. 
```

lib/new_relic/traced_thread.rb
```diff
@@ -0,0 +1,38 @@
...
+  # In order to have this functionality inserted into all threads automatically,
+  # enable the `instrumentation.thread.tracing` configuration option in your newrelic.yml
```
kaylareopelle, line 10: What happens if `instrumentation.thread` is `disabled`? Will this still work?

lib/new_relic/traced_thread.rb
```diff
@@ -0,0 +1,38 @@
...
+  #
+  # @api public
+  class TracedThread < Thread
+    #
+    # Creates a new thread that will be traced by the agent.
```
kaylareopelle, line 15:
```suggestion
    # Creates a new Thread that will be traced by the agent.
```

lib/new_relic/traced_thread.rb
```diff
@@ -0,0 +1,38 @@
...
+    # Use this class the same as the Thread class
```
kaylareopelle, line 16:
```suggestion
    # Use this class as a replacement to creating Threads using `Thread.new`:
    # ```ruby
    #   TracedThread.new { execute_some_code }
    # ```
```

kaylareopelle: Do you think users will be confused by the instrumentation being on `#initialize` rather than `#new`?

kaylareopelle (commented):
I have a few suggestions with documentation feedback! Great foundation :)

lib/new_relic/traced_thread.rb
tannalynn: Good point. It will still work, but the span parents will be a little wonky. I'll add something about that config setting as well

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
...
+
+    Previously, the agent was not able to record events and metrics inside Threads created inside of an already running transaction. This release includes 2 new configuration options to support multithreaded applications to automatically instrument threads. To support applications that only want some threads instrumented by New Relic, a new class is available `NewRelic::TracedThread` that will create a thread that includes New Relic instrumentation, see our [API documentation](https://rubydoc.info/github/newrelic/newrelic-ruby-agent/NewRelic) for more details. 
+
+    New configuration options included in this release:
+    | Configuration name | Default | Behavior |
+    | ----------- | ----------- |----------- |
+    | `instrumentation.thread`  | `auto` (enabled) | Allows the agent to correctly nest spans inside of an asyncronous transaction   |
+    | `instrumentation.thread.tracing` | `false` (disabled)   |  Automatically add tracing to all Threads created in the application. This may be enabled by default in a future release. |
+
+
```
kaylareopelle, line 15: I think it would be nice to thank the customer who shared his gist with us in this entry as long as Erik thinks it's ok.

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # New Relic Ruby Agent Release Notes #
 
+  ## v8.7.0
+
+  * **Improved async support and Thread instrumentation**
+
+    Previously, the agent was not able to record events and metrics inside Threads created inside of an already running transaction. This release includes 2 new configuration options to support multithreaded applications to automatically instrument threads. To support applications that only want some threads instrumented by New Relic, a new class is available `NewRelic::TracedThread` that will create a thread that includes New Relic instrumentation, see our [API documentation](https://rubydoc.info/github/newrelic/newrelic-ruby-agent/NewRelic) for more details.
```
kaylareopelle, line 7: In addition to the rubydoc.info link, what do you think about adding `NewRelic::TracedThread` to our New Relic docs Ruby Custom Instrumentation docs too? https://docs.newrelic.com/docs/apm/agents/ruby-agent/api-guides/ruby-custom-instrumentation

lib/new_relic/traced_thread.rb
```diff
@@ -0,0 +1,46 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  #
+  # This class allows the state of a transaction to be passed to a Thread so that nested segments can be created from the operations performed within the Thread's block.
+  # To have the New Relic Ruby agent automatically trace all of your applications threads,
+  # enable the `instrumentation.thread.tracing` configuration option in your newrelic.yml
```
kaylareopelle, line 10:
```suggestion
  # enable the `instrumentation.thread.tracing` configuration option in your newrelic.yml.
```

lib/new_relic/traced_thread.rb
```diff
@@ -0,0 +1,46 @@
...
+  #
+  # Note: disabling the configuration option `instrumentation.thread` while using this class can cause incorrectly nested spans
```
kaylareopelle, line 12:
```suggestion
  # Note: disabling the configuration option `instrumentation.thread` while using this class can cause incorrectly nested spans.
```

lib/new_relic/traced_thread.rb
```diff
@@ -0,0 +1,46 @@
...
+  #
+  # @api public
+  class TracedThread < Thread
+    #
+    # Creates a new Thread whose work will be traced by New Relic.
+    # Use this class as a replacement for the native Thread class.
+    # Example: instead of using `Thread.new`, use
```
kaylareopelle, line 19:
```suggestion
    # Example: Instead of using `Thread.new`, use:
```

lib/new_relic/traced_thread.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      instrumentation = ::Thread.current[:newrelic_tracer_state]
+      Proc.new do
+        begin
+          ::Thread.current[:newrelic_tracer_state] = instrumentation
+          segment = NewRelic::Agent::Tracer.start_segment(name: "Thread#{::Thread.current.object_id}")
+          block.call(*args) if block.respond_to?(:call)
+        ensure
+          segment.finish if segment
+        end
```
kaylareopelle, line 42: Since these lines duplicate the code in `lib/new_relic/agent/instrumentation/thread/instrumentation.rb`, I wonder if a separate Helper/Concern/Class could be created that held the logic for both? This might help avoid problems down the line if/when this code needs to be updated so that it can be maintained in one location.

test/new_relic/agent/stats_engine_test.rb
```diff
@@ -178,22 +178,54 @@ def test_record_scoped_and_unscoped_metrics_records_multiple_unscoped_metrics
...
-      threads << Thread.new do
-        iterations.times do
-          in_transaction('txn') do
-            @engine.tl_record_scoped_and_unscoped_metrics('m1', ['m3'], 1)
-            @engine.tl_record_scoped_and_unscoped_metrics('m2', ['m4'], 1)
+    with_config(:'instrumentation.thread.tracing' => false) do
+      nthreads.times do |tid|
+        threads << Thread.new do
+          iterations.times do
+            in_transaction('txn') do
```
kaylareopelle, line 190: This reminds me, should we write somewhere that it's no longer supported to create transactions inside of threads if `instrumentation.thread.tracing` is true?

test/performance/suites/external_segment.rb
```diff
@@ -39,6 +39,17 @@ def test_external_request
     stop_server io_server
   end
 
+  def test_external_request_in_thread
```
kaylareopelle, line 42: How do these performance tests look with the latest updates?

kaylareopelle (commented):
I'm so happy to see this coming together. I have a few documentation-related questions and an implementation suggestion. 

Also, could you add the new config options to `newrelic.yml`? Thanks!

CHANGELOG.md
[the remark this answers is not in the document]
tannalynn: Yeah we could totally do that if you think it's a good idea, I just didn't include it since it came through support instead of from github

test/new_relic/agent/stats_engine_test.rb
tannalynn: I haven't seen anything in our documentation that says that we definitely do support it, or any recommendations for doing it. If you have seen anything like that we should definitely update that, but I don't think people are actually doing this very often, it's probably pretty rare (if anyone is doing it), so it might just be confusing for most people

kaylareopelle: No need to take action then! :)

CHANGELOG.md
tannalynn: Oh interesting, I hadn't considered that. Most of our API methods aren't described there, it seems like that is primarily describing how to do add_method_tracer and add_transaction_tracer stuff, which i think is the most common custom instrumentation usage. I'm not sure if this is something that should be added there or just in the API documentation. what do you think?

test/performance/suites/external_segment.rb
tannalynn: I haven't run them yet! Thanks for the reminder, I definitely want to do that.

CHANGELOG.md
kaylareopelle: Hmm... maybe this is a better question for the docs hero? 

@bradleycamacho - Should all of the Ruby Agent API methods be documented in two places -- once in the more Ruby-standard location and one on the New Relic site? Does this feature seem like it would fit within the scope of [this article](https://docs.newrelic.com/docs/apm/agents/ruby-agent/api-guides/ruby-custom-instrumentation)?

bradleycamacho: I'm open to either option, but my preferred method is keeping the info in the external ruby docs as long as:

1. `NewRelic::TracedThread` is more of an advanced configuration than a basic custom instrumentation
2. We add a more prominent link to the external ruby docs in our NR doc. Possibly in the intro or in a callout right below it.

The other option is creating a small section right under our [advanced configuration section](https://docs.newrelic.com/docs/apm/agents/ruby-agent/api-guides/ruby-custom-instrumentation/#advanced-tracing). If we do this, we should ensure that it follows that pattern of being more of a full example/procedure instead of just defining it. What I'd want to avoid is just replicating content that may get out of sync (someone updates our doc and not the ruby docs or vice-versa).

kaylareopelle: @tannalynn @fallwith - As a developer, it's more intuitive to me to update the RubyDoc site through `@api public` comments than adding something to the newrelic docs site. I'm leaning more toward #2, adding a more prominent link to the external ruby docs site in our NR doc. 

What are your thoughts?

I think I'm caring extra here because I want people to use this and know we have it!

tannalynn: I'd say this is definitely more of an advanced custom instrumentation, I don't expect it would be one of the first things people would be looking for. So I think I also like the suggestion of adding the link to the rubydocs in a place that is more prominent. Also, since rubydocs are automatically updated when we release, it should always be fully up to date which is nice.

fallwith: I'm nodding along with all that's been brought up so far. Personally I'd find it useful to have a dedicated new section titled "Multithreading" / "Concurrency" / "Asynchronous Transactions" / "Threads" / etc. that clearly lay out the fact that we have 2 options: all threads, or just those we're explicitly told about, with source/configuration code block examples for each. It being a part of the advanced configuration section seems good.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -901,6 +901,21 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :default => 'auto',
+          :public => true,
+          :type => String,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => "Controls auto-instrumentation of the Thread class at start up to enable accurate parent tracking for spans in applications manually inserting New Relic tracing into threads. This does not enable the agent to automatically trace all threads created (see `instrumentation.thread.tracing`). May be one of [auto|prepend|chain|disabled]."
```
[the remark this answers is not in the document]
tannalynn: I will do that! these were ones i originally wrote and never looked at again. I'll get that updated.

CHANGELOG.md
kaylareopelle: Excellent point! Perhaps that could be the new purpose for https://github.com/newrelic/newrelic-ruby-agent/issues/1057?

fallwith: Certainly!

lib/new_relic/agent/transaction.rb
```diff
@@ -261,6 +261,30 @@ def initialize(category, options)
         end
       end
 
+      def parent_thread_id
+        Thread.current.nr_parent_thread_id if Thread.current.respond_to?(:nr_parent_thread_id)
+      end
+
+      def current_segment
+        # if this thread id has a current segment use that
```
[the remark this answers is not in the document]
tannalynn: No not really, those were mostly from when i was figuring out what I wanted to do. I'll remove some of those that seem obvious now.

lib/new_relic/agent/stats.rb
```diff
@@ -122,17 +129,35 @@ def ==(other)
...
-        else
-          ::NewRelic::Agent.logger.warn("Attempted to set apdex_t to #{apdex_t.inspect}, backtrace = #{caller.join("\n")}")
+      end
+
+      # Override marshalling methods to exclude @lock from being included in marshalled data
+      def marshal_dump
+        instance_variables.reject { |name| SKIP_MARSHALLING.include?(name) }.inject({}) do |instance_copy, name|
+          instance_copy[name] = instance_variable_get(name)
+          instance_copy
+        end
```
fallwith, line 152: Two things here: 1) I think we should use `#each_with_object`  (available since Ruby 1.9) instead of `#inject`, and 2) I think we should simply perform the skipping of names ourselves, as `#reject` will end up looping over every single value a second time otherwise.

Here's a suggested rewrite:

```ruby
def marshal_dump
  instance_variables.each_with_object({}) do |name, instance_copy|
    next if SKIP_MARSHALLING.include?(name)
    instance_copy[name] = instance_variable_get(name)
  end
end
```

tannalynn: I do like that a lot better, it looks a lot nicer, I'll get that updated, thank you!

lib/new_relic/traced_thread.rb
tannalynn: Yeah that's a good idea. I ended up putting the new method inside of Tracer, since that is where the other methods that deal with the state live.

lib/new_relic/agent/transaction.rb
```diff
@@ -261,6 +261,30 @@ def initialize(category, options)
         end
       end
 
+      def parent_thread_id
+        Thread.current.nr_parent_thread_id if Thread.current.respond_to?(:nr_parent_thread_id)
```
fallwith, line 265: Maybe `::Thread` here

lib/new_relic/agent/transaction.rb
fallwith, line 272: No need for the explicit `return` on the default case here.

tannalynn: True, i just liked the symmetry with the lines above it when it has the return. I can totally remove it though, since that is the standard way to do it.

lib/new_relic/agent/transaction/tracing.rb
```diff
@@ -6,7 +6,7 @@ module NewRelic
   module Agent
     class Transaction
       module Tracing
-        attr_reader :current_segment
+        attr_reader :current_segment_by_thread
```
fallwith, line 9: I can't find where this reader is used

tannalynn: It's used in the new methods in Transaction for current segment. Those could be updated to be @current_segment_by_thread though, and remove this reader since those are the only places using it.

fallwith: Oh, I see what happened. I typed in "current_segment_by_id " for my search for some reason. Keeping it is ok by me.

tannalynn: Ahh okay that makes sense! Sounds good!

lib/new_relic/agent/transaction/tracing.rb
```diff
@@ -34,7 +34,12 @@ def add_segment segment, parent = nil
         end
 
         def segment_complete segment
-          @current_segment = segment.parent
+          # if parent was in another thread, remove the current_segment entry for this thread
+          if segment.parent && segment.parent.starting_thread_id != ::Thread.current.object_id
+            remove_current_segment_by_thread_id(::Thread.current.object_id)
```
fallwith, line 39: Is it possible for 2 threads to reach this code at the same time, and if so would there be a problem with the simultaneous deletion operation calls?

tannalynn: Hmm, that's a great question. Since it's deleting based on the thread ID, they would definitely be trying to delete _different_ things at least, but I'm not sure how hash handles that situation. i'll look into it

lib/new_relic/traced_thread.rb
```diff
@@ -0,0 +1,36 @@
...
+module NewRelic
+  #
+  # This class allows the state of a transaction to be passed to a Thread so that nested segments can be created from the operations performed within the Thread's block.
+  # To have the New Relic Ruby agent automatically trace all of your applications threads,
+  # enable the `instrumentation.thread.tracing` configuration option in your newrelic.yml.
+  #
+  # Note: disabling the configuration option `instrumentation.thread` while using this class can cause incorrectly nested spans.
+  #
+  # @api public
+  class TracedThread < Thread
```
fallwith, line 15: Grammatically I'm wondering if this is "TracedThread" then the other name should be "MonitoredThread" or if that one should stay "ThreadMonitor" and this one because "ThreadTrace". But given that they are used in an either / or fashion, it's not a blocker.

tannalynn: I like TracedThread over ThreadTrace. Since this inherits from Thread and adds tracing in it, i think TracedThread seems more obvious that it's creating a thread object. As for ThreadMonitor, I have no preference on that name really because originally it was just Thread, but that was causing annoying namespace issues in other parts of the agent so I just changed it to anything else to avoid that. I'd be down to update that one.

fallwith: Ok cool. Let's called TracedThread for sure. @kaylareopelle any preference on the naming of `ThreadMonitor`? We can name it anything but `Thread`.

test/new_relic/agent/instrumentation/rails/active_storage_subscriber.rb
```diff
@@ -1,6 +1,7 @@
 # encoding: utf-8
 # This file is distributed under New Relic's license terms.
 # See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+require File.expand_path '../../../../../test_helper', __FILE__
```
fallwith, line 4: 👍 This is one of the better implementations of this `require` that would make a good candidate for https://github.com/newrelic/newrelic-ruby-agent/issues/1094

lib/new_relic/traced_thread.rb
kaylareopelle: I think I'd prefer MonitoredThread over ThreadMonitor :) 
Another option could be InstrumentedThread, given that it's the instrumentation for the Thread. I know that's a bit redundant with the prefix, but I kind of like how `NewRelic::Agent::Instrumentation::InstrumentedThread` reads. Thoughts?

tannalynn: Hmm i'm less of a fan of that, the double 'Instrument' part looks odd to me. personally, I like MonitoredThread over InstrumentedThread i think.

kaylareopelle: Cool! that works!

lib/new_relic/agent/transaction/tracing.rb
tannalynn: I tested this behavior out in a script, and didn't see any issues... and then i remembered the issues we had with jruby, so I checked that and jruby does have an issue with that. I'll have to add a lock for it.

tannalynn: Ok. I added a lock to the methods `remove_current_segment_by_thread_id` and `set_current_segment` in Transaction, since those are the only places where we're actually modifying the hash directly.

fallwith (approved):
👍

CHANGELOG.md
kaylareopelle: #1057 updated with this scope.

newrelic.yml
```diff
@@ -395,6 +395,13 @@ common: &default_settings
 # May be one of [auto|prepend|chain|disabled].
 # instrumentation.typhoeus: auto
 
+# Controls auto-instrumentation of the Thread class at start up to allow the agent to correctly nest spans inside of an asyncronous transaction.
+# May be one of [auto|prepend|chain|disabled].
+# instrumentation.thread: auto
+
+# Controls auto-instrumentation of the Thread class at start up to automatically add tracing to all Threads created in the application.
+# instrumentation.thread.tracing: true
```
fallwith, line 403: I think the boolean value here should match what is in `default_source.rb` (false), perhaps with a note that explicitly says "set to \`true\` to enable" or whatnot.

lib/new_relic/agent/threading/internal_thread.rb
```diff
@@ -0,0 +1,14 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+
+module NewRelic
+  module Agent
+    module Threading
+      # This class is used internally by the agent to prevent agent threads from sharing a state with application threads
+      class InternalThread < ::Thread
```
fallwith, line 9: I love this solution. Do you think this new class could be used in the future to help accommodate a blocklist of app threads that should not be monitored? I'm thinking the agent could use this new class for those types of threads. If so, I propose that this class be renamed and the comment updated so that we go with something more generic than "internal". Perhaps something like "NonMonitored", "Unmonitored", "Ignored", "Unreported", etc.

tannalynn: Are you thinking that this should be a public api class like we did for TracedThread?

fallwith: Yep, exactly. But after some quick sketching out of how it would work, I've changed my mind.

It's a lot easier to rename and expose a private API later than it is to make something public now prematurely before it's wired up to anything.

The more I think about it, the more I'd prefer to work out a long term solution that allows an app to simply use `::Thread` but tell us whether or not it should be traced. Perhaps by monkeypatching `::Thread.initialize`. That would allow the app's own source code to continue to use `::Thread` instead of a NR proprietary class name.

I withdraw my suggestion for a rename. I think we should leave things as is for now before we work out the idea more.

test/performance/suites/external_segment.rb
tannalynn: Completed the perf suite tests for this PR

CHANGELOG.md
```diff
@@ -2,10 +2,22 @@
 
   ## v8.7.0
 
+  * **Improved async support and Thread instrumentation**
+
+    Previously, the agent was not able to record events and metrics inside Threads created inside of an already running transaction. This release includes 2 new configuration options to support multithreaded applications to automatically instrument threads. To support applications that only want some threads instrumented by New Relic, a new class is available `NewRelic::TracedThread` that will create a thread that includes New Relic instrumentation, see our [API documentation](https://rubydoc.info/github/newrelic/newrelic-ruby-agent/NewRelic) for more details.
```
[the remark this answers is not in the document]
tannalynn: Updated to add more info!

fallwith (dismissed):

CHANGELOG.md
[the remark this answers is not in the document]
kaylareopelle: Checking with one more person for approval before I'll have the yea or nay on this commendation.

fallwith (approved):

kaylareopelle (dismissed):
I'm so excited for this feature, Tanna! 🚀

kaylareopelle (approved):
✨ 🐐 ✨

fallwith (approved):

CHANGELOG.md
```diff
@@ -2,10 +2,22 @@
...
+
+    New configuration options included in this release:
+    | Configuration name | Default | Behavior |
+    | ----------- | ----------- |----------- |
+    | `instrumentation.thread`  | `auto` (enabled) | Allows the agent to correctly nest spans inside of an asyncronous transaction   |
+    | `instrumentation.thread.tracing` | `false` (disabled)   |  Automatically add tracing to all Threads created in the application. This may be enabled by default in a future release. |
+
```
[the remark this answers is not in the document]
mikeantonelli: 🎉 thank you @kaylareopelle!

PR #1081, opened 2022-04-12
config tests: update FakeFS usage for Ruby 3.2+

fallwith:
These changes get the config file tests working with Ruby 3.2+

* For Ruby v2.4.0+ (lowest version supported), start using the latest
  version of FakeFS, v1.4.1
* For all Rubies, don't allow Bundler to require FakeFS when `Envfile`
  gets processed. This prevents a known collision issue with `FileUtils`
  that the FakeFS README lists under "Caveats".


tannalynn (approved):
Looks good, thank you!

PR #1082, opened 2022-04-12
delayed job tests: use a newer sqlite3 version

fallwith:
for the delayed job tests, use sqlite3 v1.4.2 instead of v1.3.13

this addresses a compilation issue with Ruby 3.2 and the new version
should be backwards compatible with older Rubies as well.


tannalynn (approved):
🎉

PR #1083, opened 2022-04-12
CI: upgrade CRubies for CVE-2022-28739

fallwith:
Upgrade CRubies 2.6+ to their latest minor versions in order to address
[CVE-2022-28739](https://www.ruby-lang.org/en/news/2022/04/12/buffer-overrun-in-string-to-float-cve-2022-28739/)
which involes a String-to-Float conversion buffer overrun.

Versions 3.0.4 and 3.1.2 also deliver fixes for [CVE-2022-28738](https://www.ruby-lang.org/en/news/2022/04/12/double-free-in-regexp-compilation-cve-2022-28738/)

### DIFF
--- a/.github/actions/build-ruby/action.yml
+++ b/.github/actions/build-ruby/action.yml
@@ -5,7 +5,7 @@ inputs:
   ruby-version:
     description: 'Ruby version'
     required: true
-    default: '2.7.5'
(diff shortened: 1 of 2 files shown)


tannalynn (approved):
Thanks for getting this updates so quickly!

PR #1084, opened 2022-04-12
tests: assume Ruby 2.2.10+

fallwith:
assume Ruby version 2.2.10 or higher is used for all tests and disable
or rework any checks that previous tested for older versions


kaylareopelle (approved):
Thanks for cleaning this up! 🧹

Issue #1075, opened 2022-04-08
MLT: Conduct performance A/B tests with and without MLT enabled


fallwith:
Performance tests should be conducted against the `method_level_telemetry` branch with MLT enabled and disabled via the new configuration setting and optionally against the `dev` branch as well if the configuration setting is not thought to be sufficient.

We are seeking to determine:

* What is the difference in performance if MLT is enabled?
* Within the performance difference, how much of it is caused by introspecting the Ruby source code and determining values for the 4 MLT metrics?
* Within the performance difference, how much of it is caused by the increased segment size that gets reported with 4 additional MLT metrics?

angelatan2:
The scope of this ticket is limited to stackprof and exclude OATs.

kaylareopelle:
One more version of the output. Here "before" is with code-level metrics turned off. "after" is with them turned on. 
This test was run on the method_level_telemetry branch.
| name                                                      | before    | after     | delta  | allocs_before | allocs_after | allocs_delta | retained | retained_delta |
|-----------------------------------------------------------|-----------|-----------|--------|---------------|--------------|--------------|----------|----------------|
| MethodTracerTest#test_tracer_code_level_metrics           | 305.16 µs | 271.85 µs | -10.9% |           442 |          442 |         0.0% |        0 |           0.0% |
| MethodTracerTest#test_with_metric_code_level_metrics      | 312.52 µs | 295.89 µs |  -5.3% |           442 |          442 |         0.0% |        0 |           0.0% |
| MethodTracerTest#test_with_proc_code_level_metrics        | 324.55 µs | 276.11 µs | -14.9% |           442 |          442 |         0.0% |        0 |           0.0% |
| MethodTracerTest#test_push_scope_false_code_level_metrics | 334.80 µs | 281.63 µs | -15.9% |           442 |          442 |         0.0% |        0 |           0.0% |
| MethodTracerTest#test_metric_false_code_level_metrics     | 278.35 µs | 270.68 µs |  -2.8% |           442 |          442 |         0.0% |        0 |           0.0% |

kaylareopelle:
## Results dev (before) compared with method_level_telemetry (after)

The following tests were not found in the baseline results:
  MethodTracerTest#test_tracer_code_level_metrics_enabled
  MethodTracerTest#test_tracer_code_level_metrics_disabled
  MethodTracerTest#test_with_metric_code_level_metrics_enabled
[...]

PR #1086, opened 2022-04-13
CLM: Performance tests for :add_method_tracer

kaylareopelle:
# Overview
* This PR adds performance tests with `:'code_level_metrics.enabled'` set to true and false.
* Each of the `:add_method_tracer` options are tested.
* The test file can be run within the `test/performance` directory using `script/runner -n code_level_metrics`

# Related Github Issue
Relates to #1075

### DIFF
--- a/test/performance/suites/method_tracer.rb
+++ b/test/performance/suites/method_tracer.rb
@@ -1,0 +1,72 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+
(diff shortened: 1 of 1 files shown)


kaylareopelle:
We don't currently have a mechanism to test the change of an object's size between two scenarios. I'd like to try using `ObjectSpace.memsize_of(object)`.

kaylareopelle:
> We don't currently have a mechanism to test the change of an object's size between two scenarios. I'd like to try using `ObjectSpace.memsize_of(object)`.

If we want to test this, it can probably be in a different PR.

fallwith (approved):

PR #1088, opened 2022-04-14
Support Ruby 3.2+ cache invalidation stats

### LINKED ISSUE #1059: Ruby 3.2 support - Report on the new VM's `:constant_cache` stat instead of the old `:global_constant_state`
As per [Ruby PR #5433](https://github.com/ruby/ruby/pull/5433), which satisfies [feature #18589](https://bugs.ruby-lang.org/issues/18589), the CRuby VM now provides finer grained control over the constants cache.

For Ruby 2.1.0 through 3.1.x, the Ruby agent has been making use of `RubyVM.stat[:global_constant_state]` to tally up cache invalidations (see: `lib/new_relic/agent/vm/mri_vm.rb`).

For Ruby 3.2+, the `RubyVM.stat` hash no longer has a `:global_constant_state` key. Instead it now has a `:constant_cache` key, and a hash as the corresponding value. That inner hash keeps track of cache on a per constant basis, with the constants themselves serving as the inner hash keys.

My current guess is that we need to replace the existing call to `RubyVM.stat[:global_constant_state]` with something like `RubyVM.stat[:constant_cache].values.sum`, but I have not taken the time yet to thoroughly research the RubyVM change that took place.

fallwith:
Ruby 3.2 uses a new `RubyVM` stat to track more fine grained constant
cache invalidation data. Sum the fine grained values together to produce
a Ruby 3.1 compatible value for now and leave a TODO for possibly
reporting the individual constant key/value pairs later.

### DIFF
--- a/lib/new_relic/agent/vm/mri_vm.rb
+++ b/lib/new_relic/agent/vm/mri_vm.rb
@@ -53,7 +53,7 @@ module NewRelic
           end
         end
 
-        def gather_constant_cache_invalidatiions
+        def gather_constant_cache_invalidations
           # Ruby >= 3.2 uses :constant_cache
           # see: https://github.com/ruby/ruby/pull/5433 and https://bugs.ruby-lang.org/issues/18589
           # TODO: now that 3.2+ provides more granual cache invalidation data, should we report it instead of summing?
--- a/lib/new_relic/agent/vm/mri_vm.rb
+++ b/lib/new_relic/agent/vm/mri_vm.rb
@@ -49,7 +49,7 @@ module NewRelic
           end
 
           if supports?(:constant_cache_invalidations)
-            snap.constant_cache_invalidations = gather_constant_cache_invalidatiions
+            snap.constant_cache_invalidations = gather_constant_cache_invalidations
           end
         end
 
--- a/lib/new_relic/agent/vm/mri_vm.rb
+++ b/lib/new_relic/agent/vm/mri_vm.rb
@@ -49,7 +49,19 @@ module NewRelic
           end
 
           if supports?(:constant_cache_invalidations)
-            snap.constant_cache_invalidations = RubyVM.stat[:global_constant_state]
+            snap.constant_cache_invalidations = gather_constant_cache_invalidatiions
+          end
+        end
+
+        def gather_constant_cache_invalidatiions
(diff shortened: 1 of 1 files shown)


lib/new_relic/agent/vm/mri_vm.rb
```diff
@@ -49,7 +49,19 @@ def gather_ruby_vm_stats(snap)
           end
 
           if supports?(:constant_cache_invalidations)
-            snap.constant_cache_invalidations = RubyVM.stat[:global_constant_state]
+            snap.constant_cache_invalidations = gather_constant_cache_invalidatiions
```
kaylareopelle, line 52: small typo: 
```suggestion
            snap.constant_cache_invalidations = gather_constant_cache_invalidations
```

lib/new_relic/agent/vm/mri_vm.rb
```diff
@@ -49,7 +49,19 @@ def gather_ruby_vm_stats(snap)
...
+          end
+        end
+
+        def gather_constant_cache_invalidatiions
```
kaylareopelle, line 56: small typo: 
```suggestion
        def gather_constant_cache_invalidations
```

tannalynn (approved):
Looks good! Glad you noticed this change in ruby 3.2

kaylareopelle (dismissed):
Would you mind creating an issue for the 3.2 todo as part of this merge so that we can track it when we start 3.2 work?

Other than that and the typos, this looks good to me!

lib/new_relic/agent/vm/mri_vm.rb
tannalynn: 🦅 👁️

lib/new_relic/agent/vm/mri_vm.rb
fallwith: So my hardware apparently let me down with a bit of key chatter and then my own laziness in copy and pasting and not proofreading let me down further! Thanks for having my back!

fallwith: > Would you mind creating an issue

Good call. I've gone ahead and updating the original issue and marked this PR as not resolving it so that it can continue to track efforts to enhance our reporting with 3.2 specific functionality.

tannalynn (approved):

PR #1093, opened 2022-04-15
Use transaction_event_data to set interval

kaylareopelle:
# Overview
Previously, the 5-second harvest cycle value for `transaction_events.max_samples_stored` was being applied every 60 seconds because the `NewRelic::Agent::StartWorkerThread#interval_for` method was still using `analytic_event_data` as a reference point. Since that value was no longer being set with the updates made in https://github.com/newrelic/newrelic-ruby-agent/pull/952, the transaction events were falling back to the default 60-second cycle. This caused the rate of sampling to be too high, so customers reaching the limit would receiving only 100 transactions per minute instead of 1200 transactions per minute as a default.

# Related Github Issue
[GTSE-14220 (Internal)](https://newrelic.atlassian.net/browse/GTSE-14220)

# Testing
WIP


eiskrenkov:
Hey, @kaylareopelle, sorry to disturb you, especially about such old commit, but my colleagues and I are struggling to figure out one thing, related to this.

So after upgrading `newrelic_rpm `to `8.7.0` our NewRelic ingest grew **12** times bigger than on `8.6.0`. We tried to investigate the whole `8.6.0...8.7.0` diff, but the only thing that is at least somehow related is this commit. I wanted to ask you to help us to understand, how exactly changes, introduced in this commit affect `max_samples_stored`. We really dug into it, tried to understand the code for hours, but still no luck. Our NewRelic config never modified `analytic_event_data`, neither `transaction_event_data`, so that we have values set to 60 for both of those keys, according to the config. It seems like this commit shouldn't have change anything for us, but it seems like it did.

I would be really grateful if you will find a minute to explain this to us, we really want to figure this out. Thank you in advance!

hannahramadan:
Hi @eiskrenkov! In agent version 8.5.0, we introduced a bug with [PR#952](https://github.com/newrelic/newrelic-ruby-agent/pull/952) that caused the agent to over-sample transactions (only sending 100 transactions/minute instead of the intended 1200 transactions/minute). 

The reason this happened is that in 8.5.0, we deprecated the term `analytic_events` in favor of `transaction_events`, but missed that update for the method `NewRelic::Agent::StartWorkerThread#interval_for`, which caused us to fallback to the default 60-second cycle (instead of the intended 5-second cycle). PR#1093 corrects this and returns sampled transactions to 1200 transactions/minute. The sampled transactions/minute going from 100 to 1200 explains the 12x increase you are seeing.

By default, the agent is supposed to sample 1200 transactions/minute. But if you would like to return the agent to the behavior you were seeing before the bugfix, you can set the config option `transaction_events.max_samples_stored` to 100.

eiskrenkov:
Thank you @hannahramadan! Will try to set `transaction_events.max_samples_stored` to 100

PR #1095, opened 2022-04-16
Improve configuration documentation for rake.tasks

### LINKED ISSUE #1087: Default behaviour of rake instrumentation is not clear

It has recently come to my attention that none of our rake tasks are being instrumented. Trying to understand the expected default behavior is difficult from newrelic.yml.

Between a clean newrelic.yml and looking at the defaults you get:
* autostart.denylisted_rake_tasks: (a whole load of common tasks specified by `AUTOSTART_DENYLISTED_RAKE_TASKS`)
* instrumentation.rake: auto
* rake.tasks: [] _(this bit of config is furthest down the file!)_
* disable_rake: false _(not in newrelic.yml)_

I'd expected leaving `instrumentation.rake: auto` and customising `autostart.denylisted_rake_tasks` to automatically instrument rake tasks that were not on the deny-list, but this isn't the case.

Looking in the code, it looks like you are supposed to manually specify a list of rake tasks to instrument "automatically", which is surprising in two ways:

1. Manually specifying a list of tasks isn't really what I'd call "automatic" instrumentation. Certainly not in the same way as, say, the ActiveRecord instrumentation (where I don't need to list out all the models explicitly)
2. It's quite common to either deny by default and have a defined allow-list, or to allow (instrument) by default and have a defined deny-list. However having an allow-list and then applying a deny-list against the allow-list is both unusual and unintuitive. If you have to list out the tasks in `rake.tasks`, it's not clear why you would then deny a task via `autostart.denylisted_rake_tasks` rather than just removing it from `rake.tasks`.

### Feature Description

Some possible actions I'd like to see at least one of:
* Updating the clean `newrelic.yml` comments to make it clear that _no_ rake tasks will be instrumented without them being added to `rake.tasks` first
* Updating the clean `newrelic.yml` comments to include the "allow all" regex from [this documentation](https://docs.newrelic.com/docs/apm/agents/ruby-agent/background-jobs/rake-instrumentation/)
* Updating the clean `newrelic.yml` comments to include a link to the [rake setup documentation](https://docs.newrelic.com/docs/apm/agents/ruby-agent/background-jobs/rake-instrumentation/) if it's not intuitive without reading it.
* Changing the default behaviour of "auto" to be to instrument all rake tasks except those on the deny-list. (I appreciate this is probably a step too far)

### Describe Alternatives

I think this is covered above

### Additional context

I think the above is clear enough

### Priority
Nice to Have

kaylareopelle:
# Overview
Improves documentation in `newrelic.yml` and the source code for the [Ruby Agent Configuration docs](https://docs.newrelic.com/docs/apm/agents/ruby-agent/configuration/ruby-agent-configuration) to include more information about the behavior and usage.

Many thanks to @robotfelix for suggesting these changes.

# Related Github Issue
Relates to #1087

# Testing
Documentation update, no tests required.


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -518,7 +518,7 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :type => Array,
           :allowed_from_server => false,
           :transform => DefaultSource.method(:convert_to_regexp_list),
-          :description => 'Specify an array of Rake tasks to automatically instrument.'
+          :description => 'Specify an array of Rake tasks to automatically instrument. This configuration option converts the Array to a RegEx list. If you\'d like to allow all tasks by default, use "[.+]". No rake tasks will be instrumented unless they\'re added to this list. For more information, visit our (Rake Instrumentation)[/docs/apm/agents/ruby-agent/background-jobs/rake-instrumentation] docs.'
```
fallwith, line 521: The single backslash looks fine here. If you prefer getting rid of it, you can replace the outermost single quotes with `%q{}`

newrelic.yml
```diff
@@ -452,6 +452,11 @@ common: &default_settings
 # rake.connect_timeout: 10
 
 # Specify an array of Rake tasks to automatically instrument.
+# This configuration option converts the Array to a RegEx list.
+# If you'd like to allow all tasks by default, use [.+].
```
fallwith, line 456: It's hard to correct for my bias of knowing how this works to decide what others would benefit from seeing. Maybe the use of quotes or the inclusion of `rake.tasks` would be help, but I'm not sure. Here are all 3 side by side for comparison:

- `, use [.+].`
- `, use "[.+]".`
- `, use "rake.tasks: [.+]"`

fallwith (approved):

newrelic.yml
kaylareopelle: Excellent point. I think my favorite is a riff on the last: 
```suggestion
# If you'd like to allow all tasks by default, use `rake.tasks: [.+]`.
```

What do you think about back tics vs. quotes?

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: What do you think about putting this in a HEREDOC so that the line is a little shorter?

fallwith: Zooming out to consider this entire file, I think it's rather silly that we have 255 descriptions stored in an .rb file in the first place. Perhaps we could source them from a file more suited to holding plain text. And then at that point I'd be curious to see if we could launch the agent without loading up all of that text when we don't need it.

newrelic.yml
fallwith: Good stuff! 👍

fallwith (dismissed):

fallwith (dismissed):

kaylareopelle:
Hi @robotfelix! 👋 

This PR addresses the documentation recommendations you made in issue #1087. Would you like to review and let us know if these changes alleviate your concerns?

kaylareopelle:
> Hi @robotfelix! 👋
> 
> This PR addresses the documentation recommendations you made in issue #1087. Would you like to review and let us know if these changes alleviate your concerns?

I'm going to merge this in as part of our next release. Feel free to recommend changes in a new issue!

fallwith (approved):

PR #1099, opened 2022-04-18
Logging: Turn log forwarding on by default

### LINKED ISSUE #1022: Turn application_logging.forwarding on by default
When it's time to turn the `application_logging.forwarding.enabled` setting on by default, we'll need to take care of the following: 

- [x] Update `default_source.rb` with the appropriate value
- [x] Update `newrelic.yml` file with the new default
- [x] Add [post-install message](https://brandonhilkert.com/blog/ruby-gem-post-install-message/) to the release about the feature being turned on
- [x] Update configuration docs with the new value

_More tasks TBD_

kaylareopelle:
# Overview
Sets the default for `application_logging.forwarding.enabled` to true.

# Related Github Issue
Closes #1022

# Testing
Updat[ing] tests to reflect the new default

# Todos
~~- [ ] Add CHANGELOG entry (once available)~~ - Moved to #1109

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1965,7 +1965,7 @@ A map of error classes to a list of messages. When an error of one of the classe
           :description => 'If `true`, enables log decoration and the collection of log events and metrics.'
         },
         :'application_logging.forwarding.enabled' => {
-          :default => false,
+          :default => true,
           :public => true,
           :type => Boolean,
           :allowed_from_server => false,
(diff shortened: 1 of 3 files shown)


fallwith (approved):
Given that we might want to drop the post install announcement after a few releases, perhaps it'd be good to list the change in CHANGELOG as well.

kaylareopelle:
> Given that we might want to drop the post install announcement after a few releases, perhaps it'd be good to list the change in CHANGELOG as well.

Absolutely! I am holding off on the CHANGELOG entry until I get the specific verbiage from Erik.

Issue #1046, opened 2022-03-27
Verify "big red button" works


kaylareopelle:
The "big red button" is an account-wide configuration setting to turn log forwarding off, regardless of the configuration setting in each individual account. To support this setting, we need to make sure the LogEventAggregator and the PrioritySampledBuffer will change the `application_logging.forwarding.max_samples_stored` value to zero when this value is sent by the collector.

For additional details, read [APM Agent Log Forwarding: Account-Level Shutoff Control (Internal)](https://newrelic.atlassian.net/wiki/spaces/TDP/pages/2431516817/APM+Agent+Log+Forwarding+Account-Level+Shutoff+Control)

kaylareopelle:
#1103 adds a unit test to our suite to verify the functionality.

The remaining work will be done in a pairing session with members of other teams to evaluate how the new setting responds end-to-end.

kaylareopelle:
After my User ID was added to the feature flag and I disabled account logs for the Ruby Agent Test account, I saw the `event_harvest_config.harvest_limit.log_event_data` value get reported as zero. The agent upheld this value and forwarded zero logs: https://staging.onenr.io/0DvwBK1GVRp

After this, I re-enabled account logs and had the response as the expected 833 and logs were forwarded to NR1: https://staging.onenr.io/0LZQWgb3BQW

PR #1103, opened 2022-04-19
Logging: Test agent response when connect reply is zero

kaylareopelle:
# Overview
The "big red button" sets the harvest limits for log_event_data to zero for all applications in an account. This PR adds a test that uses a connect reply of 0 for `event_harvest_config.harvest_limits.log_event_data` to see how the agent responds.

# Related Github Issue
Relates to #1046 

# Testing
Adds unit test to evaluate response to a connect reply of zero.

### DIFF
--- a/test/new_relic/agent/configuration/event_harvest_config_test.rb
+++ b/test/new_relic/agent/configuration/event_harvest_config_test.rb
@@ -88,5 +88,23 @@ module NewRelic::Agent::Configuration
       }
       assert_equal expected, EventHarvestConfig.to_config_hash(connect_reply)
     end
+
+    def test_to_config_hash_with_zero_response_for_log_event_data
+      connect_reply = {
+        'event_harvest_config' => {
+          'report_period_ms' => 5000,
+          'harvest_limits' => {
+            'log_event_data' => 0
+          }
+        }
+      }
+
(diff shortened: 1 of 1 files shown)


test/new_relic/agent/configuration/event_harvest_config_test.rb
```diff
@@ -88,5 +88,23 @@ def test_to_config_hash_with_omitted_event_type
...
+        }
+      }
+
+      expected = {
+        :'application_logging.forwarding.max_samples_stored' => 0,
+        :'event_report_period.log_event_data' => 5,
+        :event_report_period => 5
+      }
+      assert_equal expected, EventHarvestConfig.to_config_hash(connect_reply)
+    end
```
kaylareopelle, line 108: @tannalynn  - we chatted a bit about how this test would work last week. Do you think it's in the right spot?

tannalynn: Yeah, this seems good to me! I do think this is the right place for it

kaylareopelle: Thanks!

fallwith (approved):

Issue #1100, opened 2022-04-18
Update supported gem versions 


kaylareopelle:
### Feature Description
Some gems instrumented by the agent have a very low adoption rate. Others have not received a new release for many years or have very old versions still being tested. 

The goal for this ticket is to update our multiverse test suite and documented supported libraries to be more up to date so that we can remove little-used code and pave the way to support new libraries

- [x] Deprecate instrumented gems with a very low adoption rate
- [x] Remove old versions of gems from their multiverse suite
- [x] Update public documentation to display the new lowest supported version

[This spreadsheet (Internal)](https://docs.google.com/spreadsheets/d/1AIbDDKOasinWC4H2OXrkktuzNOIsVa4pQXw8vN_UhNY/edit#gid=0) has the actions that need to be taken for each gem.

PR #1104, opened 2022-04-19
Deprecate gems and update tested versions

kaylareopelle:
# Overview
Deprecates older versions of gems and instrumented gems with low customer adoption. Raises the lowest tested version for some gems.

# Related Github Issue
Relates to #1100 

# Testing
Updates and removes tests as needed


kaylareopelle:
Closing in favor of #1114

PR #1114, opened 2022-04-25
Deprecate gems and update Envfiles

### LINKED ISSUE #1100: Update supported gem versions

kaylareopelle:
* Deprecates older versions of gems and instrumented gems with low customer adoption. 
* Raises the lowest tested version for some gems.
* Adds TODOs to remove code in major version 9.0.0

Created from #1104 
Closes #1100

### DIFF
--- a/lib/new_relic/agent/agent.rb
+++ b/lib/new_relic/agent/agent.rb
@@ -460,6 +460,7 @@ module NewRelic
           # requests, we need to wait until the children are forked
           # before connecting, otherwise the parent process sends useless data
           def using_forking_dispatcher?
+            # TODO: MAJOR VERSION - remove :rainbows
             if [:puma, :passenger, :rainbows, :unicorn].include? Agent.config[:dispatcher]
               ::NewRelic::Agent.logger.info "Deferring startup of agent reporting thread because #{Agent.config[:dispatcher]} may fork."
               true
--- a/lib/new_relic/agent/instrumentation/active_merchant.rb
+++ b/lib/new_relic/agent/instrumentation/active_merchant.rb
@@ -34,4 +34,19 @@ DependencyDetection.defer do
       end
     end
   end
+
+  executes do
+    if ActiveMerchant::VERSION < Gem::Version.new('0.65.0')
+      deprecation_msg = 'The Ruby Agent is dropping support for ActiveMerchant versions below 0.65.0 ' \
+        'in version 9.0.0. Please upgrade your ActiveMerchant version to continue receiving full support. ' \
+
+      ::NewRelic::Agent.logger.log_once(
+        :warn,
+        :deprecated_active_merchant_version,
+        deprecation_msg
+      )
+
+      ::NewRelic::Agent.record_metric("Supportability/Deprecated/ActiveMerchant", 1)
+    end
+  end
 end
--- a/lib/new_relic/agent/instrumentation/active_merchant.rb
+++ b/lib/new_relic/agent/instrumentation/active_merchant.rb
@@ -37,7 +37,7 @@ DependencyDetection.defer do
 
   executes do
     if ActiveMerchant::VERSION < Gem::Version.new('1.65.0')
-      deprecation_msg = 'The Ruby Agent is dropping support for ActiveMerchant versions below 0.65.0 ' \
+      deprecation_msg = 'The Ruby Agent is dropping support for ActiveMerchant versions below 1.65.0 ' \
         'in version 9.0.0. Please upgrade your ActiveMerchant version to continue receiving full support. ' \
 
       ::NewRelic::Agent.logger.log_once(
--- a/lib/new_relic/agent/instrumentation/active_merchant.rb
+++ b/lib/new_relic/agent/instrumentation/active_merchant.rb
@@ -36,7 +36,7 @@ DependencyDetection.defer do
   end
 
   executes do
-    if ActiveMerchant::VERSION < Gem::Version.new('0.65.0')
+    if ActiveMerchant::VERSION < Gem::Version.new('1.65.0')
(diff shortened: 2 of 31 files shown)


kaylareopelle:
FYI - Changelog entry is on a different PR: https://github.com/newrelic/newrelic-ruby-agent/pull/1116/commits/c28f409b946a748e33c85ca594396620a9ff8446 (I wanted to avoid another CI run)

lib/new_relic/agent/instrumentation/active_merchant.rb
```diff
@@ -34,4 +34,19 @@ class ActiveMerchant::Billing::Gateway
       end
     end
   end
+
+  executes do
+    if ActiveMerchant::VERSION < Gem::Version.new('1.65.0')
```
fallwith, line 39: If this were a method we'd do something like this to avoid the giant whole-method `if`:

```ruby
def executes
  return unless ActiveMerchant::VERSION < Gem::Version.new('1.65.0')

  # deprecation warning
end
```

We can pull off the same thing with blocks by using `next`:

```ruby
executes do
  next unless ActiveMerchant::VERSION < Gem::Version.new('1.65.0')

  # deprecation warning
end
```

lib/new_relic/agent/instrumentation/active_merchant.rb
```diff
@@ -34,4 +34,19 @@ class ActiveMerchant::Billing::Gateway
...
+      deprecation_msg = 'The Ruby Agent is dropping support for ActiveMerchant versions below 1.65.0 ' \
```
fallwith, line 40: The code base has loads of "Ruby Agent" and "Ruby agent" instances all over the place, both in comments and in printed out content. No need to address it with this PR though, I'll create an issue.

lib/new_relic/agent/instrumentation/rainbows_instrumentation.rb
```diff
@@ -12,6 +12,17 @@
...
+    deprecation_msg = 'The dispatcher rainbows is deprecated. It will be removed ' \
+     'in version 9.0.0. Please use a supported dispatcher instead. ' \
+     'Visit https://docs.newrelic.com/docs/apm/agents/ruby-agent/getting-started/ruby-agent-requirements-supported-frameworks for options.'
+
+    ::NewRelic::Agent.logger.log_once(
+      :warn,
+      :deprecated_rainbows_dispatcher,
+      deprecation_msg
+    )
+    ::NewRelic::Agent.record_metric("Supportability/Deprecated/Rainbows", 1)
```
fallwith, line 25: > Like with climate change, you know? It used to rain here all of the time and now it's almost all gone. I don't even know if it's raining somewhere else, or if it's dry there too. Dry prolly. And that's when it hit me. These rainbows are the last ones, man. They won't be here for much longer. So that's when I decided to name the band 'Deprecated Rainbows'.

lib/new_relic/agent/instrumentation/sinatra.rb
```diff
@@ -32,18 +32,29 @@
   end
 
   executes do
-    if Sinatra::Base.respond_to?(:build)
```
fallwith, line 35: Would you please let us know about why we don't have to worry about this check anymore?

test/multiverse/suites/activemerchant/Envfile
```diff
@@ -1,6 +1,6 @@
 if RUBY_VERSION >= '3.0.0'
   gemfile <<-RB
-    gem 'activemerchant', '~>1.121.0'
+    gem 'activemerchant' # latest
```
fallwith, line 3: I suggest going without this comment unless there's something specific that warrants it. It's redundantly describing default behavior and it's not repeated for all of the others in the same boat within this same heredoc block.

lib/new_relic/agent/agent.rb
```diff
@@ -460,6 +460,7 @@ def correct_license_length
           # requests, we need to wait until the children are forked
           # before connecting, otherwise the parent process sends useless data
           def using_forking_dispatcher?
+            # TODO: MAJOR VERSION - remove :rainbows
```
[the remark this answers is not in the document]
kaylareopelle: I like that! I'll update with `# TODO: MAJOR VERSION -` for all related

lib/new_relic/agent/instrumentation/active_merchant.rb
kaylareopelle: Thank you! I'll make this adjustment to all the version conditions :)

lib/new_relic/agent/instrumentation/sinatra.rb
kaylareopelle: The `:build` method was added in Sinatra 1.2.1. The agent stopped supporting versions below 1.4.x in version 6.15.0. This condition will never be false for the versions we support.

lib/new_relic/agent/instrumentation/active_merchant.rb
fallwith: Created #1118

fallwith (approved):
👍

PR #1120, opened 2022-04-26
Add alt-text to README banner

kaylareopelle:
# Overview
The Community Plus header on our README does not have alt text describing the image. This PR adds alt-text describing the banner and duplicating the text that appears so that it is more accessible.


fallwith (approved):
I applaud this! Thank you.

PR #1123, opened 2022-04-27
Fix comparison for string vs Gem::Version

tannalynn:

# Overview
While I was looking into an unrelated issue, I discovered that there were errors in the agent log for these lines, since it doesn't like the string VS Gem::Version comparison. This pr converts the strings to version objects to avoid that issue.

### DIFF
--- a/lib/new_relic/agent/instrumentation/active_merchant.rb
+++ b/lib/new_relic/agent/instrumentation/active_merchant.rb
@@ -36,7 +36,7 @@ DependencyDetection.defer do
   end
 
   executes do
-    next unless ActiveMerchant::VERSION < Gem::Version.new('1.65.0')
+    next unless Gem::Version.new(ActiveMerchant::VERSION) < Gem::Version.new('1.65.0')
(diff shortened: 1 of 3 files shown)


lib/new_relic/agent/instrumentation/active_merchant.rb
```diff
@@ -36,7 +36,7 @@ class ActiveMerchant::Billing::Gateway
   end
 
   executes do
-    next unless ActiveMerchant::VERSION < Gem::Version.new('1.65.0')
+    next unless Gem::Version.new(ActiveMerchant::VERSION) < Gem::Version.new('1.65.0')
```
fallwith, line 39: 👍 NOTE that this will work if `ActiveMerchant::VERSION` is a String OR an instance of `Gem::Version` already.

fallwith (approved):
Good catch

lib/new_relic/agent/instrumentation/active_merchant.rb
kaylareopelle: This is so great. Thank you for finding this, Tanna.

PR #1125, opened 2022-04-28
perf tests need webrick for Ruby 3

fallwith:
perf tests: bundle webrick for Ruby 3+


kaylareopelle (approved):
Thanks for catching this!

PR #1127, opened 2022-04-28
docker-compose.yml: supply a default Ruby version

fallwith:
`docker-compose.yml` was inadvertently made to require a 'RUBY_VERSION'
environment variable because passing an empty value to `Dockerfile`
caused the `Dockerfile` build argument default value to be overridden.

Now when the `RUBY_VERSION` environment variable is not set, a default
version will be defined by `docker-compose.yml`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,7 +53,7 @@ services:
     build:
       context: .
       args:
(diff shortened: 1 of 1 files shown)


kaylareopelle (approved):
Thanks, James!

PR #1128, opened 2022-04-28
Bugfix: Error events missing priority when created outside of a transaction

### LINKED ISSUE #817: NoMethodError: undefined method `<' for nil:NilClass leads to "Failure when capturing error 'READONLY You can't write against a read only slave.':"
## Description

👋 I was investigating a report in our Honeybadger exception catcher, which we are using in conjunction with NewRelic.

In the stack of warnings and exception backtrace, I could see that NewRelic Ruby client had tried to capture the exception, but emitted an error:

    Failure when capturing error 'READONLY You can't write against a read only ...':

This "Failure when capturing" message comes from newrelic_rpm. That's `exception` in this line: https://github.com/newrelic/newrelic-ruby-agent/blob/24bf088f4fb2536cb0753b72960bbe5b5a420031/lib/new_relic/agent/error_collector.rb#L246

Now, just next to that was _WARN_ message, and this is the `e` in the code above

    NoMethodError: undefined method `<' for nil:NilClass

bild

Could it be that the code which is capturing runs into some trouble?

The `Redis::CommandError` exception being managed is derived from RuntimeError, nothing special about it:
 https://github.com/redis/redis-rb/blob/6542934f01b9c390ee450bd372209a04bc3a239b/lib/redis/errors.rb#L20

## Expected Behavior

I expect no specific messages to be emitted.

## Steps to Reproduce

- Have a Redis READONLY instance
- Stack both newrelic_rpm and honeybadger-ruby to do exception notifications
- Try connecting to Redis using the redis-4.5.1 gem

## Your Environment

- `ruby 2.7.4p191 (2021-07-07 revision a21a3b7d23) [x86_64-linux]`
- this gem: 8.0.0

tannalynn:

# Overview
This fixes a bug in how we're creating error events when not in a transaction.
Currently, when notice_error is called while not in a transaction, the priority is not set on the event because we are only taking it from the transaction payload passed in, which does not exist in this situation. This only begins to cause errors once the error event aggregator is full and we begin sampling.

I did check other event types to make sure this situation isn't possible for any other events.
- Span events are not created outside of a transaction
- Transaction events don't have this issue because they are the transaction
- Custom Events don't appear to have this issue since they are constructed differently and the priority is set directly in the custom even aggregator. 

# Related Github Issue
closes #817 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,11 @@
 
     The `transaction_events.max_samples_stored` capacity value within the TransactionEventAggregator did not match up with its expected harvest cycle interval, causing TransactionEvents to be over-sampled. This bugfix builds upon the updates made in [#952](https://github.com/newrelic/newrelic-ruby-agent/pull/952) so that the interval and capacity behave as expected for the renamed `transaction_events*` configuration options.
 
+  * **Bugfix: Error events missing attributes when created outside of a transaction**
+
+    Previously the agent was not assigning a priority to error events that were created by calling notice_error outside the scope of a transaction. This caused issues with sampling when the error event buffer was full, resulting in a `NoMethodError: undefined method '<' for nil:NilClass` in the newrelic_agent.log. This bugfix ensures that a priority is always assigned on error events so that the agent will be able to sample these error events correctly. Thank you to @olleolleolle for bringing this issue to our attention.
+    
+
 
   ## v8.6.0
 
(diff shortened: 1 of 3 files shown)


fallwith (approved):
👍

kaylareopelle (approved):
🕵️

kaylareopelle (approved):

PR #1129, opened 2022-04-29
Deprecate Ruby 2.2 and Rails 3.2

kaylareopelle:
* Deprecate Ruby 2.2 
* Deprecate Rails 3.2
* Add entry for constant changes in CHANGELOG

### DIFF
--- a/lib/new_relic/control/instrumentation.rb
+++ b/lib/new_relic/control/instrumentation.rb
@@ -84,7 +84,6 @@ module NewRelic
     end
 
     def ruby_22_deprecation
-      binding.irb
       if RUBY_VERSION <= '3.2.0'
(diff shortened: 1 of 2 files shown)


lib/new_relic/control/instrumentation.rb
```diff
@@ -62,9 +62,42 @@ def _install_instrumentation
...
+          :deprecated_rails_version,
+          deprecation_msg
+        )
+
+        ::NewRelic::Agent.record_metric("Supportability/Deprecated/Rails32", 1)
+      end
+    end
+
+    def ruby_22_deprecation
+      if RUBY_VERSION <= '3.2.0'
```
fallwith, line 87: - The `'3.2.0'` string doesn't match the `22` in the method name
- Maybe use an early return here instead of a giant `if`?

fallwith (approved):

PR #1137, opened 2022-05-05
supportability metrics: record uncompressed bytes

### LINKED ISSUE #1135: Supportability metrics: data usage reported should be scoped to uncompressed payload size only
Our cross agent spec for data usage in Supportability Metrics (see `Data-Usage-Supportability-Metrics.md` internally) calls for agents to always report the uncompressed payload size.

The Ruby agent currently calculates the size after compression, which does not itself take place for payloads below 64KiB. As a result, it is currently possible that we sometimes report a compressed size and sometimes report an uncompressed size within the same metric. We should always calculate the raw size prior to compression and report on that value.

fallwith:
For supportability metrics, record the uncompressed payload byte size
prior to sending the payload off for possible compression, so that we
are guaranteed to a) be consistent with always reporting the
uncompressed value to guarantee that two instances of the same metric
can reliably be compared, and b) be aligned with New Relic's cross agent
specification describing supportability metrics byte sizes.

resolves #1135

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # New Relic Ruby Agent Release Notes #
 
+  ## v8.8.0
+
+  * **Supportability Metrics will always report uncompressed payload size**
+
+    New Relic's cross agent specifications call for Supportability Metrics to always reference the uncompressed payload byte size. Previously, the Ruby agent was calculating the byte size after compression. Furthermore, compression is only performed on payloads of a certain size. This means that sometimes the value could have represented a compressed size and sometimes an uncompressed one. Now the uncompressed value is always used, bringing consistency for comparing two instances of the same metric and alignment with the New Relic cross agent specifications.
+
+
   ## v8.7.0
 
   * **APM logs-in-context log forwarding on by default**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
   * **Supportability Metrics will always report uncompressed payload size**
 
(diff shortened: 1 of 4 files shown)


kaylareopelle (approved):
👍

kaylareopelle:
One thing -- JIC someone is using the Supportability metrics, would you be open to adding a CHANGELOG entry for this?

fallwith:
> One thing -- JIC someone is using the Supportability metrics, would you be open to adding a CHANGELOG entry for this?

Good call. Addressed with 5028ba8

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # New Relic Ruby Agent Release Notes #
 
+  ## v8.8.0
+
+  * **Supportability Metrics will always report uncompressed payload size**
+
+    New Relic's cross agent specifications call for Supportability Metrics to always reference the uncompressed payload byte size. Previously, the Ruby agent was calculating the byte size after compression. Furthermore, compression is only performed on payloads of a certain size. This means that sometimes the value could have represented a compressed size and sometimes an uncompressed one. Now the uncompressed value is always used, bringing consistency for comparing two instances of the same metric and alignment with the New Relic cross agent specifications.
```
```suggestion
    New Relic's agent specifications call for Supportability Metrics to always reference the uncompressed payload byte size. Previously, the Ruby agent was calculating the byte size after compression. Furthermore, compression is only performed on payloads of a certain size. This means that sometimes the value could have represented a compressed size and sometimes an uncompressed one. Now the uncompressed value is always used, bringing consistency for comparing two instances of the same metric and alignment with the New Relic agent specifications.
```

Nit: the Repo is called "agent specs", but the tests are called "cross-agent tests". Not sure what's appropriate here. Feel free to disregard.

kaylareopelle (approved):

tannalynn (approved):
Thanks for getting this updated!

tannalynn (approved):

PR #1138, opened 2022-05-05
rework cross agent pull/push rake tasks

### LINKED ISSUE #924: cross_agent_tests:pull rake task is not functioning properly
## Description
The `cross_agent_tests:pull` rake task runs and appears to succeed, but does not actually update the local copy of the cross agent tests. Our current hypothesis is that the Ruby portion of the task is fine, but the external `rsync` command is not accomplishing the desired result. It is possible that the `-u` switch passed to `rsync` (instructing it to not update the local copy if the local copy's timestamps are newer) is a factor.

## Expected Behavior

New changes available in the remote source should arrive locally when the task is ran. Ideally, we would allow the remote copy to clobber any unmodified (from a git perspective) local file regardless of the timestamps involved.

n/a

## Steps to Reproduce
Make a change to the remote source (on disk, without committing via git), run the task, and confirm the change appears locally.

## Your Environment
macOS

## Additional context
n/a

fallwith:
The previous passing of `-u` to `rsync` (don't overwrite the destination
content if it has a newer timestamp than the source content) was
problematic because it assumed that a developer would have both the Ruby
agent and the cross agent specs cloned, then have new specs authored and
pulled down, and then perform the sync. If the cross agent specs haven't
been updated in a while, then the clone order of the repos becomes a
factor in the timestamp comparison.

Now, the pull and push rake targets will always overwrite the
destination content with the source content. To compensate for this, the
user is now prompted to continue or cancel with a warning that any
unsaved changes will be overwritten, and the previous source code
comment based warning is now presented to stdout as well.

resolves #924

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -116,8 +116,8 @@ task :update_ca_bundle do |t|
 end
 
 namespace :cross_agent_tests do
-  CROSS_AGENT_TESTS_UPSTREAM_PATH = File.expand_path(File.join('..', 'cross_agent_tests'), __FILE__).freeze
-  CROSS_AGENT_TESTS_LOCAL_PATH = File.expand_path(File.join('test', 'fixtures', 'cross_agent_tests'), __FILE__).freeze
+  CROSS_AGENT_TESTS_UPSTREAM_PATH = File.expand_path(File.join('..', 'cross_agent_tests')).freeze
+  CROSS_AGENT_TESTS_LOCAL_PATH = File.expand_path(File.join('test', 'fixtures', 'cross_agent_tests')).freeze
 
   def prompt_to_continue(command, destination = 'local')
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -116,24 +116,38 @@ task :update_ca_bundle do |t|
...
+    if destination.eql?('local')
+      puts 'CAUTION 2: Before continuing, make sure your local repo is on the correct, synced branch!'
+    end
+    puts
+    print "Do you wish to continue? ('y' to continue, return to cancel) [n] "
+    continue = STDIN.gets.chomp
+    if continue.downcase.eql?('y')
+      system(command)
+    else
+      puts 'Cancelled'
```
fallwith, line 137: From a Scottish upbringing. Let me know if the American version is preferred.

kaylareopelle (approved):
What a great find. Thanks, James!

PR #1139, opened 2022-05-05
default to Code Level Metrics support disabled

fallwith:
The test results and value add for CLM are in favor of enabling the
functionality by default, but we will follow the same approach taken for
logging and default to having the option off for an initial opt-in
rollout.


fallwith:
NOTE: the demo app we delivered explicitly enables CLM via newrelic.yml

PR #1145, opened 2022-05-09
Sync up with the latest cross agent test data (March 23, 2022)

### LINKED ISSUE #925: Latest cross-agent test suite produces failures
## Description
(related to #924) As of 8f9210b1662226b95897a20c8830445bf4e41e57, this repo features the June 2019 edition of the cross-agent test suite. The June 2020 edition produces 35 (mostly not unique) failures when tested against that version of this repo. We should get the local (this repo based) copy of the tests updated and get them to all pass.

## Expected Behavior
The June 2020 edition of the cross-agent tests (described in the release process docs) pass when ran from this repo.

n/a

## Steps to Reproduce
Update the local (this repo based) copy of the cross-agents with the latest one available, and run the tests.

## Your Environment
Include as many relevant details about your environment as possible including the running version of the Ruby agent, the Ruby version being used and any other relevant environment information
macOS

## Additional context
Add any other context about the problem here. For example, relevant community posts or support tickets.
n/a

fallwith:
As per #924, the local dev workflow for pulling in new cross agent tests wasn't working as desired. With the workflow now fixed up, here are the new tests and local agent test specific modifications to support those new tests.

resolves #925


test/fixtures/cross_agent_tests/README.md
```diff
@@ -17,7 +17,7 @@ Push access to this repository is granted via membership in the cross-agent-team
...
 | [rules.json](rules.json) | Describe how url/metric/txn-name rules should be applied. |
 | [rum_client_config.json](rum_client_config.json) | These tests dictate the format and contents of the browser monitoring client configuration.  For more information see: [SPEC](https://newrelic.atlassian.net/wiki/display/eng/BAM+Agent+Auto-Instrumentation) |
-| [sql_parsing.json](sql_parsing.json) | These tests show how an SQL string should be parsed for the operation and table name. |
+| [sql_parsing.json](sql_parsing.json) | These tests show how an SQL string should be parsed for the operation and table name. *Java Note*: The Java Agent is [out-of-sync with these tests](https://source.datanerd.us/java-agent/java_agent/blob/master/newrelic-agent/src/main/java/com/newrelic/agent/database/DefaultDatabaseStatementParser.java), [has its own tests](https://source.datanerd.us/java-agent/java_agent/blob/master/newrelic-agent/src/test/java/com/newrelic/agent/database/DatabaseStatementResponseParserTest.java), and cannot implement these without a breaking change. |
```
kaylareopelle, line 20: This is an interesting inclusion related to our discussion earlier about the attributes tests. I wonder if we should add something similar for the places where Ruby is out-of-sync with the tests?

Java may also have advice about how they avoid overwriting their unique changes.

kaylareopelle (dismissed):
Thanks for fixing this! I'm so relieved your computer caught the diff when we paired months ago :)

lib/new_relic/agent/datastores/metric_helper.rb
```diff
@@ -101,7 +101,9 @@ def self.metrics_from_sql(product, sql)
         end
 
         def self.operation_from_sql(sql)
-          NewRelic::Agent::Database.parse_operation_from_query(sql) || OTHER
+          operation = NewRelic::Agent::Database.parse_operation_from_query(sql)
+          operation = OTHER if operation.eql?('other')
```
tannalynn, line 105: Would it be possible to return `OTHER` from `parse_operation_from_query`, instead of the string `'other'`? Then maybe we could just call the method and not need to do any checking? could decrease object allocation a little as well

fallwith: Different tests seem to require `'OTHER'`, `'Other'`, or `'other'` which gave me trouble when I was trying to standardize on things.

I suppose it's possible to use constants for all three cases to prevent instantiating more than one copy of any of these 3 strings though, yeah. I'll take a look.

fallwith: Addressed with [2e60b55](https://github.com/newrelic/newrelic-ruby-agent/pull/1145/commits/2e60b55723f1c28d6ac1901b15c6c48fb5b8a889)

tannalynn (approved):
🎉

test/fixtures/cross_agent_tests/README.md
fallwith: I like this. I've created #1152 to track it.

PR #1149, opened 2022-05-10
drop mocha in favor of minitest stubs for trace context tests

fallwith:
For these couple of tests, use vanilla MiniTest stubs and remove our dependency on mocha.


kaylareopelle (approved):
Good progress toward removing our dependency on Mocha! I hope we have the capacity in the future to remove it entirely.

PR #1156, opened 2022-05-11
Fix TracedThread yard documentation

tannalynn:

# Overview
I noticed that our rubydocs was [missing](https://www.rubydoc.info/gems/newrelic_rpm/NewRelic/TracedThread) the new TracedThread class we added in the last version. I figured out how to generate it locally and I was able to get it updated it to make it show up and also fixed some syntax since markdown is not what it uses.

# Related Github Issue

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/.yardopts
+++ b/.yardopts
@@ -1,6 +1,7 @@
 --non-transitive-tag api
 --api public
 lib/new_relic/agent.rb
+lib/new_relic/traced_thread.rb
 lib/new_relic/agent/method_tracer.rb
 lib/new_relic/agent/distributed_tracing.rb
 lib/new_relic/agent/distributed_trace_payload.rb
--- a/lib/new_relic/traced_thread.rb
+++ b/lib/new_relic/traced_thread.rb
@@ -7,19 +7,18 @@ module NewRelic
   #
   # This class allows the current transaction to be passed to a Thread so that nested segments can be created from the operations performed within the Thread's block.
   # To have the New Relic Ruby agent automatically trace all of your applications threads,
-  # enable the `instrumentation.thread.tracing` configuration option in your newrelic.yml.
+  # enable the +instrumentation.thread.tracing+ configuration option in your newrelic.yml.
   #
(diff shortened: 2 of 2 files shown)


fallwith (approved):

kaylareopelle (approved):

.yardopts
```diff
@@ -1,6 +1,7 @@
 --non-transitive-tag api
 --api public
 lib/new_relic/agent.rb
+lib/new_relic/traced_thread.rb
```
hannahramadan, line 4: Let's talk about yard later? I want to understand it better.

lib/new_relic/traced_thread.rb
```diff
@@ -7,19 +7,18 @@ module NewRelic
...
   class TracedThread < Thread
     #
     # Creates a new Thread whose work will be traced by New Relic.
     # Use this class as a replacement for the native Thread class.
-    # Example: Instead of using `Thread.new`, use:
-    # ```ruby
-    #   NewRelic::TracedThread.new { execute_some_code }
-    # ```
+    # Example: Instead of using +Thread.new+, use:
+    #
```
hannahramadan, line 20: I thought this block needed the ``` ruby``` to designate the code as ruby. Is that no longer needed?

hannahramadan (approved):

lib/new_relic/traced_thread.rb
tannalynn: It actually wasn't needed originally, I had just done this incorrectly. I thought it was using markdown which does use that, but turns out it doesn't actually use markdown! Yard uses a totally different syntax for documentation, so that is what this change is using instead. So instead of doing a code block like ` ```ruby `, a code block is supposed to just be indented. and instead of doing a snippet with `, you use +, like
```ruby
    ```ruby
    # wow a code block
    this = code.wow
    ```
    # VS 
         # wow a code block
         this = code.wow


    `code snippet`
    # VS
    +code snippet+
```

PR #1161, opened 2022-05-13
8.8.0 changelog entry for CLM

fallwith:
mlt/clm changelog entry for 8.8.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
   * **Provide Code Level Metrics for New Relic CodeStream**
 
-    For Ruby on Rails applications and/or those with manually traced methods, the agent is now capable of reporting metrics with Ruby method level granularity. When the new `code_level_metrics.enabled` configuration parameter is set to a `true` value, the agent will associate source code related metadata with the metrics for things such as Rails controller methods. Then when the corresponding Ruby class file that defines the methods is loaded up in a New Relic [CodeStream](https://www.codestream.com/) powered IDE, data for each method will be presented to the developer directly.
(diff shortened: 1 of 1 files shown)


kaylareopelle (commented):
This is a little different from what the Python team did: https://docs.newrelic.com/docs/release-notes/agent-release-notes/python-release-notes/python-agent-71000175 

Did you get any feedback from CodeStream or Docs folks about the content?

fallwith:
> This is a little different from what the Python team did: https://docs.newrelic.com/docs/release-notes/agent-release-notes/python-release-notes/python-agent-71000175
> 
> Did you get any feedback from CodeStream or Docs folks about the content?

This is just for the repo CHANGELOG. From what I can tell Python doesn't have one of those so I didn't have anything to go on there. I was only planning on syncing with other agents for the docs site content. Did you have any specific concerns or suggested rewrites specific to this entry?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
   ## v8.8.0
 
+  * **Provide Code Level Metrics for New Relic CodeStream**
+
+    For Ruby on Rails applications and/or those with manually traced methods, the agent is now capable of reporting metrics with Ruby method level granularity. When the new `code_level_metrics.enabled` configuration parameter is set to a `true` value, the agent will associate source code related metadata with the metrics for things such as Rails controller methods. Then when the corresponding Ruby class file that defines the methods is loaded up in a New Relic [CodeStream](https://www.codestream.com/) powered IDE, data for each method will be presented to the developer directly.
```
kaylareopelle, line 7: Apologies for not putting my recommendations in the review off the bat! 

I have a few content recommendations: 

```suggestion
    For Ruby on Rails applications and/or those with manually traced methods, the agent is now capable of reporting metrics with Ruby method-level granularity. When the new `code_level_metrics.enabled` configuration parameter is set to a `true` value, the agent will associate source-code-related metadata with the metrics for things such as Rails controller methods. Then, when the corresponding Ruby class file that defines the methods is loaded up in a [New Relic CodeStream](https://www.codestream.com/)-powered IDE, [the four golden signals](https://sre.google/sre-book/monitoring-distributed-systems/) for each method will be presented to the developer directly.
```

For logs-in-context we used a link directly to the docs page that described how to use the feature. I know that isn't quite ready yet, but imagine it will be before the release. Would you be open to replacing/adding a link to that page?

kaylareopelle (approved):
👍

Issue #802, opened 2021-10-01
Padrino Multiverse tests do not accurately test instrumentation installation


tannalynn:
A bug was reported (#737) that the padrino instrumentation was not working correctly. After looking into it we found that the padrino instrumentation was not being installed in such a way that it would work. The instrumentation was being installed too late for us to prepend, as the ancestor chain had already been established on the objects using the modules we prepended our instrumentation to. 
While we were able to correct the issues with instrumentation installation, we were not able to figure out a way to update our padrino tests to load the same way as running a padrino app outside of multiverse does. The tests do still test the functionality of the instrumentation, however it is not able to accurately test **IF** the instrumentation actually gets installed. 
We should update our multiverse tests to load padrino in the same way an application using padrino does.

kaylareopelle:
A similar issue was raised during the resolution of an issue related to Rails load order and initializers. 

Issue: https://github.com/newrelic/newrelic-ruby-agent/issues/662 
PR: https://github.com/newrelic/newrelic-ruby-agent/pull/806

angelatan2:
The original failure no longer occurs.  Improvements have been made to our test and merged.

PR #1164, opened 2022-05-13
improve padrino testing

fallwith:
enhance our padrino testing to confirm `#invoke_route_with_tracing` is
executed

### DIFF
--- a/test/multiverse/suites/padrino/padrino_test.rb
+++ b/test/multiverse/suites/padrino/padrino_test.rb
@@ -40,7 +40,8 @@ class PadrinoRoutesTest < Minitest::Test
   end
 
   def test_tracing_is_involved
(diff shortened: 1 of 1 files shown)


fallwith:
I attempted to address #802 but was not able to reproduce the core problem. By this I mean that I was unable to find the existing tests to be lacking in their exposure of the problem. After reverting #803, the 2 existing padrino tests immediately fail, either when ran alone or in the context of a large suite of tests that includes sinatra tests which may have helped cloak the underlying issue.

For now I've added a very specific test to confirm the exact desired tracing method is called.

tannalynn (approved):
Thanks for looking into this ticket and adding that additional test! Weird that the original issue, where tests weren't failing if it there were issues with installation, doesn't seem to be a problem anymore. 
Looks good to me!

kaylareopelle (approved):

PR #1165, opened 2022-05-13
lower min compression threshold to 2048 bytes

### LINKED ISSUE #1136: Consider compressing all payloads en route to the collector
Currently `NewRelic::Agent::NewRelicService#compress_request_if_needed` will only perform compression on a payload if it is found to be greater than 64KiB (64 * 1024 bytes) in size. The source code comment for that method states that compressing smaller payloads constitutes aggressive compression behavior that will run afoul of related bugs present in some versions of Ruby.

We do not currently know:

- If the choice of 64KiB was arbitrary or based on some code or testing
- What the related Ruby bugs are
- Which Ruby versions are thought to have the bugs (NOTE: the comment was written in 2012)
- If there are bugs present for both our raw DEFLATE stream and GZIP stream compression approaches (configurable per agent instance)

We should research these things and conduct testing with an optimal goal of having the Ruby agent compress data in all cases without condition. If that goal can not be achieved, and any part of the 10 year old source comment still holds true for currently supported Ruby versions, we should update the comment and any relevant internal knowledge bases.

fallwith:
lower the compression byte threshold:
`> 64 KiB`  ->  `>= 2 KiB`

previously the agent had a minimum byte size threshold of 64KiB. for any
payload equal to or below that size, compression would not be applied.
this relatively high threshold was originally established to prevent
segfaults from taking place in older versions of Ruby trying trying to
wrap C zlib/gzip libraries to compress small payloads.

with the currently supported Rubies (2.2 - 3.1), segfaults aren't a
problem for even the tiniest of payloads, so we are free to come up with
a value that balances cpu usage and bandwidth savings.

resolves #1136

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
   ## v8.8.0
 
-  * **Lowered the minimum payload size to apply compression at**
+  * **Lowered the minimum payload size to compress**
 
     Previously the Ruby agent used a particularly large payload size threshold of 64KiB that would need to be met before the agent would compress data en route to New Relic's collector. The original value stems from segfault issues that very old Rubies (< 2.2) used to encounter when compressing smaller payloads. This value has been lowered to 2KiB (2048 bytes), which should provide a more optimal balance between the CPU cycles spent on compression and the bandwidth savings gained from it.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
   ## v8.8.0
 
+  * **Lowered the minimum payload size to apply compression at**
+
+    Previously the Ruby agent used a particularly large payload size threshold of 64KiB that would need to be met before the agent would compress data en route to New Relic's collector. The original value stems from segfault issues that very old Rubies (< 2.2) used to encounter when compressing smaller payloads. This value has been lowered to 2KiB (2048 bytes), which should provide a more optimal balance between the CPU cycles spent on compression and the bandwidth savings gained from it.
+
   * **Supportability Metrics will always report uncompressed payload size**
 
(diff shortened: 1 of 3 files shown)


kaylareopelle (commented):
Thanks for your thorough research and comment! Could you add a CHANGELOG entry for this?

lib/new_relic/agent/new_relic_service.rb
```diff
@@ -22,8 +22,15 @@ class NewRelicService
...
-      # greater than this number of bytes
-      MIN_BYTE_SIZE_TO_COMPRESS = 64 * 1024
+      # greater than or equal to this number of bytes. In testing with
+      # Ruby 2.2 - 3.1, we determined an absolute minimum value for ASCII to be
+      # 535 bytes to obtain at least a 10% savings in size. It is recommended
+      # that this value be kept above that 535 number. It is also important to
+      # consider the CPU cost involved with performing compression and to find
+      # a balance between CPU cycles spent and bandwidth saved. A good
+      # reasonable default here is 2048 bytes, which is a tried and true Apache
+      # Tomcat default (as of v8.5.78)
```
tannalynn, line 32: I really appreciate the detailed comment explaining the WHY of the new minimum value 👍

tannalynn (commented):
+1 about the changelog entry, other than that it looks great to me! TY!

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
   ## v8.8.0
 
+  * **Lowered the minimum payload size to apply compression at**
```
```suggestion
  * **Lowered the minimum payload size to compress**
```

kaylareopelle (approved):
Small recommendation, otherwise LGTM! I'm looking forward to seeing how this impacts our users.

Issue #1132, opened 2022-05-02
Adopt latest `setup-ruby` GitHub Actions action, and see if we can cut down on manual customizations


fallwith:
The latest [setup-ruby](https://github.com/ruby/setup-ruby) action seems to support all of the Ruby versions that we would need to test against. We should see if we could adopt it into our CI workflow and eliminate as much of our custom logic found in `.github/actions/build-ruby/index.js` (our fork) as possible.

angelatan2:
Closing this issue with the  merge of #1178

PR #1178, opened 2022-05-23
Use ruby/setup-ruby for CI

tannalynn:

# Overview
This PR removes our custom build ruby action and instead now uses the official ruby action https://github.com/ruby/setup-ruby. Our custom build ruby script was also doing additional steps including installing packages needed for some native gems, installing older mysql version for older rails tests, and making sure the right bundler versions are installed. I've moved these additional steps into the steps for the jobs. 
Installing mysql55 typically takes about 8 minutes, so I have also added a cache step so that we can avoid installing this every time. 

# Related Github Issue
https://github.com/newrelic/newrelic-ruby-agent/issues/1132

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


.github/workflows/ci.yml
```diff
@@ -59,212 +45,254 @@ jobs:
...
               },
               "jruby-9.3.4.0": {
-                "rails": "norails,rails61,rails52,rails42"
+                "rails": "norails,rails61,rails60,rails52,rails51,rails50,rails42"
               }
             }
 
+      - run: pwd
+
+      - run: if [[ ${{ env.rails }} =~ .*"4".* ]]; then sudo ./test/script/install_mysql55; echo "mysql55 installed"; fi
```
tannalynn, line 81: This seems to take ~7 minutes to do. In our custom build ruby, we are doing this but also cacheing it to avoid building it every time. That's something I'd like to add to this step as well

.github/workflows/scripts/setup_bundler
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+# This script sets up the bundler versions we need for our older unit tests
+# and sets up the bundler config for when we are using the older mysql version
```
fallwith, line 3: Only 2 of the semicolons in this script are needed - the ones that smoosh `if` and `then` into a one-liner. The others are all redundant with the newline on the end of each line already signifying an ending.

```shell
if [[ SOMETHING == "value" ]]; then
                             ^-- keep these ones
```

.github/workflows/scripts/setup_bundler
```diff
@@ -0,0 +1,23 @@
...
+  if [[ $RUBY_VERSION != "2.2.10" ]]; then 
+    # 2.2 does not install any 2+ bundler version, so nothing to uninstall
+    # but we do need to remove higher bundler versions for when we're running rails 4 tests
+    gem uninstall bundler; 
+    echo "did gem uninstall bundler"; 
+  fi ;  
+  echo "use bundler 1.17.3"; 
+fi
+
+bundle install
```
fallwith, line 23: It'd be good to address these "no newline on the end of the file" warnings.

fallwith (approved):
Astonishing! This huge effort really simplifies and modernizes things. Terrific work!

Screen Shot 2022-06-09 at 11 18 21 PM

.github/workflows/scripts/setup_bundler
tannalynn: Ahh that makes sense, I originally had most of these all squished on one line in the ci.yml as steps, and then forgot about it once I moved this into its own file. Got it updated!

fallwith (approved):

PR #1179, opened 2022-05-24
Rubocop Performance

hannahramadan:
# Overview
Added two new Rubocop Performance updates to rubocop.yml: [concurrent_monotonic_time](https://github.com/rubocop/rubocop-performance/blob/master/lib/rubocop/cop/performance/concurrent_monotonic_time.rb) and [string_identifier_argument](https://github.com/rubocop/rubocop-performance/blob/master/lib/rubocop/cop/performance/string_identifier_argument.rb). 

Edited files where Rubocop was upset about string_identifier_argument. Codebase was already looking good for concurrent_monotonic_time.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -962,7 +962,7 @@ Performance/CollectionLiteralInLoop:
   Enabled: false
 
 Performance/CompareWithBlock:
-  Enabled: false
+  Enabled: true
 
 Performance/ConcurrentMonotonicTime:
   Enabled: true
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -962,6 +962,9 @@ Performance/CollectionLiteralInLoop:
   Enabled: false
 
 Performance/CompareWithBlock:
+  Enabled: false
+
+Performance/ConcurrentMonotonicTime:
   Enabled: true
 
 # Disabling for now
@@ -1068,6 +1071,9 @@ Performance/Squeeze:
 # Disabling for now
 Performance/StartWith:
   Enabled: false
+
+Performance/StringIdentifierArgument:
+  Enabled: true
 
 Performance/StringInclude:
   Enabled: false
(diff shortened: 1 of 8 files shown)


test/new_relic/agent/rpm_agent_test.rb
```diff
@@ -46,7 +46,7 @@ def test_startup_shutdown_real
   def test_manual_start
     NewRelic::Agent.instance.expects(:connect).once
     NewRelic::Agent.instance.expects(:start_worker_thread).once
-    NewRelic::Agent.instance.instance_variable_set '@started', nil
+    NewRelic::Agent.instance.instance_variable_set :@started, nil
```
fallwith, line 49: The use of a symbol instead of a string here is a good change!

For the very specific situation of wanting to blank out a variable as we're doing here, I think `remove_instance_variable(:@variable_name)` would be even better than `instance_variable_set(:@variable_name, nil)`.

hannahramadan: Thank you! As chatted about, will explore making this nifty change across codebase in separate PR 🌈

.rubocop.yml
```diff
@@ -962,6 +962,9 @@ Performance/CollectionLiteralInLoop:
   Enabled: false
 
 Performance/CompareWithBlock:
+  Enabled: false
```
fallwith, line 965: In addition to the 2 new cops being added, this existing cop `Performance/CompareWithBlock` seems to have been disabled by these changes. Would you please either reenable it or update the PR description to explain why it's a good idea to disable it?

hannahramadan: Oh shoot. Thank you, was testing something. Not supposed to be disabled.

fallwith (approved):
👍

kaylareopelle (approved):
🦕

PR #1180, opened 2022-05-26
support db adapter suffixes, such as '_makara'

fallwith:
* route ActiveRecord database adapter names through a filter to perform
  vendor specific transformations. for now, only makara based adapters
  are supported, and the '_makara' suffix is stripped away. thus
  'postgresql_makara' will be converted to 'postgresql' so that SQL
  obfuscation will work with makara
* for Rails <= 5, seek the database info needed for New Relic agent
  segment creation from the first known connection pool's spec config

Many thanks to @lucasklaassen for the original implementation of these
concepts, which will take makara off of our unsupported gem list and
also potentially help with other similar database related gems in
future.

### DIFF
--- a/lib/new_relic/agent/instrumentation/active_record_helper.rb
+++ b/lib/new_relic/agent/instrumentation/active_record_helper.rb
@@ -93,9 +93,21 @@ module NewRelic
 
         ACTIVE_RECORD = "ActiveRecord".freeze
         OTHER = "other".freeze
+        MAKARA_SUFFIX = "_makara".freeze
+
+        # convert vendor (makara, etc.) wrapper names to their bare names
+        # ex: postgresql_makara -> postgresql
+        def bare_adapter_name(adapter_name)
+          # TODO: OLD RUBIES - RUBY_VERSION < 2.5
+          # With Ruby 2.5+ we could use #delete_suffix instead of #chomp for a
+          # potential speed boost
+          return adapter_name.chomp(MAKARA_SUFFIX) if adapter_name && adapter_name.end_with?(MAKARA_SUFFIX)
+
+          adapter_name
+        end
 
         def product_operation_collection_for name, sql, adapter_name
-          product = map_product(adapter_name)
+          product = map_product(bare_adapter_name(adapter_name))
           splits = split_name(name)
           model = model_from_splits(splits)
           operation = operation_from_splits(splits, sql)
@@ -194,8 +206,7 @@ module NewRelic
         ACTIVE_RECORD_DEFAULT_PRODUCT_NAME = "ActiveRecord".freeze
 
         def map_product(adapter_name)
-          PRODUCT_NAMES.fetch(adapter_name,
-            ACTIVE_RECORD_DEFAULT_PRODUCT_NAME)
+          PRODUCT_NAMES.fetch(adapter_name, ACTIVE_RECORD_DEFAULT_PRODUCT_NAME)
         end
 
         module InstanceIdentification
@@ -221,11 +232,16 @@ module NewRelic
           SLASH = "/".freeze
           LOCALHOST = "localhost".freeze
 
+          def adapter_from_config(config)
+            bare_name = NewRelic::Agent::Instrumentation::ActiveRecordHelper.bare_adapter_name(config[:adapter])
+            PRODUCT_SYMBOLS[bare_name]
+          end
+
           def host(config)
             return UNKNOWN unless config
 
             configured_value = config[:host]
-            adapter = PRODUCT_SYMBOLS[config[:adapter]]
+            adapter = adapter_from_config(config)
             if configured_value.nil? ||
                 postgres_unix_domain_socket_case?(configured_value, adapter)
 
@@ -243,7 +259,7 @@ module NewRelic
           def port_path_or_id(config)
             return UNKNOWN unless config
 
-            adapter = PRODUCT_SYMBOLS[config[:adapter]]
+            adapter = adapter_from_config(config)
             if config[:socket]
               config[:socket].empty? ? UNKNOWN : config[:socket]
             elsif postgres_unix_domain_socket_case?(config[:host], adapter) || mysql_default_case?(config, adapter)
@@ -263,7 +279,7 @@ module NewRelic
           SUPPORTED_ADAPTERS = [:mysql, :postgres].freeze
 
           def supported_adapter? config
-            config && SUPPORTED_ADAPTERS.include?(PRODUCT_SYMBOLS[config[:adapter]])
+            config && SUPPORTED_ADAPTERS.include?(adapter_from_config(config))
           end
 
           private
--- a/lib/new_relic/agent/instrumentation/active_record_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/active_record_subscriber.rb
@@ -91,17 +91,24 @@ module NewRelic
           end
 
           return unless connection_id = payload[:connection_id]
-          connection = nil
 
           ::ActiveRecord::Base.connection_handler.connection_pool_list.each do |handler|
-            connection = handler.connections.detect do |conn|
-              conn.object_id == connection_id
-            end
+            connection = handler.connections.detect { |conn| conn.object_id == connection_id }
 
-            break if connection
+            return connection.instance_variable_get(:@config) if connection
+
+            # when using makara, handler.connections will be empty, so use the
+            # spec config instead.
+            # https://github.com/newrelic/newrelic-ruby-agent/issues/507
+            # thank you @lucasklaassen
+            return handler.spec.config if use_spec_config?(handler)
           end
 
-          connection.instance_variable_get(:@config) if connection
+          nil
+        end
+
+        def use_spec_config?(handler)
+          handler.spec && handler.spec.config && handler.spec.config[:adapter].end_with?('makara')
         end
 
         def start_segment(config, payload)
--- a/test/new_relic/agent/instrumentation/active_record_helper_test.rb
+++ b/test/new_relic/agent/instrumentation/active_record_helper_test.rb
@@ -62,5 +62,23 @@ module NewRelic::Agent::Instrumentation
       assert_equal "other", operation
       assert_nil collection
     end
+
+    def test_suffixes_are_stripped_away_from_the_adapter_name
+      assert_equal 'postgresql', ActiveRecordHelper.bare_adapter_name('postgresql_makara')
+    end
+
+    def test_product_operation_collection_for_handles_suffixes
+      mock = MiniTest::Mock.new
+      mock.expect('postgresql', [])
+      NewRelic::Agent::Instrumentation::ActiveRecordHelper.stub :map_product, mock do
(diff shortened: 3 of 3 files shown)


lucasklaassen:
Fantastic thank you!

Rakefile
```diff
@@ -11,15 +11,6 @@ task :test => ['test:newrelic']
...
   desc "Run all tests"
   task :all => %w[newrelic multiverse all_compatible_envs]
-
-  begin
-    require 'test_bisect'
-    TestBisect::BisectTask.new do |t|
-      t.test_task_name = 'test:newrelic'
-    end
-  rescue LoadError
-  end
```
fallwith, line 21: We don't currently use `test_bisect` (untouched since 2013), so let's stop rescuing its load error on every single Rake task run.

lib/new_relic/agent/instrumentation/active_record_helper.rb
```diff
@@ -93,9 +93,21 @@ def pluck(*args, &blk)
 
         ACTIVE_RECORD = "ActiveRecord".freeze
         OTHER = "other".freeze
+        MAKARA_SUFFIX = "_makara".freeze
+
+        # convert vendor (makara, etc.) wrapper names to their bare names
+        # ex: postgresql_makara -> postgresql
+        def bare_adapter_name(adapter_name)
```
fallwith, line 100: Route all adapter name look ups through this choke point. For now it only has support for Makara, to strip away the `_makara` suffix on an adapter name, but in future we could leverage this method to support other gems / adapters as well.

lib/new_relic/agent/instrumentation/active_record_helper.rb
```diff
@@ -153,20 +165,16 @@ def map_operation(raw_operation)
         PRODUCT_NAMES = {
           "mysql" => "MySQL",
           "mysql2" => "MySQL",
-          "mysql2_makara" => "MySQL",
```
fallwith, line 156: Use the new chokepoint method, `bare_adapter_name` instead of having every possible suffix based adapter name defined in this hash.

lib/new_relic/agent/instrumentation/active_record_helper.rb
```diff
@@ -198,8 +206,7 @@ def map_operation(raw_operation)
         ACTIVE_RECORD_DEFAULT_PRODUCT_NAME = "ActiveRecord".freeze
 
         def map_product(adapter_name)
-          PRODUCT_NAMES.fetch(adapter_name,
-            ACTIVE_RECORD_DEFAULT_PRODUCT_NAME)
+          PRODUCT_NAMES.fetch(adapter_name, ACTIVE_RECORD_DEFAULT_PRODUCT_NAME)
```
fallwith, line 209: No change, just eliminated a newline.

lib/new_relic/agent/instrumentation/active_record_helper.rb
```diff
@@ -208,14 +215,10 @@ module InstanceIdentification
           PRODUCT_SYMBOLS = {
             "mysql" => :mysql,
             "mysql2" => :mysql,
-            "mysql2_makara" => :mysql,
```
fallwith, line 211: Same deal here, the `bare_adapter_name` chokepoint method will be used instead of having every suffix based name defined.

lib/new_relic/agent/instrumentation/active_record_helper.rb
```diff
@@ -229,11 +232,16 @@ module InstanceIdentification
           SLASH = "/".freeze
           LOCALHOST = "localhost".freeze
 
+          def adapter_from_config(config)
+            bare_name = NewRelic::Agent::Instrumentation::ActiveRecordHelper.bare_adapter_name(config[:adapter])
```
fallwith, line 236: We're currently in the `InstanceIdentification` namespace here, so explicitly back up a level to get at `bare_adapter_name`.

lib/new_relic/agent/instrumentation/active_record_helper.rb
```diff
@@ -229,11 +232,16 @@ module InstanceIdentification
...
+            PRODUCT_SYMBOLS[bare_name]
+          end
+
           def host(config)
             return UNKNOWN unless config
 
             configured_value = config[:host]
-            adapter = PRODUCT_SYMBOLS[config[:adapter]]
+            adapter = adapter_from_config(config)
```
fallwith, line 244: Move the repeated use of this pattern out into a method, which the bare adapter name logic applied.

lib/new_relic/agent/instrumentation/active_record_subscriber.rb
```diff
@@ -91,21 +91,26 @@ def active_record_config(payload)
...
           ::ActiveRecord::Base.connection_handler.connection_pool_list.each do |handler|
-            # RE: https://github.com/newrelic/newrelic-ruby-agent/issues/507
-            spec_config = handler.respond_to?(:spec) && handler.spec && handler.spec.config
-            return spec_config if using_makara?(spec_config)
-
-            connection = handler.connections.detect do |conn|
-              conn.object_id == connection_id
-            end
-
-            break if connection
```
fallwith, line 105: NOTE that this loop through the connection pool list is only reached for Rails 5 and below.

These changes do the following:

- Instead of returning directly from the loop OR breaking and then returning, simply return from the loop in both early exit cases. This saves us from needing to have `connection` defined outside of the loop as well.
- Check for `#connections` - as this code always has - before `#spec`.
- Perform a `use_spec_config?` method to encompass ALL cases that leverage the handler spec config. For now, only Makara in supported in this way, but there could be others added in future.

lib/new_relic/agent/instrumentation/active_record_subscriber.rb
```diff
@@ -91,21 +91,26 @@ def active_record_config(payload)
...
+            # https://github.com/newrelic/newrelic-ruby-agent/issues/507
+            # thank you @lucasklaassen
+            return handler.spec.config if use_spec_config?(handler)
           end
 
-          connection.instance_variable_get(:@config) if connection
+          nil
+        end
+
+        def use_spec_config?(handler)
```
fallwith, line 109: This is pretty much `using_makara?` renamed with an additional `respond_to?` check. It's possible that the `#respond_to?` is only needed for the tests and not in real world scenarios, but I'm choosing to err on the safe side by including the check for now.

test/environments/lib/environments/runner.rb
```diff
@@ -51,15 +52,6 @@ def run_and_report
     end
 
     def tests_to_run
-      dirs = potential_directories
-
-      version = RUBY_VERSION
-      version = "jruby-#{JRUBY_VERSION[0..2]}" if defined?(JRUBY_VERSION)
```
fallwith, line 57: For some reason we define a local variable, `version`, don't use it for anything, and then call another method. I've simply collapsed these two methods into 1.

test/new_relic/agent/instrumentation/active_record_helper_test.rb
```diff
@@ -62,5 +62,20 @@ def test_product_operation_collection_for_name_with_integer_returns_nil
       assert_equal "other", operation
       assert_nil collection
     end
+
```
fallwith, line 65: New tests to confirm that `<adapter name>_makara` adapters work as desired.

test/new_relic/agent/instrumentation/rails/active_record_subscriber.rb
```diff
@@ -183,7 +183,7 @@ def test_active_record_config_for_event_with_connection_id
     connection_handler, connection_pool_handler = mock(), mock()
     connection_pool_handler.expects(:connections).returns([@connection])
     connection_handler.expects(:connection_pool_list).returns([connection_pool_handler])
-    ActiveRecord::Base.stubs(:connection_handler).returns(connection_handler)
+    ::ActiveRecord::Base.stubs(:connection_handler).returns(connection_handler)
```
fallwith, line 186: Just being explicit here for consistency's sake.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
   ## v8.8.0
 
+  * **Support Makara database adapters with ActiveRecord**
+
+    Thanks to a community submission from @lucasklaassen with [PR #1177](https://github.com/newrelic/newrelic-ruby-agent/pull/1177), the Ruby agent will now correctly work well with the [Makara gem](https://github.com/instacart/makara). Functionality such as SQL obfuscation should now work when Makara database adapaters are used with ActiveRecord.
```
```suggestion
    Thanks to a community submission from @lucasklaassen with [PR #1177](https://github.com/newrelic/newrelic-ruby-agent/pull/1177), the Ruby agent will now correctly work well with the [Makara gem](https://github.com/instacart/makara). Functionality such as SQL obfuscation should now work when Makara database adapaters are used with Active Record.
```

(Rails's official style uses two words)

lib/new_relic/agent/instrumentation/active_record_subscriber.rb
kaylareopelle: I'm a little worried this method's relation to Makara is buried by naming it `use_spec_config?` What are your thoughts about renaming it to `use_makara_spec_config?`? As it's written right now, the `makara` string in the final clause leads me to believe it can only be used in that context.

Rakefile
kaylareopelle: If we need functionality like this in the future, I suggest we use the Minitest Bisect gem - https://rubygems.org/gems/minitest-bisect

lib/new_relic/agent/instrumentation/active_record_subscriber.rb
fallwith: I very much wanted the method to be generic and not specific to Makara. If we fail to obtain a match from `handler.connections` (which will be an empty array in the case of Makara), but we happen to be able to call `.spec.config` on the `handler` object instead, then that should be all we need to create a segment from. I could foresee other gems also working with this `.spec.config` approach, and so I went with the generic `use_spec_config?` name.

I am envisioning in the future us being able to either drop the Makara specific check from `use_spec_config?` or having it "or"ed with another gem specific check. I decided to leave the Makara specific check in place just because it's currently the only gem we support going to the spec config for.

Both this `use_spec_config?` method and the `bare_adapter_name` method currently contain Makara specific logic, with the intention being that that logic can be added to over time. This gives us easily accessible chokepoints for all gems.

test/new_relic/agent/instrumentation/active_record_helper_test.rb
fallwith: Hmm... indeed. I think I saw the existence of that dedicated subscriber class in the test directory and just decided to forget I'd ever seen it rather than try to figure out how it worked. Will try to get some updates going.

fallwith: Addressed with [5d6d93b](https://github.com/newrelic/newrelic-ruby-agent/pull/1180/commits/5d6d93b5e64b4d667b57d2734566f95b9c311c5f)

kaylareopelle (approved):
Thanks for adding the test!

PR #1185, opened 2022-05-31
Fix performance test file

hannahramadan:
Fixed require statement inside performance test.

### DIFF
--- a/test/performance/script/runner
+++ b/test/performance/script/runner
@@ -7,8 +7,7 @@
 require 'optparse'
 require 'rubygems'
 require 'json'
-
-require File.expand_path('../../lib/performance', __FILE__)
+require_relative '../lib/performance'
 $: << File.expand_path('../../../../lib', __FILE__)
 
 options = {}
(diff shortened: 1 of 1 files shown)


kaylareopelle:
Thanks for finding this! What situation were you in when this issue came up?

tannalynn (approved):
Thanks for getting the PR created for this fix!

hannahramadan:
> Thanks for finding this! What situation were you in when this issue came up?

@kaylareopelle Found while running performance tests. This was actually an error I made in an earlier PR.

fallwith (approved):
Looks like this was the only straggler. 👍

```ruby
# Usage: 1. save as rr_checker.rb in the root of the project to be checked
#        2. run 'ruby rr_checker.rb'
Dir.glob('*/**/*.rb').each do |path|
  File.open(path).each do |line|
    next unless line =~ %r{require_relative ['"]([a-zA-Z0-9./]+)['"]}

    absolute = "#{path.split('/')[0..-2].join('/')}/#{Regexp.last_match(1)}.rb"
    next if File.exist?(absolute)

    puts 'Problem'
    puts '======='
    puts "file = #{path}"
    puts "invalid relative require: #{line}"
    puts "expected location: #{absolute}"
    puts
  end
end
```

PR #1197, opened 2022-06-03
rework PipeChannelManager test for nonblock

fallwith:
rework the PipeChannelManager test for verifying that an
`IO::WaitReadable` based exception will be rescued properly


tannalynn (approved):
Thanks for going over how this works with us!

Issue #974, opened 2022-02-18
Replace calls to empty objects with their constants


kaylareopelle:
### Is your feature request related to a problem? Please describe.
Performance is very important to the New Relic Ruby Agent. Even small adjustments can make a huge difference in our customers' environments. `lib/new_relic/constants.rb` includes constant definitions for EMPTY_HASH, EMPTY_ARRAY, and EMPTY_STRING. There are some calls to the values these constants hold that could be replaced with the constants themselves.

### Feature Description
Replace all references to {}, [], and "" with their EMPTY_* constant where appropriate in the agent.

### Additional context
Areas where an array or hash could be mutated are not appropriate to make this change. 

### Priority
[Nice to Have]

PR #1199, opened 2022-06-06
Replace calls to empty arrays with constant

hannahramadan:

# Overview
Replace [] with constant EMPTY_ARRAY where appropriate (not being mutated.)

# Related Github Issue
Related to #974

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/agent_logger.rb
+++ b/lib/new_relic/agent/agent_logger.rb
@@ -83,11 +83,9 @@ module NewRelic
       # Allows for passing exceptions in explicitly, which format with backtrace
       def format_and_send(level, *msgs, &block)
         if block
-          if @log.send("#{level}?")
-            msgs = Array(block.call)
-          else
-            msgs = NewRelic::EMPTY_ARRAY
-          end
+          return unless @log.send("#{level}?")
+
+          msgs = Array(block.call)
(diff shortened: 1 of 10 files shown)


lib/new_relic/agent/agent_logger.rb
```diff
@@ -83,11 +83,9 @@ def backtrace_from_exception(e)
...
       def format_and_send(level, *msgs, &block)
         if block
-          if @log.send("#{level}?")
-            msgs = Array(block.call)
-          else
-            msgs = []
-          end
+          return unless @log.send("#{level}?")
+
+          msgs = Array(block.call)
```
fallwith, line 88: We mobbed as a team on this one. In the case of a block being present and the logger's level being below the threshold, we ultimately just return `nil`, so we may as well do it early here. Doing so saves us from spawning an empty array object after `#flatten` is called.

fallwith (approved):
👍

lib/new_relic/agent/agent_logger.rb
hannahramadan: Thank you for going over this live with us!

PR #1201, opened 2022-06-08
Replace empty {} calls

### LINKED ISSUE #974: Replace calls to empty objects with their constants

hannahramadan:

# Overview
Replace calls to immutable `{}` with the constant `EMPTY_HASH`, thus reducing object allocations.

Also in this PR is a refactor of a the now removed method `get_processor_info`, which is now broken up into 5 methods: `processor_info`, `processor_info_darwin`, `processor_info_linux`, `processor_info_bsd`, `remove_bad_values`. Tests were also added for each of these methods. 

There is also a slight refactor of the method `sysctl_value`, which now turns its return value into an integer inside the method vs calling `.to_i` whenever the method is called. 

Closes #974

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


lib/new_relic/agent/system_info.rb
```diff
@@ -81,7 +81,7 @@ def self.get_processor_info
 
         @processor_info
       rescue
-        {}
+        NewRelic::EMPTY_HASH
```
fallwith, line 84: It looks like this one replaces a mutable hash with a frozen one, but from looking at the consumption of the return value, that should be totally fine. 👍 

The `get_processor_info` method itself is looking pretty long and gnarly though. If you'd ever like to take a crack at modernizing its approach to memoization and also breaking it up into several smaller methods (one per OS seems appropriate here), we can pair on doing so.

hannahramadan: Yes from what I could tell, we are only reporting on that hash in the case of an error or unknown processor. This was a crazy long method that I needed Tanna to go through with me, so I'd love to pair on cleaning this up!

kaylareopelle (approved):
Awesome refactor! Thanks for taking this on, Hannah!

Issue #1183, opened 2022-05-31
Update Envfiles to test the latest release of all gems automatically


kaylareopelle:
Our multiverse suites have a filetype called "Envfile" that holds heredocs that describe the Gemfiles that a given suite should be tested against. Gemfile syntax allows gem declarations without versions (ex: `gem 'newrelic_rpm'`) to install the most recent version when a Gemfile.lock is not present. 

Manually update all Envfiles for core technologies (see: [Gem Research - Internal](https://docs.google.com/spreadsheets/d/1AIbDDKOasinWC4H2OXrkktuzNOIsVa4pQXw8vN_UhNY/edit?usp=sharing) to test the latest release by adding a Gemfile without a version constraint if such a heredoc block is not already present.

kaylareopelle:
This work is in progress on PR #1204. The core technologies spreadsheet includes a column with checks to see if we're testing the latest [(Internal)](https://docs.google.com/spreadsheets/d/1AIbDDKOasinWC4H2OXrkktuzNOIsVa4pQXw8vN_UhNY/edit#gid=0&range=K:K). 

With the exception of Unicorn (still untested), cRuby, jRuby, and the Ruby-internal instrumentation, PR #1204 should add latest version testing for the core technologies that needed updates.

PR #1204, opened 2022-06-10
Test unpinned gem version of core technologies

kaylareopelle:
# Overview
Adds an unspecified gem install Gemfile entry to the multiverse suite Envfiles of Core Technologies.

Closes: #1183

# Testing
This ticket is about expanding the versions tested by our multiverse suites. No tests need to be added.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,7 @@ jobs:
 
       - name: Setup bundler
         run: ./.github/workflows/scripts/setup_bundler
-        env: 
+        env:
           RUBY_VERSION: ${{ matrix.ruby-version }}
           RAILS_VERSION: ${{ env.rails }}
 
@@ -155,7 +155,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby-version }}
 
-      - if: matrix.ruby-version == '2.2.10' || matrix.ruby-version == '2.3.8' 
+      - if: matrix.ruby-version == '2.2.10' || matrix.ruby-version == '2.3.8'
         name: Cache mysql55
         id: mysql55-cache
         uses: john-shaffer/cache@sudo-tar
@@ -169,7 +169,7 @@ jobs:
 
       - name: Setup bundler
         run: ./.github/workflows/scripts/setup_bundler_multiverse
-        env: 
+        env:
           RUBY_VERSION: ${{ matrix.ruby-version }}
 
       - name: Test MySQL
--- a/test/multiverse/suites/delayed_job/before_suite.rb
+++ b/test/multiverse/suites/delayed_job/before_suite.rb
@@ -18,7 +18,8 @@ rescue LoadError
   # Let it fail, might be working with another library
 end
 
-# Deprecated on some versions, required on others. Hurray!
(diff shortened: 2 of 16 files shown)


test/multiverse/suites/delayed_job/before_suite.rb
```diff
@@ -43,7 +44,7 @@ class CreateDelayedJobs
     @connection = $db_connection
   end
 
-  class CreatePelicans < ActiveRecord::Migration
+  class CreatePelicans < ActiveRecord::VERSION::STRING >= "5.0.0" ? ActiveRecord::Migration["#{ActiveRecord::VERSION::STRING[0]}.0"] : ActiveRecord::Migration
```
kaylareopelle, line 47: Leaving like this for now, but there are lots of other references to the added ternary in other migrations. After this PR is merged, I'll create another that abstracts this out into a helper method and updates all references.

test/multiverse/suites/sidekiq/sidekiq_instrumentation_test.rb
```diff
@@ -77,7 +77,11 @@ def test_delayed_with_malformed_yaml
       YAML.stubs(:load).raises(RuntimeError.new("Ouch"))
       run_delayed
       assert_metric_and_call_count(ROLLUP_METRIC, JOB_COUNT)
-      assert_metric_and_call_count(DELAYED_FAILED_TXN_NAME, JOB_COUNT)
+      if RUBY_VERSION >= '3.0.0'
+        assert_metric_and_call_count(DELAYED_TRANSACTION_NAME, JOB_COUNT)
+      else
+        assert_metric_and_call_count(DELAYED_FAILED_TXN_NAME, JOB_COUNT)
+      end
```
kaylareopelle, line 84: Psych's `#unsafe_load` method for YAML causes Ruby 3+ to return the correct transaction name rather than the failed transaction name. This PR adds 4.2.0, a now-deprecated version of Sidekiq, to the Envfile. This test is only run on versions under 5.0.0, so I believe this is why we missed this test issue when we initially added `#unsafe_load` support to our Sidekiq instrumentation.

tannalynn (approved):
Looks good, thanks for getting this updated!

.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: PR Continuous Integration
```
kaylareopelle, line 1: Temporary change -- Making sure this passes on all Ruby versions before merge.

.github/workflows/ci.yml
```diff
@@ -26,9 +26,10 @@ jobs:
         ports:
           - "3306:3306"
     strategy:
-      fail-fast: false
+      fail-fast: true
```
kaylareopelle, line 29:
```suggestion
      fail-fast: false
```

.github/workflows/ci.yml
```diff
@@ -199,9 +221,9 @@ jobs:
     needs: run-rubocop
     runs-on: ubuntu-20.04
     strategy:
-      fail-fast: false
+      fail-fast: true
```
kaylareopelle, line 224:
```suggestion
      fail-fast: false
```

kaylareopelle (commented):
Fail fast revert

PR #1205, opened 2022-06-10
Remove verbose = true from Rake::TestTask

kaylareopelle:
I've had a hard time reading test output since I started working on this repository. One culprit seems to be the inclusion of all files being tested in each run's output. Removing `verbose = true` from `Rake::TestTask(:newrelic)` seems to remove the file list output.


kaylareopelle:
Update: Taking a look at the output, it seems like the list of files is no longer present on the CI! 🎉

tannalynn (approved):
🎉 This is awesome! So glad you figured out how to get rid of all that cluttered output!

PR #1206, opened 2022-06-10
remove license test

fallwith:
remove our antiquated automated license audit test that predates our
open sourcing of the gem


kaylareopelle (commented):
I don't see anything within the newly released Security and Privacy Standards for Agents that would conflict with this change. However, should we still check in with the open source team or SLC before removing this file?

test/new_relic/license_test.rb
```diff
@@ -1,130 +0,0 @@
...
-      # skip multiverse error log output
-      path =~ %r{/test/multiverse/lib/multiverse/errors.txt$} ||
-      # skip the artifacts directory
-      path =~ %r{/artifacts/} ||
-      # skip the changelog
-      path =~ %r{CHANGELOG.md}
-    )
-  end
-
-  def test_all_rb_and_js_files_have_license_header
```
tannalynn, line 88: I remember talking about the other test (test_for_scary_license_terms), but I don't remember this one being mentioned. Since this is checking to make sure the license header is on all files, is that something we should be checking or is similarly unnecessary as the other test?

fallwith:
Closed in favor of #1214 

- @tannalynn #1214 retains the check for the license header and only gets rid of the "scary" terms test
- @kaylareopelle If we want to revisit the removal of the need for a license header at the top of each file (something I'd like to revisit with all agent teams), we can bring in the open source team at that time. For now, I think #1214 is ready to go.

Thanks to you both for your help!

kaylareopelle:
@fallwith - I think discussing the need for a license header would be timely for an upcoming Agents council, given this week's meeting's focus on the [WIP Quality Standards Document (Internal)](https://docs.google.com/document/d/1AqYQuC20oXt30d2fI8Yt1l8FEXNc7WHo8cOgU9Aoa-c/edit?usp=sharing). 

That document also raises a concern about removing the "scary" license terms. I've tagged you in a comment on that doc with the reference. 

I also have a vague memory of you mentioning there could be another way to automate "scary" license terms checking. Do you remember that too, and/or do you have other ideas?

fallwith:
@kaylareopelle thanks, I've responded on that internal doc in agreement with the recommended approach.

There are 3 license related concerns in play:

- New Relic's desire to put a license blurb at the top of every single distributed file, in addition to having a repo wide LICENSE file
- [Distribution] Preventing content with an incompatible license from hitting the repository
- [Distribution] Preventing content with an incompatible license from hitting the distributed payload at build/package time

For item 1, I think it's a bit silly to do so but we are arguably in good company by doing so because Apache does this. Apache's own internal guidelines for distributing content calls for a commented out blurb to be included with each file. A wild guess is that New Relic borrowed this guideline from Apache when it came time to open source the agents. We'll keep doing this for now and keep automating checks that enforce doing so consistently. #1214 delivers a rewrite of the check. One example of an improvement there is that it no longer slurps in the content of every single Ruby file into memory, only the comments at the top of each.

For item 2, this specific Ruby agent repository doesn't currently contain any relevant third party content. It did in the past when we included an SSL cert. Should we ever start including such content in the repo again in the future, we can revisit the need for an automated check. But we definitely don't need one now and my expectation is that our current PR review process is sufficient for keeping any such content out unless it is properly disclosed.

For item 3, the Ruby agent does not bring in any third party content during the packaging process of creating our 2 Ruby gems. So for us, we're completely non-applicable here. Our dependencies declared by the 2 gemspec files (a combined list that currently only contains "grpc") is the closest we come to distributing content, but given that the consumers of our gems will themselves pull down that content, we're safe here.

My long ago suggestion of using an automated tool was essentially a call to use a pre-existing, open source tool to enforce items 2 and 3. I think doing so would be helpful for New Relic on the whole but for this repo specifically we're okay until we start distributing anything that isn't of our own authorship.

kaylareopelle:
Thanks for explaining this, @fallwith! Now I understand. The idea of a pre-existing, open source tool escaped my imagination. I agree that adding a tool like this could be beneficial for New Relic on the whole.

PR #1216, opened 2022-06-23
update Puma multiverse tests

### LINKED ISSUE #1189: Enhance Puma instrumentation tests
Review the existing instrumentation tests for Puma (they're nested within the Rack suite) and add tests as needed to make the suite comprehensive. This analysis should also include tests related to Puma's role as a web dispatcher. Dispatcher tests live in the unit test directories.

fallwith:
update the testing of our Puma instrumentation:

- point to the latest package of each of the 3 most recently released
  major Puma versions
- stop having newer Rubies test multiple versions of Rack. Now simply
  test with the latest Rack for those Rubies that can (Ruby 2.3+) and
  the latest Ruby 2.2 compatible Rack for Ruby 2.2
- add a test for Puma URLMap handling which was previously instrumented
  but not covered by multiverse testing

Pumas are less than half as fast as cheetahs and over 1.5 times as
heavy.

But Pumas can live 19 years as opposed to 11 for cheetahs. If the Ruby
Puma project had been named Cheetah when it began in 2011, it might have
been on its last legs now in 2022.

resolves #1189

### DIFF
--- a/test/multiverse/suites/rack/Envfile
+++ b/test/multiverse/suites/rack/Envfile
@@ -1,25 +1,16 @@
 instrumentation_methods :chain, :prepend
 
-puma_versions = [nil, '~>3.12.0', '~>4.3.8', '~> 5.2.2', '~>5.3.2']
-ruby_rack_versions = {'2.3.0' => '~>2.2.3', '2.2.2' => '~>2.0.3'}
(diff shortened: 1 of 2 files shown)


kaylareopelle (approved):
😹 

This is exactly what I hoped for, thank you!

PR #1217, opened 2022-06-23
Create Unicorn multiverse suite

### LINKED ISSUE #1102: Add tests for Unicorn
We currently do not have a multiverse suite to cover the assignment of unicorn as the http server/dispatcher. Let's write tests to confirm our integration with different versions of this gem.

This gem is referenced in:
* lib/new_relic/local_environment.rb
* lib/new_relic/agent/agent.rb

Our documentation states we support versions 5.3.0 and above.

kaylareopelle:
# Overview
This PR adds a multiverse test suite for our supported versions of the `unicorn` server gem.

Resolves #1102


test/multiverse/suites/unicorn/Envfile
```diff
@@ -0,0 +1,16 @@
+SUPPORTED_UNICORN_VERSIONS = [
```
tannalynn, line 1: Are these the only supported versions, or is it just what we're going to be testing? Just wanted to clarify because of the name here confused me a little.

tannalynn (dismissed):
Looks great! 🎉 Just had one question for clarification

test/multiverse/suites/unicorn/Envfile
kaylareopelle: Great question! These are the supported versions once we get to major release 9.0. I made the mistake of consulting the EOL Gem Spreadsheet instead of our public documentation for the versions here. We should really have something from the 4.x series right now to make sure we're reflecting the public docs!

kaylareopelle: @tannalynn  - I've added 4.9.0 to the supported versions list. Strangely, it doesn't have a Ruby 3.0 restriction like Unicorn 5.3.0! 😅 

Please re-review when you can!

test/multiverse/suites/unicorn/Envfile
```diff
@@ -0,0 +1,17 @@
...
+  nil,
+  ", '6.1.0'",
+  ", '4.9.0'"
+]
+
+SUPPORTED_UNICORN_VERSIONS << ", '5.3.0'" if RUBY_VERSION < '3.0.0'
+
+SUPPORTED_UNICORN_VERSIONS.each do |version|
+  next if RUBY_PLATFORM == 'java'
```
fallwith, line 10: It'd be nice to be able to declare the "JRuby doesn't get tested here at all" bit in a higher level way for Envfiles. As-is, the array gets defined, appended to, and then iterated over with fully all 4 supported versions being examined.

Perhaps for now (shy of an Envfile overhaul) we could do one of these things:

- Change the instantiation of `SUPPORTED_UNICORN_VERSIONS` and the appending to it to result in `[]` as the value for JRuby.
- Have `break` hit once instead of `next` hit four times.

test/multiverse/suites/unicorn/config/newrelic.yml
```diff
@@ -0,0 +1,18 @@
+---
```
fallwith, line 1: Should we include the license header for YAML files? Currently only .rb and .rake files are checked for it.

test/multiverse/suites/unicorn/unicorn_test.rb
```diff
@@ -0,0 +1,41 @@
...
+
+require 'unicorn'
+require 'newrelic_rpm'
+
+class UnicornTest < Minitest::Test
+  include MultiverseHelpers
+  attr_accessor :server
+
+  def setup
+    File.truncate('log/newrelic_agent.log', 0)
```
fallwith, line 14: My hunch is that this operation is more costly than `File.unlink`. Is there a risk of losing the write access required to replace the deleted file that has led us to truncate instead of deleting?

test/multiverse/suites/unicorn/unicorn_test.rb
```diff
@@ -0,0 +1,41 @@
...
+  def test_unicorn_set_as_discovered_dispatcher
+    assert_logged('Dispatcher: unicorn')
+  end
+
+  def test_defer_message_logged_when_unicorn_in_use
+    assert_logged('Deferring startup of agent reporting thread because unicorn may fork.')
+  end
+
+  def assert_logged(expected_message)
+    messages = File.readlines('log/newrelic_agent.log')
```
fallwith, line 35: Does the startup of Unicorn happen in such a way that we can't hang an expectation on the logger itself, necessitating the reading of the log file instead?

fallwith (dismissed):

test/multiverse/suites/unicorn/Envfile
kaylareopelle: Ah, great catch. While investigating the Envfile refactor, I found this option that already exists to skip JRuby: 

```ruby
# test/multiverse/suites/bunny/Envfile

suite_condition("bunny not supported for JRuby") do
  RUBY_PLATFORM != 'java'
end
```

I forgot to mention this during our last meeting about Envfiles. I'd like to integrate it into all our suites that skip JRuby. Would this work for you?

test/multiverse/suites/unicorn/config/newrelic.yml
kaylareopelle: Looking at New Relic's guidelines for Open Source repositories ([this one is the most recent - Internal](https://docs.google.com/document/d/1g4v0LAro5dqEsKqr3DiIuFIhhWvEJhuPu_jgM6lCFS0/edit#)), it seems the license header is needed only on source files. What qualifies as a source file is not defined. 

My hunch would be that source files are what we distribute in the `lib` directory to our customers. This would lead me to believe that we don't need license headers in any of our test suite files. 

I'd be happy to add them to the YAML files if we prefer to err on the side of over-commenting.

test/multiverse/suites/unicorn/unicorn_test.rb
kaylareopelle: That's what I've noticed. The New Relic logger, though it is the object adding these items to the agent log, cannot find the reference to the printed unicorn log lines. In my tests, I may have been trying to access the logger incorrectly. I'd be interested in pairing on this if you'd like to take a look with me.

test/multiverse/suites/unicorn/unicorn_test.rb
kaylareopelle: I didn't know of `File.unlink`'s existence! I'll give that a try.

test/multiverse/suites/unicorn/unicorn_test.rb
kaylareopelle: Turns out, I was missing a stub in the setup block to allow me to connect with the logger object. This should be fixed now!

test/multiverse/suites/unicorn/Envfile
fallwith: Yep, that sounds great. I was thinking you'd have a great long term fix like that as part of the Envfile overhaul work. That's even better than the short-term stopgap measures I was proposing, definitely.

test/multiverse/suites/unicorn/config/newrelic.yml
fallwith: Having recently looked at the automated license checker, our license header is fresh in mind.

Good thoughts, thanks. I personally would prefer no headers at all on any files, and simply use the existence of the `LICENSE` file in the root of the project. For any type of relaying of individual code (if we were to attach it to an email, put it in a gist, etc.), then we might want to put the header at the top in lieu of also relaying the entire `LICENSE` file, but as long as everything is in the repo I think the header is overkill. It's especially bothersome that our header simply says to open up `LICENSE` as opposed to letting people with a standalone file-with-header in front of them to at least know that it's Apache 2 that's involved.

fallwith (approved):

PR #1218, opened 2022-06-23
Automate Envfiles

### LINKED ISSUE #1184: Create automated Envfiles
Our multiverse suites have a filetype called "Envfile" that holds heredocs that describe the Gemfiles that a given suite should be tested against. 

All Envfiles share some similarities, however, there is enough difference between each implementation to cause friction when trying to determine whether all desired combinations of gems are being tested.

### Task
Create a strategy to automate Envfile creation and standardize the format of these files.

kaylareopelle:
# Overview
WIP

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Install required packages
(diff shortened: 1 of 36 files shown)


kaylareopelle:
Closing, as we decided to take a different approach

PR #1226, opened 2022-06-27
Add initial structure for gem slack notifications

### LINKED ISSUE #1195: Create a tool to notify us about new releases of core technologies
To help us keep up to date on the latest releases of the gems we instrument, let's create a new tool to help us: 
* Search all libraries instrumented by the Ruby agent
* Search all Ruby versions and interpreters supported by the Ruby agent 
* Receive notifications/alerts when a new version of a gem is released
* Receive notifications/alerts when a new version of a Ruby interpreter/version is released 

Beyond the above criteria being met, the implementation is completely up to the owner of this ticket.

Some ideas for implementation:
* Have this run as part of our CI. Have the CI record what version of a gem was previously tested. Now that we're always testing the latest version, have it tell us when there's a different version available. (James)
* Set this up as a Slackbot that runs on a cron using Bundle outdated to check the contents of a Gemfile (Hannah/Kayla)

hannahramadan:
# Overview

A ruby script that checks for and notifies the Ruby agent team slack channel of new gem version releases.
closes #1195 

Inside `ci_cron.yml`, a newly added`gem_notifications` job stores a list of gems the team has determined to be core technologies. When the nightly CI runs, this list of gems is passed to the script, `notifications_script`. 

Gem updates are determined by calling the Ruby Gems API and using the `created_at` timestamp. If the release is within the past 24 hours, we know a new version is our and send a POST request to an internal Slackbot.

Files Added:
- `.github/workflows/scripts/slack_gem_notifications/notifications_methods`
- `.github/workflows/scripts/slack_gem_notifications/notifications_script`
- `test/gem_notifications_tests`

Files Edited:
- `ci_cron.yml`

Submitter Checklist:

# Testing
exercise your changes.

### DIFF
--- a/.github/workflows/ci_cron.yml
+++ b/.github/workflows/ci_cron.yml
@@ -338,7 +338,7 @@ jobs:
       - uses: actions/checkout@v2
       - run: gem install httparty
       - name: Check for outdated gems 
-        run: ruby ./.github/workflows/scripts/slack_gem_notifications/notifications_script.rb ${{ env.gems }}
+        run: ruby .github/workflows/scripts/slack_gem_notifications/notifications_script.rb ${{ env.gems }}
         env:
           SLACK_GEM_NOTIFICATIONS_WEBHOOK: ${{ secrets.SLACK_GEM_NOTIFICATIONS_WEBHOOK }}
           gems:
--- a/.github/workflows/ci_cron.yml
+++ b/.github/workflows/ci_cron.yml
@@ -328,3 +328,35 @@ jobs:
       - name: Annotate errors
         if: ${{ failure() }}
         uses: ./.github/actions/annotate
+
+  gem_notifications:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: ruby/setup-ruby@v1.90.0
+        with:
+          ruby-version: 2.7
+      - uses: actions/checkout@v2
+      - run: gem install httparty
+      - name: Check for outdated gems 
+        run: ruby ./.github/workflows/scripts/slack_gem_notifications/notifications_script.rb ${{ env.gems }}
+        env:
+          SLACK_GEM_NOTIFICATIONS_WEBHOOK: ${{ secrets.SLACK_GEM_NOTIFICATIONS_WEBHOOK }}
+          gems:
+            "activerecord 
+            bunny
+            dalli
+            delayed_job
+            excon
+            http
+            httpclient
+            mongo
+            puma
+            sidekiq
+            sinatra
+            tilt
+            rack
+            rails
+            rake 
+            redis
+            resque
+            unicorn"
--- a/.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
+++ b/.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
@@ -1,0 +1,48 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'time'
+require 'httparty'
+
+def check_for_updates(watched_gems)
+  return if gem_list_empty(watched_gems)
+  watched_gems.each do |gem_name|
+    verify_gem(gem_name) ? gem_info = verify_gem(gem_name) : return
+    versions = gem_versions(gem_info)
+    send_bot(gem_name, versions) if gem_updated?(versions)
+  end
+end
+
+def gem_list_empty(watched_gems)
+  if watched_gems.empty?
+    abort "Nothing to see here! The 'watched_gems' array cannot be empty"
+  end
+end
+
+def verify_gem(gem_name)
+  gem_info = HTTParty.get("https://rubygems.org/api/v1/versions/#{gem_name}.json")
+
+  gem_info if gem_info.success?
+end
+
+# Gem entries are ordered by release date. The break limits versions to two versions: newest and previous.
+def gem_versions(gem_info)
+  versions = gem_info.each_with_object([]) do |gem, arr|
+    arr << gem if gem['platform'] == 'ruby'
+    break arr if arr.size == 2
+  end
+end
+
+def gem_updated?(versions)
+  Time.now.utc - Time.parse(versions[0]['created_at']) < 24 * 60 * 60
+end
+
+def send_bot(gem_name, versions)
+  abort "Expected exactly 2 version numbers in the 'versions' array" unless versions.size == 2
+  newest, previous = versions[0]['number'], versions[1]['number']
+  HTTParty.post(ENV['SLACK_GEM_NOTIFICATIONS_WEBHOOK'],
+    headers: {'Content-Type' => 'application/json'},
+    body: {text: "A new gem version is out! Gem: #{gem_name}, #{previous} -> #{newest}."}.to_json)
+end
--- a/.github/workflows/scripts/slack_gem_notifications/notifications_script.rb
+++ b/.github/workflows/scripts/slack_gem_notifications/notifications_script.rb
@@ -1,0 +1,10 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+# This script runs on a 24 hour cycle via ci_cron.yml and sends slack updates for new gem version releases.
+
+require_relative 'notifications_methods'
+
+check_for_updates(ARGV)
--- a/test/new_relic/gem_notifications_tests.rb
+++ b/test/new_relic/gem_notifications_tests.rb
@@ -1,0 +1,63 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+# require_relative '../test_helper'
+require 'minitest/autorun'
+require_relative '../../.github/workflows/scripts/slack_gem_notifications/notifications_methods'
+
+class GemNotifications < Minitest::Test
+  def invalid_gem_response
+    response = MiniTest::Mock.new
+    response.expect :success?, false
+  end
+
+  def valid_gem_response
+    response = MiniTest::Mock.new
+    response.expect :success?, true
+  end
+
+  def http_get_response
+    [{"created_at" => "2001-07-18T16:15:29.083Z", "platform" => "ruby", "number" => "3.0.0"},
+      {"created_at" => "1997-05-23T16:15:29.083Z", "platform" => "java", "number" => "2.0.0"},
+      {"created_at" => "1993-06-11T16:15:29.083Z", "platform" => "ruby", "number" => "1.0.0"}]
+  end
+
+  def test_valid_gem_name
+    response = valid_gem_response()
+    HTTParty.stub :get, response do
+      assert verify_gem("puma!")
+    end
+  end
+
+  def test_invalid_gem_name
+    response = invalid_gem_response()
+    HTTParty.stub :get, response do
+      assert_nil verify_gem("TrexRawr!")
+    end
+  end
+
+  def test_get_gem_info_returns_array
+    versions = gem_versions(http_get_response())
+    assert_instance_of Array, versions
+  end
+
+  def test_get_gem_info_max_size
+    versions = gem_versions(http_get_response())
+    assert_equal true, versions.size == 2
+  end
+
+  def test_gem_updated
+    assert_equal true, gem_updated?([{"created_at" => "#{Time.now}"}])
+    assert_equal false, gem_updated?([{"created_at" => "1993-06-11T17:31:14.298Z"}])
+  end
+
+  def test_send_bot_input_size
+    assert_raises(ArgumentError) { send_bot() }
+    assert_raises(ArgumentError) { send_bot("tyrannosaurus") }
+    HTTParty.stub :post, nil do
+      assert_nil send_bot("tyrannosaurus", [{"number" => "83.6"}, {"number" => "66.0"}])
+    end
+  end
+end
--- a/newrelic_rpm.gemspec
+++ b/newrelic_rpm.gemspec
@@ -64,4 +64,5 @@ https://github.com/newrelic/newrelic-ruby-agent/
     s.add_development_dependency 'simplecov'
     s.add_development_dependency 'simplecov_json_formatter'
   end
+  s.add_development_dependency 'httparty'
 end


.github/workflows/ci_cron.yml
```diff
@@ -328,3 +328,18 @@ jobs:
...
         uses: ./.github/actions/annotate
+
+  gem_notifications:
+    runs-on: ubuntu-20.04
+    env:
+      gems: "activerecord bunny dalli delayed_job excon http httpclient mongo puma sidekiq sinatra tilt rack rails rake redis resque unicorn"
+    steps:
+      - uses: ruby/setup-ruby@v1.90.0
+        with:
+          ruby-version: 2.7
```
kaylareopelle, line 339: I'm curious, why did you decide to use Ruby 2.7?

.github/workflows/ci_cron.yml
```diff
@@ -328,3 +328,18 @@ jobs:
       - name: Annotate errors
         if: ${{ failure() }}
         uses: ./.github/actions/annotate
+
+  gem_notifications:
+    runs-on: ubuntu-20.04
+    env:
+      gems: "activerecord bunny dalli delayed_job excon http httpclient mongo puma sidekiq sinatra tilt rack rails rake redis resque unicorn"
```
kaylareopelle, line 335: Could this accept an array argument? It might be nicer to maintain something like: 
```yaml
  env:
    gems:
      - activerecord
      - bunny
      - dalli
      - etc... 
```

.github/workflows/scripts/slack_gem_notifications.rb
```diff
@@ -0,0 +1,42 @@
...
+  warn 'WATCHED_GEMS cannot be empty'
+  exit
+end
+
+def watch_gem(gem_name)
+  gem_info = HTTParty.get("https://rubygems.org/api/v1/versions/#{gem_name}.json")
+  latest_versions = []
+  gem_info.each do |gem|
+    latest_versions << gem if gem['platform'] == 'ruby'
+    break if latest_versions.size == 2
```
kaylareopelle, line 23: Why is this break statement here? What happens if `latest_versions.size == 2` that would make us want to exit?

.github/workflows/scripts/slack_gem_notifications.rb
```diff
@@ -0,0 +1,42 @@
...
+
+if WATCHED_GEMS.empty?
+  warn 'WATCHED_GEMS cannot be empty'
+  exit
+end
+
+def watch_gem(gem_name)
+  gem_info = HTTParty.get("https://rubygems.org/api/v1/versions/#{gem_name}.json")
+  latest_versions = []
+  gem_info.each do |gem|
```
kaylareopelle, line 21: There are other methods in Ruby that return a new array containing the block's return values as items. You may be able to refactor this to use `#map` or `#collect` and avoid the array definition on line 20.

kaylareopelle (commented):
I really like this refactor, Hannah! It's getting so much cleaner and easier to read. 

I have a few in-line comments. 

I also want to suggest creating one overall method that calls `watch_gems`, `compare_gems` and `send_bot` and call that method on line 41 instead of `watch_gems` so that the responsibility of the individual methods remains more clear. 

Also, what are your thoughts about adding some tests here? We talked about this before your vacation, but I'm not sure where you currently stand.

.github/workflows/ci_cron.yml
fallwith, line 335: Thank you for the alphabetical ordering!

.github/workflows/scripts/slack_gem_notifications.rb
fallwith: ^-- This would be a fun exercise. NOTE that the value returned by `break` will be used by `map`, so you'd have to do something like `break gem if gem['platform'] == 'ruby' && !latest_versions.empty?` to pull this off.

.github/workflows/scripts/slack_gem_notifications.rb
fallwith: ^-- This is a great question for which it would be good to have a comment based answer to.

.github/workflows/scripts/slack_gem_notifications.rb
```diff
@@ -0,0 +1,42 @@
...
+  compare_gems(gem_name, latest_versions)
+end
+
+def compare_gems(gem_name, latest_versions)
+  newest, previous = latest_versions[0]['number'], latest_versions[1]['number']
+  send_bot(gem_name, previous, newest)
+end
+
+def send_bot(gem, current_version, latest_version)
+  `curl -X POST -H 'Content-type: application/json' --data '{"text":"Hey Ruby team! There is a new gem version release.
```
fallwith, line 36: After the work done to bring in HTTParty, perhaps it could be used instead of backticking `curl` here to do this POST.

As-is, the `curl` command is problematic, because on a multi-user unix system, the publicly available process list will display the sensitive webhook environment variable. This is possibly not an issue with GHA, but would definitely be addressed with a switch to HTTParty.

.github/workflows/ci_cron.yml
hannahramadan: I will explore this. I agree that looks cleaner!

.github/workflows/ci_cron.yml
hannahramadan: I went with ruby 2.7 because it was the version we use for the rubocop and simplecov jobs. I just liked the consistency of it, but happy to change it if you think that is a better move.

.github/workflows/scripts/slack_gem_notifications.rb
hannahramadan: It is an answer I will comment! But also, the [Ruby Gems API](https://guides.rubygems.org/rubygems-org-api/) returns a large array of hashes (which each represents a release). Array entries are in order by the date they were released, the latest appearing first.

The break limits us to the only 2 ruby-platform items/hashes we want: the newest version and the previous version.

There's a TODO to revisit limiting the API response to how much we're getting back, but right now that isn't supported.

.github/workflows/ci_cron.yml
hannahramadan: I've tried a couple of different formats, but none that GA likes. Maybe we can look at this together? I'm sure there's something...

.github/workflows/ci_cron.yml
```diff
@@ -328,3 +328,36 @@ jobs:
...
+        puma
+        sidekiq
+        sinatra
+        tilt
+        rack
+        rails
+        rake 
+        redis
+        resque
+        unicorn"
```
fallwith, line 353: 1. Putting one gem on each line is a cool way to do it. 👍 
2. I didn't realize earlier that the "Check for outdated gems" step already has its own `env:` section. The "gems" env var can be moved down there and this global-to-all-steps `env:` list can be deleted.

.github/workflows/ci_cron.yml
```diff
@@ -328,3 +328,36 @@ jobs:
...
+    steps:
+      - uses: ruby/setup-ruby@v1.90.0
+        with:
+          ruby-version: 2.7
+      - uses: actions/checkout@v2
+      - run: gem install httparty
+      - name: Check for outdated gems 
+        run: ruby ./.github/workflows/scripts/slack_gem_notifications/notifications_script.rb ${{ env.gems }}
```
fallwith, line 361: The leading `./` is unnecessary (though it doesn't hurt anything if you deliberately included it for style).

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
```diff
@@ -0,0 +1,55 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+
+require 'time'
```
fallwith, line 5: What happens if you don't require 'time'?

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
```diff
@@ -0,0 +1,55 @@
...
+require "httparty"
+
+def check_for_updates(watched_gems)
+  return if gem_list_empty?(watched_gems)
+  watched_gems.each do |gem_name|
+    return unless is_valid_gem?(gem_name)
```
fallwith, line 11: Verifying that a good HTTP response was received before attempting to look at the version info is a great addition! Nice job. There is a drawback to the current validation approach, though. The exact same GET request is being made twice; once for validation and once to go through the version data. Can you think of a way to still perform the important validation while only performing a single GET request?

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+end
+
+def gem_list_empty?(watched_gems)
+  if watched_gems.empty?
+    raise 'Nothing to see here! watched_gems cannot be empty'
+    exit
+  end
+
+  watched_gems.empty?
```
fallwith, line 23: Given that the array has already been checked with an `#empty?` call and the script would have exited if it returned `true`, there's no need to call `#empty?` again. You can just put a literal `false` here or not put anything to return `nil` (Ruby will consider both `false` and `nil` to be falsey).

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
```diff
@@ -0,0 +1,55 @@
...
+end
+
+def is_valid_gem?(gem_name)
+  res = HTTParty.get("https://rubygems.org/api/v1/versions/#{gem_name}.json")
+
+  res.code == 200
```
fallwith, line 29: `res` here is an instance of the class `HTTParty::Response`. Does that class offer us any way to confirm that the response was successful apart from us having to manually check the numeric HTTP response code?

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
```diff
@@ -0,0 +1,55 @@
...
+  versions
+end
+
+def gem_updated?(versions)
+  Time.now.utc - Time.parse(versions[0]['created_at']) < 24 * 60 * 60
+end
+
+def send_bot(gem_name, versions)
+  newest = versions[0]['number'] if versions[0]
+  previous = versions[1]['number'] if versions[1]
```
fallwith, line 51: Confirming that the 2 versions exist is a great idea! Currently if one or both are missing, this method will still proceed to fire off the post to Slack. Is that ok, or are at least one of the version numbers required?

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
```diff
@@ -0,0 +1,55 @@
...
+  HTTParty.post(ENV['SLACK_GEM_NOTIFICATIONS_WEBHOOK'],
```
fallwith, line 52: Hooray! You figured out a way to leverage HTTParty instead of backticking `curl`. Great work!

.github/workflows/scripts/slack_gem_notifications/notifications_script.rb
```diff
@@ -0,0 +1,13 @@
...
+
+# This script runs on a 24 hour cycle via ci_cron.yml and sends slack updates for new gem version releases.
+
+require_relative 'notifications_methods'
+
+# TODO: Add support for comparing latest and previous ruby versions
+
+WATCHED_GEMS = ARGV
+
+check_for_updates(watched_gems)
```
fallwith, line 13: There's a constant called `WATCHED_GEMS` (all caps) and then a variable being passed to `check_for_updates` called `watched_gems` (all lowercase). Is this a mismatch? If so, maybe you can get rid of the constant and just do `check_for_udpates(ARGV)`.

test/gem_notifications_tests.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def test_gem_list_empty
+    error = assert_raises(Exception) { gem_list_empty?([]) }
+    assert_equal("Nothing to see here! watched_gems cannot be empty", error.message)
+    assert_equal false, gem_list_empty?(['tyrannosaurus'])
+  end
+
+  def test_valid_gem_name
+    assert_equal true, is_valid_gem?("puma")
+    assert_equal false, is_valid_gem?("TrexRawr!")
```
fallwith, line 27: I see 2 problems here:

- The HTTParty gem appears to be used when these assertions are made, but `httparty` is not declared as a dev dependency for the project.
- We want to avoid making any real network calls to third party servers when the tests are ran.

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
```diff
@@ -0,0 +1,55 @@
...
+def get_gem_info(gem_name)
+  gem_info = HTTParty.get("https://rubygems.org/api/v1/versions/#{gem_name}.json")
+  versions = []
+  gem_info.each do |gem|
+    versions << gem if gem['platform'] == 'ruby'
+    break if versions.size == 2
+  end
+
+  versions
+end
```
kaylareopelle, line 43: What happened to our refactor of this method?

.github/workflows/scripts/slack_gem_notifications/notifications_script.rb
```diff
@@ -0,0 +1,13 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+
+# This script runs on a 24 hour cycle via ci_cron.yml and sends slack updates for new gem version releases.
+
+require_relative 'notifications_methods'
+
+# TODO: Add support for comparing latest and previous ruby versions
```
kaylareopelle, line 9: Do you still need this todo?

test/gem_notifications_tests.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def test_get_gem_info_returns_array
+    versions = get_gem_info('puma')
+    assert_instance_of Array, versions
+  end
+
+  def test_get_gem_info_max_size
+    versions = get_gem_info('puma')
+    assert_equal true, versions.size <= 2
+  end
```
kaylareopelle, line 38: Nice! 😄

kaylareopelle (commented):
Huge improvements since the last review! 🌈 I really like the comments James made. Keep up the good work!

.github/workflows/ci_cron.yml
hannahramadan: Oh cool! I didn't know that. Will remove.

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
hannahramadan: My tests failed without requiring Time. I didn't understand this as I thought it was an included Ruby Class already.

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
hannahramadan: sure enough I can do `res.success?` ✨ nice call out, thank you

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
hannahramadan: I  couldn't get the refactor to work :( We added a slice but I realized that doesn't work because we first need to grab two items that have gem['platform'] == 'ruby'.

I'm interested in making this smoother but have run out of ideas.

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
```diff
@@ -0,0 +1,55 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+
+require 'time'
+require "httparty"
```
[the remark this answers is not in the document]
hannahramadan: Oops! Will edit this, thank you!

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
fallwith: We want to grab the first 2 hashes in the `gem_info` array whose `'platform'` values are `'ruby'`.

If we were to use `#map` or `#select`, we run into inefficiencies because we have to check the `'platform'` value for every single hash and only afterwards cut down the list to `2`.

```ruby
# inefficient - checks on every entry in the array and then grabs 2 afterwards
gem_info.select { |gem| gem['platform'] == 'ruby' }.take(2)
```

Kayla's original comment pointed to not having to declare an empty array and then fill it as two separate steps. We can shrink things down with `#each_with_object` and still retain the efficiency that comes with an early break. It just looks a little different, because `break` in that context will need a return value:

```ruby
versions = gem_info.each_with_object([]) do |gem, arr|
  arr << gem if gem['platform'] == 'ruby'
  break arr if arr.size == 2
end
```

.github/workflows/ci_cron.yml
fallwith: This consistency with the other jobs will be helpful when it comes time to switch to Ruby 3.1 as part of the migration to Ubuntu 22. 👍

test/new_relic/gem_notifications_tests.rb
```diff
@@ -0,0 +1,73 @@
...
+  def http_get_response
+    [{"created_at": "2022-03-30T16:15:29.083Z", "platform": "ruby"},
+      {"created_at": "2021-03-30T16:15:29.083Z", "platform": "java"},
+      {"created_at": "2020-03-30T16:15:29.083Z", "platform": "ruby"}]
+  end
+
+  def test_check_for_updates_if_empty
+    response = MiniTest::Mock.new
+    response.expect :empty?, true
+    [].stub :empty?, response do
```
fallwith, line 28: From what I can tell, this test will instruct every instance of `Array` to return `true` when asked `#empty?`. But then the only array you really put into play is `[]` when it gets passed to `check_for_updates()`, and `[]` would return `true` for `#empty?` by default without the need for stubbing.

You should be able to remove the first 3 lines of this test and the `end` block (remove all mocking and stubbing).

test/new_relic/gem_notifications_tests.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def valid_gem_response
+    response = MiniTest::Mock.new
+    response.expect :success?, true
+  end
+
+  def http_get_response
+    [{"created_at": "2022-03-30T16:15:29.083Z", "platform": "ruby"},
+      {"created_at": "2021-03-30T16:15:29.083Z", "platform": "java"},
+      {"created_at": "2020-03-30T16:15:29.083Z", "platform": "ruby"}]
```
fallwith, line 22: Great use of realistic response data, and shared use of it in a method!

Two requests here:

**request 1**
The real response (yielded after HTTParty gives it a parse) uses strings for each gem entry's hash key. We should match that format with this test data.

So instead of `"created_at":` which uses a symbol for the key, go with `'created_at' =>` to ensure the use of a string for the hash key.

**request 2**
The current version of the code being tested (to be changed later when the TODOs get satisfied) looks for fully 3 separate keys in each gem entry hash: `'created_at'`, `'platform'`, and `'number'`). Even though your tests don't use this response data for testing the code that looks for the `'number'` key, it would be useful to have the `'number'` key and a corresponding realistic Rubygem version number value represented in the canned response hash here. That way a human reviewer of the code can understand the data better.

test/new_relic/gem_notifications_tests.rb
```diff
@@ -0,0 +1,73 @@
...
+    [].stub :empty?, response do
+      error = assert_raises(Exception) { check_for_updates([]) }
+      assert_equal("Nothing to see here! watched_gems cannot be empty", error.message)
+    end
+  end
+
+  def valid_gem_name
+    response = valid_gem_response()
+    HTTParty.stub :get, response do
+      assert_nil is_valid_gem?("puma!")
```
fallwith, line 37: I'm confused by this one. If a valid gem response is being returned, then shouldn't the `is_valid_gem?` call return the response object instead of `nil`?

test/new_relic/gem_notifications_tests.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def test_get_gem_info_returns_array
+    versions = gem_versions(http_get_response())
+    assert_instance_of Array, versions
+  end
+
+  def test_get_gem_info_max_size
+    versions = gem_versions(http_get_response())
+    assert_equal true, versions.size <= 2
```
fallwith, line 55: Can this be `== 2`? One noble goal of the tests is to tell the story of the new or updated functionality. Given your canned response that contains 3 entries (one which is Java based) and your new code that wants to report on a version number change by looking at two different version numbers, I think `==2` would tell a better story than `<= 2`.

test/new_relic/gem_notifications_tests.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def test_gem_updated
+    assert_equal true, gem_updated?([{"created_at" => "#{Time.now}"}])
+    assert_equal false, gem_updated?([{"created_at" => "1993-06-11T17:31:14.298Z"}])
```
fallwith, line 60: Excellent use of an easter egg!

test/new_relic/gem_notifications_tests.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def test_send_bot_input_size
+    assert_raises(ArgumentError) { send_bot() }
+    assert_raises(ArgumentError) { send_bot("tyrannosaurus") }
+    HTTParty.stub :post, nil do
+      assert_nil send_bot("tyrannosaurus", [{"number" => "1.2.3"}])
+    end
+    HTTParty.stub :post, nil do
+      assert_nil send_bot("tyrannosaurus", [{"number" => "1.2.3"}, {"number" => "4.5.6"}, {"number" => "7.8.9"}])
```
fallwith, line 70: This assertion is interesting because to me it indicates that the `send_bot()` method currently doesn't something I would think we don't want it to do in that it will happily accept a version array with anything other than exactly 2 entries. Is that desired?

If not, you can guard against any array size other than 2 with something like:

```ruby
def send_bot(gem_name, versions)
  raise 'Expected exactly 2 version numbers (newest and previous) in the 'versions' array' unless versions.size == 2
 
  ...
```

or even:

```ruby
def send_bot(gem_name, newest_version, previous_version)
  ...
```

and then add the appropriate assertions to check the functionality.

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
```diff
@@ -0,0 +1,48 @@
+# encoding: utf-8
```
kaylareopelle, line 1: What do you think about adding the `# frozen_string_literal: true` magic comment to this file?

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
```diff
@@ -0,0 +1,48 @@
...
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+
+require 'time'
+require 'httparty'
+
+def check_for_updates(watched_gems)
+  return if gem_list_empty?(watched_gems)
+  watched_gems.each do |gem_name|
+    is_valid_gem?(gem_name) ? gem_info = is_valid_gem?(gem_name) : return
```
kaylareopelle, line 11: When a method ends in a question mark, I expect it to return a boolean. With the refactored scope of `is_valid_gem?`, perhaps a different name is in order?

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
```diff
@@ -0,0 +1,48 @@
...
+    versions = gem_versions(gem_info)
+    send_bot(gem_name, versions) if gem_updated?(versions)
+  end
+end
```
kaylareopelle, line 15: I love that the definition of the subsequent methods is in the order in which they're called within `check_for_updates`!

test/new_relic/gem_notifications_tests.rb
hannahramadan: Makes sense! Thank you.

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
```diff
@@ -0,0 +1,48 @@
...
+
+def gem_list_empty?(watched_gems)
+  if watched_gems.empty?
+    raise 'Nothing to see here! watched_gems cannot be empty'
+    exit
```
kaylareopelle, line 20: Nice use of `exit`! I wasn't familiar with that method and learned a bit more in, https://www.rubyguides.com/2019/03/ruby-exit/

It looks like you could accomplish what's on line 19 and 20 using `abort 'Nothing to see here! watched_gems cannot be empty'` if you'd like!

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
```diff
@@ -0,0 +1,48 @@
...
+  gem_info if gem_info.success?
+end
+
+# Gem entries are ordered by release date. The break limits versions to two versions: newest and previous.
+def gem_versions(gem_info)
+  versions = gem_info.each_with_object([]) do |gem, arr|
+    arr << gem if gem['platform'] == 'ruby'
+    break arr if arr.size == 2
+  end
+end
```
kaylareopelle, line 36: Nice!

kaylareopelle (approved):
THIS IS SO COOL! Few small things, feel free to disregard them!

test/new_relic/gem_notifications_tests.rb
hannahramadan: 🦖

test/new_relic/gem_notifications_tests.rb
hannahramadan: Good call. I went ahead and matched this to the exact response format and included the 3 keys we use at some point.

test/new_relic/gem_notifications_tests.rb
hannahramadan: Yes, this makes sense

test/new_relic/gem_notifications_tests.rb
hannahramadan: Originally I imagined a world where a watched_gem was on V1. But I'm seeing how there would be no case where we'd send a bot without the previous version, unless we somehow knew about the gem before it was even released. 

I'll refactor this.

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
hannahramadan: I think a wonderful contributor added `# frozen_string_literal: true` to all our files and it would be a shame to take that in the wrong direction! Added ✅ and thank you!

.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb
fallwith: `exit` when not passed any arguments will cause the Ruby application to exit with a code 0, signaling that everything worked fine. We should have been exiting here with a non 0 code to signal that there was an issue:

```ruby
exit false
# or
exit 42
```

Another nice thing about `abort` is that it automatically calls `exit false` for the desired non-zero exit code.

Great suggestion, @kaylareopelle! We should go with it!

test/new_relic/gem_notifications_tests.rb
hannahramadan: I was stumped on how this was passing. Finally figured out the method didn't start with `test` 🙃

fallwith (approved):
👍

fallwith (approved):

PR #1212, opened 2022-06-22
Add magic comments to all the files

radar07:
- Added frozen_string_literal: true magic comment to all the missing files.
- Added encoding, and license magic comments to all the missing rake files.

There are no breaking changes in the PR but this may cause some of the CI checks to fail. This might close issue #1198.

Issue #1198, opened 2022-06-03
All Ruby files should contain the "frozen_string_literal: true" magic comment


fallwith:
Around 74% of our code base's individual Ruby files (`*.rb`, `Rakefile`, `Gemfile`, etc.) are missing the `frozen_string_literal: true` magic comment string that tells Ruby to freeze strings by default.

The comment section at the top of each of our Ruby files should declare UTF-8 encoding, provide our license info, and then enable frozen strings. The comment section should look like this for every file:

```ruby
# encoding: utf-8
# This file is distributed under New Relic's license terms.
# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
# frozen_string_literal: true
```

With [ugrep](https://github.com/Genivia/ugrep), the files that are missing the `frozen_string_literal: true` comment are easily located:

```shell
ug --file-type=ruby -L frozen_string_literal
```

All of the files currently missing the comment should be updated to include it.

NOTE: This issue has been labeled with `good first issue` because the list of actions required is simply:

1. Find the files missing the comment
2. Add the comment to each file found

Once the files have been updated, it is possible that the newly enabled frozen string behavior may cause our CI checks to fail. Modifications to the agent code and/or the tests to get the CI checks to pass would certainly be welcome, but are not at all required as part of submitting a PR for this issue. Only the steps listed above are needed. New Relic's dev team can take things from there.

radar07:
@fallwith Can you assign this to me? I'd like to work on this.

kaylareopelle:
Hi @radar07! 👋  

@fallwith is out right now, so I went ahead and assigned you to this ticket! Let us know if you need any help. Thanks for your interest in contributing!

radar07:
@kaylareopelle Thanks!!

radar07:
Hi @kaylareopelle, I have a couple of questions.

1. When I tried adding the frozen string literal comment in the file and executed bundle exec rake, [database_test.rb](https://github.com/newrelic/newrelic-ruby-agent/blob/553c5a9ef02cc3a7ccd7c0af0d4f9cafdefff195/test/new_relic/agent/database_test.rb) this line   **INVALID_UTF8_STRING = (''.respond_to?(:force_encoding) ? "select \x80".force_encoding('UTF-8') : "select \x80")**
 is throwing the error **can't modify frozen String: "select \\x80" (FrozenError)**. 

[...]

PR #1229, opened 2022-06-28
Freeze all strings by default

fallwith:
Thanks to the wonderful contribution by @radar07 via #1212, all Ruby files in the project now make use of the `frozen_string_literal: true` magic comment to have Ruby freeze all strings by default.

This PR intends to deliver @radar07's changes to `dev` along with a few select alterations made to strings that need to actually remain mutable.


lib/new_relic/agent/new_relic_service.rb
```diff
@@ -625,11 +626,10 @@ def decompress_response(response)
...
-        ruby_description = ''
-        # note the trailing space!
-        ruby_description << "(ruby #{::RUBY_VERSION} #{::RUBY_PLATFORM}) " if defined?(::RUBY_VERSION) && defined?(::RUBY_PLATFORM)
-        zlib_version = ''
-        zlib_version << "zlib/#{Zlib.zlib_version}" if defined?(::Zlib) && Zlib.respond_to?(:zlib_version)
+        if defined?(::RUBY_VERSION) && defined?(::RUBY_PLATFORM)
+          ruby_description = "(ruby #{::RUBY_VERSION} #{::RUBY_PLATFORM}) " # note the trailing space!
+        end
+        zlib_version = "zlib/#{Zlib.zlib_version}" if defined?(::Zlib) && Zlib.respond_to?(:zlib_version)
         "NewRelic-RubyAgent/#{NewRelic::VERSION::STRING} #{ruby_description}#{zlib_version}"
```
kaylareopelle, line 633: If everything is eventually getting mushed into a single string, would it be plausible to remove the space from the end of `ruby_description` and add it on line 633 before `#{zlib_version}`?

kaylareopelle (dismissed):
Wow! What a PR! Thanks for bringing us up to date @radar07  and @fallwith! Approved as-is, with one tiny suggestion you may choose to integrate. **Correction: I failed to notice the JRuby failures. We have a little more to do before crossing the finish line on this one**

lib/new_relic/agent/new_relic_service.rb
fallwith: Well... the short answer is that we can't do it that way because the setting of `ruby_description` is guarded by a conditional. If we moved the trailing space to the big string on the last line of the method, we'd run the risk of having 2 spaces showing up together between the agent version number and the zlib version number when `ruby_description` is `nil`.

But the long answer is that I think we might want to reevaluate the conditional that allows `ruby_version` to potentially be `nil`. It seems like the check would always evaluate to `true` for our currently supported Rubies.

kaylareopelle (dismissed):
👍

tannalynn (approved):
:tada:

tannalynn (approved):

fallwith:
> @fallwith it looks like this issue is complete(?) Can I move [#1198](https://github.com/newrelic/newrelic-ruby-agent/issues/1198) to 'Done' on our board?

@hannahramadan yep, this one is done. We held off on getting the code into dev until release v8.9.0 went out, but everything is in dev now and this one can be marked as done.

PR #1230, opened 2022-06-29
MiniTest: optionally fail fast

fallwith:
To have MiniTest "fail fast" and halt after encountering the first
failed test, set the "MT_FAIL_FAST" environment variable and then
proceed to run the tests as normal.

### DIFF
--- a/test/helpers/minitest.rb
+++ b/test/helpers/minitest.rb
@@ -25,3 +25,22 @@ class Minitest::Test
     super
   end
 end
+
+class MiniTest::Unit
+  def puke klass, meth, e
+    if fail_fast? && !e.is_a?(MiniTest::Skip)
+      puke_fast klass, meth, e
+    else
+      super klass, meth, e
+    end
+  end
+
+  def puke_fast klass, meth, e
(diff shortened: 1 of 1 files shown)


tannalynn (commented):
Will this change apply to just one Minitest run? Like with multiverse, would one suite failing stop the rest of the suites after that? I see that its using an env variable, but I wasn't sure how this is working

fallwith:
> Will this change apply to just one Minitest run? Like with multiverse, would one suite failing stop the rest of the suites after that? I see that its using an env variable, but I wasn't sure how this is working

Great question. While I had intended this to only be used for local dev for now, I expected it to work for both the unit tests and the multiverse tests. But it seems that it currently only works for the unit tests.

I still get value from it as is, but I'll look into a solution that works across both sets of tests.

tannalynn:
> > Will this change apply to just one Minitest run? Like with multiverse, would one suite failing stop the rest of the suites after that? I see that its using an env variable, but I wasn't sure how this is working
> 
> Great question. While I had intended this to only be used for local dev for now, I expected it to work for both the unit tests and the multiverse tests. But it seems that it currently only works for the unit tests.
> 
> I still get value from it as is, but I'll look into a solution that works across both sets of tests.
That makes sense, thank you for explaining that!

tannalynn (approved):

PR #1252, opened 2022-07-09
perf tests: constant bugfix, README updates

fallwith:
* require 'constants' so that the new use of `EMPTY_ARRAY` will work
* update `README.md`, especially to make it clear that the tests have
  their own `Gemfile`.

### DIFF
--- a/test/performance/README.md
+++ b/test/performance/README.md
@@ -15,45 +15,57 @@ There are two main goals driving the development of this framework:
 
 ### Invoking via the runner directly
 
-More advanced options can be specified by invoking the runner script directly.
-See `./test/performance/script/runner -h` for a full list of options.
+The performance testing specific content lives in the `test/performance`
+directory beneath the repository's root. That directory includes its own
+`Gemfile` which specifies dependencies needed for testing.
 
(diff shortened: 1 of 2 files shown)


test/performance/README.md
```diff
@@ -15,45 +15,57 @@ There are two main goals driving the development of this framework:
...
+#### Initial setup
 
+```shell
+cd test/performance
+bundle install
 ```
-$ cd test/performance
-$ script/runner -l
+
+#### Running tests
```
kaylareopelle, line 29: This is a bit unrelated to the other updates, but I wonder if it may make sense to add something our team's new preference for running the tests using docker for enhanced consistency?

kaylareopelle (dismissed):
Approved, as-is! Thanks for fixing these things. One comment you can address if you'd like.

test/performance/README.md
fallwith: Good idea. Added Docker usage documentation with [f2dc766](https://github.com/newrelic/newrelic-ruby-agent/pull/1252/commits/f2dc766ffbf36d51ab8726271ae9ada8d47c5288)

kaylareopelle (approved):
Awesome! Thanks for adding the 🐳 section!

PR #1255, opened 2022-07-14
Remove gRPC Client ResponseWrapper

### LINKED ISSUE #1240: gRPC: Determine what to do with response data to effectively define a response wrapper
Given the binary data format, we can’t exactly match what we do with HTTP clients. Let’s review what other agents are doing.

See `lib/new_relic/agent/instrumentation/grpc/client/response_wrapper.rb` for the placeholder class.

kaylareopelle:
# Overview

ResponseWrapper boilerplate code has been removed as it is not necessary for gRPC Clients.

The gRPC Client instrumentation draft was based on the HTTP Client instrumentation. HTTP Clients return a response status code attribute in their ResponseWrapper objects. gRPC Clients return a 0/OK status in all cases except errors. The gRPC error codes and statuses are captured when `NewRelic::Agent::Tracer.capture_segment_error` is called on the segment. 

Closes #1240


tannalynn (approved):
🎉 Glad you figured out this wasn't needed!

fallwith (approved):
👍

Issue #1245, opened 2022-07-01
gRPC: Implement config file driven blocklist functionality


fallwith:
The gRPC client instrumentation currently uses a constant to avoid instrumenting certain client methods based on the host they are sending data to.

- We should determine if the target host is the best way to perform blocklisting certain client calls
- We should make the blocklist logic flexible and config file based
- We should determine how to accommodate exceptions or whatnot for internal use (don't assume all traffic is 8T based just because the target host is newrelic.com)
- We should determine a blocklist strategy for the instrumented server methods as well

Issue #1242, opened 2022-07-01
gRPC: Author complete unit test coverage for the new instrumentation code


fallwith:
As with other instrumented libraries, we should leverage our multiverse system, bring in the `grpc` gem via an `Envfile` file, and perform exhaustive unit test coverage of all new gRPC instrumentation.

kaylareopelle:
@fallwith - Is this to add tests to the code that was written as part of the spike? I would expect tests to be included in the code submitted for the other issues in this milestone.

fallwith:
> @fallwith - Is this to add tests to the code that was written as part of the spike? I would expect tests to be included in the code submitted for the other issues in this milestone.

I wanted to be flexible with an overall gRPC branch and require the tests for any merge to dev. But I like your idea here. Let's do that.

hannahramadan:
@kaylareopelle can I pair with you on this ticket?

kaylareopelle:
@hannahramadan - Sure! I'll put some time on the calendar.

kaylareopelle:
We've made a lot of progress on the test coverage for gRPC client code.

The following areas remain: 
* Create a helper test file to test `#cleaned_name`
* Write tests that exercise the following gRPC methods call our instrumentation: 
  * bidi_streaming
  * request_response
  * server_streaming
  * client_streaming 

I started consulting gRPC's specs and got a few ideas that involve building out a more robust gRPC client for the sake of the tests.

fallwith:
Resolved with the current state of the `grpc` branch

PR #1256, opened 2022-07-14
Add configuration for grpc blocklist

tannalynn:

# Overview
Adds the configuration option `instrumentation.grpc.host_denylist` as an array of regexes that the agent will use to ignore certain grpc connections. Currently is only set up for the grpc client. We have a few other configs that use `denylist`, so I opted to use that over blocklist for consistency. 
#1245

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1095,7 +1095,7 @@ If `true`, disables agent middleware for Sinatra. This middleware is responsible
           :description => 'Controls auto-instrumentation of gRPC clients at start up.  May be one of [auto|prepend|chain|disabled].'
         },
         :'instrumentation.grpc.host_denylist' => {
-          :default => ['tracing\.(staging-)?edge\.nr-data'],
+          :default => ['tracing\.(?:staging-)?edge\.nr-data'],
           :public => true,
           :type => Array,
           :allowed_from_server => false,
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1094,6 +1094,14 @@ If `true`, disables agent middleware for Sinatra. This middleware is responsible
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of gRPC clients at start up.  May be one of [auto|prepend|chain|disabled].'
         },
+        :'instrumentation.grpc.host_denylist' => {
+          :default => ['tracing\.(staging-)?edge\.nr-data'],
+          :public => true,
+          :type => Array,
+          :allowed_from_server => false,
+          :transform => DefaultSource.method(:convert_to_regexp_list),
+          :description => 'Define grpc connections you want the agent to ignore by specifying a list of patterns matching the URI of the hosts to be ignored'
+        },
         :'instrumentation.grpc_server' => {
           :default => instrumentation_value_of(:disable_grpc_server),
           :documentation_default => 'auto',
--- a/lib/new_relic/agent/instrumentation/grpc/client/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/grpc/client/instrumentation.rb
@@ -70,14 +70,14 @@ module NewRelic
             return do_trace unless do_trace.nil?
 
             host ||= @host
-            return false unless host && !host_denylisted(host)
+            return false unless host && !host_denylisted?(host)
 
             true
           end
 
-          def host_denylisted(host)
+          def host_denylisted?(host)
             NewRelic::Agent.config[:'instrumentation.grpc.host_denylist'].any? do |regex|
-              host.match(regex)
+              host.match?(regex)
             end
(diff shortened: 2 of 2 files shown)


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1094,6 +1094,14 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of gRPC clients at start up.  May be one of [auto|prepend|chain|disabled].'
         },
+        :'instrumentation.grpc.host_denylist' => {
+          :default => ['tracing\.(staging-)?edge\.nr-data'],
+          :public => true,
+          :type => Array,
+          :allowed_from_server => false,
```
fallwith, line 1101: Out of curiosity, what goes into the true/false determination for allowing a value to be set from the server?

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1094,6 +1094,14 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of gRPC clients at start up.  May be one of [auto|prepend|chain|disabled].'
         },
+        :'instrumentation.grpc.host_denylist' => {
+          :default => ['tracing\.(staging-)?edge\.nr-data'],
```
fallwith, line 1098: Given that we are not intending to ever perform a capture with the regex, `(staging-)` should be `(?:staging)` to make our non-capturing intent clear to Ruby (and future human observers).

lib/new_relic/agent/instrumentation/grpc/client/instrumentation.rb
```diff
@@ -73,10 +70,16 @@ def trace_with_newrelic?(host = nil)
...
             host ||= @host
-            return false unless host && !host.match?(HOST_BLOCKLIST_PATTERN)
+            return false unless host && !host_denylisted(host)
 
             true
           end
+
+          def host_denylisted(host)
+            NewRelic::Agent.config[:'instrumentation.grpc.host_denylist'].any? do |regex|
+              host.match(regex)
```
fallwith, line 80: I think `#match?` would be better than `#match` here. Giving `#any?` a simple boolean seems better than a match object, given that we don't actually perform any captures or anything that the match object would be used for.

fallwith (changes requested):

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: We should document this answer in the file! I've often had that question. My understanding is that it should be allowed_from_server if it can be set via server-side config, or if the value could be manipulated by the collector or another out of agent response.

lib/new_relic/agent/instrumentation/grpc/client/instrumentation.rb
```diff
@@ -73,10 +70,16 @@ def trace_with_newrelic?(host = nil)
             return do_trace unless do_trace.nil?
 
             host ||= @host
-            return false unless host && !host.match?(HOST_BLOCKLIST_PATTERN)
+            return false unless host && !host_denylisted(host)
 
             true
           end
+
+          def host_denylisted(host)
```
kaylareopelle, line 78: Since we're aiming to return a boolean here, would you mind adding a question mark to the end of the method name?

kaylareopelle (commented):
Nice work, Tanna! I made a few small comments. Even though I'm writing tests for the client in #1242, what do you think about writing a test for these changes as part of this PR?

lib/new_relic/agent/configuration/default_source.rb
tannalynn: I think the only time we would set this to true would be based on agent specs probably. We would set it to true if we knew that the collector was going to be sending us info for this setting, but most settings the collector doesn't care about or give us back things to use.

lib/new_relic/agent/instrumentation/grpc/client/instrumentation.rb
tannalynn: resolved: b05ee085cfd6e726be0285827c325be5a5aca9cd

lib/new_relic/agent/configuration/default_source.rb
tannalynn: resolved: b05ee085cfd6e726be0285827c325be5a5aca9cd

lib/new_relic/agent/instrumentation/grpc/client/instrumentation.rb
tannalynn: resolved: 4040d521ecd26241c7c92c96cfa89d96cc6ed147

fallwith (approved):
👍

PR #1258, opened 2022-07-15
Party like it s 2016

fallwith:
Ensure that all Ruby files instruct Ruby to freeze string literals by
default


.rubocop.yml
```diff
@@ -1361,9 +1361,6 @@ Style/FormatString:
 Style/FormatStringToken:
   Enabled: false
 
-Style/FrozenStringLiteralComment:
-  Enabled: false
-
```
fallwith, line 1366: `frozen_string_literal` was introduced on 2015-12-15 and the RuboCop style cop to enforce its usage with a `true` value was introduced soon after in early 2016. By simply no longer disabling the cop, we'll benefit from the default behavior of it enforcing the use of `frozen_string_literal: true`. Party like it's 2016!

test/new_relic/agent/database_test.rb
```diff
@@ -182,9 +183,9 @@ def test_explain_sql_one_select_with_pg_explain_string_result
     config = {:adapter => 'postgresql'}
     sql = 'select count(id) from blogs limit 1'
 
-    plan = "Limit  (cost=11.75..11.76 rows=1 width=4)
+    plan = String.new("Limit  (cost=11.75..11.76 rows=1 width=4)
```
fallwith, line 186: This string gets mutated, needs `String.new`

test/new_relic/agent/database_test.rb
```diff
@@ -453,7 +454,7 @@ def test_capture_query_long_query
     assert_equal('a' * (NewRelic::Agent::Database::MAX_QUERY_LENGTH - 3) + '...', truncated_query)
   end
 
-  INVALID_UTF8_STRING = (''.respond_to?(:force_encoding) ? "select \x80".force_encoding('UTF-8') : "select \x80")
+  INVALID_UTF8_STRING = String.new("select \x80").force_encoding('UTF-8')
```
fallwith, line 457: This string gets mutated, needs `String.new`. NOTE: we no longer test with rubies that don't respond to `#force_encoding`.

kaylareopelle (approved):
Thanks for taking care of this, James!

PR #1259, opened 2022-07-15
minitest: fix fail fast implementation

fallwith:
for the custom minitest fail fast monkeypatch, use prepend to allow
`super` to work properly


kaylareopelle (approved):
👍

tannalynn (approved):
👍

PR #1265, opened 2022-07-18
Execute @@after_run check only for Ruby >= 2.7

kaylareopelle:
## Overview

This block was being run on versions of Ruby that do not generate SimpleCov coverage results, causing an error to be raised on the CI run. This if statement should prevent the errors from being raised.

Also, the rescue has been limited to NameError instead of globally rescuing any exception.

### DIFF
--- a/test/multiverse/lib/multiverse/suite.rb
+++ b/test/multiverse/lib/multiverse/suite.rb
@@ -538,10 +538,16 @@ module Multiverse
       end
 
       load(@after_file) if @after_file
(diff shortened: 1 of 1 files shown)


kaylareopelle:
Warnings are still raised for the non-SimpleCov checks, but this PR brings us one step closer. I haven't figured out how to run steps in a workflow conditionally by Ruby version yet.

fallwith (approved):

Issue #1264, opened 2022-07-18
newrelic_rpm slows down requests 10x when used with graphql dataloader


bmulholland:
See https://github.com/rmosolgo/graphql-ruby/issues/3945 for some info about the error

Note the investigation at https://github.com/rmosolgo/graphql-ruby/issues/3945#issuecomment-1186232935 which shows where in newrelic_rpm the slowdown occurs

And the repro at https://github.com/rmosolgo/graphql-ruby/issues/3945#issuecomment-1186210895

Thanks to @hkdnet for moving this along

## For Maintainers Only or Hero Triaging this bug
*Suggested Priority (P1,P2,P3,P4,P5):*
*Suggested T-Shirt size (S, M, L, XL, Unknown):*

fallwith:
@bmulholland @hkdnet I opened up `lib/new_relic/agent/range_extensions.rb` and rewrote `merge` and `merge_append` like so:

```ruby
  # CHANGES: 1. instead of instantiating new Integer objects with 'range_min'
  #             and 'range_max' variables, just use a one-liner to build a new
  #             Range object directly (object allocation savings)
  #          2. remove the intersects? check and assume that it has already been
  #             performed elsewhere
  def merge(r1, r2)
    (r1.begin < r2.begin ? r1.begin : r2.begin)..(r1.end > r2.end ? r1.end : r2.end)
  end

  # CHANGES: 1. instead of using Array#each_with_index and finding the first
  #             range with overlap, use Array#index and get the index for the
  #             range instead
  #          2. instead of calling merge on every range, perform the intersects?
  #             check in the Array#index block itself
  def merge_or_append(range, ranges)
    i = ranges.index { |r| intersects?(r, range) }
    (ranges[i] = merge(ranges[i], range) and return ranges) if i
    ranges << range
  end
```

On my development machine with Ruby 3.1.2, these two changes produce about a 27% speed boost.

It appears that the remaining time is being spent inside the `intersects?` method. That method checks to see if 2 ranges have an intersection. For example:

- `1..3` and `2..4` => true, these 2 intersect
- `1..3` and `4..5` => false, these 2 do not intersect

ActiveSupport's core extensions for `Range` define an `#overlaps?` method which uses `#cover?` on each of 2 ranges to see if there's overlap. Our `intersects?` method is instead using `#include?`. `#include?` ought to be slower in general, but with the specific code in question, it seems to be slightly faster than `#cover?`.

[...]

PR #1266, opened 2022-07-19
Avoid calling merge_or_append to speed up 

hkdnet:

I've read them 👍 

# Overview

newrelic_rpm gets slow when `NewRelic::Agent::RangeExtensions.merge_or_append` is called so many times.

https://github.com/newrelic/newrelic-ruby-agent/issues/1264

`newrelic_rpm@8.9.0` slows the response, 446ms -> 33670ms. (See https://github.com/hkdnet/graphql-ruby-nr/tree/ccefc10dcae5a1ed90eaba4cc6f8ea3bcb6e38a4#result )
But the current patch slows the same response, 403ms -> 626ms. (See https://github.com/hkdnet/graphql-ruby-nr/tree/perf-test-for-lazy-merge#result )

Though the result is not stable (it depends on my machine's status), it shows a significant improvement.

Submitter Checklist:
    - I think we don't need it.

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/range_extensions.rb
+++ b/lib/new_relic/agent/range_extensions.rb
@@ -12,26 +12,6 @@ module NewRelic
         r1.include?(r2.begin) || r2.include?(r1.begin)
       end
 
-      def merge(r1, r2)
-        return unless intersects?(r1, r2)
-        range_min = r1.begin < r2.begin ? r1.begin : r2.begin
-        range_max = r1.end > r2.end ? r1.end : r2.end
-        range_min..range_max
-      end
-
-      # Takes an array of ranges and a range which it will
-      # merge into an existing range if they intersect, otherwise
-      # it will append this range to the end the array.
-      def merge_or_append(range, ranges)
-        ranges.each_with_index do |r, i|
-          if merged = merge(r, range)
-            ranges[i] = merged
-            return ranges
-          end
-        end
-        ranges.push(range)
-      end
-
       # Computes the amount of overlap between range and an array of ranges.
       # For efficiency, it assumes that range intersects with each of the
       # ranges in the ranges array.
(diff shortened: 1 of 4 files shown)


hkdnet:
Oops, some tests failed. I will check them soon. Let me convert this PR to a draft

fallwith:
Hi @hkdnet! Thank you so much for your efforts to get the `graphql` and `newrelic_rpm` gems to better harmonize. We love GraphQL and we want to provide GraphQL adopters with useful information without incurring performance overhead.

The `newrelic_rpm` gem keeps track of several Range objects in order to help with multiple concurrently running traces. In particular, the [test_total_time_metrics_async_sync_children_non_web](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/test/new_relic/agent/transaction/tracing_test.rb#L810) has comments with an illustration to explain the desired functionality.

It seems like the `merge_or_append` method is the main cause of performance problems for `graphql`. If that method goes away, it seems that some core `newrelic_rpm` trace functionality stops working and some tests fail.

Thank you very much for moving your PR to draft status and looking at the tests. We will also continue looking at this problem.

One current hypothesis that we have is that `graphql`'s use of [trace_execution_scoped](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/new_relic_tracing.rb#L33) is possibly getting called too frequently, especially when Dataloader is involved. If that method is called too frequently, it could cause there to be too many Range objects concurrently being tracked by `newrelic_rpm` and performance problems take place.

We definitely want to see what can be done to improve the performance of handling Range objects in `newrelic_rpm`, and if possible we would also like some help from the `graphql` community in evaluating ways to possibly improve the use of `trace_execution_scoped` when Dataloader is engaged.

fallwith:
Pinging @bmulholland just to make sure you're in the loop with the comments and code present in this PR designed to address #1264.

hkdnet:
https://github.com/newrelic/newrelic-ruby-agent/pull/1274 covers this 👋

PR #1268, opened 2022-07-19
Add create_gemfiles calls

kaylareopelle:
# Overview

This PR adds `Multiverse::Envfile#create_gemfiles` calls to some Envfiles that adopted the new style, but did not contain the call. `Multiverse::Envfile#create_gemfiles` calls `Multiverse::Envfile#gemfile` to turn the Heredocs and versions passed to the method into Gemfiles for testing. Also refactored `twiddle_wakka` to `add_version` to use more dynamically in more places.


.github/workflows/ci.yml
```diff
@@ -149,7 +149,7 @@ jobs:
       fail-fast: false
       matrix:
         multiverse: [agent, background, background_2, database, frameworks, httpclients, httpclients_2, rails, rest]
-        ruby-version: [2.2.10, 3.1.2]
+        ruby-version: [2.2.10, 2.3.8, 2.4.10, 2.5.9, 2.6.10, 2.7.6, 3.0.4, 3.1.2, 3.2.0-preview1]
```
fallwith, line 152: please revert this change before merging (this is the very old `ci.yml` syntax before we split into per-PR and cron)

test/multiverse/lib/multiverse/envfile.rb
```diff
@@ -112,6 +112,11 @@ def add_twiddle_wakka(version)
       ", '~> #{version}'"
     end
 
+    def add_version(version)
```
fallwith, line 115: This is so close to `add_twiddle_waka` that I wonder if we ought to just combine them:

```ruby
def add_version(version, twiddle_wakka: true)
  return unless version  # alternative way to write this guard clause

  ", '#{'~> ' if twiddle_wakka}#{version}'"
end
```

then all existing `add_twiddle_waka` calls could be renamed to `add_version` without any other changes.

test/multiverse/suites/active_record/Envfile
```diff
@@ -18,14 +18,16 @@ RAILS_VERSIONS = [
   ['3.2.0', 2.2, 2.2]
 ]
 
-def jbdc_version(activerecord_version)
+def jdbc_version(activerecord_version)
   return unless RUBY_PLATFORM.eql?('java') || activerecord_version.nil?
-  activerecord_version.slice('1')
+  # remove the '.' between the major and minor versions
+  activerecord_version_jdbc = activerecord_version.sub('.', '')
+  activerecord_version_jdbc
```
fallwith, line 25: No need for a variable here, just a line with `activerecord_version.sub('.', '')` by itself will work

test/multiverse/suites/unicorn/Envfile
```diff
@@ -20,3 +20,6 @@ def gem_list(unicorn_version = nil)
     gem 'rack'
   RB
 end
+
+create_gemfiles(SUPPORTED_UNICORN_VERSIONS, gem_list)
+# vim: ft=ruby
```
fallwith, line 25: Please remove the Vim comment to match the #1289 changes

fallwith (dismissed):

fallwith (approved):

PR #1274, opened 2022-07-21
Segments: rework child timing range merging

fallwith:
For a given segment's children, each child's start and finish time pair
needs to be checked against every other child's pair of the same to
discover overlap and merge the timings when overlap is found.

Searching an array of ranges for an entry that overlaps with a given
range can be quite performance heavy, even when directly leaning on
`Range#cover?`. To avoid the performance penalty as much as possible, do
not incur it every time a new child's timings pair is seen. Instead,
keep the children timing pairs as 2 item arrays for as long as possible
without looking for overlap or merging any values. Then, only when it
comes time to record the timing, incur the hit a single time (and
memoize the result) to produce an array of ranges whose overlapping
entries have all been merged.


tannalynn (commented):
This looks great! I really appreciate the detailed comments you added to help explain this complicated functionality! 
Could you add a changelog entry for this?

fallwith:
> Could you add a changelog entry for this?

Absolutely. Added with [c187c4f](https://github.com/newrelic/newrelic-ruby-agent/pull/1274/commits/c187c4fc03c498f307d7316698a9ca52b9a4c073)

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 
   ## v8.10.0
 
+  * **Performance: Rework timing range overlap calculations for multiple transaction segments
```
tannalynn, line 6: The formatting here is missing the end **
```suggestion
  * **Performance: Rework timing range overlap calculations for multiple transaction segments**
```

fallwith: addressed with [e546d82](https://github.com/newrelic/newrelic-ruby-agent/pull/1274/commits/e546d82fe271429682a131d477f9a820e5dc60e5), thanks!

tannalynn (approved):
🎉 Thank you!

PR #1277, opened 2022-07-22
More gRPC Client Tests! 

kaylareopelle:
## Overview

* Add `dt_enabled?` helper to DistributedTracer class 
* Add two new dt-related assertions: 
  *  `assert_newrelic_metdata_present`
  * assert_distributed_tracing_payload_created_for_transaction(transaction)'
* Refactor some of the repeated code in our tests into constants and helpers
* Break out tests for: 
  * error segment 
  * DT headers
  * span attributes
  * metric name
  * possibly others that I'm forgetting

### DIFF
--- a/lib/new_relic/agent/transaction/distributed_tracer.rb
+++ b/lib/new_relic/agent/transaction/distributed_tracer.rb
@@ -75,6 +75,7 @@ module NewRelic
 
         def insert_headers(headers)
           return unless NewRelic::Agent.agent.connected?
+          return unless dt_enabled?
           insert_trace_context_header(headers)
           insert_distributed_trace_header(headers)
           insert_cross_app_header(headers)
@@ -91,7 +92,7 @@ module NewRelic
         end
 
         def assign_intrinsics
-          if Agent.config[:'distributed_tracing.enabled']
+          if dt_enabled?
             DistributedTraceAttributes.copy_to_attributes(transaction.payload, transaction.attributes)
           elsif is_cross_app?
             assign_cross_app_intrinsics
@@ -99,7 +100,7 @@ module NewRelic
         end
 
         def insert_distributed_trace_header(headers)
-          return unless Agent.config[:'distributed_tracing.enabled']
+          return unless dt_enabled?
           return if Agent.config[:'exclude_newrelic_header']
           payload = create_distributed_trace_payload
           headers[NewRelic::NEWRELIC_KEY] = payload.http_safe if payload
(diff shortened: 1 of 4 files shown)


lib/new_relic/agent/transaction/distributed_tracer.rb
```diff
@@ -75,6 +75,7 @@ def log_request_headers(headers, direction = "OUTGOING")
 
         def insert_headers(headers)
           return unless NewRelic::Agent.agent.connected?
+          return unless dt_enabled?
```
kaylareopelle, line 78: This line was not present previously, so it could change the behavior of the other operations that call #insert_headers. I'm a bit shocked that we would still insert headers on something if distributed tracing isn't enabled, so I would advocate for keeping this new LOC, however this is open to discussion!

test/multiverse/suites/grpc/grpc_client_test.rb
```diff
@@ -9,107 +9,184 @@
...
 
-  def setup
+  # Helpers
+  TRACE_WITH_NEWRELIC = :@trace_with_newrelic
+  HOST = '0.0.0.0:5000'
+  CHANNEL = :this_channel_is_insecure
+  METHOD = 'routeguide.RouteGuide/GetFeature'
+
+  def exception_class
+    GRPC::Unknown
```
fallwith, line 19: This one uses `GRPC` and `basic_grpc_client` uses `::GRPC`. I think they should both use the same approach.

fallwith (changes requested):

test/agent_helper.rb
```diff
@@ -295,6 +295,15 @@ def assert_metrics_recorded_exclusive(expected, options = {})
   assert_equal(0, unexpected_metrics.size, "Found unexpected metrics: #{format_metric_spec_list(unexpected_metrics)}")
 end
 
+def assert_newrelic_metdata_present(metadata)
+  assert_includes metadata.keys, 'newrelic'
```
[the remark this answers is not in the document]
kaylareopelle: Great suggestion!

test/multiverse/suites/grpc/grpc_client_test.rb
kaylareopelle: Great catch!

lib/new_relic/agent/transaction/distributed_tracer.rb
kaylareopelle: I think that's a great point. I've removed the line and will put this on our list for a future innovation meeting.

fallwith (approved):
👍

fallwith (approved):

PR #1283, opened 2022-07-26
gRPC updates

fallwith:
* share `host_denylisted?` between client and server
* fix server instrumentation chaining (2 separate classes are having
  their methods chained)
* update server host and port parsing
* update server hostname denylist checking
* add helper tests
* add server tests
* add the host denylist param to the config file


lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
```diff
@@ -28,7 +29,7 @@ def handle_with_tracing(active_call, mth, inter_ctx)
             )
             yield
           ensure
-            finishable.finish
+            finishable.finish if finishable
```
kaylareopelle, line 32: 👍

lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
```diff
@@ -38,18 +39,23 @@ def add_http2_port_with_tracing(*args)
...
 
-          def set_host_and_port_on_server_instace(host_string)
+          def host_and_port_from_host_string(host_string)
+            return [nil, nil] unless host_string
+
             info = host_string.split(':')
 
-            # TODO:
-            # raise if info.size != 2
+            return [nil, nil] unless info.size == 2
```
kaylareopelle, line 52: Since `[nil, nil]` is referenced twice, would it make sense to set it as a variable?

test/multiverse/suites/grpc/grpc_client_test.rb
```diff
@@ -6,7 +6,7 @@
 require 'grpc'
 require 'newrelic_rpm'
 
-class GrpcTest < Minitest::Test
+class GrpcClientTest < Minitest::Test
```
kaylareopelle, line 9: nice!

lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
fallwith: Great question. The arrays are both hidden behind conditionals that we currently do not expect to evaluate as true very often. My hope was that the happy path would not bear the cost of allocating the object array at all. Switching to a constant would have all paths always defining the array.

But now that I look over this code again, I think we can refactor things to just perform a `return` (without params). For some reason I wanted to explicitly set the two instance vars even if they had `nil` as a value. But Ruby doesn't complain when you ask for a non-existent instance variable and just gives you back the same `nil`, so I think we can clean this up a bit. Will update!

newrelic.yml
```diff
@@ -413,6 +413,14 @@ common: &default_settings
 # May be one of [auto|prepend|chain|disabled].
 # instrumentation.grpc_server: auto
 
+# Specifies a list of hostname patterns separated by commas
+# for gRPC hosts to not trace requests for
+# By default, New Relic's own infinite tracing servers are
+# listed with this pattern: 'tracing\.(?:staging-)?edge\.nr-data'
+# It is recommended that that pattern be kept and added to if
+# additional host patterns need to be defined.
+# instrumentation.grpc.host_denylist: ""
```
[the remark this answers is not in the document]
fallwith: @kaylareopelle idea: Let's put the 'nr-data' pattern into a constant that can not be altered. Then we permit the configuration to specify additional ignore patterns. By default, there are no additional patterns. When present, the additional patterns will be added to the list that always contains at least the 'nr-data' pattern. This would simplify the config and documentation and prevent any unwanted unblocking of infinite tracing monitoring.

Thoughts?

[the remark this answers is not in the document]
kaylareopelle: I like that a lot. If we like how this pans out, there may be other settings that could benefit from similar measures.

lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
fallwith: Addressed with [5f30f9e](https://github.com/newrelic/newrelic-ruby-agent/pull/1283/commits/5f30f9e0794726d7b2bad11287f7874c678a2fa5)

newrelic.yml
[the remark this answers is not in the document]
fallwith: Addressed with [5f30f9e](https://github.com/newrelic/newrelic-ruby-agent/pull/1283/commits/5f30f9e0794726d7b2bad11287f7874c678a2fa5)

newrelic.yml
```diff
@@ -23,628 +23,630 @@ common: &default_settings
   # debug.
   log_level: info
 
-# All of the following configuration options are optional. Review them, and
-# uncomment or edit them if they appear relevant to your application needs.
+  # All of the following configuration options are optional. Review them, and
```
kaylareopelle, line 26: I think this would be much shorter if grpc was merged with dev

test/multiverse/lib/multiverse/suite.rb
```diff
@@ -265,10 +265,11 @@ def generate_gemfile(gemfile_text, env_index, local = true)
         f.puts minitest_line unless gemfile_text =~ /^\s*gem .minitest[^_]./
         f.puts "gem 'rake'" unless gemfile_text =~ /^\s*gem .rake[^_]./ || suite == 'rake'
 
-        f.puts "gem 'mocha', '~> 1.9.0', :require => false"
+        f.puts "gem 'mocha', '~> 1.9.0', require: false"
+        f.puts "gem 'minitest-stub-const', '~> 0.6', require: false"
 
         if debug
-          f.puts "gem 'pry', '~> 0.10.0'"
+          f.puts "gem 'pry', '~> 0.14'"
```
kaylareopelle, line 272: yay, upgrades!

lib/new_relic/agent/instrumentation/grpc/helper.rb
```diff
@@ -16,9 +18,11 @@ def cleaned_method(method)
...
 
           def host_denylisted?(host)
-            NewRelic::Agent.config[:'instrumentation.grpc.host_denylist'].any? do |regex|
-              host.match?(regex)
-            end
+            ignore_patterns.any? { |regex| host.match?(regex) }
+          end
+
+          def ignore_patterns
+            ([NR_8T_HOST_PATTERN] + NewRelic::Agent.config[:'instrumentation.grpc.host_denylist']).freeze
```
kaylareopelle, line 25: I like what you've done with this file!

newrelic.yml
fallwith: Good call. Dev merged in with [d7756ab](https://github.com/newrelic/newrelic-ruby-agent/pull/1283/commits/d7756aba505709c581070f7ed16dc5eb54c3684d)

lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
```diff
@@ -38,20 +39,26 @@ def add_http2_port_with_tracing(*args)
...
+
+            info = host_string.split(':').freeze
+            return unless info.size == 2
+
+            info
+          end
 
-            # TODO:
-            # raise if info.size != 2
+          def set_host_and_port_on_server_instace(host_string)
```
kaylareopelle, line 56: There's a slight misspelling: 
```suggestion
          def set_host_and_port_on_server_instance(host_string)
```

kaylareopelle: The call on line 36 needs to be updated too.

kaylareopelle (dismissed):
There's a tiny misspelling in one of the method names. After that gets resolved, I think this is good!

lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
fallwith: Suggested change committed, thank you! 36 updated with [7669e0b](https://github.com/newrelic/newrelic-ruby-agent/pull/1283/commits/7669e0b43f028d5f6f2f03fd5bd44bc2459eda30)

test/multiverse/suites/grpc/helper_test.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  class HelpedClass
+    include NewRelic::Agent::Instrumentation::GRPC::Helper
+  end
+
+  def unwanted_host_patterns
+    [/unwanted/.freeze].freeze
+  end
+
+  def test_class
```
hannahramadan, line 19: Just wondering—I know MiniTest looks at methods starting with `test_`, and even though this doesn't look like its testing anything, its still cool to name like this? I also could be misunderstanding this.

fallwith: That's a great point, @hannahramadan - I don't want MiniTest to see that helper method as a test. With [3f53ed2](https://github.com/newrelic/newrelic-ruby-agent/pull/1283/commits/3f53ed2a2a820052276411d6d60c679bc4845f24) I've renamed the method to `helped_class`.

PR #1286, opened 2022-07-27
gRPC server: distributed tracing fixes

fallwith:
- fix distributed tracing by flipping the order of operations and having the
  code that creates a transaction fire first before the code that processes
  distributed tracing data, which requires a transaction
- in the gRPC server instrumentation, move the distributed tracing logic out
  into a dedicated method
- in the gRPC server instrumentation, do not check if distributed tracing is
  enabled before invoking a distributed tracing method, as that method will
  itself perform the very same check
- send the entire metadata headers hash (not just a single value) over to the
  distributed tracing method, which will short circuit if not given a hash
- in the gRPC server instrumentation, rescue, notice, and re-raise exceptions
- in #handle_with_tracing, prefix the last argument with an underscore to denote
  that it is not used by New Relic (only by the actual underlying gRPC server
  handling process)
- add tests for all of these changes and everything else on the gRPC server
  instrumentation side to bring coverage up to 100% of branches

### DIFF
--- a/lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
@@ -18,13 +18,14 @@ module NewRelic
 
           def handle_with_tracing(active_call, mth, _inter_ctx)
             return yield unless trace_with_newrelic?
+            metadata = metadata_for_call(active_call)
 
             finishable = NewRelic::Agent::Tracer.start_transaction_or_segment(
               name: mth.original_name,
               category: :web,
-              options: server_options(active_call.metadata)
+              options: server_options(metadata)
             )
-            process_distributed_tracing_headers(active_call)
+            process_distributed_tracing_headers(metadata)
 
             begin
               yield
@@ -48,10 +49,16 @@ module NewRelic
 
           private
 
-          def process_distributed_tracing_headers(active_call)
-            return unless active_call.metadata
+          def metadata_for_call(active_call)
+            return NewRelic::EMPTY_HASH unless active_call && active_call.metadata
 
-            ::NewRelic::Agent::DistributedTracing::accept_distributed_trace_headers(active_call.metadata, 'Other')
+            active_call.metadata
+          end
+
+          def process_distributed_tracing_headers(metadata)
+            return unless metadata && !metadata.empty?
+
+            ::NewRelic::Agent::DistributedTracing::accept_distributed_trace_headers(metadata, 'Other')
           end
 
           def host_and_port_from_host_string(host_string)
--- a/lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
@@ -16,18 +16,22 @@ module NewRelic
           INSTANCE_VAR_PORT = :@port_nr
           INSTANCE_VAR_METHOD = :@method_nr
 
-          def handle_with_tracing(active_call, mth, inter_ctx)
+          def handle_with_tracing(active_call, mth, _inter_ctx)
             return yield unless trace_with_newrelic?
-
-            trace_headers = active_call.metadata.delete(NewRelic::NEWRELIC_KEY)
-            ::NewRelic::Agent::DistributedTracing::accept_distributed_trace_headers(trace_headers, 'Other') if ::NewRelic::Agent.config[:'distributed_tracing.enabled']
 
             finishable = NewRelic::Agent::Tracer.start_transaction_or_segment(
               name: mth.original_name,
               category: :web,
               options: server_options(active_call.metadata)
             )
-            yield
+            process_distributed_tracing_headers(active_call)
+
+            begin
+              yield
+            rescue => e
+              NewRelic::Agent.notice_error(e)
+              raise
+            end
           ensure
             finishable.finish if finishable
           end
@@ -43,6 +47,12 @@ module NewRelic
           end
 
           private
+
+          def process_distributed_tracing_headers(active_call)
+            return unless active_call.metadata
+
+            ::NewRelic::Agent::DistributedTracing::accept_distributed_trace_headers(active_call.metadata, 'Other')
+          end
 
           def host_and_port_from_host_string(host_string)
             return unless host_string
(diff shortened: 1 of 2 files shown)


lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
```diff
@@ -16,18 +16,22 @@ module Server
...
             )
-            yield
+            process_distributed_tracing_headers(active_call)
+
+            begin
+              yield
+            rescue => e
+              NewRelic::Agent.notice_error(e)
+              raise
+            end
```
kaylareopelle, line 34: Not something to solve in this issue, but while looking into these lines, I noticed we inconsistently raise the error after calling `NewRelic::Agent.notice_error`. 

These files don't raise:
* lib/new_relic/agent/instrumentation/action_cable_subscriber.rb
* lib/new_relic/agent/instrumentation/sidekiq.rb
* lib/new_relic/agent/instrumentation/sinatra/instrumentation.rb

These files do raise: 
* lib/new_relic/agent/instrumentation/controller_instrumentation.rb
* lib/new_relic/agent/instrumentation/middleware_tracing.rb
* lib/new_relic/agent/instrumentation/active_support_logger/instrumentation.rb
* lib/new_relic/agent/instrumentation/tilt/instrumentation.rb 

A method with a similar purpose, `NewRelic::Agent::Tracer.capture_segment_error`, includes a raise in its definition. Maybe we should follow that pattern unless we truly don't want the raise to occur in some instances? 

I'll transfer this to an issue for further discussion.

lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
```diff
@@ -44,6 +48,12 @@ def run_with_tracing(*args)
 
           private
 
+          def process_distributed_tracing_headers(active_call)
+            return unless active_call.metadata
```
kaylareopelle, line 52: Out of curiosity, have you found any cases where active_call.metadata was absent?

lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
```diff
@@ -44,6 +48,12 @@ def run_with_tracing(*args)
...
+
+            ::NewRelic::Agent::DistributedTracing::accept_distributed_trace_headers(active_call.metadata, 'Other')
```
kaylareopelle, line 54: I'm not familiar with the syntax of using `::` before a method. Is this another way to call a class method? I noticed this is also used in our Sidekiq instrumentation.

kaylareopelle (approved):

lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
kaylareopelle: Issue: https://github.com/newrelic/newrelic-ruby-agent/issues/1290

fallwith: We should definitely follow a consistent pattern. 👍  For the 3 specific examples you gave that don't raise, I have confirmed that it is appropriate for them to not do so. It all comes down to whether or not we are rescuing an exception that would otherwise be left for the instrumented library to handle or not. If we steal an exception like that, we need to re-raise. If we're only operating on a copy of the exception, then no need to re-raise.

Here are specific notes on each of the files you listed:

`lib/new_relic/agent/instrumentation/action_cable_subscriber.rb`
  ^-- this one is finding an exception object in a payload and not actually performing a rescue, so there's no need to re-raise

`lib/new_relic/agent/instrumentation/sidekiq.rb`
  ^-- this one involves a dedicated agent specific error handler that has been registered with Sidekiq. So we're operating on a copy of the underlying exception. The actual exception itself is being handled by a primary Sidekiq error handler and there's no need for our handler to re-raise.

`lib/new_relic/agent/instrumentation/sinatra/instrumentation.rb`
  ^-- this one is operating on a recorded copy of an exception having occured that is stored in an environment hash. We grab the error from the hash and report on it, but we're not actually involved in rescuing the error and don't need to re-raise.

lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
fallwith: I haven't. Given that the underlying gRPC source code (an obvious port) isn't particularly strict about type checking or input validation, I wanted to be as cautious as possible. But now that you mention it, I see that we do not check `active_call` itself for `nil`, which is far more likely I'd imagine, and we also don't perform this same check when we're spawning a `finishable` segment. I'll make some updates.

lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
fallwith: Oh neat, I hadn't noticed that. This syntax was copied from `lib/new_relic/agent/instrumentation/sidekiq.rb` so that's why they're aligned.

`DistributedTracing` is a module that extends `self`, so the `accept_distributed_trace_headers` can be directly invoked like that. But as far as I know, it's equivalent to just using a `.` with `DistributedTracing.accept_distributed_trace_headers`. I wonder if there's a RuboCop cop for preferring one style over the other.

lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
kaylareopelle: Thanks for investigating!

lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
kaylareopelle: Taking a cursory look at the RuboCop docs didn't reveal such a cop to me. I searched using `class method`, `module`, `::`, and `call`. Perhaps there's an opportunity for a contribution!

lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb
fallwith: With [cb8b5dc](https://github.com/newrelic/newrelic-ruby-agent/pull/1286/commits/cb8b5dcb08e4b3aadddd810b8a6b96782854033c) I've added a new method for obtaining the metadata that lets us be consistent about it.

kaylareopelle (approved):
👍

PR #1289, opened 2022-07-27
CI: treat Envfile files as Ruby

fallwith:
- give each `Envfile` file a standard Ruby file header, specifying utf-8
  encoding, the license info, and that string literals ought to be
  frozen
- remove the '# vim:' comments present on only some `Envfile` instances
  that would tell Vim to treat the file as a Ruby file. Vimmers can
  simply tell Vim that an 'Envfile' pattern encountered at buffer read
  time should trigger a 'set filetype=ruby' operation, like so:

```lua
-- 'Envfile' files are Ruby
local envfileGroup = vim.api.nvim_create_augroup('envfile', { clear = true })
vim.api.nvim_create_autocmd(
  { "Bufread" },
  { pattern = "Envfile", command = "set filetype=ruby", group = envfileGroup }
)
```

### DIFF
--- a/test/multiverse/suites/active_record/Envfile
+++ b/test/multiverse/suites/active_record/Envfile
@@ -1,3 +1,8 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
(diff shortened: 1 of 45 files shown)


kaylareopelle (approved):

PR #1291, opened 2022-07-27
Don't modify frozen Logger instance

### LINKED ISSUE #1288: Can't modify frozen Logger
## Description
Standard logger instrumentation try to modify the Logger instance, which, as in our case, can be frozen (redirecting to `/dev/null`).

## Expected Behavior
The instrumentation should not modify the logger from other libraries, or clone it, or catching the `FrozenError`.

## Steps to Reproduce
The issue happened when using `newrelic_rpm 8.8.0` together with the [braintree_ruby](https://github.com/braintree/braintree_ruby) gem, which allows to pass an external logger in the configuration. 
We use a frozen instance of the standard logger, whicih works well with previous version of the `newrelic_rpm` gem (ie 8.0.0):

```ruby
    NULL_LOGGER = Logger.new('/dev/null', level: Logger::INFO).freeze

    @client ||= Braintree::Gateway.new(environment: environment,
                                       merchant_account_id: merchant_account_id,
                                       merchant_id: merchant_id,
                                       public_key: public_key,
                                       private_key: private_key,
                                       logger: NULL_LOGGER)
``` 

The above code throws the following error:
```ruby
FrozenError: can't modify frozen Logger: #<Logger:0x00007f4fda9b1be0 @level=1, @progname=nil, @default_formatter=#<Logger::Formatter:0x00007f4fda9b1b18 @datetime_format=nil>, @formatter=nil, @logdev=#<Logger::LogDevice:0x00007f4fda9b1a50 @shift_period_suffix="%Y%m%d", @shift_size=1048576, @shift_age=0, @filename="/dev/null", @dev=#<File:/dev/null>, @binmode=false, @mon_data=#<Monitor:0x00007f4fda9b1a00>, @mon_data_owner_object_id=76420>> /app/vendor/bundle/ruby/2.7.0/gems/newrelic_rpm-8.9.0/lib/new_relic/agent/instrumentation/logger/instrumentation.rb:29:in `clear_skip_instrumenting'
```

## Your Environment
The following relevant gems was used:
* `newrelic_rpm` (8.9.0)
* `braintree` (4.4.0)

tannalynn:

# Overview
This PR adds checks for if the logger is frozen before attempting to set/change the `@skip_instrumenting` instance variable. Since this variable is only used internally in the agent and we know those instances of Logger are not frozen, this shouldn't interfere with the purpose of this variable. 
closes #1288

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
@@ -15,18 +15,22 @@ module NewRelic
         # instrumentation installed yet. This lets us disable in AgentLogger
         # and AuditLogger without them having to know the inner details.
         def self.mark_skip_instrumenting(logger)
+          return if logger.frozen?
           logger.instance_variable_set(:@skip_instrumenting, true)
         end
 
         def self.clear_skip_instrumenting(logger)
+          return if logger.frozen?
           logger.instance_variable_set(:@skip_instrumenting, false)
         end
 
         def mark_skip_instrumenting
+          return if self.frozen?
           @skip_instrumenting = true
         end
 
         def clear_skip_instrumenting
+          return if self.frozen?
           @skip_instrumenting = false
         end
 
--- a/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/logger/instrumentation.rb
@@ -25,12 +25,12 @@ module NewRelic
         end
 
         def mark_skip_instrumenting
-          return if self.frozen?
+          return if frozen?
           @skip_instrumenting = true
         end
 
         def clear_skip_instrumenting
-          return if self.frozen?
+          return if frozen?
           @skip_instrumenting = false
         end
 
--- a/test/multiverse/suites/logger/logger_instrumentation_test.rb
+++ b/test/multiverse/suites/logger/logger_instrumentation_test.rb
@@ -174,6 +174,15 @@ class LoggerInstrumentationTest < Minitest::Test
     assert_metrics_recorded_exclusive([])
   end
 
+  def test_dont_modify_frozen_logger
+    @logger.freeze
+    @logger.mark_skip_instrumenting
+    @logger.clear_skip_instrumenting
+    NewRelic::Agent::Instrumentation::Logger.mark_skip_instrumenting(@logger)
+    NewRelic::Agent::Instrumentation::Logger.clear_skip_instrumenting(@logger)
+    assert !@logger.instance_variable_defined?(:@skip_instrumenting), "instance variable should not be defined"
+  end
+
   def assert_logging_instrumentation(level, count = 1)
     # We count on Logger calls but actually write metrics on harvest to
     # minimize impact in the hot path
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # New Relic Ruby Agent Release Notes #
 
 
+  ## v8.10.0
+
+  * **Bugfix: Don't modify frozen Logger**
+
+    Previously the agent would modify each instance of the Logger class by adding a unique instance variable as part of the instrumentation. This could cause the error `FrozenError: can't modify frozen Logger` to be thrown if the Logger instance had been frozen. The agent will now check if the object is frozen before attempting to modify the object. Thanks to @mkcosta for bringing this issue to our attention.
+
+
+
   ## v8.9.0
   
   


kaylareopelle (changes requested):
I like where this is headed! Could you add tests for the new guard statements?

lib/new_relic/agent/instrumentation/logger/instrumentation.rb
```diff
@@ -15,18 +15,22 @@ def skip_instrumenting?
...
           logger.instance_variable_set(:@skip_instrumenting, true)
         end
 
         def self.clear_skip_instrumenting(logger)
+          return if logger.frozen?
           logger.instance_variable_set(:@skip_instrumenting, false)
         end
 
         def mark_skip_instrumenting
+          return if self.frozen?
```
fallwith, line 28: These can just be `return if frozen?`

tannalynn:
> I like where this is headed! Could you add tests for the new guard statements?

Resolved in 134f6e441b1bfd483d3f18308beb65055b947e33

fallwith:
The instrumentation code has this comment:

```ruby
# It's critical we don't instrument logging from metric recording
# methods within NewRelic::Agent, or we'll stack overflow!!
mark_skip_instrumenting
```

Is being able to apply the mark critical only for the agent's own Logger instances?

If it is only critical for the agent's own Logger instances, then why are we bothering to apply the instance var to any other instances in the first place? We could define a method on only our instances and then have the instrumentation call that method only if it is defined.

If it is critical for other instances, then how can we accomplish the same mark/clear functionality when a logger is frozen? Would we keep an internal hash of Logger instance object ids and their current skip/don't skip status?

kaylareopelle:
@fallwith  - The instance var is also used in the ActiveSupportLogger instrumentation to prevent double logging.

test/multiverse/suites/logger/logger_instrumentation_test.rb
```diff
@@ -174,6 +174,15 @@ def refute_any_logging_instrumentation
     assert_metrics_recorded_exclusive([])
   end
 
+  def test_dont_modify_frozen_logger
+    @logger.freeze
+    @logger.mark_skip_instrumenting
+    @logger.clear_skip_instrumenting
+    NewRelic::Agent::Instrumentation::Logger.mark_skip_instrumenting(@logger)
+    NewRelic::Agent::Instrumentation::Logger.clear_skip_instrumenting(@logger)
+    assert !@logger.instance_variable_defined?(:@skip_instrumenting), "instance variable should not be defined"
```
kaylareopelle, line 183: `refute` would do the same thing as `assert !` here!

kaylareopelle (approved):
Thank you for adding the test! There's one optional suggestion.

fallwith:
> @fallwith - The instance var is also used in the ActiveSupportLogger instrumentation to prevent double logging.

OK cool, that means that there are indeed some third party loggers that need to be marked for skipping. My concern then is that a frozen logger might need to be marked for skipping too.

There's also no current way to configure the agent to enable instrumentation only for certain logger instances.

Questions...

- Should frozen loggers have their output forwarded or decorated?
    - If so, do we need to develop a mark/unmark system for frozen loggers that doesn't use instance vars?
    - If not, the prepended/chained method could just `yield` early when called by a frozen logger
- Regardless of the logger being frozen or not, should loggers targetting `/dev/null` have their output forwarded or decorated?

Here's my current take:

- The contents of this PR look great and I absolutely think they should be merged in as they are a great improvement over what we currently have
- I'm not yet sure we're ready to have the current contents alone close out #1288

tannalynn:
> > @fallwith - The instance var is also used in the ActiveSupportLogger instrumentation to prevent double logging.
> 
> OK cool, that means that there are indeed some third party loggers that need to be marked for skipping. My concern then is that a frozen logger might need to be marked for skipping too.
> 
> There's also no current way to configure the agent to enable instrumentation only for certain logger instances.
> 
> Questions...
> 
> * Should frozen loggers have their output forwarded or decorated?
>   
>   * If so, do we need to develop a mark/unmark system for frozen loggers that doesn't use instance vars?
>   * If not, the prepended/chained method could just `yield` early when called by a frozen logger
> * Regardless of the logger being frozen or not, should loggers targetting `/dev/null` have their output forwarded or decorated?
> 
> Here's my current take:
> 
> * The contents of this PR look great and I absolutely think they should be merged in as they are a great improvement over what we currently have
> * I'm not yet sure we're ready to have the current contents alone close out [Can't modify frozen Logger #1288](https://github.com/newrelic/newrelic-ruby-agent/issues/1288)

Thats a good question. I originally thought that we knew that the loggers we would care about would be not frozen, but I'm not as familiar with ActiveSupportLogger. Is that something that just rails uses for itself, or is that something anyone can create and use (and freeze)?  The ActiveSupportLogger instrumentation is only to prevent double logging, correct? So at least if it WAS frozen we wouldn't have issues with recursion (like we would with the agent logger and metrics), but it would double up how many logs we're grabbing which isn't great. 

As for loggers targeting `dev/null`, that's a great question, I'm not sure. I don't remember that coming up during the original logging work (although perhaps @kreopelle might have some thoughts about that).

fallwith:
After some thought and group discussion, we really ought to get this fix out in the next release. We can revisit alternative non-instance-var based behavior toggles and per-logger-output-destination filtration later with separate issues should they arise beyond the theoretical in future.

PR #1292, opened 2022-07-28
gabbi checks for Ruby CVEs

### LINKED ISSUE #1281: Gabbi: Routinely check for Ruby CVEs
It would neat to have Gabbi automatically relay information about Ruby CVEs to us.

I'm thinking it could grab the [feedjira](https://github.com/feedjira/feedjira) gem, use HTTParty or net/http to grab the RubySec atom feed at [https://rubysec.com/atom.xml](https://rubysec.com/atom.xml) and report on any activity less than 24 hours old.

hannahramadan:
This PR enables gabbi to notify the team of Ruby CVEs, as reported by rubysec.com. Running on the same 24-hour cycle as when we get gem version update notifications, the addition of a CVE script checks the rubysec XML feed for Ruby security vulnerability updates and sends a Slack message to our dev-notifications channel.

closes #1281

Updates include:
- Addition of a `cve_methods.rb` file and related tests in `gem_notifications_tests.rb`
- Expansion of `gem_notifications.yml` to run a new job/script: `cve_script.rb`

# Overview

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/.github/workflows/gem_notifications.yml
+++ b/.github/workflows/gem_notifications.yml
@@ -36,3 +36,17 @@ jobs:
             redis
             resque
             unicorn"
+
+  cve_notifications:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: ruby/setup-ruby@v1.90.0
+        with:
+          ruby-version: 2.7
+      - uses: actions/checkout@v2
+      - run: gem install httparty
+      - run: gem install feedjira
+      - name: Check for CVEs
+        run: ruby .github/workflows/scripts/slack_gem_notifications/cve_script.rb
+        env:
+          SLACK_GEM_NOTIFICATIONS_WEBHOOK: ${{ secrets.SLACK_GEM_NOTIFICATIONS_WEBHOOK }}
--- a/.github/workflows/scripts/slack_gem_notifications/README.md
+++ b/.github/workflows/scripts/slack_gem_notifications/README.md
@@ -1,14 +1,19 @@
-# Slack Gem Notifications
+# Slack Notifications
 
-An internal tool for alerting the Ruby agent team of Ruby gem version updates. 
+An internal tool for alerting the Ruby agent team of gem version updates and Ruby CVEs. 
 
 ## Description
 
-This script runs every 24 hours via GitHub Actions and uses the RubyGems API to see if gems were updated during that time. A slack alert is sent to the Ruby agent team's notification channel when an update is found. This alert contains the updated gem name, the gem's previous and latest version, and how to view more information on the update.
+This script runs every 24 hours via GitHub Actions and sends notifications for 1) gem version updates and 2) Ruby security vulnerabilities.
 
+Gem Version Updates
+The RubyGems API is used to see if gems were updated during that time. A slack alert is sent to the Ruby agent team's notification channel when an update is found. This alert contains the updated gem name, the gem's previous and latest version, and how to view more information on the update.
+
+Ruby CVE Updates
+An XML feed from rubysec.com reports updates on Ruby CVEs. When a vulnerability is found, a slack alert is sent to the Ruby agent team's notification channel. This alert contains the security alert and related rubysec.com link.
 ## Helpful Information
 
-### Adding or Removing Gems
+### Adding or Removing Gem Version Updates
 
 The `gem_notifications.yml` file contains a list of gems that are checked for updates. Gems can be added or removed from this list to adjust notifications.
 
@@ -18,7 +23,14 @@ This project uses the internal Ruby Gem Updates slackbot to post notifications. 
 
 ## Relevant Project Files
 
+Gem Version Updates
 * `.github/workflows/scripts/slack_gem_notifications/notifications_script.rb`
 * `.github/workflows/scripts/slack_gem_notifications/notifications_methods.rb`
+
+Ruby CVE Updates
+* `.github/workflows/scripts/slack_gem_notifications/cve_methods.rb`
+* `.github/workflows/scripts/slack_gem_notifications/cve_script.rb`
+
+Shared
 * `.github/workflows/gem_notifications.yml`
 * `test/new_relic/gem_notifications_tests.rb`
--- a/.github/workflows/scripts/slack_gem_notifications/README.md
+++ b/.github/workflows/scripts/slack_gem_notifications/README.md
@@ -6,11 +6,11 @@ An internal tool for alerting the Ruby agent team of gem version updates and Rub
 
 This script runs every 24 hours via GitHub Actions and sends notifications for 1) gem version updates and 2) Ruby security vulnerabilities.
 
-Gem Version Updates
-The RubyGems API is used to see if gems were updated during that time. A slack alert is sent to the Ruby agent team's notification channel when an update is found. This alert contains the updated gem name, the gem's previous and latest version, and how to view more information on the update.
+### Gem Version Updates
+The RubyGems API is used to see if gems were updated during that time. A Slack alert is sent to the Ruby agent team's notification channel (ruby-agent-notifications) when an update is found. This alert contains the updated gem name, the gem's previous and latest version, and how to view more information on the update.
 
-Ruby CVE Updates
-An XML feed from rubysec.com reports updates on Ruby CVEs. When a vulnerability is found, a slack alert is sent to the Ruby agent team's notification channel. This alert contains the security alert and related rubysec.com link.
+### Ruby CVE Updates
+An XML feed from [RubySec](https://rubysec.com/) reports updates on Ruby CVEs. When a vulnerability is found, a Slack alert is sent to the Ruby agent team's notification channel (ruby-agent-notifications). This alert contains the security alert and related RubySec link.
 ## Helpful Information
 
 ### Adding or Removing Gem Version Updates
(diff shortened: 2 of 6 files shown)


.github/workflows/scripts/slack_gem_notifications/cve_methods.rb
```diff
@@ -0,0 +1,32 @@
...
+require 'time'
+require 'feedjira'
+require 'httparty'
+
+# limiting to 5 checks since more than 5 security concerns in 24 hours is unlikely. 
+def check_for_cves
+  xml = HTTParty.get('https://rubysec.com/atom.xml').body
+  feed = Feedjira.parse(xml)
+  feed.entries.take(5).each do |entry|
+    if Time.now.utc - entry.updated < 24 * 60 * 60
```
fallwith, line 15: The cool thing about the RubySec feed in particular is that it is sorted in reverse chronological order (giving you the newest entry first). As a result, we should be able to do a `break` to stop processing any more feeds as soon as we find one older than the timeframe, rather than testing every single feed entry for being recent.

fallwith (commented):
This work looks great! I have some thoughts on organizing things now that there are 2 separate bits of functionality that the tool handles, but I think we should just wait until packaging things into a gem before we worry about them.

.github/workflows/scripts/slack_gem_notifications/cve_methods.rb
hannahramadan: A much better way to do it! [Addressed here](https://github.com/newrelic/newrelic-ruby-agent/pull/1292/commits/af0ee976fa859d6338e6fb22391749df36f714ce).

.github/workflows/gem_notifications.yml
```diff
@@ -36,3 +36,17 @@ jobs:
             redis
             resque
             unicorn"
+
+  cve_notifications:
```
fallwith, line 40: The benefit of going with a separate entry here is that it will run in parallel with the gem notification job and both won't be impacted by an issue with the other.

The drawback is with cost and more moving parts (for example, there are now two separate internet calls to fetch 'httparty)'.

We'll want to give some thought to these trade-offs as we continue to add more functionality to Gabbi.

Some thoughts...

- Maybe a single entrypoint invokes Gabbi and it goes through a list of work to do.
    - Each separate task will need to be hardened against crashing Gabbi itself, so that the following tasks are guaranteed to run
- Once Gabbi itself is a gem, a single `gem install gabbi` will simplify the fetching of all of Gabbi's own dependencies
- If gem fetching over the internet ever becomes an issue, we can always switch to GHA caching or prebuilt and stored `.gem` files
- Design-wise, we'll want to think of mitigating unhappy path scenarios. Knowing about a CVE is pretty important. Currently if the Gabbi run to discover them fails, there's a chance the CVE won't be picked up by the next Gabbi run 24 hours later and we'll never see the notification.

(Just thoughts to bear in mind as Gabbi continues to evolve - no related requests for change for these things with this PR)

.github/workflows/scripts/slack_gem_notifications/README.md
```diff
@@ -1,14 +1,19 @@
...
-An internal tool for alerting the Ruby agent team of Ruby gem version updates. 
+An internal tool for alerting the Ruby agent team of gem version updates and Ruby CVEs. 
 
 ## Description
 
-This script runs every 24 hours via GitHub Actions and uses the RubyGems API to see if gems were updated during that time. A slack alert is sent to the Ruby agent team's notification channel when an update is found. This alert contains the updated gem name, the gem's previous and latest version, and how to view more information on the update.
+This script runs every 24 hours via GitHub Actions and sends notifications for 1) gem version updates and 2) Ruby security vulnerabilities.
 
+Gem Version Updates
+The RubyGems API is used to see if gems were updated during that time. A slack alert is sent to the Ruby agent team's notification channel when an update is found. This alert contains the updated gem name, the gem's previous and latest version, and how to view more information on the update.
```
fallwith, line 10: "Slack" ought to be capitalized everywhere.

.github/workflows/scripts/slack_gem_notifications/cve_methods.rb
```diff
@@ -0,0 +1,31 @@
...
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'time'
+require 'feedjira'
+require 'httparty'
+
+def check_for_cves
+  xml = HTTParty.get('https://rubysec.com/atom.xml').body
```
fallwith, line 11: It would be good to have the RubySec feed URL live in a constant.

.github/workflows/scripts/slack_gem_notifications/cve_methods.rb
```diff
@@ -0,0 +1,31 @@
...
+  feed = Feedjira.parse(xml)
+  feed.entries.each do |entry|
+    break if Time.now.utc - entry.updated > 24 * 60 * 60
```
fallwith, line 14: It would be good to have the 24 hours value live in a constant. That way Ruby isn't performing the math over and over again on every single feed item.

fallwith: I recommend calculating `Time.now.utc` just once and storing it in a local variable before starting to loop over the feeds. Otherwise, it will keep representing a later and later time as the feed iteration keeps running.

For this exact scenario in which we're expecting to be done in a single second, it's not a big deal. But if iterating was to take multiple minutes and `Time.now` kept advancing, there would be a risk of missing out on data that was 23 hours and 59 minutes old after the iteration was left running for 2 minutes.

.github/workflows/scripts/slack_gem_notifications/cve_methods.rb
```diff
@@ -0,0 +1,31 @@
...
+    
+    cve_send_bot(entry.title, entry.entry_id)
+  end
+end
+
+def cve_bot_message(title, url)
+  alert_message = ":rotating_light: #{title}\n<#{url}|More info here>"
```
fallwith, line 21: There's no need to use a variable here that will just get thrown away immediately afterwards on the next line which ends the method (`alert_message = ` can be removed).

.github/workflows/scripts/slack_gem_notifications/cve_methods.rb
```diff
@@ -0,0 +1,31 @@
...
+end
+
+def cve_send_bot(title, url)
+  path = ENV['SLACK_GEM_NOTIFICATIONS_WEBHOOK']
```
fallwith, line 25: Given that this value is only needed exactly 1 time, there's no need to waste a variable here. The `ENV['...']` code can just be placed directly as the first argument to `post`.

.github/workflows/scripts/slack_gem_notifications/cve_methods.rb
```diff
@@ -0,0 +1,31 @@
...
+  options = {headers: {'Content-Type' => 'application/json'},
+             body: {
+              text: cve_bot_message(title, url)}.to_json}
```
fallwith, line 28: Do we ever want the CVE bot message to _not_ be in a JSON format? If we always want JSON, perhaps the `cve_bot_message` method itself should return JSON.

test/new_relic/gem_notifications_tests.rb
```diff
@@ -86,4 +87,20 @@ def test_interpolate_rubygems_url
...
     assert_kind_of String, interpolate_rubygems_url(gem_name)
   end
+
+  def test_cve_bot_message
+    assert_raises(ArgumentError) { cve_bot_message() }
+    assert_raises(ArgumentError) { cve_bot_message("allosaurus") }
+    message = cve_bot_message("allosaurus", "dinotracker.com")
+    assert_equal message, ":rotating_light: allosaurus\n<dinotracker.com|More info here>"
+    assert_kind_of String, message
+  end
```
fallwith, line 97: We haven't yet introduced the MiniTest specific RuboCop rules to this project and the existing test files are filled with blocks like this one, so no need to change anything at this time.

But to be thinking of patterns for us to standardize on a team...

Some unit testing style rules forbid the use of multiple expectations / assertions within a single test. For us - especially when the mocking or stubbing is particularly tricky to set up - it has been helpful to use multiple assertions to validate a response. For example, you might perform a web request and want to validate both the HTTP response code AND the response body.

My take that I would like input on from the team would be to only use multiple assertions for the exact same testing scenario.

Looking at `test_cve_bot_message`, I see fully 3 separate scenarios:

- What happens if zero arguments are given?
- What happens if the wrong number of arguments are given?
- What happens if the happy path version of arguments are given?

I'd use 3 separate `test_*` methods to cover these scenarios, even if the first 2 just end up being one liners.

For the last test, using both the `assert_equal` and `assert_kind_of` checks in the same `test_*` method would be ok, because both are validating the same output of the same scenario.

fallwith (commented):
It's fantastic to see this idea not only realized, but bringing in cool things I wouldn't have thought of!

test/new_relic/gem_notifications_tests.rb
kaylareopelle: I've been thinking the same thing, @fallwith. I agree with your take and am all for adding this to our Dev Team Best Practices.

hannahramadan: Thank you! This is quite helpful and a best practice to take moving forward :) I've refactored the current and previous tests through [this commit](https://github.com/newrelic/newrelic-ruby-agent/pull/1292/commits/4052615af1e88f7ead3746b7526c03cdc43c73e6).

.github/workflows/scripts/slack_gem_notifications/README.md
hannahramadan: Made both these changes as well as the other README changes you suggested, Kayla!

.github/workflows/scripts/slack_gem_notifications/cve_methods.rb
hannahramadan: Done ✅

.github/workflows/scripts/slack_gem_notifications/cve_methods.rb
hannahramadan: Super great feedback. I was wondering about the little gap in time from starting to run this file to the end. I like this suggestion a lot. [4052615](https://github.com/newrelic/newrelic-ruby-agent/pull/1292/commits/4052615af1e88f7ead3746b7526c03cdc43c73e6)

.github/workflows/scripts/slack_gem_notifications/cve_methods.rb
hannahramadan: Always JSON, so I've made this update.

fallwith (approved):
Good stuff.

Some ideas for the future:

- Turn the `*_methods.rb` files into Ruby classes, and have the `*_script.rb` files call public class methods
    - With classes, local variables like `ruby_sec_feed` and `cycle` could become constants
- Implement error handling and possibly retry logic for each operation that could fail
- Break up some of the complex operations (`Feedjira.parse(HTTParty.get(ruby_sec_feed).body)`) into multiple methods
- Add `newrelic_rpm` to the project!

hannahramadan:
> Good stuff.
> 
> Some ideas for the future:
> 
> * Turn the `*_methods.rb` files into Ruby classes, and have the `*_script.rb` files call public class methods
>   
>   * With classes, local variables like `ruby_sec_feed` and `cycle` could become constants
> * Implement error handling and possibly retry logic for each operation that could fail
> * Break up some of the complex operations (`Feedjira.parse(HTTParty.get(ruby_sec_feed).body)`) into multiple methods
> * Add `newrelic_rpm` to the project!

Thanks for the review and feedback! I've saved these ideas to address in the future 🌈

kaylareopelle (approved):

PR #1298, opened 2022-08-02
Make ActiveRecord::Migration class name dynamic

### LINKED ISSUE #1208: Refactor ActiveRecord::Migration dynamic class naming
Starting in Active Record 5.0, migrations are expected to have version strings corresponding with the Rails version they were created with in square brackets at the end of the inheriting `ActiveRecord::Migration` class. Since the Ruby agent covers multiple ActiveRecord versions below and above 5.0, we've addressed this using the following snippet:

```
ActiveRecord::VERSION::STRING >= "5.0.0" ? ActiveRecord::Migration["#{ActiveRecord::VERSION::STRING[0]}.0"] : ActiveRecord::Migration
```

This snippet is used in a few areas of the codebase. Rather than continuing to copy/paste the code, let's rewrite it as a method and call the method instead.

### Tasks
- [ ] Write a test_helper method that includes the above code in its body (this may live in our unit test helper file or may need to live within the multiverse suite, some experimentation required!)
- [ ] Replace the calls to this code with the method 
- [ ] Verify the tests relying on these sample migrations still pass

hannahramadan:
Previously, when testing ActiveRecord::Migration in multiverse, a static line determined the correct class syntax depending on the ActiveRecord version. This commit replaces where these static calls were made with a method call, which was added to test_helper.rb

closes #1208

# Overview

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


test/multiverse/lib/multiverse/suite.rb
```diff
@@ -266,6 +266,7 @@ def generate_gemfile(gemfile_text, env_index, local = true)
         f.puts "gem 'rake'" unless gemfile_text =~ /^\s*gem .rake[^_]./ || suite == 'rake'
 
         f.puts "gem 'mocha', '~> 1.9.0', :require => false"
+        f.puts "gem 'minitest-stub-const', '~> 0.6', require: false"
```
hannahramadan, line 269: This line is already included in the grpc branch. It was needed to have tests pass in this PR, so I've included it in this commit.

test/multiverse/suites/active_record/Envfile
```diff
@@ -48,3 +48,5 @@ def gem_list(activerecord_version = nil)
     #{mysql_gem(activerecord_version)}
   RB
 end
+
+create_gemfiles(ACTIVERECORD_VERSIONS, gem_list)
```
hannahramadan, line 52: The `create_gemfiles` call is currently missing from dev, but will be addressed in PR https://github.com/newrelic/newrelic-ruby-agent/pull/1268. Needed this update for testing, so it is included here.

`RAILS_VERSIONS` -> `ACTIVERECORD_VERSIONS` for more appropriate naming.

test/multiverse/suites/active_record/db/migrate/20150414084400_create_groups.rb
```diff
@@ -3,7 +3,9 @@
 # See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
 # frozen_string_literal: true
 
-class CreateGroups < ActiveRecord::VERSION::STRING >= "5.0.0" ? ActiveRecord::Migration[5.0] : ActiveRecord::Migration
```
hannahramadan, line 6: All `active_record/db/migrate` test files had this line, which was added 5 years ago when ActiveRecord 5.0+ started requiring the version. Rather than pin the version to 5.0, I've included it in this update to match the dynamic versioning we're adding elsewhere.

For comparison, the new line/method call reads:
ActiveRecord::VERSION::STRING >= "5.0.0" ? ActiveRecord::Migration["#{ActiveRecord::VERSION::STRING[0]}.0"] : ActiveRecord::Migration

test/multiverse/suites/active_record_pg/Envfile
```diff
@@ -27,7 +27,7 @@ def gem_list(activerecord_version = nil)
   RB
 end
 
-
+create_gemfiles(ACTIVERECORD_VERSIONS, gem_list)
```
hannahramadan, line 30: The create_gemfiles call is currently missing from dev, but will be addressed in PR https://github.com/newrelic/newrelic-ruby-agent/pull/1268. Needed this update for testing, so it is included here.

test/multiverse/suites/active_record/Envfile
```diff
@@ -20,7 +15,9 @@ RAILS_VERSIONS = [
 
 def jbdc_version(activerecord_version)
   return unless RUBY_PLATFORM.eql?('java') || activerecord_version.nil?
-  activerecord_version.slice('1')
+  # remove the '.' between the major and minor versions
+  activerecord_version[1] = ''
+  activerecord_version
```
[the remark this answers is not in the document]
hannahramadan: `[Addressed here](https://github.com/newrelic/newrelic-ruby-agent/pull/1298/commits/41f0f9720f756ec16d189109af9ececa20d18581)`

test/multiverse/suites/active_record/Envfile
```diff
@@ -1,14 +1,9 @@
-# encoding: utf-8
-# This file is distributed under New Relic's license terms.
-# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
-# frozen_string_literal: true
-
 # not testing the latest of activerecord in this suite because
 # mysql2 has issues with newer Rubies
 
 # don't change the first index to a number ending in other than zero
 # it is used by jbdc_version
```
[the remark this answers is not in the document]
hannahramadan: `[Fixed those](https://github.com/newrelic/newrelic-ruby-agent/pull/1298/commits/41f0f9720f756ec16d189109af9ececa20d18581)`

test/multiverse/suites/active_record/Envfile
```diff
@@ -40,11 +37,25 @@ def mysql_gem(activerecord_version)
   end
 end
 
+def minitest_version(rails_version = nil)
+  if rails_version && rails_version.include?('4.0')
```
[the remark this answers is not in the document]
hannahramadan: I assume, but I'll need to check with @kaylareopelle. This comes from PR #1268. I found that Rails 4 sets the dependency on Minitest to “~> 4.2” so I think that's whats happening here

test/multiverse/suites/active_record/Envfile
```diff
@@ -18,33 +13,50 @@ RAILS_VERSIONS = [
   ['3.2.0', 2.2, 2.2]
 ]
 
-def jbdc_version(activerecord_version)
+def jdbc_version(activerecord_version)
   return unless RUBY_PLATFORM.eql?('java') || activerecord_version.nil?
-  activerecord_version.slice('1')
+  # remove the '.' between the major and minor versions
+  activerecord_version_jdbc = activerecord_version.sub('.', '')
+  activerecord_version_jdbc
```
fallwith, line 20: My guess here is that this method is intended to determine which version of the `activerecord-jdbc-adapter` to use for a given ActiveRecord version.

From looking over the list of that gem's [versions](https://rubygems.org/gems/activerecord-jdbc-adapter/versions), it looks like for ActiveRecord version 5.2.0, there's an activerecord-jdbc-adapter version 52.0.

So following that logic, you'd want to convert the input of 5.2.0 into an output of 52.0 (not 520).

Unfortunately, this logic won't work for ActiveRecord versions 3.x and 4.x.

I'm wondering if we as a team should come up with a better way of deciding which gem version to use with each ActiveRecord version. Maybe it's as simple as "use this version for AR versions up to v5, this version for 5-6, and this version for 7+" or whatnot.

hannahramadan: Darn. That would be great to come up with something as a team. For this particular case, I think we would get for example, 52.0 vs 520, as `sub` just removes the first occurrence.

hannahramadan:
Moving a portion of this work to PR #1268 [PR 1268] in order to address core CI failures that are outside the scope of this PR and the related ticket #1208

test/multiverse/suites/active_record_pg/Envfile
hannahramadan: No longer relevant comment as of PR #1268

test/multiverse/suites/active_record_pg/Envfile
```diff
@@ -28,7 +28,6 @@ def gem_list(activerecord_version = nil)
 end
 
 create_gemfiles(ACTIVERECORD_VERSIONS, gem_list)
-
```
hannahramadan, line 31: This wasn't really a line removal as much as it was adding our license & heading to the top of this file

fallwith (approved):

PR #1300, opened 2022-08-02
add tests for methods in test/test_helper.rb

fallwith:
now that `test/test_helper.rb` has started to contain methods, let's set
up tests for it

yodawg

### DIFF
--- a/test/test/test_helper_test.rb
+++ b/test/test/test_helper_test.rb
@@ -1,0 +1,35 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require_relative '../test_helper'
+
+# test_helper works without Rails, so fake a Rails class for testing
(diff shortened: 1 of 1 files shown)


kaylareopelle (approved):
🚗

tannalynn (approved):

test/test/test_helper_test.rb
```diff
@@ -0,0 +1,35 @@
...
+        version
+      end
+    end
+  end
+end
+
+class TestHelperTest < Minitest::Test
+  # BEGIN current_active_record_migration_version
+  def test_use_a_version_with_ar_gte_5
+    version = [7, 0, 3, 1]
```
hannahramadan, line 23: Why do we declare `version` as an array, and then join it later as a stringed version?

fallwith: I think I wanted to be able to easily grab the first digit (the major version number) so I used an array. I may have been trying to avoid repeating the same code as the method being tested. Definitely feel free to change it around - I was mainly looking for approval on the overall approach of faking an AR class when Rails is absent.

PR #1306, opened 2022-08-03
no stubs

fallwith:
we need to revert #1149 because introducing MiniTest to the perf tests
causes issues. but we certainly don't want to go back to Mocha. so let's
just do without stubs entirely.

> ... A stub is a placeholder that thinks it's fly
> And is also known as a busta
> Always talkin' about how it'll respond
> And never does anything real
>
> ... So no, I don't need your code
> No, I don't want to trust you with mine and
> No, I don't want your return value
> No, I don't want none of your time and
>
> ... No, I don't want no stub
> A stub is a dummy that can't get no love from me
> Hangin' out on the unit test side
> Of the CI ride
> Trying to holla at me
> I don't want no stub
> A stub is a dummy that can't get no love from me
> Hanging' out on the unit test side
> Of the CI ride
> Trying to holla at me

no-scrubs

### DIFF
--- a/test/new_relic/agent/monitors/trace_context_request_monitor_test.rb
+++ b/test/new_relic/agent/monitors/trace_context_request_monitor_test.rb
@@ -101,14 +101,9 @@ module NewRelic
 
       def build_parent_transaction_headers
         carrier = {}
-
-        # stubbing contexted allows trace_context_active? to pass
-        # which, in turn, allows us to insert a trace_context header here.
         parent_txn = in_transaction('referring_txn') do |txn|
-          Agent.instance.stub(:connected?, true) do
-            txn.sampled = true
-            txn.distributed_tracer.insert_trace_context_header(carrier, NewRelic::FORMAT_RACK)
-          end
+          txn.sampled = true
+          txn.distributed_tracer.insert_trace_context_header(carrier, NewRelic::FORMAT_RACK)
         end
         [parent_txn, carrier]
       end
--- a/test/performance/suites/trace_context.rb
+++ b/test/performance/suites/trace_context.rb
@@ -52,9 +52,7 @@ class TraceContext < Performance::TestCase
     with_config(CONFIG) do
       in_transaction do |txn|
         measure do
-          NewRelic::Agent.agent.stub(:connected?, true) do
-            txn.distributed_tracer.insert_trace_context_header(carrier: {})
-          end
+          txn.distributed_tracer.insert_trace_context_header(carrier: {})
         end
       end
     end


test/new_relic/agent/monitors/trace_context_request_monitor_test.rb
```diff
@@ -101,14 +101,9 @@ def test_does_not_accept_malformed_trace_context
 
       def build_parent_transaction_headers
         carrier = {}
-
-        # stubbing contexted allows trace_context_active? to pass
-        # which, in turn, allows us to insert a trace_context header here.
         parent_txn = in_transaction('referring_txn') do |txn|
-          Agent.instance.stub(:connected?, true) do
```
kaylareopelle, line 108: Without some kind of stubbing for `Agent.instance.connected?`, it seems like we're changing the functionality of this test. Is it now `connected?` by default?

kaylareopelle: (Same question for the other removal)

fallwith: I removed the stubs, expecting to have the tests fail and needing me to write a pure Ruby solution (no mocha, no stubs), but they passed. So I just left them passing. Do you think them passing is a case of false positives?

tannalynn: Yeah, that's interesting, I would expect these not to work anymore since I think `connected?` is only true after the agent has communicated with the collector. We might want to look into why removing the stubs didn't cause it to fail, seems suspicious to me

kaylareopelle: It could be. I've needed to stub `:connected?` for a truthy value lately. I believe it was in gRPC client, but that's a multiverse test so perhaps it's different. I'm not sure if the test environment changed so that the agent is now connected by default or if it's a false positive.

fallwith: @kaylareopelle @tannalynn 

Ok, so I did the following:

1. Modified `Agent#connected?` to perform `1/0`
2. Ran the `trace_context_request_monitor_test.rb` tests with `rake test file=test/new_relic/agent/monitors/trace_context_request_monitor_test.rb`
3. Ran the `trace_context.rb` perf tests with `bundle exec script/runner -n trace_context`

Everything passed. `Agent#connected?` is never invoked. We don't want no stubs.

tannalynn: So weird, I wonder why it was there in the first place 🤔 Maybe something was changed since it was first added that makes it no longer needed? Thanks for checking that out!

tannalynn (approved):
💯

PR #1308, opened 2022-08-03
Fix for multiple multiverse failures

tannalynn:
Frozen strings in the test ViewsController were causing failures, made the
strings mutable that need to be

# Overview
Frozen strings in the test ViewController caused errors that caused several tests to fail because the controller action was failing. These 4 strings were the cause of all the failures, I was able to confirm locally that 2.3.8 no longer fails on these tests. 

See failures here: https://github.com/newrelic/newrelic-ruby-agent/runs/7648494987?check_suite_focus=true

In addition to this, a new version of 'jar-dependencies' gem was released today. This is a dependency of psych, so when attempting to install psych it would pull in the latest version of 'jar-dependencies' 0.4.2. But the previous version, 0.4.1, was already activated by bundler, resulting in an error. I pinned the jar-dependencies version to avoid this version to 0.4.1 to avoid this conflict.

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/test/multiverse/suites/config_file_loading/Envfile
+++ b/test/multiverse/suites/config_file_loading/Envfile
@@ -21,6 +21,7 @@ def gem_list(psych_version = nil)
     gem 'jruby-openssl' if RUBY_PLATFORM == 'java'
 
     gem 'psych'#{psych_version}
+    gem 'jar-dependencies', '0.4.1' if RUBY_PLATFORM == 'java' 
 
     # don't start the agent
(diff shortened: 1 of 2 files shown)


kaylareopelle (dismissed):
Thanks for sleuthing this out!

fallwith (dismissed):

fallwith (approved):

PR #1312, opened 2022-08-04
CI: temporary Sidekiq fix

fallwith:
while waiting on Sidekiq v6.5.4 to remove the unwanted dependency on
`concurrent-ruby`, pin to v6.5.1 for now


kaylareopelle (dismissed):
Thanks for finding the fix!

kaylareopelle (approved):
Thanks for improving #add_twiddle_wakka!

PR #1334, opened 2022-08-19
CVE-2022-35954: update @actions/core to 1.9.1

fallwith:
For CVE-2022-35954 / CWE-74 / CWE-77, upgrade @actions/core to v1.9.1


kaylareopelle (approved):
Thanks for finding this and applying the upgrade!

Issue #1225, opened 2022-06-27
CI: Troubleshoot SimpleCov branch testing


hannahramadan:
### Is your feature request related to a problem? Please describe.
SimpleCov was added to the Ruby agent to provide code coverage analysis. SimpleCov originally reported line coverage when our CI runs, but a feature was added to report on branch coverage as well. However, reported branch coverage results seem too low, and further investigation is needed to understand how we might change SimpleCov in our agent to provide more accurate testing results. The issue likely is related to multiverse testing interacting with unit tests.

Our branch coverage threshold is currently set to 30% so that we aren't continuously failing CI (reported branch coverage is ~35%). Once branch coverage is reporting realistic results, we will raise SimpleCov's`failedThresholdBranch` inside `ci.yml` to about 5 less than the reported results. 

### Feature Description
Accurate SimpleCov branch testing.

### Priority
Please help us better understand this feature request by choosing a priority from the following options: 
Really Want

kaylareopelle:
Some additional context: We're concerned that the mechanism that stitches together our multiple test runs and/or multiple CI steps is not accurately reporting the simplecov coverage numbers.

First step: Understand if this hypothesis is correct.

PR #1336, opened 2022-08-22
Simplecov Branch Coverage

kaylareopelle:
# Overview

## Unique command names
* SimpleCov overwrote chain test run results with the prepend results because the two shared a `command_name` value in multiverse
* Subsequent multiverse test suite runs tried to get merged into the same `.resultset.json` file at the end of each run with varied results
* Creating unique `command_name` values for each test run and saving those results to a `coverage_dir` with the same name ensures neither of the above issues can overwrite previous results

## rake coverage:report
* The task can now be run locally
* Collated results will be saved to a a directory named coverage_result
* The default format will be the only one generated, as we did not use the JSON results. The gem for JSON results has been removed from the gemspec.

## CI optimization
* Only the .resultset.json files used in the rake task are uploaded as CI artifacts, lowering our data usage
* SimpleCov CI artifacts were saved for 90 days by default. We rarely refer to them beyond the immediate completion of a PR test run, so the retention period has been reduced to two days
* The `failedThreshold` and `failedBranchThreshold` have been increased to values closer to our new coverage percentages 😄 

Closes: #1225 

Submitter Checklist:

# Testing
This ticket is all about testing 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


.github/workflows/ci.yml
```diff
@@ -338,14 +339,15 @@ jobs:
...
-          path: lib/coverage
+          path: lib/coverage_results
+          retention-days: 2
       - name: Simplecov Report
         uses: ./.github/actions/simplecov-report
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          resultPath: lib/coverage/.last_run.json
-          failedThreshold: 55
-          failedThresholdBranch: 20
```
kaylareopelle, line 351: TODO: Update the threshold levels once we get the new numbers from a full run

.github/workflows/ci.yml
```diff
@@ -204,10 +204,11 @@ jobs:
         uses: ./.github/actions/annotate
 
       - name: Save coverage results
-        uses: actions/upload-artifact@master
```
kaylareopelle, line 207: TODO: Revert these changes, use @master instead

.simplecov
```diff
@@ -1,4 +1,10 @@
 # frozen_string_literal: true
+require 'securerandom'
+
+random = SecureRandom.uuid
+SimpleCov.command_name(random)
+SimpleCov.coverage_dir("coverage_#{random}")
```
kaylareopelle, line 6: With this change, a new coverage directory will be created every time a test run is complete. This prevents results from being overwritten by subsequent runs. Inspo from [ENOSUCHBLOG](https://blog.yossarian.net/2018/04/01/Code-coverage-with-Simplecov-across-multiple-processes). 

The folder names, however, look a bit gnarly: `/coverage_9ea424f4-ce16-4c07-ae67-2d15d42f66621`

We haven't been doing anything with them though. Are we comfortable with using this approach to create unique names?

lib/tasks/coverage_report.rake
```diff
@@ -7,17 +7,15 @@ namespace :coverage do
...
-      return
-    end
-    SimpleCov.collate(Dir["coverage*/.resultset.json"]) do
-      formatter SimpleCov::Formatter::MultiFormatter.new([
-        SimpleCov::Formatter::JSONFormatter,
-        SimpleCov::Formatter::HTMLFormatter
-      ])
-      refuse_coverage_drop
+
+    SimpleCov.coverage_dir('coverage_results')
```
kaylareopelle, line 11: To make it easier to identify the collated results, call the new directory `coverage_results`

lib/tasks/coverage_report.rake
```diff
@@ -7,17 +7,15 @@ namespace :coverage do
...
-    require 'simplecov_json_formatter'
-    unless ENV['CI']
-      puts 'This task is intended to be run only on the CI.'
-      return
-    end
-    SimpleCov.collate(Dir["coverage*/.resultset.json"]) do
-      formatter SimpleCov::Formatter::MultiFormatter.new([
-        SimpleCov::Formatter::JSONFormatter,
-        SimpleCov::Formatter::HTMLFormatter
-      ])
```
kaylareopelle, line 19: My understanding is that these were used while debugging the original SimpleCov results and we don't need them for anything now. HTML formatter is the default, so we'll still get a nice `index.html` file to search.

lib/tasks/coverage_report.rake
```diff
@@ -7,17 +7,15 @@ namespace :coverage do
...
-      refuse_coverage_drop
+
+    SimpleCov.coverage_dir('coverage_results')
+
+    if ENV['CI']
+      SimpleCov.collate(Dir["*/coverage_*/.resultset.json"]) do
+        refuse_coverage_drop
+      end
+    else
+      SimpleCov.collate(Dir["lib/coverage_*/.resultset.json"])
```
kaylareopelle, line 18: This condition enables us to use `rake coverage:report` locally!

.github/workflows/ci.yml
kaylareopelle: Changing my mind here, let's stick to a version. We do for actions/download-artifact

test/simplecov_test_helper.rb
```diff
@@ -9,9 +9,3 @@
 SIMPLECOV_MIN_RUBY_VERSION = '2.7.0'
 
 require 'simplecov' if RUBY_VERSION >= SIMPLECOV_MIN_RUBY_VERSION
```
kaylareopelle, line 11: I decided to leave this file so that we have a single location where the simplecov minimum version constraint is enforced.

lib/tasks/coverage_report.rake
tannalynn: I believe that's accurate based on what I remember about these

.simplecov
tannalynn: That makes sense to me. We have so many different runs that this does seem like the most straightforward way to do that.

tannalynn (approved):
🎉 This is great, I'm so glad you were able to get this working fully so we can see the actual coverage of the agent!

fallwith (approved):

PR #1339, opened 2022-08-23
gRPC client: don't touch initialize

### LINKED ISSUE #1337: GRPC Wrong Number of arguments
When we upgraded the newrelic gem and deployed, ran into an issue with it seeming to fail when trying to prepend to GRPC when integrating with a google GRPC gem.

## Description
Screen Shot 2022-08-22 at 1 13 59 PM

Screen Shot 2022-08-22 at 1 10 56 PM

Using newrelic 8.9.0 and  google-ads-googleads 18.0.0 and grpc 1.48.0

## Expected Behavior
For it not to blow up by upgrading newrelic by default

## Steps to Reproduce
```
# this is one of the ways specified in https://github.com/googleads/google-ads-ruby
Google::Ads::GoogleAds::GoogleAdsClient.new('config/google_ads_config.rb').service.google_ads
```
Tried auto, prepend, and chain arguments for grpc_client. only got it to work once i disabled it.

## Your Environment
ruby: 3.1.2
Using newrelic 8.9.0 and  google-ads-googleads 18.0.0 and grpc 1.48.0

fallwith:
for the gRPC client instrumentation, we had been prepending/chaining
`initialize` calls to get at the host argument in order to apply our
gRPC host denylist behavior. when other libraries are themselves
bringing in the gRPC client library (via the `grpc` gem) and using their
own `initialize` chaining and `super` calls, we run the risk of passing
the wrong number of arguments to the `initialize` method in the chain
that we are handing things off to.

notably, this causes `google-ads-googleads` based apps to encounter
issues.

to address this, let's get out of the business of touching `initialize`
entirely, and source the host information from the gRPC client
instance's own `@host` ivar.

resolves #1337

### DIFF
--- a/lib/new_relic/agent/instrumentation/grpc/client/chain.rb
+++ b/lib/new_relic/agent/instrumentation/grpc/client/chain.rb
@@ -13,14 +13,6 @@ module NewRelic::Agent::Instrumentation
           ::GRPC::ClientStub.class_eval do
             include NewRelic::Agent::Instrumentation::GRPC::Client
 
-            def initialize_with_newrelic_trace(*args)
-              @trace_with_newrelic = trace_with_newrelic?(args.first)
-              initialize_without_newrelic_trace(*args)
-            end
-
-            alias initialize_without_newrelic_trace initialize
-            alias initialize initialize_with_newrelic_trace
-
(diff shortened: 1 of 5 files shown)


kaylareopelle (approved):
Nice pivot!

PR #1340, opened 2022-08-23
Add coverage:clear task to remove coverage dirs

kaylareopelle:
Every multiverse test run will accumulate new coverage directories. This task removes all directories in `lib` that begin with "coverage_"


tannalynn (approved):
Thanks for getting that set up!

PR #1341, opened 2022-08-23
Use SimpleFormatter for test runs

kaylareopelle:
`SimpleCov::Formatter::SimpleFormatter` does not add a line about coverage
percentages at the end of every test run. It also does not create an HTML
output file. Instead, it only generates JSON files with the result
information and the last run data.

We don't use the HTML information very frequently for individual test runs, we can get that by running `coverage:report` if it's needed.


tannalynn (approved):

fallwith (approved):
👍

PR #1343, opened 2022-08-23
Update multiverse run message

kaylareopelle:
The message that read something like: 
```
Running "/Users/jbunch/git/public/newrelic-ruby-agent/test/multiverse/suites/grpc" in 2 environments in serial
``` 

was slightly misleading, as the `2 environments` would be from a single Envfile-produced Gemfile for both `chain` and `prepend` instrumentation permutations. It would also get re-printed when the next instrumentation permutation would be executed.
    
Now, the message reads something like:
```
Running "logger" suite in 1 CHAIN environment and 1 PREPEND environment in serial
```

and will only be printed at the top of the suite's run.

### DIFF
--- a/test/multiverse/lib/multiverse/suite.rb
+++ b/test/multiverse/lib/multiverse/suite.rb
@@ -422,10 +422,6 @@ module Multiverse
       return unless check_environment_condition
       configure_instrumentation_method(instrumentation_method)
 
-      label = should_serialize? ? 'serial' : 'parallel'
-      env_count = filter_env ? 1 : environments.size
-      puts yellow("\nRunning #{directory.inspect} in #{env_count} environments in #{label}")
-
       environments.before.call if environments.before
       if should_serialize?
         execute_serial(instrumentation_method)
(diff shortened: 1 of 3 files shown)


test/multiverse/lib/multiverse/suite.rb
```diff
@@ -681,6 +677,16 @@ def verbose?
...
     end
+
+    def execution_message
+      label = should_serialize? ? 'serial' : 'parallel'
+      env_count = filter_env ? 1 : environments.size
+      env_plural = env_count > 1 ? 'environments' : 'environment'
+      opening = "\nRunning \"#{suite}\" suite in"
+      body = environments.instrumentation_permutations.map { |p| "#{env_count} #{p.upcase} #{env_plural}" }
+      ending = "in #{label}"
+      message = [opening, body.join(' and '), ending].join(' ')
```
fallwith, line 688: You're about 1 byte away from me recommending that we introduce the [new HAML](https://github.com/haml/haml/releases/tag/v6.0.0.beta.1) for this much formatting! 😆

fallwith (approved):
Thanks for this, much appreciated!

PR #1346, opened 2022-08-24
Use assert_nil for compatibility with Minitest 6

kaylareopelle:
# Overview
Addresses deprecation warning when running using Minitest 5.16.3: 

```
DEPRECATED: Use assert_nil if expecting nil from /Users/kreopelle/dev/newrelic-ruby-agent/test/new_relic/agent/system_info_test.rb:219. This will fail in Minitest 6.
DEPRECATED: Use assert_nil if expecting nil from /Users/kreopelle/dev/newrelic-ruby-agent/test/new_relic/agent/system_info_test.rb:220. This will fail in Minitest 6.
```

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/test/new_relic/agent/system_info_test.rb
+++ b/test/new_relic/agent/system_info_test.rb
@@ -216,8 +216,8 @@ class NewRelic::Agent::SystemInfoTest < Minitest::Test
     NewRelic::Agent::SystemInfo.stub(:sysctl_value, 1) do
       NewRelic::Agent::SystemInfo.processor_info_bsd
       info = NewRelic::Agent::SystemInfo.instance_variable_get(:@processor_info)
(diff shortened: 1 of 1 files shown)


tannalynn (approved):

fallwith (commented):
The test in question doesn't appear to differentiate between the info hash key/val pair missing entirely and there being a key present with a nil value. Is that ok given what we know about the info hash?

kaylareopelle:
@fallwith - Sorry I missed this yesterday! I looked into it a bit and it seems that a nil value caused by either scenario is ok for this hash. Great question!

PR #1348, opened 2022-08-25
CI: set the Mongo tag to v5

fallwith:
with Mongo 6 out and causing issues with the older Rubies and older
Mongo gem versions, use 5 for now


fallwith:
closing in favor of #1350

PR #1350, opened 2022-08-25
CI: lock mongo tag to a known good v5 version

fallwith:
with mongo:latest now pointing to v6, we have some work to do to figure
out our compatibility stance on Mongo v6 given that some of our older
Ruby versions are failing CI checks with it. it's possible that the
failures are exclusively limited to very old versions of the mongo gem
that are no longer supported anyhow. Until we determine if that's the
case, let's lock the mongo image tag to v5.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,7 +124,7 @@ jobs:
           --health-timeout 5s
           --health-retries 5
       mongodb:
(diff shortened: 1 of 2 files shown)


tannalynn (approved):
Thanks for getting this changed so quick!

kaylareopelle (approved):

PR #1359, opened 2022-08-27
CI: Use Ubuntu 22.04 LTS w/Ruby v3.1

fallwith:
updates to have our CI based entirely on Ubuntu 22, with Ruby v3.1 used
by default

- Update the CI Bundler setup scripts
  - Ruby 2.2 needs `update_rubygems`, not `gem update --system`
  - The setup-ruby action appears to leave us with the desired Bundler
    version already installed. Check the version and don't bother with
    installations and uninstallations if it exists
  - Add debugging statements everywhere
  - Don't attempt `bundle config` until the desired Bundler version is
    known to be present
- Stop testing Datamapper with Ruby 2.2, to avoid Ruby crashes
- Stop testing Rails 2.3 with Ruby 2.5 (Ruby 2.2, 2.3, and 2.4 will still test it)
- Update the Mongo suite to fall back to a default Mongo version if a
  `mongo` executable can't be found in `PATH`
- Don't test `redis` v5 yet

resolves #1112

PR #1360, opened 2022-08-29
3 2 1 lets jammy jellyfish

fallwith:
Angels banished from heaven have no choice but to become devils.


fallwith:
Closing in favor of #1359

PR #1363, opened 2022-08-30
CI: refine apt-get package installation

fallwith:
- Refine the apt-get based installation of Ubuntu packages
    - Don't waste time performing `apt-get update`
    - Only install those packages that we require that are not already present on the image
    - Annotate why the packages are needed
 - Curb: stop installing the `addressable` gem. The agent no longer uses it, and installing it causes crashes for certain Rubies
 - httprb: Testing versions 2 and 3 with Ruby 2.2 on Ubuntu 22 causes Ruby 2.2 to segfault occasionally. Stop testing httprb with Ruby 2.2 altogether.
 - Rails: Rails 4.2 introduces a requirement on Nokogiri, and Ruby 2.2 requires a very old version of Nokogiri that has OS package dependencies. Restrict Ruby 2.2 to testing Rails 4.1 and below.


fallwith:
let's see how many of these we actually need. maybe we get lucky and we don't need any...

.github/workflows/ci_cron.yml
```diff
@@ -38,9 +38,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Install required packages
```
tannalynn, line 41: Have you forced this ci_cron to be run yet? Since it doesn't run on PRs, I'm worried that there could be failures after merging from changing this that we haven't seen yet. If you haven't yet, maybe adding a PR trigger to here temporarily would give us a better idea of how this will affect all the ruby versions

fallwith: Good call. Originally I was only changing the initial `apt-get` call for the unit tests. I went ahead and updated the `apt-get` call for the multiverse tests as well, and got everything (ci_cron included) working with #1365. I have updated this PR's description to accurately reflect everything that changed. We ended up being able to remove the packages needed to compile Ruby from source, but had to retain 3 packages for instrumented gems.

tannalynn (approved):
Thanks for digging into this!

PR #1364, opened 2022-08-31
CI: address warnings

fallwith:
- ActionControllerSubscriber instrumentation: don't attempt to grab @req
  unless it's known to exist
- gemspec: remove the unused listen, rb-inotify, pry-nav,
  pry-stack_explorer, guard, and guard-minitest. less surface area =
  fewer issues
- update mocha in the hopes of addressing warnings from v1.9
- replace the removed pry gems with pry itself, but only for local dev
- stop (redundantly?) monkeypatching MiniTest assert_match
- stop shadowing outer variables


kaylareopelle (approved):
🪓

Issue #1309, opened 2022-08-03
CPU utilization and Memory usage by host missing on frontend pods


texpert:
^^ Provide a general summary of the issue in the title above. ^^ 

## Description
In the APM Summary dashboard, CPU utilization and Memory usage by host missing on frontend pods, though this data is present for the backend ones - see the screenshot:

<img width="1356" alt="Screenshot 2022-08-04 at 02 45 28" src="https://user-images.githubusercontent.com/2990033/182733866-9e890391-d3a0-46f7-975c-97c6b4408519.png">

## Expected Behavior
Expected to see the charts for the frontend pods as well

## [Troubleshooting]

Here's a suspected failure on Puma middleware tracing try from the NR agent debug log:

```
[2022-08-03 23:04:09 +0000 frontend-f45dc9c7f-l6wfn (1)] DEBUG : Creating Ruby Agent worker thread.
[2022-08-03 23:04:09 +0000 frontend-f45dc9c7f-l6wfn (1)] DEBUG : Creating New Relic thread: Worker Loop
[2022-08-03 23:04:09 +0000 frontend-f45dc9c7f-l6wfn (1)] DEBUG : Connecting Process to New Relic: /bundle/ruby/2.7.0/bin/puma
[2022-08-03 23:04:09 +0000 frontend-f45dc9c7f-l6wfn (1)] DEBUG : NameError : wrong constant name rails - when accessing config key agent_enabled
[2022-08-03 23:04:10 +0000 frontend-f45dc9c7f-l6wfn (1)] ERROR : Exception during Transaction#finish
[2022-08-03 23:04:10 +0000 frontend-f45dc9c7f-l6wfn (1)] ERROR : NoMethodError: undefined method `name' for nil:NilClass
[2022-08-03 23:04:10 +0000 frontend-f45dc9c7f-l6wfn (1)] DEBUG : Debugging backtrace:
/bundle/ruby/2.7.0/gems/newrelic_rpm-8.9.0/lib/new_relic/agent/transaction/abstract_segment.rb:231:in `force_finish'
  /bundle/ruby/2.7.0/gems/newrelic_rpm-8.9.0/lib/new_relic/agent/transaction/abstract_segment.rb:89:in `finalize'
  /bundle/ruby/2.7.0/gems/newrelic_rpm-8.9.0/lib/new_relic/agent/transaction/tracing.rb:52:in `block in finalize_segments'
[...]

Issue #1238, opened 2022-07-01
RubyVM and Logging Metrics not appearing


kaylareopelle:
## Description
After setting `instrumentation.thread.tracing` to `true`, the API Evolution team noticed key information missing from the New Relic platform. The Logs graph on the APM Summary page appears empty. The RubyVMs page also shows 0 for its metrics. Other information, such as that related to transactions, appears as expected.

## Expected Behavior
* Expect Logging Summary table to be populated
* Expect Logging/lines/{severity} metric to be collected
* Expect Ruby
* Expect RubyVM metrics to be collected 

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) 
* Debug-level logs turned on - No errors being raised, Samplers correctly ingested, some metric events still being sent
* Audit logging turned on - Confirmed that no `Logging/lines` metrics are being gathered, no `RubyVM/*` metrics being gathered

## Steps to Reproduce
* Have not been able to reproduce the issue locally, however, the problem can be observed on ASGS (staging), an internal app.
  * permalink to summary page: https://staging.onenr.io/08dQeboLowe (Internal)
  * permalink to RubyVM page: https://staging.onenr.io/04ERPNql6QW (Internal)
 
## Your Environment
* Rails 7.0.3
* Ruby 3.1.1
* newrelic-infinite_tracing 8.8.0 + bugfix branch
* application_logging.local_decorating.enabled: true
* instrumentation.thread.tracing: true
* error_collector.max_backtrace_frames: 250

## Additional context

<img width="1792" alt="Screen Shot 2022-07-01 at 9 23 01 AM" src="https://user-images.githubusercontent.com/87386821/176933388-d790b4ae-7c66-444a-a740-e5ba2b608469.png">
<img width="1004" alt="Screen Shot 2022-07-01 at 9 24 22 AM" src="https://user-images.githubusercontent.com/87386821/176933577-e7f1ca04-d85f-4db7-9420-7497e87b9720.png">

## For Maintainers Only or Hero Triaging this bug
*Suggested Priority (P1,P2,P3,P4,P5): P3
*Suggested T-Shirt size (S, M, L, XL, Unknown): Unknown at this time

fallwith:
Linking this one to #1309

PR #1373, opened 2022-09-01
Bugfix: thread tracing causes missing unscoped metrics

tannalynn:

# Overview
Missing unscoped metrics were found to be erroneously marked as an unscoped metric on a long running puma transaction that does not finish. This pr makes the following changes to prevent any unscoped metrics from being attached to transactions incorrectly. 
- All Threads created by the agent now use AgentThread (infinite tracing was not previously)
- AgentThread disables tracing when spawning threads
- When tracing threads
   -  the agent will no longer copy over the current_transaction if tracing is disabled
   - the agent will not attempt to create a segment if there is no current_transaction
- Added test that confirms any AgentThread spawned during the execution of a transaction will not be considered parto f the transaction

closes https://github.com/newrelic/newrelic-ruby-agent/issues/1238
closes https://github.com/newrelic/newrelic-ruby-agent/issues/1309

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,17 @@
 # New Relic Ruby Agent Release Notes #
 
 
+  ## v8.11.0
+
+
+  * **Bugfix: Missing unscoped metrics when instrumentation.thread.tracing is enabled**
+    
+    Previously, when `instrumentation.thread.tracing` was set to true, some puma applications encountered a bug where a varying number of unscoped metrics would be missing. The agent now will correctly store and send all unscoped metrics.
+    
+    Thank you to @texpert for providing details of their situation to help resolve the issue.
+  
+  
+
   ## v8.10.0
 
 
--- a/infinite_tracing/lib/infinite_tracing/worker.rb
+++ b/infinite_tracing/lib/infinite_tracing/worker.rb
@@ -52,7 +52,7 @@ module NewRelic::Agent
 
       def start_thread
         NewRelic::Agent.logger.debug("starting worker #{@name} thread...")
-        @worker_thread = Thread.new do
+        @worker_thread = NewRelic::Agent::Threading::AgentThread.create("infinite_tracing_worker") do
           catch(:exit) do
             begin
               @job.call
--- a/infinite_tracing/test/infinite_tracing/worker_test.rb
+++ b/infinite_tracing/test/infinite_tracing/worker_test.rb
@@ -25,6 +25,12 @@ module NewRelic::Agent::InfiniteTracing
       assert_metrics_recorded "Supportability/InfiniteTracing/Worker"
     end
 
+    def test_worker_uses_agentthread
+      NewRelic::Agent::Threading::AgentThread.expects(:create).returns(Thread.new {})
+      worker = Worker.new("simple") {}
+      worker.join
+    end
+
     def test_worker_handles_errors
       worker = Worker.new("error") do
         NewRelic::Agent.record_metric("Supportability/InfiniteTracing/Worker", 0.0)
--- a/lib/new_relic/agent/threading/agent_thread.rb
+++ b/lib/new_relic/agent/threading/agent_thread.rb
@@ -8,18 +8,18 @@ module NewRelic
     module Threading
       class AgentThread
         def self.create(label, &blk)
-          ::NewRelic::Agent.logger.debug("Creating New Relic thread: #{label}")
+          ::NewRelic::Agent.logger.debug("Creating AgentThread: #{label}")
           wrapped_blk = Proc.new do
             ::NewRelic::Agent.logger.warn("AgentThread created with current transaction #{::Thread.current[:newrelic_tracer_state].current_transaction.best_name}") if ::Thread.current[:newrelic_tracer_state] && ::Thread.current[:newrelic_tracer_state].current_transaction
             begin
               blk.call
             rescue => e
-              ::NewRelic::Agent.logger.error("Thread #{label} exited with error", e)
+              ::NewRelic::Agent.logger.error("AgentThread #{label} exited with error", e)
             rescue Exception => e
-              ::NewRelic::Agent.logger.error("Thread #{label} exited with exception. Re-raising in case of interrupt.", e)
+              ::NewRelic::Agent.logger.error("AgentThread #{label} exited with exception. Re-raising in case of interrupt.", e)
               raise
             ensure
-              ::NewRelic::Agent.logger.debug("Exiting New Relic thread: #{label}")
+              ::NewRelic::Agent.logger.debug("Exiting AgentThread: #{label}")
             end
           end
           thread = nil
--- a/lib/new_relic/agent/threading/agent_thread.rb
+++ b/lib/new_relic/agent/threading/agent_thread.rb
@@ -10,7 +10,10 @@ module NewRelic
         def self.create(label, &blk)
           ::NewRelic::Agent.logger.debug("Creating AgentThread: #{label}")
           wrapped_blk = Proc.new do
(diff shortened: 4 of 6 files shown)


lib/new_relic/agent/threading/agent_thread.rb
```diff
@@ -10,6 +10,7 @@ class AgentThread
         def self.create(label, &blk)
           ::NewRelic::Agent.logger.debug("Creating New Relic thread: #{label}")
           wrapped_blk = Proc.new do
+            ::NewRelic::Agent.logger.warn("AgentThread created with current transaction #{::Thread.current[:newrelic_tracer_state].current_transaction.best_name}") if ::Thread.current[:newrelic_tracer_state] && ::Thread.current[:newrelic_tracer_state].current_transaction
```
kaylareopelle, line 13: Why did you decide to print to the `warn` level log here?

kaylareopelle: I like that you called this AgentThread since that's the name of the class. Even though it's not part of your changes, if you're up for it, I'd recommend changing the references in the other log messages from `New Relic thread` to `AgentThread`.

test/new_relic/agent/threading/agent_thread_test.rb
```diff
@@ -157,6 +157,17 @@ def test_scrub_backtrace_handles_nil_backtrace
...
+        in_transaction do |txn|
+          t = AgentThread.create("label") do
+            refute ::Thread.current[:newrelic_tracer_state] && ::Thread.current[:newrelic_tracer_state].current_transaction, "Agent thread should not contain a current transaction"
+          end
+          t.join
+        end
+      end
+    end
+
     DONT_CARE = true
```
kaylareopelle, line 171: lol wut?

test/new_relic/agent/threading/agent_thread_test.rb
```diff
@@ -157,6 +157,17 @@ def test_scrub_backtrace_handles_nil_backtrace
...
+    def test_agent_thread_creation_ignores_current_transaction
+      with_config(:'instrumentation.thread.tracing' => true) do
+        in_transaction do |txn|
+          t = AgentThread.create("label") do
+            refute ::Thread.current[:newrelic_tracer_state] && ::Thread.current[:newrelic_tracer_state].current_transaction, "Agent thread should not contain a current transaction"
+          end
+          t.join
+        end
+      end
+    end
```
kaylareopelle, line 169: Yay! Nice test! Does this fail without your changes?

CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
...
 
 
+  ## v8.11.0
+
+
+  * **Bugfix: Missing unscoped metrics when instrumentation.thread.tracing is enabled**
+    
+    Previously, when `instrumentation.thread.tracing` was set to true, some puma applications encountered a bug where a varying number of unscoped metrics would be missing. The agent now will correctly store and send all unscoped metrics.
+    
+    Thank you to @texpert for providing details of their situation to help resolve the issue.
```
kaylareopelle, line 11: Might be nice to find Chuck and Kinjal's public GH handles and thank them here too! They did a lot of early debugging with me on this.

infinite_tracing/lib/infinite_tracing/worker.rb
```diff
@@ -52,7 +52,7 @@ def stop
 
       def start_thread
         NewRelic::Agent.logger.debug("starting worker #{@name} thread...")
-        @worker_thread = Thread.new do
+        @worker_thread = NewRelic::Agent::Threading::AgentThread.create("infinite_tracing_worker") do
```
kaylareopelle, line 55: Could 8T use a new test too?

kaylareopelle (commented):
I'm so happy for you! This is amazing, Tanna. WALUIGI FOREVER! 💜

test/new_relic/agent/threading/agent_thread_test.rb
tannalynn: LOL i know right, i have no idea why some of the tests were set up with that

lib/new_relic/agent/threading/agent_thread.rb
tannalynn: I did warn level because if it does happen it's not immediately breaking anything, which is what I would use error for. But it can still cause issues in some applications, so seemed more important than info or debug level. 

And sure, I can make that change, I think that makes sense!

test/new_relic/agent/threading/agent_thread_test.rb
tannalynn: It does! I switched between dev and this branch and was able to confirm it fails on dev.

lib/new_relic/agent/threading/agent_thread.rb
tannalynn: resolved in 32ca4edde262e868cc7098c2774cb691d69826b9

infinite_tracing/lib/infinite_tracing/worker.rb
tannalynn: added in e76793206b045c5a16d3295891aa4f783916e47b

infinite_tracing/test/infinite_tracing/worker_test.rb
```diff
@@ -25,6 +25,12 @@ def test_processes_simple_task
       assert_metrics_recorded "Supportability/InfiniteTracing/Worker"
     end
 
+    def test_worker_uses_agentthread
```
fallwith, line 28: I'd like for this test to guarantee that calling `Worker.new` actually triggers ` NewRelic::Agent::Threading::AgentThread.create`.

From what I can tell, the use of Mocha's `expects` here will yield `Thread.new` as desired in the event that `create` gets called, but there's no guarantee or assertion that it's being called.

To look at it another way, this test should fail when ran against an earlier version of the code base that did not cause `Worker.new` to create a new agent thread.

kaylareopelle (approved):
Thanks for the updates!

lib/new_relic/agent/threading/agent_thread.rb
fallwith: Do we not have a line length linter check? I think for now it'd be nice to stick to at least an upper boundary of 120 columns.

infinite_tracing/test/infinite_tracing/worker_test.rb
tannalynn: That is how `.expects` works., it will fail if the method you are expecting is not called. Here is an example failure if I revert the worker change:
```
  1) Failure:
NewRelic::Agent::InfiniteTracing::ResponseHandlerTest#test_worker_uses_agentthread [/Users/tmcclure/projects/ruby-agent/infinite_tracing/test/infinite_tracing/worker_test.rb:29]:
not all expectations were satisfied
unsatisfied expectations:
- expected exactly once, not yet invoked: NewRelic::Agent::Threading::AgentThread.create(any_parameters)
```

lib/new_relic/agent/threading/agent_thread.rb
tannalynn: updated in 011b14e6ec9f9bc4567aa3d849866c7fc876021e

infinite_tracing/test/infinite_tracing/worker_test.rb
fallwith: arg! Before I submitted my comment, I took the new test and added it to `worker_test.rb` in dev, ran the 8T tests, and everything passed. So I thought the test wasn't sufficient. My apologies - I must have not ran the tests correctly. As long as the test is dependent on the change being present (as it is), we're good. Thank you.

fallwith (approved):

PR #1374, opened 2022-09-01
Set failedThreshold to 90

kaylareopelle:
# Overview

We've had pretty consistent 92% coverage since the latest SimpleCov updates. Since we can expect that number, increase the threshold to 90% to ensure coverage does not drop too far.


fallwith (approved):
👍

PR #1375, opened 2022-09-01
RuboCop: stop permitting infinitely long lines

fallwith:
Enforce a line length max with RuboCop, with a TODO to actually have the
value be reasonable in the future


kaylareopelle (approved):
Thanks for getting this process started!

PR #1377, opened 2022-09-01
Use HTMLFormatter in rake task

kaylareopelle:
# Overview

We use the SimpleFormatter by default for regular test runs. The HTML
format is most useful to use when we generate a collated report.
This PR will make sure HTML is created when the rake task is called.


fallwith (approved):
Fantastic! I look forward to seeing what #1378 brings with this improvement!

PR #1380, opened 2022-09-01
condense failures to one notification

hannahramadan:
As it currently stands, each job failure produces a notification. This could lead to multiple of the same slack notification per CI run. To reduce noise, this commit turns notifications into a separate job that relies on the completion of all test runs.

# Overview

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
         # - xslt is needed for older Nokogiris, RUBY_VERSION < 2.5
         # - sasl is needed for memcached
       - name: Install OS packages
-        run: sudo apt-get install -y --no-install-recommends libcurl4-nss-dev libsasl2-dev libxslt1-dev
+        run: sudo apt-get update; sudo apt-get install -y --no-install-recommends libcurl4-nss-dev libsasl2-dev libxslt1-dev
 
       - name: Install Ruby ${{ matrix.ruby-version }}
         uses: ruby/setup-ruby@v1
@@ -160,7 +160,7 @@ jobs:
         # - xslt is needed for older Nokogiris, RUBY_VERSION < 2.5
         # - sasl is needed for memcached
       - name: Install OS packages
-        run: sudo apt-get install -y --no-install-recommends libcurl4-nss-dev libsasl2-dev libxslt1-dev
+        run: sudo apt-get update; sudo apt-get install -y --no-install-recommends libcurl4-nss-dev libsasl2-dev libxslt1-dev
 
       - name: Install Ruby ${{ matrix.ruby-version }}
         uses: ruby/setup-ruby@v1
--- a/.github/workflows/ci_cron.yml
+++ b/.github/workflows/ci_cron.yml
@@ -343,7 +343,7 @@ jobs:
         uses: ./.github/actions/annotate
 
   notify_slack_fail:
-    needs: unit_tests multiverse infinite_tracing jruby_multiverse
+    needs: [unit_tests, multiverse, infinite_tracing, jruby_multiverse]
     runs-on: ubuntu-18.04
     steps:
     - name: Notify slack fail
--- a/.github/workflows/ci_cron.yml
+++ b/.github/workflows/ci_cron.yml
@@ -343,8 +343,10 @@ jobs:
         uses: ./.github/actions/annotate
 
   notify_slack_fail:
+    needs: unit_tests multiverse infinite_tracing jruby_multiverse
+    runs-on: ubuntu-18.04
+    steps:
     - name: Notify slack fail
-      needs: unit_tests multiverse infinite_tracing jruby_multiverse
       if: ${{ success() }}
       env:
         SLACK_BOT_TOKEN: ${{ secrets.RUBY_GITHUB_ACTIONS_BOT_WEBHOOK }}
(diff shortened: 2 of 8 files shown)


fallwith (commented):
- Being able to condense things is fantastic
- The YAML syntax looks great
- Should we also notify on RuboCop failures?
- Have you tested this by artificially causing one of the jobs to fail and seeing the notification come through?

For that last point, my concern is that `${{ failure() }}` in the context of a job would be scoped to that job itself, and not to other jobs. Now that we have a new `notify_slack_fail` job, we don't want the condition to be related to that new job itself failing - we want the failure check to still be triggered by any of the other jobs.

.github/workflows/ci_cron.yml
```diff
@@ -1,12 +1,7 @@
 name: Scheduled Continuous Integration
 
 on:
-  push:
```
kaylareopelle, line 4: Reminder! Revert this change before merging!

.github/workflows/ci_cron.yml
```diff
@@ -377,12 +342,16 @@ jobs:
...
+    dependsOn: [unit_tests, multiverse, infinite_tracing, jruby_multiverse]
+    condition: ${{ failure() }}
+    runs-on: ubuntu-22.04
+    steps:
+    - name: Notify slack fail
+      env:
+        SLACK_BOT_TOKEN: ${{ secrets.RUBY_GITHUB_ACTIONS_BOT_WEBHOOK }}
+      uses: voxmedia/github-action-slack-notify-build@v1
+      with:
+        channel: hannah-private-testing-channel
```
kaylareopelle, line 355: Reminder! Update channel name!

lib/new_relic/constants.rb
```diff
@@ -6,7 +6,7 @@
 module NewRelic
   PRIORITY_PRECISION = 6
 
-  EMPTY_ARRAY = [].freeze
+  # EMPTY_ARRAY = [].freeze
```
kaylareopelle, line 9: What is this change related to?

kaylareopelle (commented):
I think this concept looks great and will make for much more effective notifications! Great catch that different test runs will product different notifications. 

I'm not quite sure what exactly you were looking for in the review, so I added a few reminders and a question.

lib/new_relic/constants.rb
kaylareopelle: Oh, I get it now! Very clever! Great way to make multiple suites fail 😄

hannahramadan:
> * Being able to condense things is fantastic
> * The YAML syntax looks great
> * Should we also notify on RuboCop failures?
> * Have you tested this by artificially causing one of the jobs to fail and seeing the notification come through?
> 
> For that last point, my concern is that `${{ failure() }}` in the context of a job would be scoped to that job itself, and not to other jobs. Now that we have a new `notify_slack_fail` job, we don't want the condition to be related to that new job itself failing - we want the failure check to still be triggered by any of the other jobs.

Hey @fallwith you were right about my scoping issue! I've added an action that tracks the 'conclusions' of every job and has an overall 'conclusion' variable. Now the determining pass/fail is decided by `if: env.WORKFLOW_CONCLUSION == 'failure'`. And `if: always()` should ensure the Notify Slack Fail job always runs. Tested and it should work. 

As for RuboCop, since any typical PR will fail if RuboCop does, not sure that we need to include it (but I supposed it wouldn't hurt).

.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         # - xslt is needed for older Nokogiris, RUBY_VERSION < 2.5
         # - sasl is needed for memcached
       - name: Install OS packages
-        run: sudo apt-get install -y --no-install-recommends libcurl4-nss-dev libsasl2-dev libxslt1-dev
+        run: sudo apt-get update; sudo apt-get install -y --no-install-recommends libcurl4-nss-dev libsasl2-dev libxslt1-dev
```
hannahramadan, line 38: I merged a CI fix branch #1380 before it was merged into Dev so that my branch would pass.

.github/workflows/ci_cron.yml
```diff
@@ -377,11 +347,17 @@ jobs:
...
-        if: ${{ failure() }}
+  notify_slack_fail:
+    name: Notify slack fail
+    needs: [unit_tests, multiverse, infinite_tracing, jruby_multiverse]
+    runs-on: ubuntu-22.04
+    if: always()
+    steps:
+      - uses: technote-space/workflow-conclusion-action@v3
+      - uses: voxmedia/github-action-slack-notify-build@v1
+        if: env.WORKFLOW_CONCLUSION == 'failure'
```
hannahramadan, line 358: The 'technote' action tracks the 'conclusions' of every job and has an overall 'WORKFLOW_CONCLUSION' variable. Now the determining pass/fail is decided by if: env.WORKFLOW_CONCLUSION == 'failure'. And if: always() should ensure the Notify Slack Fail job always runs. 

This change is to limit failure notifications to 1 if the CI fails. VS multiple notifications if multiple jobs have failures.

fallwith (approved):
I'm glad you were able to get a condensed solution working, and thanks for the apt-get fix!

PR #1384, opened 2022-09-04
CI: deprecation warning fixes

fallwith:
address CI deprecation warnings

- upgrade actiona/checkout to v3
- actions/checkout performs `git init` before the checkout, which a) defaults to using a `master` branch, and b) emits a stream of "hint" warnings letting us know how to set a default branch name. Globally set a `main` branch name to address these items.
- Mocha: use an explicit `nil` response
- RabbitMQ: stop using the deprecated health check, switch to a modern "fine grained" check

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
         image: mysql:5.7
         env:
           MYSQL_ALLOW_EMPTY_PASSWORD: yes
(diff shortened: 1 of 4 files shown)


kaylareopelle (approved):
This looks great, James!

Your changes make me wonder if there's anything we can do to reduce the duplication across some of our jobs 🤔

PR #1386, opened 2022-09-06
Gabby: report only stable versions for 'previous'

fallwith:
When Gabby notifies us about a new gem version that has recently become
available, it would be preferable to:

- mention rc1/beta/preview releases for the newly released version to
  give the team a heads up about development work being done on the gem
- only report stable versions for the previously released version.

This way, if a gem goes `v1.5 -> v2.0.preview -> v2.0.rc1 -> v2.0.rc2`
then the notifications will always tell us about the previous stable
version, not the previous beta version.

Currently, the notifications look like this:

- v1.5 to v2.0.preview
- v2.0.preview to v2.0.rc1
- v2.0.rc1 to v2.0.rc2

With the included changes, the notifications would instead read:

- v1.5 to v2.0.preview
- v1.5 to v2.0.rc1
- v1.5 to v2.0.rc2

This way we are notified of beta releases and don't have to manually
search for the previous stable version.

### DIFF
--- a/newrelic_rpm.gemspec
+++ b/newrelic_rpm.gemspec
@@ -47,7 +47,7 @@ https://github.com/newrelic/newrelic-ruby-agent/
   s.homepage = "https://github.com/newrelic/rpm"
   s.require_paths = ["lib"]
   s.summary = "New Relic Ruby Agent"
-  s.add_development_dependency 'feedjira', '3.2.1' # for Gabby
+  s.add_development_dependency 'feedjira', '3.2.1' unless ENV['CI'] # for Gabby
   s.add_development_dependency 'rake', '12.3.3'
   s.add_development_dependency 'minitest', '5.3.3'
   s.add_development_dependency 'minitest-stub-const', '0.6'
@@ -58,5 +58,5 @@ https://github.com/newrelic/newrelic-ruby-agent/
   s.add_development_dependency 'rubocop'
   s.add_development_dependency 'rubocop-performance'
   s.add_development_dependency 'simplecov' if RUBY_VERSION >= '2.7.0'
-  s.add_development_dependency 'httparty' # for perf tests and Gabby
+  s.add_development_dependency 'httparty' unless ENV['CI'] # for perf tests and Gabby
 end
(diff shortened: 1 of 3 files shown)


newrelic_rpm.gemspec
```diff
@@ -47,6 +47,7 @@ https://github.com/newrelic/newrelic-ruby-agent/
   s.homepage = "https://github.com/newrelic/rpm"
   s.require_paths = ["lib"]
   s.summary = "New Relic Ruby Agent"
+  s.add_development_dependency 'feedjira', '3.2.1' unless ENV['CI'] # for Gabby
```
hannahramadan, line 50: The 'feedjira' and 'httparty' addition are saying that we don't want them installed if we're running the CI? But I would've thought that was the only place we'd need them. Can you help explain?

kaylareopelle (changes requested):

newrelic_rpm.gemspec
kaylareopelle: I'm also curious about this! I'll hold off final approval until you have a chance to answer.

fallwith: > The 'feedjira' and 'httparty' addition are saying that we don't want them installed if we're running the CI? But I would've thought that was the only place we'd need them. Can you help explain?

Yeah, good callout here @hannahramadan.

- Developers iterating on Gabby need to run the unit tests located at `test/new_relic/gem_notifications_tests.rb`. These tests require both `httparty` and `feedjira` to work. As such, it's appropriate that they exist as dev dependencies in the `.gemspec` file.
    - `httparty` was already present in `.gemspec` due to some fetching behavior that the perf tests are apparently capable of
    - `feedjira` became needed after Gabby started tracking CVEs
- The installation of `feedjira` causes issues with Ruby 2.2. For now, given that the gem notification tests are not currently exercised by our CI, I've decided to save us on trouble and time by not having the gems get installed.
- Gabby's own work performed as GitHub Actions jobs has no dependency on the Ruby agent, and never ends up installing the agent. As a result, the `.gemspec` file is never brought in, and its declaration of `httparty` and `feedjira` as dev dependencies is ignored. Gabby's GHA jobs use `gem install` commands directly to bring these gems in.
- Long term, I believe that Gabby should be moved out into a separate gem, complete with a separate `.gemspec` file, and a separate CI setup that exercises the unit tests. For now though, these changes give us a tiny boost in performance and simplification by stopping the installation of `httparty` when running CI, and gives devs the necessary `feedjira` gem for local dev.

kaylareopelle (approved):

PR #1396, opened 2022-09-07
CI: for rake + rails, use rack v2

fallwith:
when rake is tested with Rails, sprockets is brought in and has a `rack < 3` dependency. the multiverse suite helpers see the unpinned rack version and bring in `rackup`, which won't work with `rack < 3`.

lock rake + rails to rack v2 to let the multiverse suite helpers know to not bring in `rackup`.


fallwith:
This is a fix for the recent scheduled CI jobs failing for Ruby 2.4 and 2.5. Failure reproduced and fix confirmed with `rubypants 2.4` testing.

tannalynn (approved):
🎉

PR #1402, opened 2022-09-08
Create instrumentation generator

kaylareopelle:
# Overview

* Installs [thor](https://github.com/rails/thor) gem as a development dependency
* Adds `instrumentation` tasks:
  * `scaffold`: creates the skeleton of files needed to instrument a library
  * `add_new_method`: updates existing instrumentation to include a new method

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


newrelic_rpm.gemspec
```diff
@@ -47,16 +47,17 @@ https://github.com/newrelic/newrelic-ruby-agent/
...
-  s.add_development_dependency 'yard'
   s.add_development_dependency 'pry' unless ENV['CI']
-  s.add_development_dependency 'bundler'
+  s.add_development_dependency 'rake', '12.3.3'
   s.add_development_dependency 'rubocop'
   s.add_development_dependency 'rubocop-performance'
   s.add_development_dependency 'simplecov' if RUBY_VERSION >= '2.7.0'
-  s.add_development_dependency 'httparty' unless ENV['CI'] # for perf tests and Gabby
+  s.add_development_dependency 'thor'
+  s.add_development_dependency 'yard'
```
kaylareopelle, line 62: As our development dependencies grow, I thought alphabetizing them might be handy

lib/tasks/instrumentation_generator/README.md
```diff
@@ -0,0 +1,42 @@
...
+## Outcome
+A prototype outside the agent has been created that generates the required files to create new instrumentation. This prototype accepts three arguments: name (name of the library), method (method to instrument), args (arguments for the method).
+
+The Ruby gem Thor, a toolkit for building powerful command-line interfaces used in Bundler, Vagrant, Rails and others powers this CLI.
+
+## Usage
+
+`thor list` can be passed to get a list of all available thor commands
+
+### instrumentation:scaffold
```
tannalynn, line 28: I'm not familiar with thor. Would the command we'd run be `thor instrumentation:scaffold gem_name --method method_name --args first,second`? It might be nice to add an example of exactly what we'd want to run somewhere?

lib/tasks/instrumentation_generator/instrumentation.thor
```diff
@@ -0,0 +1,86 @@
...
+          :type => String,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => 'Controls auto-instrumentation of the #{name.capitalize} library at start up. May be one of [auto|prepend|chain|disabled].'
+        },
+    CONFIG
+    insert_into_file(
+      'lib/new_relic/agent/configuration/default_source.rb',
+      config,
+      after: ":description => 'Controls auto-instrumentation of bunny at start up.  May be one of [auto|prepend|chain|disabled].'
```
tannalynn, line 80: This is very neat!

lib/tasks/instrumentation_generator/templates/newrelic.yml.tt
```diff
@@ -0,0 +1,19 @@
+---
+development:
+  error_collector:
+    enabled: true
+  apdex_t: 0.5
+  monitor_mode: true
+  license_key: bootstrap_newrelic_admin_license_key_000
+  instrumentation:
+    <%= @name.downcase %>: <%= $instrumentation_method %>
```
tannalynn, line 9: I'm curious about this line. Does thor know to put something in the name spot, but also knows to skip the instrumentation_method part?

Thorfile
```diff
@@ -0,0 +1,6 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
```
fallwith, line 4: I was curious to know if RuboCop recognizes `Thorfile` as Ruby and it does! Nice.

lib/tasks/instrumentation_generator/README.md
```diff
@@ -0,0 +1,42 @@
...
+This task requires one parameter by default: the name of the library or class you are instrumenting. This task generates the basic file structure needed to add new instrumentation to the Ruby agent.
+
+It accepts two optional parameters:
+* `--method`: the name of a method to instrument. Only one method can be accepted at this time. Defaults to `method_to_instrument`
+* `--args`: the arguments for the instrumented method. A comma-separated list can be given if there are multiple arguments. Defaults to `*args`.
+
+### WIP: instrumentation:add_new_method
+This task inserts a new method existing piece of instrumentation. It requires one argument, the name of the existing instrumentation to add the method to. If the instrumented library does not already exist, it will create a new scaffold for that library.
```
fallwith, line 36: I'm not clear on what this first sentence is trying to say.

lib/tasks/instrumentation_generator/README.md
fallwith: Good call. And Kayla's got it covered in `TODO.md` for the future.

lib/tasks/instrumentation_generator/instrumentation.thor
```diff
@@ -0,0 +1,86 @@
...
+    desc: 'The method you would like to prepend or chain instrumentation onto'
+  option :args,
+    default: '*args',
+    desc: 'The arguments associated with the original method'
+
+  def scaffold(name)
+    @name = name
+    @method = options[:method] if options[:method]
+    @args = options[:args] if options[:args]
+    base_path = "lib/new_relic/agent/instrumentation/#{name.downcase}"
```
fallwith, line 28: Perhaps an `INSTRUMENTATION_ROOT` constant would be useful here.

lib/tasks/instrumentation_generator/instrumentation.thor
fallwith, line 80: `bunny` => `#{name.capitalize}`

lib/tasks/instrumentation_generator/templates/Envfile.tt
```diff
@@ -0,0 +1,5 @@
+instrumentation_methods :chain, :prepend
```
fallwith, line 1: needs the frozen string literal magic comment

lib/tasks/instrumentation_generator/templates/chain.tt
```diff
@@ -0,0 +1,21 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
```
fallwith, line 3: needs the frozen string literal magic comment

lib/tasks/instrumentation_generator/templates/dependency_detection.tt
```diff
@@ -0,0 +1,29 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
```
fallwith, line 3: needs the frozen string literal magic comment

lib/tasks/instrumentation_generator/templates/dependency_detection.tt
```diff
@@ -0,0 +1,29 @@
...
+require_relative '<%= @name.downcase %>/chain'
+require_relative '<%= @name.downcase %>/prepend'
+
+DependencyDetection.defer do
+  named :<%= @name.downcase %>
+
+  depends_on do
+    # The class that needs to be defined to prepend/chain onto. This can be used
+    # to determine whether the library is installed.
+    defined?(::<%= @name.capitalize %>)
```
fallwith, line 15: We might need the equivalent of ActiveSupport's `#classify` here. The agent itself maintains something of a fork of that method that you could use.

```ruby
irb(main):001:0> 'an_example'.capitalize
=> "An_example"
irb(main):002:0> 'an_example'.classify
=> "AnExample"
```

lib/tasks/instrumentation_generator/templates/dependency_detection.tt
```diff
@@ -0,0 +1,29 @@
...
+    # should be installed
+  end
+
+  executes do
+    ::NewRelic::Agent.logger.info "Installing <%= @name.capitalize %> instrumentation"
+
+    if use_prepend?
+      prepend_instrument ::<%= @name.capitalize %>, NewRelic::Agent::Instrumentation::<%= @name.capitalize %>::Prepend
+    else
+      chain_instrument NewRelic::Agent::Instrumentation::<%= @name.capitalize %>
```
fallwith, line 26: Hmm... thinking about my preference for something like `#classify` over `#capitalize`, perhaps we could just perform the operation once and stick it in an instance var like `@class_name`

lib/tasks/instrumentation_generator/templates/newrelic.yml.tt
fallwith, line 9: Puzzled by the use of global here. I'm curious as well.

lib/tasks/instrumentation_generator/templates/instrumentation.tt
```diff
@@ -0,0 +1,12 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
```
fallwith, line 3: needs the frozen string literal magic comment

lib/tasks/instrumentation_generator/templates/prepend.tt
```diff
@@ -0,0 +1,13 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
```
fallwith, line 3: needs the frozen string literal magic comment

newrelic_rpm.gemspec
fallwith: Thank you

fallwith (changes requested):

lib/tasks/instrumentation_generator/README.md
kaylareopelle: Great point! I think this seems like a pretty crucial item to add to the README at this stage. I'll incorporate it as part of this review.

lib/tasks/instrumentation_generator/instrumentation.thor
kaylareopelle: I thought the same earlier today, and forgot what this line is doing! We need to reference a line that exists in the `default_source.rb` file to append our new config value after it. Bunny was the instrumentation that came to mind, though something like `thread` or `logger` may be better to replace it with. It was pretty difficult to append to the end of the overall hash, so sticking the config block in the middle seemed like a good-enough solution.

lib/tasks/instrumentation_generator/templates/dependency_detection.tt
kaylareopelle: I didn't realize the agent had a `#classify` overwrite! Great suggestion! I was contemplating making a `.camelize` dupe, but decided against it and added a TODO instead. I'll prioritize this change.

lib/tasks/instrumentation_generator/templates/dependency_detection.tt
kaylareopelle: Ooh, that's a nice idea!

lib/tasks/instrumentation_generator/templates/newrelic.yml.tt
kaylareopelle: Great catch! This isn't working as expected. I'll fix it 😄

kaylareopelle: The global is used in our multiverse suite to swap between `chain` and `prepend`. 

https://github.com/newrelic/newrelic-ruby-agent/blob/1d2016d66e95a2d09c233f6cd2cdae0f696398c2/test/multiverse/lib/multiverse/suite.rb#L634-L636

As mentioned above, it's not working with this template (it's totally blank when `instrumentation:scaffold` is run)

lib/tasks/instrumentation_generator/templates/Envfile.tt
fallwith: I took a crack at getting the templates linted automatically with:

```yaml
AllCops:
  Include:
    - 'lib/tasks/**/templates/*.tt'
```

only to be reminded that RuboCop still can't handle stumbling upon an ERB tag inside of a file. 😞

I suppose for this PR's scope (the templates are in Ruby and render even more Ruby!), that inner layer of Ruby wouldn't be getting linted anyhow.

lib/tasks/instrumentation_generator/templates/dependency_detection.tt
kaylareopelle: I'm not finding the agent's fork. Could you help point me in the right direction?

fallwith: Ah, I apologize - I got `#constantize` mixed up with `#classify`. `lib/new_relic/language_support.rb` has a fork of `#constantize`.

Here, you want to be able to produce a camel cased constant string without needing to actually be able to retrieve a constant from the constants table (because the instrumented gem is likely not loaded), so `#classify` is better.

We likely don't need the table stuff from `#classify`, so maybe a fork of `#camelize` would suffice.

lib/tasks/instrumentation_generator/templates/prepend.tt
kaylareopelle: Resolved in: 94a9deec5307f21a2490f7a4e12e4276e411999b

lib/tasks/instrumentation_generator/templates/instrumentation.tt
kaylareopelle: Resolved in: 94a9deec5307f21a2490f7a4e12e4276e411999b

lib/tasks/instrumentation_generator/templates/dependency_detection.tt
kaylareopelle: Resolved in: 94a9deec5307f21a2490f7a4e12e4276e411999b

lib/tasks/instrumentation_generator/templates/chain.tt
kaylareopelle: Resolved in: 94a9deec5307f21a2490f7a4e12e4276e411999b

lib/new_relic/language_support.rb
```diff
@@ -73,6 +73,11 @@ def constantize(const_name)
       end
     end
 
+    def camelize(string)
+      camelized = string.downcase
+      camelized.split(/\-|\_/).map(&:capitalize).join
```
fallwith, line 78: Hmm... I was just about to say that you don't need to use escaping here and just go with `/-|_/`, but according to `Regexp.escape`'s output, it does like to escaping the `-` character. TIL.

lib/tasks/instrumentation_generator/instrumentation.thor
```diff
@@ -0,0 +1,95 @@
...
+    desc: 'The method you would like to prepend or chain instrumentation onto'
+  option :args,
+    default: '*args',
+    desc: 'The arguments associated with the original method'
+
+  def scaffold(name)
+    @name = name
+    @method = options[:method] if options[:method]
+    @args = options[:args] if options[:args]
+    @class_name = ::NewRelic::LanguageSupport.camelize(name)
```
kaylareopelle, line 36: @fallwith - your earlier comment about adding `@class_name` and creating a `camelize` method resolved by: 0150cc311ddc10051192bbb6a92a61af5a68d0bc and eea682478e8d1eb60efcf7e8bd53bd57b1c9c4c5

lib/tasks/instrumentation_generator/instrumentation.thor
```diff
@@ -0,0 +1,95 @@
...
+
+require_relative '../../new_relic/language_support'
+require 'thor'
+
+class Instrumentation < Thor
+  include Thor::Actions
+
+  INSTRUMENTATION_ROOT = 'lib/new_relic/agent/instrumentation/'
+  MULTIVERSE_SUITE_ROOT = 'test/multiverse/suites/'
+  DEFAULT_SOURCE_LOCATION = 'lib/new_relic/agent/configuration/default_source.rb'
```
kaylareopelle, line 14: Constant recommendation added with: eea682478e8d1eb60efcf7e8bd53bd57b1c9c4c5

lib/tasks/instrumentation_generator/templates/Envfile.tt
kaylareopelle: Literal comment added in: 7b628d969328a75e5059dd59529eb763958bd8ff

lib/tasks/instrumentation_generator/README.md
kaylareopelle: Resolved in: 5e4deea121cd519bdad166717f6467a65162ff9a

lib/tasks/instrumentation_generator/README.md
```diff
@@ -1,42 +1,63 @@
...
+Run it using:
+
+`thor instrumentation:scaffold gem_name`
+
 It accepts two optional parameters:
-* `--method`: the name of a method to instrument. Only one method can be accepted at this time. Defaults to `method_to_instrument`
-* `--args`: the arguments for the instrumented method. A comma-separated list can be given if there are multiple arguments. Defaults to `*args`.
+
+- `--method`: the name of a method to instrument. Only one method can be accepted at this time. Defaults to `method_to_instrument`
+- `--args`: the arguments for the instrumented method. A comma-separated list can be given if there are multiple arguments. Defaults to `*args`.
```
fallwith, line 18: I might be misreading it here, but I can't find where the

> Defaults to `*args`.

statement given here is actually fulfilled in code. It seems like the actual default is `nil`, and the generated code ends up leaving of the method's parentheses entirely.

lib/tasks/instrumentation_generator/README.md
```diff
@@ -1,42 +1,63 @@
...
 
 `thor list` can be passed to get a list of all available thor commands
 
 ### instrumentation:scaffold
+
 This task requires one parameter by default: the name of the library or class you are instrumenting. This task generates the basic file structure needed to add new instrumentation to the Ruby agent.
 
+Run it using:
+
+`thor instrumentation:scaffold gem_name`
```
fallwith, line 13: How can I obtain this `thor` binary? Should the Thor gem be added as a dev dependency to our .gemspec file?

lib/tasks/instrumentation_generator/instrumentation.thor
```diff
@@ -1,10 +1,18 @@
...
 
+require_relative '../../new_relic/language_support'
 require 'thor'
 
 class Instrumentation < Thor
   include Thor::Actions
 
+  INSTRUMENTATION_ROOT = 'lib/new_relic/agent/instrumentation/'
+  MULTIVERSE_SUITE_ROOT = 'test/multiverse/suites/'
+  DEFAULT_SOURCE_LOCATION = 'lib/new_relic/agent/configuration/default_source.rb'
```
fallwith, line 14: these look great!

lib/tasks/instrumentation_generator/instrumentation.thor
fallwith: Ah, my bad. You're giving a target to a searcher, not rendering new content. "bunny" is just the needle to find in the haystack.

lib/tasks/instrumentation_generator/templates/newrelic.yml.tt
kaylareopelle: Resolved by: c75c8010404817ca14b55798000a97a889efc7dd

kaylareopelle: Resolved by: c75c8010404817ca14b55798000a97a889efc7dd

lib/tasks/instrumentation_generator/README.md
kaylareopelle: Thor will be added as a dev dependency to our .gemspec file as part of this PR! It's a bit hard to see with the alphabetizing changes. 

See: https://github.com/newrelic/newrelic-ruby-agent/pull/1402/files#diff-1099f92de9acf3d6d134894bd17f60b5be5b96668f0d30a36785fec0c37dae0bR61

lib/tasks/instrumentation_generator/instrumentation.thor
```diff
@@ -0,0 +1,96 @@
...
+    file structure needed to add new instrumentation to the Ruby agent.
+  LONGDESC
+
+  source_root(File.dirname(__FILE__))
+
+  option :method,
+    default: 'method_to_instrument',
+    desc: 'The method you would like to prepend or chain instrumentation onto'
+  option :args,
+    default: '*args',
```
kaylareopelle, line 29: @fallwith  - this is where `args` has its default value assigned.

lib/tasks/instrumentation_generator/instrumentation.thor
```diff
@@ -0,0 +1,96 @@
...
+    desc: 'The arguments associated with the original method'
+
+  def scaffold(name)
+    @name = name
+    @method = options[:method] if options[:method]
+    @args = options[:args] if options[:args]
```
kaylareopelle, line 35: @fallwith  - this is where the `args` option becomes `@args`

lib/tasks/instrumentation_generator/templates/chain.tt
```diff
@@ -0,0 +1,22 @@
...
+  module <%= @class_name %>
+    def self.instrument!
+      ::<%= @class_name %>.class_eval do
+        include NewRelic::Agent::Instrumentation::<%= @class_name %>
+
+        alias_method :<%= @method.downcase %>_without_new_relic, :<%= @method.downcase %>
+
+        def <%= @method.downcase %><%= "(#{@args})" if @args %>
+          <%= @method.downcase %>_with_tracing<%= "(#{@args})" if @args %> do
+            <%= @method.downcase %>_without_new_relic<%= "(#{@args})" if @args %>
```
kaylareopelle, line 16: @fallwith  - Here's an example of where `@args` is used in a template to pass the arguments.

lib/tasks/instrumentation_generator/README.md
kaylareopelle: I've added a few comments elsewhere in the code to show you how the `--args` option is working. Please let me know if there's anything about the option that seems off!

lib/tasks/instrumentation_generator/README.md
fallwith: whoops! I definitely saw and liked the alphabetization change when I first looked over the PR and then forgot about it being present. Sorry for that.

lib/tasks/instrumentation_generator/README.md
fallwith: I'm referring to the literal use of `*args` with a `*`, not a `@`.

From reading the documentation, I conclude that if I do not pass `--args`, then the default will be a literal `"*args"` value. So when I look at the template generated methods, I would expect to see this:

```ruby
def generated_method(*args)
  # generated body
end
```

but instead I see this:

```ruby
def generated_method
  # generated body
end
```

when `--args` is left out. I might be understanding it wrong. I just want to make sure that what is stated in the documentation as a default value is actually the real default value (or vice versa, as long as they align).

kaylareopelle: Hmm... could you give me an example of how you're running this? 

When I run `be thor instrumentation:scaffold concurrent-ruby`, I get the following in the `instrumentation.rb` file:
```
    def method_to_instrument_with_tracing(*args)
      # add instrumentation content here
    end
```

fallwith: Ah, I understand why I became confused now.

The `instrumentation.thor` file is indeed applying a default of `*args`, which completely matches the documentation. Everything is great. Sorry for missing that bit. You definitely didn't explain anything badly in comments and all of the code itself is certainly intuitive and great as-is.

When I saw these three lines:

```ruby
        def <%= @method.downcase %><%= "(#{@args})" if @args %>
          <%= @method.downcase %>_with_tracing<%= "(#{@args})" if @args %> do
            <%= @method.downcase %>_without_new_relic<%= "(#{@args})" if @args %>
```

I was picturing what would happen when the `if` checks failed and was incorrectly assuming that there must be a default of `nil` that you are guarding against here. So... is it even possible for the `if` checks to fail given that Thor will always apply a default? Are the checks to protect against a user doing something like `thor instrumentation:add_new_method gem_name --method=method_to_instrument --args=''`?

kaylareopelle: Great point! When I ran that locally, it would still add the parens, so I adjusted the guard statement to be for an empty string rather than a nil check. Take a look at f633c2b5381ec015d93309b6e3323d6de08fa892 for details.

fallwith (approved):
This is terrific innovation. It's not only a huge time saver and enforcer of consistency for us, but it has the underpinnings to allow community contributions to have a great basis to start from! Well done.

PR #1407, opened 2022-09-09
Ruby 3.2.0preview2

fallwith:
test against Ruby 3.2.0preview2 via CI

see
[changes](https://www.ruby-lang.org/en/news/2022/09/09/ruby-3-2-0-preview2-released/)


fallwith:
This PR was used simply to test changes against the nightly cron tests. The actual work went into #1436, which has now been merged.

Issue #1404, opened 2022-09-09
Quality: Evaluate and resolve rubocop-rake errors


hannahramadan:
### Description
Currently, failing `rubocop-rake` cops are disabled by `.rubocop_todo.yml`. These include:
- [ ] `Rake/Desc`
- [ ] `Rake/MethodDefinitionInTask`

Evaluate these cops and decide if it should remain disabled or if the error should be addressed. If it's decided the error should be addressed, resolve it in the code.

### Estimates
S = 1-3 days

PR #1421, opened 2022-09-14
Address rubocop-rake offenses

hannahramadan:
Our codebase had 2 offenses recognized by `rubocop-rake`, listed in `rubocop_todo`: `Rake/Desc` and `Rake/MethodDefinitionInTask`.

- 2 offenses: `Rake/Desc` says to describe the task with the `desc` method. This is useful in giving details on the task without reading the code.  
- 15 offenses: `Rake/MethodDefinitionInTask` says to not define a method in rake task, because it will be defined to the top level. [Here is a helpful blog](https://www.albertoalmagro.com/en/ruby-methods-defined-in-rake-tasks/) that dives into issues seen by defining methods in rake tasks. Note: the infinite tracing rake tasks would've been safe because they were defined in a namespace and then a task. For consistency, they were treated like other files that needed refactoring.

This PR corrects these offenses and removes them from `rubocop_todo`.

Closes [#1404](https://github.com/newrelic/newrelic-ruby-agent/issues/1404)

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


lib/tasks/config.rake
```diff
@@ -29,120 +29,11 @@ namespace :newrelic do
...
-    end
-
-    def build_erb(format)
-      require 'erb'
-      path = File.join(File.dirname(__FILE__), "config.#{format}.erb")
-      template = File.read(File.expand_path(path))
-      ERB.new(template)
-    end
-
+    desc "Describe available New Relic configuration settings."
```
hannahramadan, line 32: This `desc` technically already existed, but needed to be moved right before the actual task

Rakefile
```diff
@@ -7,6 +7,7 @@ require 'rubygems'
 require 'rake/testtask'
 require 'yard'
 require "#{File.dirname(__FILE__)}/lib/tasks/all.rb"
+require_relative 'rakefile_helpers'
```
hannahramadan, line 10: All edits that are moving methods from rake tasks follow the same pattern:
- Remove methods from inside a task
- Add methods to a new Class defined in a new file `#{rakefile_name}_helpers.rb`
- Require the new helper file in .rake file and call the methods using newly created Class

kaylareopelle (commented):
Great work @hannahramadan! 

I think I might've slightly miscommunicated the idea of the "helpers" yesterday. In Ruby, it's a best practice for the filename and the class name to be the same, though the capitalization/writing style differs based on the context (snake_case and UpperCamelCase).

Having something suffixed "helpers" seems really beneficial to understanding the purpose of these classes. What do you think about a slight refactor, so that the files that hold the classes have the same name as the classes, but each of these live in a directory with the `*_helpers` name?

kaylareopelle (approved):
This looks beautiful, @hannahramadan! 🐣

.rubocop_todo.yml
```diff
@@ -209,21 +209,6 @@ Minitest/RefutePredicate:
...
-    - 'lib/tasks/config.rake'
-
-# Offense count: 15
-Rake/MethodDefinitionInTask:
-  Exclude:
-    - 'Rakefile'
-    - 'infinite_tracing/tasks/proto.rake'
-    - 'lib/tasks/config.rake'
-    - 'lib/tasks/tests.rake'
-
```
fallwith, line 226: Seeing big removals like this are certainly a nice aspect of having the todo file. Nice!

Rakefile
```diff
@@ -7,6 +7,8 @@ require 'rubygems'
 require 'rake/testtask'
 require 'yard'
 require "#{File.dirname(__FILE__)}/lib/tasks/all.rb"
+require_relative 'lib/tasks/helpers/prompt'
```
fallwith, line 10: This `Rakefile` file lives at `<PROJECT ROOT>/lib/tasks`. So the prompt file's location relative to `Rakefile's` location should be simply `helpers/prompt`. You could also clean up the `all.rb` file if you wanted to:

```diff
- require "#{File.dirname(__FILE__)}/lib/tasks/all.rb"
+ require_relative 'all'
- require_relative 'lib/tasks/helpers/prompt'
+ require_relative 'helpers/prompt'
```

Rakefile
```diff
@@ -147,6 +130,7 @@ namespace :cross_agent_tests do
   end
 end
 
+desc 'Start Pry'
```
fallwith, line 133: In these situations, I have a preference for using generic terminology as much as possible to avoid single product/implementation lock in. Perhaps:

```diff
- desc 'Start Pry'
+ desc 'Start an interactive console session'
```

infinite_tracing/tasks/proto.rake
```diff
@@ -3,32 +3,12 @@
 # See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
 # frozen_string_literal: true
 
+require_relative '/helpers/license'
```
fallwith, line 6: You can drop the leading slash here:

```diff
- require_relative '/helpers/license'
+ require_relative 'helpers/license'
```

infinite_tracing/tasks/proto.rake
```diff
@@ -43,7 +23,7 @@ namespace :proto do
 
     if system(cmd)
       puts "Proto file generated!"
-      add_license_preamble_and_remove_requires output_path
+      add_license_preamble_and_remove_requires(output_path)
```
fallwith, line 26: Thanks for getting this code in line with the relevant linter cop

lib/tasks/config.rake
```diff
@@ -29,116 +30,7 @@ namespace :newrelic do
...
```
fallwith, line 33: We don't have a written rule (perhaps `rubocop-rake` would allow us to?) but for consistency's sake the task descriptions should not include a period on the end.

lib/tasks/helpers/prompt.rb
```diff
@@ -0,0 +1,25 @@
...
+    if destination.eql?('local')
+      puts 'CAUTION 2: Before continuing, make sure your local repo is on the correct, synced branch!'
+    end
+    puts
+    print "Do you wish to continue? ('y' to continue, return to cancel) [n] "
+    continue = STDIN.gets.chomp
+    if continue.downcase.eql?('y')
+      system(command)
+    else
+      puts 'Cancelled'
```
fallwith, line 22: Now that this content has shown up in git as newly added instead of moved, be prepared to answer why you've gone with the British double-L spelling here! 😄

lib/tasks/helpers/removers.rb
```diff
@@ -0,0 +1,34 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module Removers
```
fallwith, line 6: Great name.

lib/tasks/tests.rake
```diff
@@ -3,20 +3,16 @@
 # See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
 # frozen_string_literal: true
 
+require_relative 'helpers/matches'
+include Matches
```
fallwith, line 7: We shouldn't play with these.

.rubocop_todo.yml
hannahramadan: It is refreshing isn't it. I'm glad we when the todo file route!

Rakefile
hannahramadan: I'm seeing `Rakefile` and `all` living at the project root?

Rakefile
hannahramadan: Great great call. [Addressed](https://github.com/newrelic/newrelic-ruby-agent/pull/1421/commits/e3c43a60bacc3b12f6541cd95d0aab5b0d2ee276)!

infinite_tracing/tasks/proto.rake
hannahramadan: I need to remember this it's like the 5th time you've helped out on this type of edit. [Done](https://github.com/newrelic/newrelic-ruby-agent/pull/1421/commits/e3c43a60bacc3b12f6541cd95d0aab5b0d2ee276)!

lib/tasks/config.rake
hannahramadan: Consistency is rad. [Done](https://github.com/newrelic/newrelic-ruby-agent/pull/1421/commits/e3c43a60bacc3b12f6541cd95d0aab5b0d2ee276).

lib/tasks/helpers/prompt.rb
hannahramadan: Now I'm questioning how I normally spell cancelled (canceled?) 🥲

Rakefile
fallwith: Whoops! I got this one mixed up with the .rake files that were edited. The path does indeed need to be different here given that `Rakefile` is up 2 directory levels from the others. Sorry for the mixup.

hannahramadan: I guess `all` isn't project root whoops but in any case I'll leave this be!

fallwith (approved):
👍

PR #1432, opened 2022-09-15
Gabbi code refactor

hannahramadan:
Refactor Gabbi code to include class structure, associated tests, and better naming/file structure.


kaylareopelle (commented):
Great way to resolve this issue! I am concerned that we could randomly get a `0` sleep, which would bring us back to the state that prompted this change. Can you think of an adjustment that would put the lowest possible sleep as 1 second?

hannahramadan:
> Great way to resolve this issue! I am concerned that we could randomly get a `0` sleep, which would bring us back to the state that prompted this change. Can you think of an adjustment that would put the lowest possible sleep as 1 second?

I think [this committ](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/commits/4820f0c325c7a590e22277a6aaf2bb04bacd4814) should take care of that! :) `sleep(rand(1..5))`

kaylareopelle (approved):
Thanks for the adjustment, @hannahramadan! This is approved as-is. If you'd like to add tests for this code change, feel free to! Otherwise, cool with merging as-is 😎

fallwith (dismissed):
With a future PR or updates to this one...

- It'd be nice to know with a code comment why `1` and `5` were chosen
- Do these 2 classes share any modules? Maybe they could both call `pause_to_avoid_slack_throttling` or something and we'd only have a single place to observe and potentially update the sleep values.
- +1 to @kaylareopelle's call for test coverage of the new functionality. I'm down to pair on that.

.github/workflows/scripts/slack_notifications/cve_notifier.rb
```diff
@@ -0,0 +1,27 @@
...
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'time'
+require 'feedjira'
+require 'httparty'
+require_relative 'slack_notifier'
+
+class CveNotifier < SlackNotifier
```
hannahramadan, line 11: Added class and put send_slackbot functionality in new SlackNotifier class

.github/workflows/scripts/slack_notifications/cve_script.rb
```diff
@@ -5,6 +5,6 @@
 
 # This script runs on a 24 hour cycle via gem_notifications.yml and sends Slack updates for Ruby CVEs.
 
-require_relative 'cve_methods'
+require_relative 'cve_notifier'
```
hannahramadan, line 8: This was a name change to match new CveNotifier class

test/new_relic/slack_notifications_tests/gem_notifier_tests.rb
```diff
@@ -29,117 +28,75 @@ def http_get_response
...
   def test_gem_updated_true
-    assert_equal true, gem_updated?([{"created_at" => "#{Time.now}"}])
+    assert GemNotifier.gem_updated?([{"created_at" => "#{Time.now}"}])
   end
 
   def test_gem_updated_false
-    assert_equal false, gem_updated?([{"created_at" => "1993-06-11T17:31:14.298Z"}])
-  end
-
-  def test_send_bot_zero_args
```
hannahramadan, line 85: All sending slack test functionality was moved into a separate test file `slack_notifier_tests`

.github/workflows/scripts/slack_notifications/cve_notifier.rb
```diff
@@ -0,0 +1,32 @@
...
+      SlackNotifier.send_slack_message(cve_message(entry.title, entry.entry_id))
+    end
+  end
+
+  def self.cve_message(title, url)
+    {text: ":rotating_light: #{title}\n<#{url}|More info here>"}.to_json
+  end
+end
+
+# This file runs on a 24 hour cycle via slack_notifications.yml and sends Slack updates for Ruby CVEs.
```
fallwith, line 29: I find this comment very useful, but it's location to be a bit odd. Personally, I'd love to see this comment at the very top of the file so that I can immediately see what the file is about after first opening it. Then here, it'd be cool to have a comment explaining the conditional check, though that's optional given how common this pattern is for Rubyists.

.github/workflows/scripts/slack_notifications/gem_notifier.rb
```diff
@@ -0,0 +1,83 @@
...
+
+require 'time'
+require 'httparty'
+require_relative 'slack_notifier'
+
+class GemNotifier < SlackNotifier
+  def self.check_for_updates(watched_gems)
+    return if gem_list_empty(watched_gems)
+    watched_gems.each do |gem_name|
+      verify_gem(gem_name) ? gem_info = verify_gem(gem_name) : return
```
fallwith, line 14: This is not related to the PR, but we're going to RubyGems twice here for each gem for the very same information.

In future we should refactor to do something like this:

```diff
- verify_gem(gem_name) ? gem_info = verify_gem(gem_name) : return
+ gem_info = verify_gem(gem_name)
+ abort("Failed to obtain info for gem '#{gem_name}'") unless gem_info
+ 
versions = gem_versions(gem_info)
```

.github/workflows/scripts/slack_notifications/gem_notifier.rb
```diff
@@ -0,0 +1,83 @@
...
+      versions = gem_versions(gem_info)
+      notifier.send_slack_message(gem_message(gem_name, versions)) if gem_updated?(versions)
+    end
+  end
+
+  def self.gem_list_empty(watched_gems)
+    if watched_gems.empty?
+      abort("Nothing to see here! The 'watched_gems' array cannot be empty")
+    end
```
fallwith, line 23: Looking at the name `gem_list_empty`, it's really tempting to say "You should rename this to `gem_list_empty?` so that we know that we're performing an emptiness check". But then you might respond with "But it won't ever return `false`, it's more of a verification process".

At which point I'd then propose we use a different naming scheme entirely. "Can we rename it to `verify_gem_list` or `verify_gem_list!` then? As-is, it sounds like you want the program to take hold of something full and go and empty it."

I'd one-liner the contents of the method too:

```ruby
abort("Nothing to see here! The 'watched_gems' array cannot be empty") if watched_gems.empty?
```

.github/workflows/scripts/slack_notifications/gem_notifier.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  def self.verify_gem(gem_name)
+    gem_info = HTTParty.get("https://rubygems.org/api/v1/versions/#{gem_name}.json")
+
+    gem_info if gem_info.success?
```
fallwith, line 29: Currently an `HTTParty` GET failure will return `nil` which will result in an early `return`. There are 2 problems with this:

1. Elsewhere in this class, we use `abort(msg)` to explain a little bit about why we are stopping early
2. We have more info about what went wrong here in this method, and not back in the caller who hasn't been given anything more to go on than `nil`. The `gem_info` object could yield things like an HTTP status code.

.github/workflows/scripts/slack_notifications/gem_notifier.rb
```diff
@@ -0,0 +1,83 @@
...
+      if arr.size == 0 || !gem['number'].match?(/(?:rc|beta|preview)/)
+        arr << gem
+      end
+
+      break arr if arr.size == 2
+    end
+  end
+
+  def self.gem_updated?(versions)
+    Time.now.utc - Time.parse(versions[0]['created_at']) < 24 * 60 * 60
```
fallwith, line 49: Hmm... I've seen this `24 * 60 * 60` somewhere else...

.github/workflows/scripts/slack_notifications/gem_notifier.rb
```diff
@@ -0,0 +1,83 @@
...
+class GemNotifier < SlackNotifier
+  def self.check_for_updates(watched_gems)
+    return if gem_list_empty(watched_gems)
+    watched_gems.each do |gem_name|
+      verify_gem(gem_name) ? gem_info = verify_gem(gem_name) : return
+      versions = gem_versions(gem_info)
+      notifier.send_slack_message(gem_message(gem_name, versions)) if gem_updated?(versions)
+    end
+  end
+
```
fallwith, line 19: Same comment here as with the CVE notifier. All methods below `check_for_updates` could be marked as private.

.github/workflows/scripts/slack_notifications/gem_notifier.rb
```diff
@@ -0,0 +1,83 @@
...
+      action_message = "<#{interpolate_github_url(gem_name, newest, previous)}|See what's new.>"
+    else
+      action_message = "See what's new with gem-compare:\n`gem compare #{gem_name} #{previous} #{newest} --diff`"
+    end
+
+    alert_message + "\n\n" + action_message
+  end
+end
+
+# This file runs on a 24 hour cycle via gem_notifications.yml and sends Slack updates for new gem version releases.
```
fallwith, line 80: Same comment here as with the CVE notifier. This comment would be better at the top of the class file.

.github/workflows/scripts/slack_notifications/slack_notifier.rb
```diff
@@ -0,0 +1,19 @@
...
+
+require 'httparty'
+
+class SlackNotifier
+  def self.send_slack_message(message)
+    path = ENV['SLACK_GEM_NOTIFICATIONS_WEBHOOK']
+    options = {headers: {'Content-Type' => 'application/json'},
+               body: {text: message}.to_json}
+
+    # Pause to avoid slack throttling
```
fallwith, line 14: Or perhaps "Slack to avoid throttling" 😆

.github/workflows/scripts/slack_notifications/slack_notifier.rb
```diff
@@ -0,0 +1,19 @@
...
+    sleep(1)
```
fallwith, line 15: This approach has the great benefit of the sleeping logic residing in a central, shared place here.

It has a two drawbacks that stands out...

1. We delay notifying others about something potentially important by 1 second. This is easily fixed by swapping the `HTTParty.post()` and `sleep()` lines.
2. We sleep even if there is only 1 notification going out for today and even if it has been more than a second since the last message was sent out. This one requires a bit more architecting to fix, and can easily be tabled given the positive benefits previously mentioned.

test/new_relic/slack_notifications_tests/slack_notifier_tests.rb
```diff
@@ -0,0 +1,29 @@
...
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'minitest/autorun'
+require_relative '../../../.github/workflows/scripts/slack_notifications/slack_notifier'
+
+class SlackNotifications < Minitest::Test
+  def test_send_slack_message_zero_args
+    SlackNotifier.stub(:sleep, nil) do
```
fallwith, line 11: Does this stub get called when the test runs?

test/new_relic/slack_notifications_tests/slack_notifier_tests.rb
```diff
@@ -0,0 +1,29 @@
...
+      assert_raises(ArgumentError) { SlackNotifier.send_slack_message() }
+    end
+  end
+
+  def test_send_slack_message_too_many_args
+    SlackNotifier.stub(:sleep, nil) do
+      assert_raises(ArgumentError) { SlackNotifier.send_slack_message("I am a notification message!", "But I'm one too many") }
```
fallwith, line 18: For all NEW code, please stick to a maximum of 120 characters in length.

fallwith (changes requested):
This is terrific work and we are brought one step closer to having everything in a gem.

Requesting a single change:

- See comment about a single line being longer than 120 columns

Everything else I commented on can - as usual - can be considered entirely optional and never be done, be done in this PR, or be done in a follow up.

.github/workflows/scripts/slack_notifications/cve_notifier.rb
```diff
@@ -0,0 +1,32 @@
...
+    time = Time.now.utc
+    cycle = 24 * 60 * 60
+    feed.entries.each do |entry|
+      break if time - entry.updated > cycle
+
+      SlackNotifier.send_slack_message(cve_message(entry.title, entry.entry_id))
+    end
+  end
+
+  def self.cve_message(title, url)
```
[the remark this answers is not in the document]
hannahramadan: I just learned about `private` with Kayla! So this is a fun time to use it. I also added `private` to `gem_notifier`[Addressed here](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/commits/6b044062a798dca30a0c6649b73684c31246f934).

.github/workflows/scripts/slack_notifications/cve_notifier.rb
hannahramadan: Agree the top of the file is much better. [Done](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/commits/6b044062a798dca30a0c6649b73684c31246f934).

.github/workflows/scripts/slack_notifications/gem_notifier.rb
hannahramadan: I'm very glad you brought this up—definitely no need to make duplicate calls for the same thing. [Addressed here](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/commits/6b044062a798dca30a0c6649b73684c31246f934).

.github/workflows/scripts/slack_notifications/gem_notifier.rb
hannahramadan: I see what you're saying about the naming. I would think the same thing just reading it now. [Updated](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/commits/6b044062a798dca30a0c6649b73684c31246f934).

.github/workflows/scripts/slack_notifications/slack_notifier.rb
hannahramadan: Very nice I didn't think about this! For point #2, I can see how at scale that would be a terrible decision as-is. I've [addressed](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/commits/6b044062a798dca30a0c6649b73684c31246f934#diff-28d6e09aae9f0b8a06eaa86323bb371cd4ea0ab2b399e13d0e89ca760b7bf704R14-R15) point #1, and would love to talk about #2 in the future. I've made note of it.

test/new_relic/slack_notifications_tests/slack_notifier_tests.rb
hannahramadan: Yes. Like what we did before, I changed `sleep(1)` to `sleep(100)`. Didn't slow down.

test/new_relic/slack_notifications_tests/slack_notifier_tests.rb
hannahramadan: Hehe [yes done](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/commits/6b044062a798dca30a0c6649b73684c31246f934#diff-30d914ac7129feb2d13bf43e052cf29fe91564381ac3623355ac7f02ae77d096R18-R20). I saw it was over a little but thought maybe that little bit that didn't matter so much. Looks better now and noted for the future 😄

test/new_relic/slack_notifications_tests/slack_notifier_tests.rb
fallwith: If you temporarily change `sleep(1)` to `sleep(100)` in the code and then change `SlackNotifier.stub(:sleep)` into `SlackNotifier.stub(:doze)` here in this one test, does the test slow down by 100 seconds?

hannahramadan: It's working! The interpreter doesn't like `doze` since its not a real method. But I added a `puts` and stubbed out `puts` instead of `sleep` and the test took a long time to run.

hannahramadan: oh and importantly in this^ test, my `puts` never showed up when it was stubbed

.github/workflows/scripts/slack_notifications/gem_notifier.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  # Gem entries are ordered by release date. The break limits versions to two versions: newest and previous.
+  def self.gem_versions(gem_info)
+    versions = gem_info.each_with_object([]) do |gem, arr|
+      next unless gem['platform'] == 'ruby'
+
+      # for the "new" version (first one recorded), report any and all types of
+      #   versions (stable, preview, rc, beta, etc.)
+      # for the "previous" version, record only the newest stable version
+      if arr.size == 0 || !gem['number'].match?(/(?:rc|beta|preview)/)
```
kaylareopelle, line 44: Instead of checking equality on the left side of this condition, you could use the `zero?` method: https://ruby-doc.org/core-3.1.2/Integer.html#method-i-zero-3F

.github/workflows/scripts/slack_notifications/gem_notifier.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def self.gem_updated?(versions)
+    Time.now.utc - Time.parse(versions[0]['created_at']) < CYCLE
+  end
+
+  def self.github_diff(gem_name, newest, previous)
+    diff = HTTParty.get(interpolate_github_url(gem_name, newest, previous))
+
+    diff.success?
```
kaylareopelle, line 59: The solution found for James' quandary on line 33 could also be applied here.

.github/workflows/scripts/slack_notifications/gem_notifier.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def self.interpolate_github_url(gem_name, newest, previous)
+    "https://github.com/#{gem_name}/#{gem_name}/compare/v#{previous}...v#{newest}"
```
kaylareopelle, line 63: I think I've brought this up in other reviews, but the source code link may have the full GitHub URL, potentially providing better results for those gems that don't have the same user/org name as the gem, however, you would probably need to use a different API endpoint than `versions` to get that value.

.github/workflows/scripts/slack_notifications/slack_notifier.rb
```diff
@@ -0,0 +1,19 @@
...
+
+class SlackNotifier
+  CYCLE = 24 * 60 * 60 # Period in seconds to check for updates that need to be Slacked."
+
+  def self.send_slack_message(message)
+    path = ENV['SLACK_GEM_NOTIFICATIONS_WEBHOOK']
+    options = {headers: {'Content-Type' => 'application/json'},
+               body: {text: message}.to_json}
+
+    HTTParty.post(path, options)
```
kaylareopelle, line 16: What happens if this POST request fails? How can we discover the failure?

test/new_relic/slack_notifications_tests/slack_notifier_tests.rb
hannahramadan: [Done, thank!](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/commits/7ade7858803276a1efd309ab396e6f00a5bf4a42)!

fallwith (approved):

.github/workflows/scripts/slack_notifications/slack_notifier.rb
hannahramadan: @kaylareopelle I always assumed the CI would post the failure and the action would fail. Do you recommend doing some kind of catch here for failures?

kaylareopelle: I think a begin / rescue block wrapped around this line would be a great idea to catch the failures and possibly print them to the CI log. (reading your specific comment vs. talking about it syncronously sparked a slightly different answer, though I'm still curious to see what happens if the post fails in the test we discussed)

.github/workflows/scripts/slack_notifications/gem_notifier.rb
hannahramadan: I really liked this comment and have made getting that `diff` more reliable! Here's the [new code](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/files#diff-ddcde05fb3aba0b9770eddeb6be0d8e2a76ac7f86d69786141db8cb712412794R57-R65).

.github/workflows/scripts/slack_notifications/gem_notifier.rb
hannahramadan: [Done](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/files#diff-ddcde05fb3aba0b9770eddeb6be0d8e2a76ac7f86d69786141db8cb712412794R45-R46).

.github/workflows/scripts/slack_notifications/gem_notifier.rb
hannahramadan: Needed to do refactoring in a different place which affected this method, but I think [this](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/files#diff-ddcde05fb3aba0b9770eddeb6be0d8e2a76ac7f86d69786141db8cb712412794R76-R79) addresses your comment.

.github/workflows/scripts/slack_notifications/slack_notifier.rb
hannahramadan: This looks a little different from when we worked on it, but let me know if you have questions. [Here's the code](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/files#diff-28d6e09aae9f0b8a06eaa86323bb371cd4ea0ab2b399e13d0e89ca760b7bf704R10-R31).

.github/workflows/scripts/slack_notifications/gem_notifier.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def self.github_diff(gem_name, newest, previous)
+    info = HTTParty.get("https://rubygems.org/api/v1/gems/#{gem_name}.json") 
+    interpolated_url = interpolate_github_url(info["source_code_uri"], newest, previous)
+
+    HTTParty.get(interpolated_url).success?
+  end
+
+  def self.interpolate_github_url(source_code_uri, newest, previous)
+    "#{source_code_uri}/compare/v#{previous}...v#{newest}"
```
kaylareopelle, line 65: ❤️

.github/workflows/scripts/slack_notifications/slack_notifier.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def self.report_errors
+    return if @@errors.empty?
+    raise @@errors.first if @@errors.length == 1
+    raise @@errors.map(&:to_s).join("\n")
+  end
+
+  def self.errors_array
+    @@errors
+  end
```
kaylareopelle, line 32: Nice! I love the extension to raising the actual errors at the end and didn't know you could map through errors and raise them like that!

test/new_relic/slack_notifications_tests/slack_notifier_tests.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def test_send_slack_message
+    SlackNotifier.stub(:sleep, nil) do
+      HTTParty.stub(:post, nil) do
+        assert_nil SlackNotifier.send_slack_message("I am a notification message!")
+      end
+    end
+  end
+
+  def test_exit_handler_no_fail
```
kaylareopelle, line 28: If you'd like to, the names for this test and the next should likely be updated to include `errors_array` since that's now the name of the tested method.

Also. since the `exit_handler` method has been broken into `errors_array` and `report_errors`, could you add tests for `report_errors` as well?

test/new_relic/slack_notifications_tests/slack_notifier_tests.rb
```diff
@@ -0,0 +1,43 @@
...
+    SlackNotifier.stub(:sleep, nil) do
+      HTTParty.stub(:post, nil) do
+        SlackNotifier.send_slack_message("I am a notification message!")
+        assert_empty SlackNotifier.errors_array
+      end
+    end
+  end
+
+  def test_exit_handler_fail
```
kaylareopelle, line 37: I don't see the case with more than one error being tested in this file. Would you mind adding that case?

test/new_relic/slack_notifications_tests/slack_notifier_tests.rb
kaylareopelle: Also, when running locally, I learned two interesting things: 


1. this test failed: 
```
  1) Failure:
SlackNotifications#test_exit_handler_no_fail [test/new_relic/slack_notifications_tests/slack_notifier_tests.rb:32]:
Expected [#<ArgumentError: wrong number of arguments (given 2, expected 0)>] to be empty.
```

2. The script files aren't getting included in SimpleCov. We might want to _try_ to adjust where we look so they get included.

kaylareopelle (changes requested):
We're so close! I love the progress you and James made yesterday! I noticed a few things in the review. I think you'll have to make some changes anyway to address the current merge conflicts.

test/new_relic/slack_notifications_tests/slack_notifier_tests.rb
hannahramadan: Thank you for this catch! Looks like an intermittent failure since I wasn't clearing out the `errors` array after adding a failure to it. For SimpleCov, it could be a fun exercise! Although I know the goal is to move all this functionality into its own gem, so not sure if its worth figuring out how to include them?

hannahramadan: Fixed naming and [added tests for `report_errors`](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/files#diff-30d914ac7129feb2d13bf43e052cf29fe91564381ac3623355ac7f02ae77d096R58-R79).

test/new_relic/slack_notifications_tests/cve_notifier_tests.rb
```diff
@@ -0,0 +1,23 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'minitest/autorun'
+require_relative '../../../.github/workflows/scripts/slack_notifications/cve_notifier'
+
+class SlackNotifications < Minitest::Test
```
kaylareopelle, line 9: Ooh! Tiny thing, could this be renamed to CveNotifierTests?

hannahramadan: Ahh!! Yes thank you

test/new_relic/slack_notifications_tests/slack_notifier_tests.rb
```diff
@@ -0,0 +1,80 @@
+# encoding: utf-8
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'minitest/autorun'
+require_relative '../../../.github/workflows/scripts/slack_notifications/slack_notifier'
+
+class SlackNotifications < Minitest::Test
```
kaylareopelle: This would also by convention be called `SlackNotifierTests`

test/new_relic/slack_notifications_tests/gem_notifier_tests.rb
```diff
@@ -4,8 +4,7 @@
 # frozen_string_literal: true
 
 require 'minitest/autorun'
-require_relative '../../.github/workflows/scripts/slack_gem_notifications/notifications_methods'
-require_relative '../../.github/workflows/scripts/slack_gem_notifications/cve_methods'
+require_relative '../../../.github/workflows/scripts/slack_notifications/gem_notifier'
 
 class GemNotifications < Minitest::Test
```
kaylareopelle: `GemNotifications` => `GemNotiiferTests` by convention

test/new_relic/slack_notifications_tests/cve_notifier_tests.rb
hannahramadan: In fact, all of those test classes could've been named better. Fixed: [CveNotifierTests](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/files#diff-e23b8415de0f7014dc906b5f630621f8a757036a01de8591fb249d2b41047b78R9), [GemNotifierTests](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/files#diff-41275d956216f3157b605d345f4b89a6ed831343f8c73d1b687d5632fbb9f2c7R9), [SlackNotifierTests](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/files#diff-30d914ac7129feb2d13bf43e052cf29fe91564381ac3623355ac7f02ae77d096)

test/new_relic/slack_notifications_tests/slack_notifier_tests.rb
hannahramadan: With that method split in two, multiple errors are now being tested with `[test_errors_array_multiple_errors](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/files#diff-30d914ac7129feb2d13bf43e052cf29fe91564381ac3623355ac7f02ae77d096R49-R56)` and `[test_report_errors_multiple_errors](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/files#diff-30d914ac7129feb2d13bf43e052cf29fe91564381ac3623355ac7f02ae77d096R71-R79)`

test/new_relic/slack_notifications_tests/cve_notifier_tests.rb
kaylareopelle: 🦖 🦖 🦖

.github/workflows/scripts/slack_notifications/gem_notifier.rb
```diff
@@ -0,0 +1,89 @@
+# encoding: utf-8
```
kaylareopelle, line 1: You may want to merge dev into your branch. This line will now raise a rubocop failure. Sorry I didn't notice this earlier!

kaylareopelle (commented):
I think this looks great, Hannah! Thanks for your updates! Once the merge conflicts are resolved, we'll be good to go!

(The one comment I made will probably be addressed by the merge conflict resolution!)

.github/workflows/scripts/slack_notifications/gem_notifier.rb
hannahramadan: Ah! Thank you. I'll merge dev in and then re-request approval :)

kaylareopelle (approved):
🦕 🦖 🏆 🦖 🦕

.github/workflows/scripts/slack_notifications/gem_notifier.rb
hannahramadan: Now marked as private methods: [here for gem_notifier](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/files#diff-ddcde05fb3aba0b9770eddeb6be0d8e2a76ac7f86d69786141db8cb712412794R24) and [here for cve_notifier](https://github.com/newrelic/newrelic-ruby-agent/pull/1432/files#diff-8ca3366eee754ac705491b7f94a5df5a8a546e1280ede02134a37e2c0392158fR25)

.github/workflows/scripts/slack_notifications/cve_notifier.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      send_slack_message(cve_message(entry.title, entry.entry_id))
+    end
+    report_errors
+  end
+
+  private
+
+  def self.feed
+    Feedjira.parse(HTTParty.get(RUBY_SEC_FEED).body)
```
fallwith, line 28: For future PRs:

- HTTParty's `get` call and Feedjira's `parse` call should be separated out into their own dedicated methods
- Each of the new dedicated methods should incorporate their own error handling now that we have a pattern for doing so

.github/workflows/scripts/slack_notifications/slack_notifier.rb
```diff
@@ -0,0 +1,32 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'httparty'
+
+class SlackNotifier
+  CYCLE = 24 * 60 * 60 # Period in seconds to check for updates that need to be Slacked.
+  @@errors = []
```
fallwith, line 9: For future PRs:

- As discussed synchronously earlier, the need to go with a class variable has demonstrated that we should refactor the notifiers to use an instance of `SlackNotifier` and have `SlackNotifier`'s current static/class methods become dynamic/instance methods.
- Currently the model prevents us from running both notifiers at the same time, because they will share an errors array. That's enough of an indication to move to instance methods and an instance variable, `@errors`.

fallwith (approved):

fallwith (approved):

kaylareopelle (approved):

PR #1443, opened 2022-09-16
Infinite Tracing: support batching and compression

fallwith:
- Introduce a new `infinite_tracing.batching` config parameter with a boolean value defaulting to `false`. When `true`, Infinite Tracing's RecordSpanBatch RPC (previously existing but inaccessible) will be used to batch requests to the Trace Observer
- Introduce a new `infinite_tracing.compression_level` config parameter with a string value defaulting to `'none'`. When changed to a value of `'low'`, `'medium'`, or `'high'`, Infinite Tracing's gRPC client will use that compression level for outbound data.

### DIFF
--- a/infinite_tracing/lib/infinite_tracing/channel.rb
+++ b/infinite_tracing/lib/infinite_tracing/channel.rb
@@ -24,15 +24,6 @@ module NewRelic::Agent
         GRPC::Core::Channel.new(host_and_port, settings, credentials)
       end
 
-      # def apply_default_compression_level
-
-
-      #   require 'pry'
-      #   binding.pry
-
-      #   NewRelic::Agent.config[:'infinite_tracing.compression_level'] = DEFAULT_COMPRESSION_LEVEL
-      # end
-
       def channel_args
         return NewRelic::EMPTY_HASH unless compression_enabled?
 
--- a/infinite_tracing/lib/infinite_tracing/channel.rb
+++ b/infinite_tracing/lib/infinite_tracing/channel.rb
@@ -7,6 +7,7 @@ module NewRelic::Agent
   module InfiniteTracing
     class Channel
       COMPRESSION_LEVELS = %i[none low medium high].freeze
+      DEFAULT_COMPRESSION_LEVEL = :none
 
       def stub
         NewRelic::Agent.logger.debug("Infinite Tracer Opening Channel to #{host_and_port}")
@@ -23,6 +24,15 @@ module NewRelic::Agent
         GRPC::Core::Channel.new(host_and_port, settings, credentials)
       end
 
+      # def apply_default_compression_level
+
+
+      #   require 'pry'
+      #   binding.pry
+
+      #   NewRelic::Agent.config[:'infinite_tracing.compression_level'] = DEFAULT_COMPRESSION_LEVEL
+      # end
+
       def channel_args
         return NewRelic::EMPTY_HASH unless compression_enabled?
 
@@ -30,17 +40,28 @@ module NewRelic::Agent
       end
 
       def compression_enabled?
-        validate_compression_level
         compression_level != :none
       end
 
       def compression_level
-        NewRelic::Agent.config[:'infinite_tracing.compression_level']
+        @compression_level ||= begin
+          level = if valid_compression_level?(configured_compression_level)
+            configured_compression_level
+          else
+            DEFAULT_COMPRESSION_LEVEL
+          end
+          NewRelic::Agent.logger.debug("Infinite Tracer compression level set to #{level}")
+          level
+        end
       end
 
       def compression_options
         {default_algorithm: :gzip,
          default_level: compression_level}
+      end
+
+      def configured_compression_level
+        NewRelic::Agent.config[:'infinite_tracing.compression_level']
       end
 
       def credentials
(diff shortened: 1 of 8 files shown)


kaylareopelle (commented):
Thanks for getting this set up! One small verbiage tweak since I don't think we have something called an "Infinite Tracer"

kaylareopelle:
(welll, and Rubocop)

infinite_tracing/lib/infinite_tracing/channel.rb
```diff
@@ -35,6 +58,15 @@ def settings
           'grpc.enable_deadline_checking' => 0
         }
       end
+
+      def validate_compression_level
+        NewRelic::Agent.logger.debug("Infinite Tracer compression level set to #{compression_level}")
```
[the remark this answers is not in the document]
fallwith: "Infinite Tracer" is what we currently use. If we want to rename it, I'd prefer to do so in a separate scope with a different PR.

```shell
newrelic-ruby-agent 13:31:15 $ ug 'Infinite Tracer'
infinite_tracing/lib/infinite_tracing/agent_integrations.rb
     9:	     NewRelic::Agent.logger.debug("Integrating Infinite Tracer with Agent")
    17:	     NewRelic::Agent.logger.debug("Skipped Integrating Infinite Tracer with Agent")

infinite_tracing/lib/infinite_tracing/channel.rb
    12:	       NewRelic::Agent.logger.debug("Infinite Tracer Opening Channel to #{host_and_port}")

infinite_tracing/lib/infinite_tracing/record_status_handler.rb
    28:	               NewRelic::Agent.logger.debug("gRPC Infinite Tracer Observer saw #{messages_seen} messages")
    31:	           NewRelic::Agent.logger.debug("gRPC Infinite Tracer Observer closed the stream")

infinite_tracing/lib/infinite_tracing/agent_integrations/agent.rb
     7:	 NewRelic::Agent.logger.debug("Installing Infinite Tracer in Agent")
    15:	         NewRelic::Agent.logger.debug("Opening Infinite Tracer Stream with gRPC server")
```

fallwith:
> (welll, and Rubocop)

I've never liked this project's rejection of spaces for padding around curly braces, being so used to no spaces for square brackets and spaces for curly ones, so I'm not surprised I missed it.

For the MiniTest code, I think we should allow the omission of parens because it seems only right to match seattle.rb for those operations.

Addressed both of these with [1606a8c](https://github.com/newrelic/newrelic-ruby-agent/pull/1443/commits/1606a8c66d5b1f84a8c575210033e74739a1307e)

fallwith:
> I've never liked this project's rejection of spaces for padding around curly braces, being so used to no spaces for square brackets and spaces for curly ones, so I'm not surprised I missed it.

StandardRB says that this helps differentiate between blocks and hashes:

> Spaces inside blocks, but not hash literals - In Ruby, the { and } characters do a lot of heavy lifting. To visually distinguish hash literals from blocks, Standard enforces that (like arrays), no leading or trailing spaces be added to pad hashes

I suppose I've not had trouble in doing so and that's why I've not gone on with this rule and seeing its benefit. Conversely, I feel that the default RuboCop spacing rules help a lot with nested structures (arrays of hashes of arrays or whatnot).

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2436,6 +2436,25 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :external => :infinite_tracing,
           :description => "Configures the TCP/IP port for the Trace Observer Host"
+        },
+        :'infinite_tracing.compression_level' => {
+          :default => :none,
+          :public => true,
```
tannalynn, line 2442: After the discussion this morning, would it be best if these were left as non public for now? Once we plan to actually turn it on, we could flip it to be public. It's not clear to me if this is planned on being completed for our next release.

infinite_tracing/lib/infinite_tracing/channel.rb
```diff
@@ -35,6 +58,15 @@ def settings
...
+
+        return if COMPRESSION_LEVELS.include?(compression_level)
+
+        raise "Invalid compression level '#{compression_level}' specified! Must be one of " \
```
tannalynn, line 67: Where does this raise end up getting caught? Would this prevent infinite tracing from starting completely? I don't think we would want to break infinite tracing if someone did something incorrect for their config settings. 
It might be good if, instead of raising here, we logged a warning and then set the compression level to be whatever the default is.

infinite_tracing/lib/infinite_tracing/client.rb
```diff
@@ -143,8 +147,10 @@ def stop
 
       def start_streaming(exponential_backoff = true)
         return if suspended?
+
+        @exponential_backoff = exponential_backoff
```
tannalynn, line 151: How come we're saving this here?

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: It'd be great to get clear guidance from Erik on this one. Good catch, Tanna.

kaylareopelle (commented):
I think Tanna has some great questions. Once those are addressed, I'll approve.

lib/new_relic/agent/configuration/default_source.rb
fallwith: Good call. Set to `false` with 873ed0a48aabb6c4137d69c2bc5a20457733553c

infinite_tracing/lib/infinite_tracing/channel.rb
fallwith: Good call - we just log instead of raising. Fixed with 873ed0a48aabb6c4137d69c2bc5a20457733553c

infinite_tracing/lib/infinite_tracing/client.rb
fallwith: Whoops! This was leftover from an earlier version of the code. I was using an instance var instead of simply baton passing this method's lone argument further down to the next method, but passing the method ended up working best. Addressed with 873ed0a48aabb6c4137d69c2bc5a20457733553c

fallwith:
> I think Tanna has some great questions. Once those are addressed, I'll approve.

Ready for re-review @kaylareopelle and @tannalynn (you can jump straight to 873ed0a48aabb6c4137d69c2bc5a20457733553c if you just want to see what changed since you looked previously).

kaylareopelle (approved):
Thanks for those adjustments, James! Let's make a note to flip those settings to public => true and add a CHANGELOG entry whenever we go live 😄

fallwith:
> Thanks for those adjustments, James! Let's make a note to flip those settings to public => true and add a CHANGELOG entry whenever we go live 😄

Good idea. Added notes to do those 2 things to the (internal) MMF doc!

PR #1451, opened 2022-09-20
CI: remove file descriptor exhaustion tests

fallwith:
Using infinite loops to saturate Ruby ObjectSpace on every single CI run is not ideal.

Our guid generation logic used to be `SecureRandom` based, and used to run into issues when a systems resource limit was hit and access to `/dev/urandom` would fail.

When we
[switched](https://github.com/newrelic/newrelic-ruby-agent/commit/1294c5bb50640752378b9d9faa2f55ac3dd0dbcd) away from `SecureRandom`, a decision was made to test the new more fault tolerant guid generation logic by inducing the exhaustion of the host machine's file descriptors and proving that guid generation still works.

These tests are slow on modern cloud based devices with lots of resources, problematic when ran by multitasking devs locally, and simply do not provide us with value anymore. The `refute` lines inside the infinite loops also cause huge artificially inflated numbers of successful tests in the thousands.

Let's get rid of these.

### DIFF
--- a/test/new_relic/agent/transaction/segment_test.rb
+++ b/test/new_relic/agent/transaction/segment_test.rb
@@ -271,29 +271,6 @@ module NewRelic
           assert_equal t, segment.start_time
         end
 
-        def test_generates_guid_when_running_out_of_file_descriptors
-          # SecureRandom.hex raises an exception when the ruby interpreter
-          # uses up all of its allotted file descriptors.
-          # See also: https://github.com/newrelic/newrelic-ruby-agent/issues/303
-          file_descriptors = []
-          begin
-            # Errno::EMFILE is raised when the system runs out of file
-            # descriptors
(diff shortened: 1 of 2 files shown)


kaylareopelle (approved):
I didn't realize this was the source of our thousands of test #'s! Looks like our coverage % is still good (possibly better?) Did you get a chance to see if the lines of code these tests covered are being exercised elsewhere?

fallwith:
> I didn't realize this was the source of our thousands of test #'s!

Yeah, thousands of extra tests with unspecific numbers that can vary from machine to machine. Not ideal.

> Looks like our coverage % is still good (possibly better?) Did you get a chance to see if the lines of code these tests covered are being exercised elsewhere?

Yeah, the transaction and segment testing parts are covered elsewhere and `GuidGenerator` has its own tests as well.

Issue #1426, opened 2022-09-15
Rubocop: Turn on and address failures for Metrics/AbcSize


kaylareopelle:
### Description
One of the aspects of our codebase that makes it difficult to iterate on is the complexity of our methods. Rubocop provides a category of linters to identify complex areas of code. These linters are currently all disabled for the Ruby agent's codebase. Let's start enabling them with Metrics/AbcSize. 

[Link to Documentation for Metrics/AbcSize](https://docs.rubocop.org/rubocop/cops_metrics.html#metricsabcsize)

Update:
Some progress has been made throughout the codebase (a focus on `lib/*`)with PR #1447 

### Acceptance Criteria
* Rubocop Metrics/AbcSize is enabled for the lib directory and passes for all files

### Design Consideration/Limitations
* There are over 100 violations of this rule in our codebase
* This might be a good task to have the whole team work on in tandem
* If we work on this task in tandem, we'll need to devise a way to avoid stepping on each other's toes

### Dependencies
None 

### Additional context
* Rubocop todo files for this linter could be created and addressed on a smaller scale

### Estimates
L

fallwith:
We had multiple devs working on this extensively recently. See all related closed PRs. Closing out the issue to produce a tangible chunk of work with a cut-off point. The remaining work is now obvious with the updates made to the `.rubocop_todo.yml` file, which has been updated to not only indicate the outstanding issues but also to prioritize work done outside of the `test/` directory.

PR #1455, opened 2022-09-22
some refactorings to satisfy ABC size limits

fallwith:
refactorings with an aim of keeping the ABC size per method below the RuboCop default of 17

### DIFF
--- a/lib/new_relic/agent/new_relic_service.rb
+++ b/lib/new_relic/agent/new_relic_service.rb
@@ -402,7 +402,7 @@ module NewRelic
       end
 
       def prep_request(opts)
-        headers = prep_headers
+        headers = prep_headers(opts)
         if Agent.config[:put_for_data_send]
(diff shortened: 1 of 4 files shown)


lib/new_relic/agent/new_relic_service.rb
```diff
@@ -380,6 +394,116 @@ def control
...
+        response = nil
+        conn = http_connection
+        ::NewRelic::Agent.logger.debug("Sending request to #{opts[:collector]}#{opts[:uri]} with #{request.method}")
+        Timeout.timeout(@request_timeout) do
+          response = conn.request(request)
+        end
+        response
+      end
+
+      def handle_error_response(response, endpoint)
```
kaylareopelle, line 451: Woof. This one will be a doozy when we turn on line limits!

kaylareopelle (approved):
Thanks for bravely facing these refactors!

fallwith:
relates to #1426

Issue #1456, opened 2022-09-22
Haml 6.0 causes tests to fail


kaylareopelle:
Haml 6.0 was released yesterday 🎉 

However, Haml 6.0 made some adjustments that cause the way our metrics are calculated for view-related instrumentation to change.

This issue looks into the changes.

Here are the test suites that failed the scheduled continuous integration on 9/21/2022:
jruby_multiverse (rails)
jruby_multiverse (rest)
multiverse (database, 2.5.9)
multiverse (httpclients, 2.5.9)
multiverse (rails, 2.2.10)
multiverse (rails, 2.3.8)
multiverse (rails, 2.4.10)
multiverse (rails, 2.5.9)
multiverse (rails, 2.6.10)
multiverse (rails, 2.7.6)
multiverse (rails, 3.0.4)
multiverse (rails, 3.1.2)
multiverse (rails, 3.2.0-preview2)
multiverse (rest, 2.2.10)
multiverse (rest, 2.3.8)
multiverse (rest, 2.4.10)
multiverse (rest, 2.5.9)
multiverse (rest, 2.6.10)
multiverse (rest, 2.7.6)
multiverse (rest, 3.0.4)
multiverse (rest, 3.1.2)
multiverse (rest, 3.2.0-preview2)

fallwith:
https://github.com/haml/haml/releases/tag/v6.0.0

PR #1457, opened 2022-09-22
Haml 6.0 Release Fixes

kaylareopelle:
# Overview

Many of our Rails tests as well as tests for other instrumented gems use Haml. With Haml's new release yesterday, a few areas that have unlocked Haml versions (some areas we control, some we seemingly don't) upgraded to version 6.0 when it was released September 21, 2022. Here's an overview of the changes made to our test suite to get things passing again.

* Use Haml::Template for Tilt metric name if Haml 6.0+
* Use Haml ~> 5.2.2 for testing with Tilt 1.4.1 (this version doesn't have a lock on accepted Haml versions, but seems incompatible with Haml 6.0)
* Create a conditional value for `rails_prepend` Action View metrics based on Haml version

Closes: #1456 

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/test/multiverse/suites/tilt/Envfile
+++ b/test/multiverse/suites/tilt/Envfile
@@ -10,10 +10,16 @@ TILT_VERSIONS = [
   ['1.4.1', 2.2, 2.7]
 ]
 
+def haml_version(tilt_version)
+  if tilt_version && (Gem::Version.new(tilt_version.match(/\d|\./)) < Gem::Version.new('2.0.0'))
+    add_version('5.2.2')
+  end
+end
+
 def gem_list(tilt_version = nil)
   <<-RB
     gem 'tilt'#{tilt_version}
-    gem 'haml'
+    gem 'haml'#{haml_version(tilt_version)}
     #{ruby3_gem_webrick}
   RB
 end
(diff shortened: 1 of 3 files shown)


test/multiverse/suites/tilt/Envfile
```diff
@@ -10,10 +10,16 @@ TILT_VERSIONS = [
   ['1.4.1', 2.2, 2.7]
 ]
 
+def haml_version(tilt_version)
+  if tilt_version && (Gem::Version.new(tilt_version.match(/\d|\./)) < Gem::Version.new('2.0.0'))
```
fallwith, line 14: From looking at our possible values for `tilt_version`, which currently is always going to either be `'1.4.1'` or `nil`, I think performing `[0]` on the version string would be better for getting at the first digit than a regex which contains an "or" pipe:

```diff
- (Gem::Version.new(tilt_version.match(/\d|\./))
+ (Gem::Version.new(tilt_version[0])
```

Granted, if you feel our 15 year old project will be around for another 15 years, neither this `[0]` approach nor the original regex one will handle double digit major version based strings.

If you want to handle that, you can do this:

```diff
- (Gem::Version.new(tilt_version.match(/\d|\./))
+ (Gem::Version.new(tilt_version.match(/^\d+/))
```

fallwith (dismissed):
Thank you for keeping us ready for the absolute latest Ruby community code!

Left two optional alternatives for getting a major version number - one easier to visually understand for the human reviewer, and one more futureproof.

fallwith (approved):

PR #1458, opened 2022-09-23
Enable four new Style cops

kaylareopelle:
# Overview

Turns on four previously disabled cops: 
* Style/MultilineTernaryOperator
* Style/NilComparison
* Style/NonNilCheck
* Style/OrAssignment

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1541,7 +1541,7 @@ Style/MultilineMethodSignature:
   Enabled: false
 
 Style/MultilineTernaryOperator:
-  Enabled: false
+  Enabled: true
 
 Style/MultilineWhenThen:
   Enabled: true
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1604,10 +1604,9 @@ Style/NilComparison:
 Style/NilLambda:
   Enabled: false
 
-# Disabling for now
 Style/NonNilCheck:
-  Enabled: false
-  # IncludeSemanticChanges: false
+  Enabled: true
+  IncludeSemanticChanges: false
 
 Style/Not:
   Enabled: true
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1540,9 +1540,6 @@ Style/MultilineMemoization:
 Style/MultilineMethodSignature:
   Enabled: false
 
-Style/MultilineTernaryOperator:
-  Enabled: true
-
 Style/MultilineWhenThen:
   Enabled: true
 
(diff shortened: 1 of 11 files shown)


fallwith (dismissed):
I like all four of these.

Where did we land on being redundant with RuboCop's defaults? My desired approach would be to simply remove these 4 from `.rubocop.yml`, but I definitely feel stronger about having them enabled than _how_ they are enabled.

kaylareopelle:
> I like all four of these.
> 
> Where did we land on being redundant with RuboCop's defaults? My desired approach would be to simply remove these 4 from `.rubocop.yml`, but I definitely feel stronger about having them enabled than _how_ they are enabled.

@fallwith  - I've updated the `.rubocop.yml` for this PR to reflect our team discussion!

.rubocop.yml
```diff
@@ -1596,19 +1593,17 @@ Style/NestedTernaryOperator:
 Style/Next:
   Enabled: false
 
-# Disabling for now
 Style/NilComparison:
-  Enabled: false
-  # EnforcedStyle: predicate
+  Enabled: true
+  EnforcedStyle: predicate
```
fallwith, line 1598: This cop's config is now redundant with [these defaults](https://github.com/rubocop/rubocop/blob/master/config/default.yml#L4366). We can remove all of these lines from .rubocop.yml

.rubocop.yml
```diff
@@ -1596,19 +1593,17 @@ Style/NestedTernaryOperator:
...
 # Disabling for now
 Style/NilLambda:
   Enabled: false
 
-# Disabling for now
 Style/NonNilCheck:
-  Enabled: false
-  # IncludeSemanticChanges: false
+  Enabled: true
+  IncludeSemanticChanges: false
```
fallwith, line 1606: This cop's config is now redundant with [these defaults](https://github.com/rubocop/rubocop/blob/master/config/default.yml#L4383) and can be removed entirely from .rubocop.yml.

.rubocop.yml
kaylareopelle: Nice catch! Resolved by: 347dcbc919d0c1d5ad64891dee9e39346af654aa

.rubocop.yml
kaylareopelle: Nice catch! Resolved by: 347dcbc919d0c1d5ad64891dee9e39346af654aa

fallwith (approved):

PR #1459, opened 2022-09-23
Enable Style/Proc

kaylareopelle:
Use proc instead of Proc.new


fallwith (approved):
👍 I'll miss seeing and using `Proc.new`, but I feel much more strongly about bringing us as close as possible to vanilla RuboCop's defaults (like this PR helps do) and then having the current team all agree on and apply any drift once we know what the baseline is like.

fallwith (approved):

tannalynn (approved):

PR #1460, opened 2022-09-23
Enable Rubocop Performance linters

kaylareopelle:
# Overview

This PR turns on Rubocop Performance linters that were previously disabled. It also labels additional linters incompatible with older Rubies and creates a group of linters disabled until more care and attention can be turned toward them.

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


bin/nrdebug
```diff
@@ -212,7 +212,7 @@ class ProcessReport
 
       section(f) do
         c_backtraces, ruby_backtraces = @target.gather_backtraces
-        if c_backtraces.match(/could not attach/i)
+        if /could not attach/i.match?(c_backtraces)
```
fallwith, line 215: Hmm... I agree that `#match?` is a much better and more performant fit for what we're trying to do here (simply "does it match", not "give me the matching content"), but I don't like the re-arranging of the components and calling `#match` on the `i` of the regex pattern.

I'd prefer this if it still passes the checks:

```ruby
if c_backtraces.match?(/could not attach/i)
```

PR #1462, opened 2022-09-25
Use a traditional SimpleCov config for local dev

fallwith:
Our CI system performs tests using multiple containers and multiple instances of Ruby per container, requiring the stitching together of multiple SImpleCov JSON output files to make a cohesive overview of our test coverage.

But for local dev, I'd prefer if we went with a more traditional approach.

Here is an overview of the changes being proposed:

CURRENTLY
=
• Dev runs a rake task that runs one or more unit tests • Multiple unique `lib/coverage_<uuid>` numbers get created in JSON format
• Dev runs `rake coverage:report`
• Dev views HTML in `lib/coverage/index.html`
• Dev makes a test tweak to improve coverage
• Dev goes back to the start
• NOTE: if a dev runs tests but is currently unconcerned with code coverage, their `lib` directory stays littered with (git ignored) `lib/coverage_<uuid>` files until the `coverage:report` or `coverage:clear` tasks are ran

PROPOSED
=
• Dev runs a rake task that runs one or more unit tests • Dev views HTML in `lib/coverage/index.html`
• Dev makes a test tweak to improve coverage
• Dev goes back to the start
• NOTE: if a dev runs tests but is currently unconcerned with code coverage, they simply ignore the single `lib/coverage/index.html` file


tannalynn (approved):
🎉

kaylareopelle (approved):
👍

Issue #1318, opened 2022-08-16
CI: Identify where time is spent across all tests


fallwith:
Looking at our history of CI runs and runs that are performed locally, we should produce an outline of where time is being spent. Time should be represented both as a percentage of the whole, and in actual minutes and seconds to give us a rough idea of real time taken.

For example, we could produce something like this:

- rails61: 30 secs (1%)
- rails52: 30 secs (1%)
- background_2: 2.5 mins (4%)
- infinite_tracing: 1.5 mins (3%)

For multiverse, we can start by listing suites and then go more granular with individually nested test stats after.

Once we produce this outline, we should work as a team to produce new issues to:

a) Address performance concerns through test and/or code refactoring
b) Address no-way-around-it slowness by re-evaluating how we group the tests for better parallelism

kaylareopelle:
@fallwith - It looks like there's an existing reporter that can provide us with the information we're looking for: https://www.rubydoc.info/gems/minitest-reporters/1.5.0/Minitest/Reporters/MeanTimeReporter 

The only downside is that this reporter isn't available on the version of `minitest-reporters` compatible with `minitest` 4.7.5

As `minitest` 5.0 was released in 2013, are we comfortable spending the time we need to get the tests that fail on minitest 5.0 fixed as part of this ticket?

kaylareopelle:
We can get class-level timing data using the `minitest-reporters` gem's `MeanTimeReporter`: 

![Screen Shot 2022-08-24 at 3 43 21 PM](https://user-images.githubusercontent.com/87386821/186784691-926b362d-e774-4070-b2d5-fd30a048b58a.png)

When chatting with James, this seems to be enough to get us over the hurdle of where to start to get the best return on investment when trying to speed things up 

I'm going to spend a bit of time on trying to get the timing information for individual tests by creating my own reporter, but not too much.

kaylareopelle:
Some of the code needs to be pulled out in order to get this to pass on the CI. I'm hoping to do that this afternoon.

PR #1468, opened 2022-09-26
Create TestTimeReporter

kaylareopelle:
# Overview
* TestTimeReporter prints the ten slowest tests from the previous run
* When running unit tests, it will print at the end of the test run
* When running multiverse, a file is created within lib/minitest that holds the test output from each suite and organizes the results at the end of all suites before printing out the ten slowest results
* Currently this output can be found within CI workflow's unit_tests or multiverse jobs in the 'Run * Tests' step

Initial work for this PR was completed in #1352 
Closes: #1318 

A Google Sheet holding test time data across all unit and multiverse tests sorted by class name can be seen [here](https://docs.google.com/spreadsheets/d/1UJblncUk8HeRyD9XWeebCICRRbbCYpKUL8UvbVbHHTM/edit#gid=948861090&range=A1)(Internal)

Submitter Checklist:
- [X] Include a link to the related GitHub issue, if applicable

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


infinite_tracing/test/test_helper.rb
```diff
@@ -16,6 +16,7 @@ module NewRelic; TEST = true; end unless defined? NewRelic::TEST
 require 'minitest/autorun'
 require 'minitest/pride' unless ENV['CI']
 require 'mocha/setup'
+require_relative '../../test/minitest/test_time_plugin'
```
fallwith, line 19: Ideally I'd like to have `require` and `require_relative` lines split into 2 sections and have each section be alphabetized based on gem / path string values. Is this a dependency order thing, where minitest needs to be ready and then we need to do something before loading the agent itself?

test/minitest/test_time_reporter.rb
```diff
@@ -0,0 +1,71 @@
...
+
+class TestTimeReporter < MINITEST_CLASS_NAME
+  attr_accessor :test_times
+
+  # Keep the value for this constant the same as in Multiverse::Runner#run
+  TEST_TIME_REPORT = File.join(File.expand_path(File.dirname(__FILE__)), 'minitest_time_report')
+
+  def initialize(options = {})
+    super
+    File.open(TEST_TIME_REPORT, 'w+') unless File.exist?(TEST_TIME_REPORT)
```
fallwith, line 23: I'm concerned about open file handles being left dandling. Is this intended to simply "touch" a new file which will be appended to later on?

If so, is `File.open(filename, 'w+')` equivalent to `file = File.open(filename, 'w+'); file.close`? I just want to make sure we only have a handle open while we're actively writing.

Also if this is truly just a "touch" operation, there's `FileUtils.touch`.

fallwith (commented):
This gives us everything I'd hoped for and more! I look forward to working through the "bad test" leaderboard and getting some lengthy times down. Left some questions.

test/multiverse/lib/multiverse/runner.rb
```diff
@@ -95,6 +104,7 @@ def execute_suites(filter, opts)
       "httpclients" => ["curb", "excon", "httpclient"],
       "httpclients_2" => ["typhoeus", "net_http", "httprb"],
       "infinite_tracing" => ["infinite_tracing"],
+      "logger" => ['grpc', 'logger'],
```
tannalynn, line 107: Why are we creating a new multiverse group?
If this is intended, we should probably rename it since it would be surprising to see grpc in the 'logger' group. Also, we would need to add this group to our CI workflows for multiverse

tannalynn (changes requested):

test/multiverse/lib/multiverse/runner.rb
kaylareopelle: Whoops! That is not intended! It was a group I was using to test the behavior of the reporter. I'll remove!

test/minitest/test_time_reporter.rb
kaylareopelle: @fallwith  - `File.open` automatically closes the file when a block is passed to it and that block completes, so it wouldn't be closed in this case. I'll adjust to using `touch` instead. Thanks!

fallwith: ok, cool. using touch will definitely not even open a handle, so that's perfect.

in my testing, I was worried about `File.open`:

```ruby
> File.open('test', 'w').closed?
=> false

> file = File.open('test', 'w')
> file.close
> file.closed?
=> true
```

it was bugging me that the first example returned `false`

kaylareopelle: Resolved in: [2c9bf39](https://github.com/newrelic/newrelic-ruby-agent/pull/1468/commits/2c9bf39904a1dd9ed964e9de6d0b84abd224ad8a)

test/multiverse/lib/multiverse/runner.rb
kaylareopelle: Resolved in: [c9fe5b8](https://github.com/newrelic/newrelic-ruby-agent/pull/1468/commits/c9fe5b86039a7f8ad7ee2a66bfd8df3e413b8dc5)

infinite_tracing/test/test_helper.rb
kaylareopelle: Resolved by: 56b03baae0646ad361243f98d8ed9594277cc285

test/multiverse/lib/multiverse/runner.rb
```diff
@@ -46,6 +46,15 @@ def parse_args(args)
     end
 
     def run(filter = "", opts = {})
+      # This file is generated by TestTimeReporter.
+      # TestTimeReporter cannot be loaded until minitest has been loaded by an
+      # individual suite, so we cannot define this inside the TestTimeReporter
+      # class. Please keep this path and the paths for the reporter constants
+      # the same!
+      # rubocop:disable Lint/NonAtomicFileOperation
+      File.delete(Multiverse::TEST_TIME_REPORT_PATH) if File.exist?(Multiverse::TEST_TIME_REPORT_PATH)
```
fallwith, line 55: Satisfying the linter should be easy enough to do here if you bring in FileUtils as you've done elsewhere in the PR:

```diff
+ require 'fileutils'
  ...
- File.delete(Multiverse::TEST_TIME_REPORT_PATH) if File.exist?(Multiverse::TEST_TIME_REPORT_PATH)
+ FileUtils.rm_f(Multiverse::TEST_TIME_REPORT_PATH)
```

fallwith (changes requested):

test/multiverse/lib/multiverse/runner.rb
kaylareopelle: The line of code you suggest is what I had here originally, but an error was resolved by removing it (though which error, I can't quite remember). I'll try reverting that change and see if things still work.

kaylareopelle: It worked fine this time around! 🙃 
Resolved in: d95ad43f2acd27a98338d0fa5ea8ac32a41d445e

fallwith (approved):

tannalynn (approved):
🎉

PR #1474, opened 2022-09-27
Add Ruby version constraints to grpc Envfile

kaylareopelle:
grpc 1.49.1 is not compatible with Ruby 2.5 due to a dependency requiring Ruby 2.6+. This update should allow our CI to pass.


tannalynn (approved):
🎉 TY!

PR #1475, opened 2022-09-27
CI: add a 'method' filter

fallwith:
when performing multiverse tests, one can now specify a 'method' filter

ex: `bundle exec rake test:multiverse[rake,method=prepend]`

### DIFF
--- a/test/multiverse/README.md
+++ b/test/multiverse/README.md
@@ -106,6 +106,9 @@ You can pass these additional parameters to the test:multiverse rake task to con
 - `name=` only tests with names matching string will be executed
 - `env=` only the Nth environment will be executed (may be specified multiple times)
 - `file=` only tests in specified file will be executed
+    NOTE: the given string will be fuzzily matched against relative paths below the suite name
+          ex: "file=instrumentation" to match `test/multiverse/suites/redis/redis_instrumentation_test.rb`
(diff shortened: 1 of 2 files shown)


kaylareopelle (approved):
Love this! Great idea!

kaylareopelle:
I took a step back from my enthusiasm and have one small comment. Here's the output on my machine from running `rake "test:multiverse[grpc,method=chain]"`:

```
Running "grpc" suite in 2 CHAIN environments and 2 PREPEND environments in parallel
Running "grpc" using CHAIN for Envfile entry 1
# ... unrelated ... 
Running "grpc" using CHAIN for Envfile entry 0
# ... unrelated ...
Skipping method 'prepend' while focusing only on 'chain'
# ... unrelated ...
```

Would you be willing to update the top line of the test output so that it doesn't include the references to `PREPEND`?

kaylareopelle (commented):
Ok, I'm still very excited about this, but have one request. 

I see the nice message at the end of the log `Skipping method 'prepend' while focusing only on 'chain'`, however, I was startled by `Running "grpc" suite in 2 CHAIN environments and 2 PREPEND environments in parallel` at the top of the test run. 

Would you be able do either/both of the following? 
* Fix the top message to exclude references to PREPEND
* Change the color of the bottom message to yellow to match the other references to chain/prepend in the test output

PR #1476, opened 2022-09-27
multiverse: update "Running..." notification

fallwith:
now that #1475 has delivered a 'method' filter, we should update the notification message at the start of a run accordingly.

before:

```
$ bundle exec rake test:multiverse[grpc,method=chain]

Running "grpc" suite in 2 CHAIN environments and 2 PREPEND environments in parallel
```

after:

```
$ bundle exec rake test:multiverse[grpc,method=chain]

Running "grpc" suite in 2 CHAIN environments in parallel
```

(this is in response to PR feedback on #1475)


kaylareopelle (approved):
Thank you! The error message is a great touch.

PR #1483, opened 2022-09-28
Address Minitest/AssertIncludes offenses

hannahramadan:
Address `Minitest/AssertIncludes` offenses and remove from `rubocop_todo`


fallwith (approved):
The changes look good and provide value.

It bugs me that `includes` is not `include`.

PR #1485, opened 2022-09-28
Address AssertNil offenses

hannahramadan:
Address `Minitest/AssertNil` and remove from `rubocop_todo`

### DIFF
--- a/test/multiverse/suites/agent_only/log_events_test.rb
+++ b/test/multiverse/suites/agent_only/log_events_test.rb
@@ -28,7 +28,7 @@ class LogEventsTest < Minitest::Test
     assert_equal span_id, last_log["span.id"]
 
     common = last_logs_common
-    assert_nil(common["attributes"]["entity.type"])
+    assert_nil common["attributes"]["entity.type"]
     assert_equal NewRelic::Agent::Hostname.get, common["attributes"]["hostname"]
   end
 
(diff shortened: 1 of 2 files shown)


test/multiverse/suites/agent_only/log_events_test.rb
```diff
@@ -42,11 +42,11 @@ def test_log_event_data_sent_no_transaction
...
     assert_equal "Deadly", last_log["message"]
     assert_equal "FATAL", last_log["level"]
-    assert_equal nil, last_log["trace.id"]
-    assert_equal nil, last_log["span.id"]
+    assert_nil(last_log["trace.id"])
+    assert_nil(last_log["span.id"])
 
     common = last_logs_common
-    assert_equal nil, common["attributes"]["entity.type"]
+    assert_nil(common["attributes"]["entity.type"])
```
fallwith, line 49: I don't like the parens in seattle.rb method usage, especially when there's another assertion right below it that doesn't use them. But this change is better functionally than what we have now, so I'd approve this as-is once it's ready for review.

hannahramadan: Ah I'd been adding them. Is it better to respect consistency in these situations?

hannahramadan: Went ahead and [removed](https://github.com/newrelic/newrelic-ruby-agent/pull/1485/commits/4f0287fc7deae122ecc6c4fb7e3494052f66f932).

fallwith: I'd like to hear everyone's input on preferences and then we can decide on a standard as a team.

My personal preference is that if a method is explicitly allowed to not have parentheses via our `Style/MethodCallWithArgsParentheses` configuration, that it should by default not have them and only see them added when it really helps clarify things.

fallwith (approved):

PR #1486, opened 2022-09-28
ABC size improvements for bin/nrdebug

fallwith:
bring `bin/nrebug` in line with RuboCop's ABC size score of 17 per method


tannalynn (approved):

fallwith:
relates to #1426

PR #1488, opened 2022-09-29
Address Minitest/RefuteFalse

hannahramadan:
Address `Minitest/RefuteFalse` and remove from `rubocop_todo`


fallwith (approved):
I love these PRs. There are hundreds of examples of "this would be cleaner / simpler / faster / more consistent if you wrote it like this..." being delivered in bulk. Great stuff.

PR #1490, opened 2022-09-29
RuboCop: scope ABC size concerns to non tests

fallwith:
for now, let's TODO our ABC size violations outside of the tests

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -18,7 +18,11 @@ Layout/EmptyLinesAroundBlockBody:
 # Offense count: 426
 # Configuration parameters: AllowedMethods, AllowedPatterns, IgnoredMethods, CountRepeatedAttributes.
 Metrics/AbcSize:
-  Max: 82
+  Max: 54
+  Exclude:
+    - test/**/*
+    - infinite_tracing/test/**/*
+
 
 # Offense count: 145
 # This cop supports safe autocorrection (--autocorrect).


kaylareopelle (approved):
👍 

I was vegetarian and also trying to follow a specific fad diet when I went to Korea. No one in the group of people I was usually with spoke or read Korean. I barely knew what I ate, but after a few days of eating only kimchi and rice, I decided to give up all restrictions and point randomly to things on the menu to order. One field trip, they provided meals to us. There was a vegetarian option and I jumped for joy. It was bibimbap without the meat and I still make that meal today.

fallwith:
> I was vegetarian and also trying to follow a specific fad diet when I went to Korea. No one in the group of people I was usually with spoke or read Korean. I barely knew what I ate, but after a few days of eating only kimchi and rice, I decided to give up all restrictions and point randomly to things on the menu to order. One field trip, they provided meals to us. There was a vegetarian option and I jumped for joy. It was bibimbap without the meat and I still make that meal today.

Thank you for sharing this! I just had bibimbap at a Korean restaurant this week after learning that every available soup, stew, and noodle dish they had was meat based. Even jjajangmyeon looks like it'll be ok at first but it's normally got pork in it.

It's a shame that the absolute best most critically acclaimed vegetarian food actually comes from [a Korean in Korea](https://www.theworlds50best.com/stories/News/asia-restaurants-icon-award-jeong-kwan.html), but you'd never know it from talking to Koreans who [survey as 99.8% carnivore](https://www.koreaherald.com/view.php?ud=20220314000865).

The same seems true of [zen](https://www.shambhala.com/the-compass-of-zen-419.html).

fallwith:
relates to #1426

PR #1492, opened 2022-09-29
agent.rb: address ABC size issues

fallwith:
break large methods up into smaller pieces to keep ABC size down


kaylareopelle (approved):
👍

fallwith:
relates to #1426

PR #1494, opened 2022-09-29
Refactor NewRelic::Control::InstanceMethods

kaylareopelle:
Refactor to adhere to Rubocop's Metrics/AbcSize linter as well as other Metrics/* linters

### DIFF
--- a/lib/new_relic/control/private_instance_methods.rb
+++ b/lib/new_relic/control/private_instance_methods.rb
@@ -1,4 +1,3 @@
-# encoding: utf-8
 # This file is distributed under New Relic's license terms.
 # See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
 # frozen_string_literal: true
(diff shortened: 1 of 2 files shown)


kaylareopelle:
Relates to https://github.com/newrelic/newrelic-ruby-agent/issues/1426

tannalynn (approved):

lib/new_relic/control/private_instance_methods.rb
```diff
@@ -0,0 +1,46 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+# require 'new_relic/agent/agent_logger'
+# require 'new_relic/agent/memory_logger'
```
fallwith, line 6: We now have methods over here in this new module that definitely do need the `require` statements to have been invoked before use. But there are still some logger uses left in the original class that needs them too.

@kaylareopelle I think the best thing to do here is to uncomment these two lines to make sure that the `require`s are happening in both places and this module isn't explicitly dependent on the original class being the only caller.

PR #1500, opened 2022-09-29
Fix method return url and add test

hannahramadan:
Fix method return URL and add test for valid GitHub diff URLs.


tannalynn (approved):
🎉

PR #1504, opened 2022-09-29
Address Minitest/AssertRespondTo

hannahramadan:
Address `Minitest/AssertRespondTo` and remove from `rubocop_todo`

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -48,16 +48,6 @@ Minitest/AssertRaisesCompoundBody:
     - 'test/new_relic/agent/new_relic_service_test.rb'
     - 'test/new_relic/cli/commands/deployments_test.rb'
 
-# Offense count: 9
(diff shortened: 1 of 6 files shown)


.rubocop_todo.yml
```diff
@@ -48,16 +48,6 @@ Minitest/AssertRaisesCompoundBody:
     - 'test/new_relic/agent/new_relic_service_test.rb'
     - 'test/new_relic/cli/commands/deployments_test.rb'
 
-# Offense count: 9
-# This cop supports safe autocorrection (--autocorrect).
-Minitest/AssertRespondTo:
```
hannahramadan, line 53: I thought this was a nice one to add since we're already using `assert_respond_to` more in our codebase than not

tannalynn (approved):
🎉 nice i don't think i even knew this was a thing before

fallwith (approved):

PR #1505, opened 2022-09-29
Minitest/AssertWithExpectedArgument

hannahramadan:
Minitest/AssertWithExpectedArgument is unsafe, so a note was added to `rubocop_todo` about why we should leave it. The files that had AssertWithExpectedArgument offenses were edited.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -58,20 +58,6 @@ Minitest/AssertRespondTo:
     - 'test/new_relic/collection_helper_test.rb'
     - 'test/new_relic/data_container_tests.rb'
 
-# Offense count: 24
-# NOTE: As of September 2022, this cop is unsafe because it is not possible to determine whether 
-# the second argument of assert is a message or not. Leaving disabled.
-Minitest/AssertWithExpectedArgument:
-  Exclude:
-    - 'infinite_tracing/test/infinite_tracing/worker_test.rb'
-    - 'test/agent_helper.rb'
-    - 'test/multiverse/suites/grpc/grpc_client_test.rb'
-    - 'test/multiverse/suites/sidekiq/sidekiq_instrumentation_test.rb'
-    - 'test/new_relic/agent/configuration/default_source_test.rb'
-    - 'test/new_relic/agent/configuration/orphan_configuration_test.rb'
-    - 'test/new_relic/agent/distributed_tracing/distributed_tracing_cross_agent_test.rb'
-    - 'test/new_relic/control_test.rb'
-
 # Offense count: 4
 Minitest/DuplicateTestRun:
   Exclude:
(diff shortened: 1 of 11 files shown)


.rubocop_todo.yml
```diff
@@ -59,18 +59,17 @@ Minitest/AssertRespondTo:
     - 'test/new_relic/data_container_tests.rb'
 
 # Offense count: 24
+# NOTE: As of September 2022, this cop is unsafe because it is not possible to determine whether 
+# the second argument of assert is a message or not. Leaving disabled.
```
hannahramadan, line 63: `Minitest/AssertWithExpectedArgument` is unsafe, so a note was added to rubocop_todo about why we should leave it. The files that had valid offenses were edited.

fallwith: I am unclear on what makes the cop "unsafe", even after reading the comment.

My hunch is that the cop is perfectly safe, but performing an autocorrection of offenses would require the `-A` flag for unsafe / may break the functionality of the code operation. Is that the case? If so, would you please update the comment to say that it's the autocorrection action that is unsafe, and not the cop itself?

fallwith (changes requested):

test/new_relic/agent/agent_test.rb
```diff
@@ -475,7 +475,7 @@ def test_wait_raises_if_not_connected_once_signaled
         @agent.signal_connected
         thread.join
 
-        assert error, error
+        assert_equal error, error
```
kaylareopelle, line 478: These puzzle me a bit. Why wouldn't `error == error`? Why might we need this assertion?

fallwith: My hunch is that `assert` required truthiness before, whereas `assert_equal` does not. In the test, `error` starts out `nil` and is expected to become non-nil by the time the assertions take place.

@hannahramadan I think you can satisfy both my hunch and Kayla's concern with this change:

```diff
- assert_equal error, error
+ refute_nil error
```

fallwith: Granted `nil` will also fail the `kind_of` check the next assertion performs immediately afterwards. So if you'd prefer, you can get rid of the `error, error` assertion entirely instead of changing it.

.rubocop_todo.yml
hannahramadan: `AssertWithExpectedArgument` doesn't support autocorrect, but I called it unsafe because that's what [the docs](https://docs.rubocop.org/rubocop-minitest/cops_minitest.html#minitestassertwithexpectedargument) say about it. Maybe a better comment is:

> # NOTE: This cop detects when an 'assert' is used when it should be 'assert_equal'. As of September 2022, this cop is unable to determine whether the second argument of assert is a message or not. Disabled for now.

test/new_relic/agent/agent_test.rb
hannahramadan: I was also confused on this, but assumed I was missing something about the code 😅 I'll remove the assert entirely.

fallwith (approved):

fallwith (approved):

PR #1509, opened 2022-09-30
Address Assert/ RefutePathExists changes

hannahramadan:
Address Assert/ RefutePathExists

### DIFF
--- a/infinite_tracing/test/support/fixtures.rb
+++ b/infinite_tracing/test/support/fixtures.rb
@@ -11,7 +11,9 @@ end
 
 def span_event_fixture(fixture_name)
   fixture_filename = File.join(fixtures_path, "span_events", "#{fixture_name}.yml")
-  assert_path_exists(fixture_filename, "Missing Span Event Fixture: #{fixture_name}. Looked for #{fixture_filename}")
+  # rubocop:disable Minitest/AssertPathExists
+  assert File.exist?(fixture_filename), "Missing Span Event Fixture: #{fixture_name}. Looked for #{fixture_filename}"
+  # rubocop:enable Minitest/AssertPathExists
   if YAML.respond_to?(:unsafe_load)
     YAML::unsafe_load(File.read(fixture_filename))
   else
--- a/infinite_tracing/test/support/fixtures.rb
+++ b/infinite_tracing/test/support/fixtures.rb
@@ -3,6 +3,7 @@
 # frozen_string_literal: true
 
 require 'yaml'
+require_relative '../test_helper'
 
 def fixtures_path
   @fixtures_path ||= File.expand_path(File.join(File.dirname(__FILE__), '..', 'fixtures'))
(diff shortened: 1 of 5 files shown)


infinite_tracing/test/support/fixtures.rb
```diff
@@ -10,7 +10,9 @@ def fixtures_path
 
 def span_event_fixture(fixture_name)
   fixture_filename = File.join(fixtures_path, "span_events", "#{fixture_name}.yml")
+  # rubocop:disable Minitest/AssertPathExists
   assert File.exist?(fixture_filename), "Missing Span Event Fixture: #{fixture_name}. Looked for #{fixture_filename}"
+  # rubocop:enable Minitest/AssertPathExists
```
hannahramadan, line 15: I'm curious about this one. `assert_path_exists` raises a NoMethodError. I'm guessing because we're not bringing in Minitest—either by require or inheriting from the class. So whey then is `assert` okay?

kaylareopelle: What version of minitest does infinite tracing use? Perhaps it's a version that does not have `assert_path_exists` defined? I can find the definition for it on the main branch of minitest: https://github.com/minitest/minitest/blob/master/lib/minitest/assertions.rb#L355 

Can you compare this with the version the infinite tracing tests run on and see if it's still there?

fallwith: +1 on getting to the bottom of this one.

If it turns out that we do need to disable this cop for 8T, then I think it would be better if we do so at the file level inside of either `.rubocop.yml` or `.rubocop_todo.yml`. By disabling it inline only inside of one test block, it makes it seem as though it would work everywhere else in the file.

hannahramadan: Yup, that was it! The action that failed was Minitest 5.10.5, and `assert_path_exists` was introduced in 5.15.0. Does the current disable of the cop work as is?

hannahramadan: I see what you're saying James, it is misleading especially since it works on some 8T tests. I can disabled at the file level instead.

fallwith: @hannahramadan Glad you got to the bottom of why the assertion was missing from that MiniTest version.

Given that 8T requires Ruby 2.5+ (or even newer if users wish to use the absolute latest version of the `grpc` gem), I think we're totally good with upgrading to the latest MiniTest version for 8T.

hannahramadan: Thanks for the help on this. [The change](https://github.com/newrelic/newrelic-ruby-agent/pull/1509/files#diff-5c92585b1f81da4bb4c96bff0e84e668725467ce36ed0b1915c68e866c50bbfdR303-R304) we made to use a later minitest with 2.5+ works across the CI.

fallwith (approved):
👍

PR #1525, opened 2022-10-05
Add elasticsearch instrumentation

tannalynn:

# Overview
Adds instrumentation for the [elasticsearch](https://rubygems.org/gems/elasticsearch) gem 7.0+.
Also updates the docker-compose file and the CI to run elasticsearch 7 and 8 to support our tests.
The Mongo sql obfuscator was moved to be used more generally for nosql obfuscation.

New configuration options:
- `instrumentation.elasticsearch`
- `elasticsearch.capture_queries`
- `elasticsearch.obfuscate_queries`

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


.github/workflows/scripts/rubygems-publish.rb
```diff
@@ -21,7 +22,7 @@
 if $?.to_i.zero?
   puts "#{gem_filename} successfully pushed to rubygems.org!"
 else
-  if result =~ /Repushing of gem versions is not allowed/
+  if result.include?('Repushing of gem versions is not allowed')
```
tannalynn, line 25: Rubocop was mad about this file for some reason, so I had to make these changes to get it to commit and run the tests.

.github/workflows/ci.yml
```diff
@@ -315,12 +363,12 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
-      memcached:
-        image: memcached:latest
+      redis:
```
kaylareopelle, line 366: Love the alphabetizing!

newrelic.yml
```diff
@@ -459,6 +459,12 @@ common: &default_settings
   # If true, the agent obfuscates Mongo queries in transaction traces.
   # mongo.obfuscate_queries: true
```
kaylareopelle, line 461: The instrumentation generator should probably be updated to add the config entry to this file too. I can take that todo if you'd like!

test/multiverse/suites/elasticsearch/elasticsearch_instrumentation_test.rb
```diff
@@ -0,0 +1,176 @@
...
+class ElasticsearchInstrumentationTest < Minitest::Test
+  def setup
+    # Keeping the log off prevents noisy test output
+    @client = ::Elasticsearch::Client.new(
+      log: false,
+      hosts: "localhost:#{port}"
+    )
+    # Ensure the client is running before the tests start
+    @client.cluster.health
+    # TODO: Update this to use constants, perhaps a hash?
```
kaylareopelle, line 17: I think we can probably remove this TODO since the index arguments are only passed once

lib/new_relic/agent/instrumentation/elasticsearch.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  depends_on do
+    defined?(::Elasticsearch)
+  end
+
+  executes do
+    ::NewRelic::Agent.logger.info('Installing elasticsearch instrumentation')
+    # why didn't this work when we looked at defined?(::Elastic)
+    # why was the name changed?
+    # perform_request is also defined in Transport::Base
```
kaylareopelle, line 20: We can probably remove these comments now

lib/new_relic/agent/database.rb
```diff
@@ -42,6 +42,7 @@ def capture_query(query)
       end
 
       def truncate_query(query)
+        return unless query
```
kaylareopelle, line 45: I was surprised to see this covered by SimpleCov. We may want to add a test somewhere for this functionality.

newrelic.yml
kaylareopelle:
```suggestion

# Controls auto-instrumentation of the elasticsearch library at start up. 
# May be one of [auto|prepend|chain|disabled].
# instrumentation.elasticsearch: auto

```

kaylareopelle: (Might be too many spaces?? I can't tell)

.github/workflows/ci.yml
```diff
@@ -101,6 +101,43 @@ jobs:
...
+          --health-cmd "curl http://localhost:9200/_cluster/health"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 10
+      elasticsearch8:
+        image: elasticsearch:8.4.2
+        env:
+          discovery.type: single-node
+          xpack.security.enabled: false
+          ES_JAVA_OPTS: "-Xms1g -Xmx1g"
```
fallwith, line 120: I'm curious about this one.

- I'm not sure where we tell GHA that we'd like to have a gig of memory available to be allocated here
- My hunch is that in a docker-compose / GHA services context, the `java` process has already been started and this env var might be getting ignored.

docker-compose.yml
```diff
@@ -1,5 +1,25 @@
 version: "3.9"
 services:
+  elasticsearch7:
+    image: elasticsearch:7.16.2
+    ports:
+      - "9200:9200"
+    environment:
+      - discovery.type=single-node
+      - bootstrap.memory_lock=true
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
```
fallwith, line 10: It looks like for docker-compose, the `ES_JAVA_OPTS` was applied to both v7 and v8, but for GHA it was only used for one of the versions. Should it be used for both?

docker-compose.yml
```diff
@@ -1,5 +1,25 @@
...
+    mem_limit: 1g
```
fallwith, line 11: Does the `mem_limit` setting override the `ES_JAVA_OPTS` setting, which looks like it may be "mem limit" related as well?

lib/new_relic/agent/datastores/mongo/obfuscator.rb
```diff
@@ -1,43 +0,0 @@
-# This file is distributed under New Relic's license terms.
```
fallwith, line 1: Looks like this change (`obfuscator.rb` -> `nosql_obfuscator.rb`) is appearing as a deletion and an addition rather than a rename. Hopefully the new file hasn't lost the git history of changes made to it.

lib/new_relic/agent/datastores/nosql_obfuscator.rb
```diff
@@ -0,0 +1,41 @@
...
+                obfuscated[key] = obfuscate_value(value, allowlist)
+              end
+            end
+            obfuscated
+          else
+            obfuscate_value(source, allowlist)
+          end
+        end
+
+        QUESTION_MARK = '?'.freeze
```
fallwith, line 27: I realize this isn't a new line, but just showed up as such from the rename - but this `#freeze` operation is redundant. The magic comment at the top is already freezing the string.

lib/new_relic/agent/instrumentation/elasticsearch.rb
```diff
@@ -0,0 +1,31 @@
...
+
+DependencyDetection.defer do
+  named :elasticsearch
+
+  depends_on do
+    defined?(::Elasticsearch)
+  end
+
+  executes do
+    ::NewRelic::Agent.logger.info('Installing elasticsearch instrumentation')
```
fallwith, line 17: The human facing text elsewhere seems to use "Elasticsearch" with a capital "E". Should we use the capitalized "E" here as well, or is it deliberately lowercase to match the `named` value?

lib/new_relic/agent/instrumentation/elasticsearch.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    to_instrument = if ::Gem::Version.create(::Elasticsearch::VERSION) < ::Gem::Version.create("8.0.0")
```
fallwith, line 19: Do you happen to know if `.create` is an alias for `.new` or if it does anything different? (Asking because I usually use `.new` and was curious when I saw `.create`)

lib/new_relic/agent/instrumentation/elasticsearch/instrumentation.rb
```diff
@@ -0,0 +1,55 @@
...
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+require_relative '../../datastores/nosql_obfuscator'
+
+module NewRelic::Agent::Instrumentation
+  module Elasticsearch
+    PRODUCT_NAME = 'Elasticsearch'
+    OPERATION = 'query'
+
+    def perform_request_with_tracing(method, path, params = {}, body = nil, headers = nil)
```
fallwith, line 11: Do we anticipate Elasticsearch ever modifying the params hash downstream? If not, we could use `NewRelic::EMPTY_HASH` here on the `def` line as an optimisation.

fallwith (changes requested):

docker-compose.yml
tannalynn: From what I understand, the amount of memory given to the ES_JAVA_OPTS for elasticsearch should be half of the memory limit given to docker compose with mem_limit. I don't remember there being any explanation as for _why_ it should be half though, but since that did solve the out of memory docker errors, I just stuck with it.

.github/workflows/ci.yml
tannalynn: Yeah, I'm not sure it's really needed here. I included it with 8 here because of an example for 8 I found when I was figuring it out. But after working on the issues with docker compose I ran into with docker running out of memory, and since the CI doesn't ever have that problem, this is probably either ignored or pointless in a GHA context.

docker-compose.yml
tannalynn: I'm gonna take it out of GHA since it's probably not actually needed there. We do need it on both in docker compose though, since both 7 and 8 would fail with the out of memory error.

lib/new_relic/agent/datastores/mongo/obfuscator.rb
tannalynn: hmm that's a good point. if github treats it as an add/remove, it probably will lose the history prior to the move.

lib/new_relic/agent/datastores/nosql_obfuscator.rb
tannalynn: I bet that's originally from before that was a thing. I believe 2.2 doesn't recognize the frozen_string_literal comment though, so we might want to keep it around a little while longer till we drop support for 2.2?

lib/new_relic/agent/instrumentation/elasticsearch.rb
tannalynn: I'm not sure, I think I've only seen create being used in the agent. 
[Looking](https://ruby-doc.org/stdlib-2.5.0/libdoc/rubygems/rdoc/Gem/Version.html#method-c-create) at the source code section though it looks like they are slightly different, but create does use new as part of it.

test/multiverse/lib/multiverse/runner.rb
```diff
@@ -101,7 +101,7 @@ def execute_suites(filter, opts)
       "agent" => %w[agent_only bare config_file_loading deferred_instrumentation high_security no_json json marshalling yajl],
       "background" => %w[delayed_job sidekiq resque],
       "background_2" => ["rake"],
-      "database" => %w[datamapper mongo redis sequel],
+      "database" => %w[datamapper mongo redis sequel elasticsearch],
```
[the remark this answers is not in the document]
tannalynn: oh i didn't even notice that, thank you!

lib/new_relic/agent/datastores/nosql_obfuscator.rb
fallwith: Fair enough. I think anyone tolerating Ruby 2.2 performance in 2022 won't gain much from having a string frozen 😄 , but I'm ok with keeping it around up until there's a linter warning about the redundancy (I think there is one that exists, but we must have it disabled for now).

lib/new_relic/agent/instrumentation/elasticsearch/instrumentation.rb
tannalynn: Hmm, I don't think it would modify it later, but I am hesitant to do that. Usually I try to match the original methods args, so I do worry that using a frozen could potentially cause something weird to happen later. I'll look into it more tomorrow

fallwith: That's a good call. If the original method was instantiating a new hash, then we are not introducing any extra allocations by being the one to spawn it and hand it off downstream. And if we stick with an unfrozen empty hash as the original code uses, we're guaranteed not to break anything. I like that - let's just keep this as is.

.github/workflows/scripts/rubygems-publish.rb
fallwith: Using `=~` produces a fresh `Regexp.last_match` object to store the overall matched text as well as any individual captures. Given that we don't want to do anything with the matched check and simply verify with a boolean that there is a match, using `#match?` or `#include?` seems more efficient. Good change!

test/multiverse/suites/elasticsearch/elasticsearch_instrumentation_test.rb
```diff
@@ -0,0 +1,175 @@
...
+  end
+
+  def search
+    txn = in_transaction do
+      @client.search(index: 'my-index', body: {query: {match: {title: 'test'}}})
+    end
+    @segment = txn.segments[1]
+  end
+
+  # TODO! TEST METRIC GENERATION!!!
```
fallwith, line 33: I think this TODO has either been satisfied or can be tracked elsewhere outside of the code now.

fallwith (dismissed):
All of my new questions / concerns for the stuff I hadn't yet synced up with you on from before have now been addressed. The instrumentation itself is still looking great and I'm excited to get it out into users' hands. Great work!

fallwith (approved):

fallwith (approved):

lib/new_relic/agent/database.rb
tannalynn: added here: 693490fbdef49afc063c70586cfb25eb570fdc91

kaylareopelle (approved):
Hooray for new instrumentation! Thanks for your leadership, Tanna!

PR #1526, opened 2022-10-05
Raise SimpleCov thresholds

fallwith:
On the road to 100%/100%, it's important to not relapse with the introduction of new pockets of uncovered code. Let's keep the forward momentum by having the thresholds creep towards 100%/100% along with the tests.


hannahramadan (approved):
.11% wiggle room 🫠 This is a great change!

PR #1528, opened 2022-10-05
perf tests need 'rackup'

fallwith:
With the split into separate 'rack' and 'rackup' gems, the perf tests will now need to bring in 'rackup'


tannalynn (approved):
🦝 🆙

PR #1541, opened 2022-10-13
Increase custom_insights_events.max_samples_stored

### LINKED ISSUE #1540: Increase Custom Event Reservoir Default Max Samples Stored to 3,000
### Description
Increase `custom_insights_events.max_samples_stored` default value from 1000 to 3000.

kaylareopelle:
# Overview 

The previous default value for `custom_insights_events.max_samples_stored` was 1000. The new default is 3000.

This new number is based on OATS results from multiple language agents in an effort to reduce the number of dropped custom insights events.

Closes #1540

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -2004,7 +2004,7 @@ A map of error classes to a list of messages. When an error of one of the classe
           :description => 'If `true`, the agent captures [custom events](/docs/insights/new-relic-insights/adding-querying-data/inserting-custom-events-new-relic-apm-agents).'
         },
         :'custom_insights_events.max_samples_stored' => {
-          :default => 1000,
+          :default => 3000,
           :public => true,
           :type => Integer,
           :allowed_from_server => true,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
   ## v8.12.0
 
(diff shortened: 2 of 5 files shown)


tannalynn (dismissed):
🎉

fallwith (approved):

tannalynn (approved):

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2004,7 +2004,7 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :description => 'If `true`, the agent captures [custom events](/docs/insights/new-relic-insights/adding-querying-data/inserting-custom-events-new-relic-apm-agents).'
         },
         :'custom_insights_events.max_samples_stored' => {
-          :default => 1000,
+          :default => 3000,
```
hannahramadan, line 2007: Is there any impact to increasing this number? I'm wondering how we can up with 1000 in the first place, and why 3000 is the new number.

kaylareopelle: Hi Hannah! Please [see this slack thread (internal)](https://newrelic.slack.com/archives/GAPN6K7T2/p1665607818802639?thread_ts=1665607325.560299&cid=GAPN6K7T2)

1000 is actually lower than what all the other agents have at this time. The spec default is 10K per limit, so we're getting closer to the spec with a smaller performance impact by changing the value to 3K.

PR #1543, opened 2022-10-14
Set Elasticsearch operation dynamically

kaylareopelle:
# Overview
* Set Elasticsearch datastore segment's operation attribute dynamically from `caller_locations`
* Elasticsearch 8 Playground tested here: https://staging.onenr.io/0dBj3VPrvRX (internal)
* Elasticsearch 7 Playground tested here: https://staging.onenr.io/0dBj3VWe2RX (internal)
* Previously, the operations were all `query`, the fallback value is proposed to become `perform_request` since the dynamic operations are set to the methods called, it seems odd to have the operation default be unrelated to any Elasticsearch method name. Totally down to revert this change, however!
* **BONUS**: Reorganizes and indents the `instrumentation.elasticsearch` config value in `newrelic.yml`

Relates to #1525

### DIFF
--- a/test/multiverse/suites/elasticsearch/elasticsearch_instrumentation_test.rb
+++ b/test/multiverse/suites/elasticsearch/elasticsearch_instrumentation_test.rb
@@ -30,8 +30,6 @@ class ElasticsearchInstrumentationTest < Minitest::Test
     @segment = txn.segments[1]
   end
 
-  # TODO! TEST METRIC GENERATION!!!
-
   def test_datastore_segment_created
     search
     assert_equal NewRelic::Agent::Transaction::DatastoreSegment, @segment.class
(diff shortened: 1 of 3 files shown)


test/multiverse/suites/elasticsearch/elasticsearch_instrumentation_test.rb
```diff
@@ -42,9 +40,26 @@ def test_segment_elasticsearch_product
...
+    txn = in_transaction do
+      @client.index(index: 'my-index', id: 1, body: {title: 'Test'})
+    end
+
+    segment = txn.segments[1]
+    assert_equal 'index', segment.operation
+  end
+
+  def test_segment_operation_returns_OPERATION_when_method_nil
+    skip('How can I stub only the Array#index method?')
```
tannalynn, line 58: I'm assuming this means that the following line trying to stub isn't working how you want. I'm not sure how to do it with minitest stubs as I'm less familiar with that, but maybe James has some ideas since he's more familiar that.
I've used mocha more often so I'm more familiar with it. I'm pretty sure we can do this with mocha stubs though, using any_instance. 
`Array.any_instance.stubs(:index).returns(nil)`
any_instance isn't really recommended generally, but we still do use it quite a lot in the agent tests already. If we do it this way, we'd also want to make sure `mocha_teardown` gets added to the teardown method in here as well, so the stub doesn't leak past the test.

kaylareopelle: Oh, I'm bummed I missed this until now! I readjusted the code and the tests a bit to make this case work in a different way. I'd be curious to see what you think!

tannalynn (approved):
This awesome, thank you!

test/multiverse/suites/elasticsearch/elasticsearch_instrumentation_test.rb
tannalynn: oh its cool, I'm glad you figured out another way to test it!

PR #1544, opened 2022-10-14
RuboCop: stop disabling EmptyLineAfterGuardClause

fallwith:
use the RuboCop default of `Layout/EmptyLineAfterGuardClause` and require an empty linen to separate a guard line (usually a `raise`, `fail`, or `return`) from other subsequent lines. Note that multiple sequential guard lines can still be grouped together and the blank line need only follow the entire group.

### DIFF
--- a/lib/new_relic/agent/configuration/dotted_hash.rb
+++ b/lib/new_relic/agent/configuration/dotted_hash.rb
@@ -36,6 +36,7 @@ module NewRelic
         def dot_flattened(nested_hash, names = [], result = {})
           nested_hash.each do |key, val|
             next if val.nil?
+
             if val.respond_to?(:has_key?)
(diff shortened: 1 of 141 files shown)


lib/new_relic/agent/configuration/dotted_hash.rb
```diff
@@ -36,6 +36,7 @@ def self.symbolize(hash)
         def dot_flattened(nested_hash, names = [], result = {})
           nested_hash.each do |key, val|
             next if val.nil?
+
```
hannahramadan, line 39: `next` is interesting to me as a guard clause! I guess here we're exiting early from a particular iteration

hannahramadan (approved):
I like! Makes code easier to read

kaylareopelle (approved):
Thank you!

PR #1547, opened 2022-10-14
Update instrumentation_generator

kaylareopelle:
# Overview
* Append to newrelic.yml when instrumentation:scaffold is run
* Rename create_configuration to append_to_default_source
* Be more specific in newrelic.yml, default_source text with letter case

These changes were noticed while reviewing the Elasticsearch PR

### DIFF
--- a/lib/tasks/instrumentation_generator/instrumentation.thor
+++ b/lib/tasks/instrumentation_generator/instrumentation.thor
@@ -11,6 +11,7 @@ class Instrumentation < Thor
   INSTRUMENTATION_ROOT = 'lib/new_relic/agent/instrumentation/'
   MULTIVERSE_SUITE_ROOT = 'test/multiverse/suites/'
   DEFAULT_SOURCE_LOCATION = 'lib/new_relic/agent/configuration/default_source.rb'
+  NEWRELIC_YML_LOCATION = 'newrelic.yml'
 
   desc('scaffold NAME', 'Scaffold the required files for adding new instrumentation')
   long_desc <<-LONGDESC
@@ -37,7 +38,8 @@ class Instrumentation < Thor
 
     empty_directory(base_path)
     create_instrumentation_files(base_path)
-    create_configuration(name)
+    append_to_default_source(name)
+    append_to_newrelic_yml(name)
     create_tests(name)
   end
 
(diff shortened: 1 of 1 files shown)


lib/tasks/instrumentation_generator/instrumentation.thor
```diff
@@ -85,18 +87,35 @@ class Instrumentation < Thor
...
           :dynamic_name => true,
           :allowed_from_server => false,
-          :description => 'Controls auto-instrumentation of the #{library} library at start up. May be one of [auto|prepend|chain|disabled].'
+          :description => 'Controls auto-instrumentation of the #{name} library at start up. May be one of [auto|prepend|chain|disabled].'
         },
     CONFIG
   end
+
+  def yaml_block(name)
+<<HEREDOC # rubocop:disable Layout/IndentationWidth
```
kaylareopelle, line 112: Cop disabling required to get the appropriate indentation in the .yml file

lib/tasks/instrumentation_generator/instrumentation.thor
```diff
@@ -85,18 +87,35 @@ class Instrumentation < Thor
...
-  def config_block(library)
+  def append_to_newrelic_yml(name)
+    insert_into_file(
+      NEWRELIC_YML_LOCATION,
+      yaml_block(name),
+      after: "# instrumentation.bunny: auto\n"
+    )
+  end
+
+  def config_block(name)
```
kaylareopelle, line 98: For both this method and `yaml_block`, I suppose their contents aren't blocks, but heredocs. Should the method names reflect that?

lib/tasks/instrumentation_generator/instrumentation.thor
fallwith: The squiggle heredoc format would be perfect here:

```ruby
  def yaml_block(name)
    <<~INSTRUMENTATION_CONFIG_INFO
      # Controls auto-instrumentation of #{name} at start up.
      # instrumentation.#{name.downcase}: auto
    INSTRUMENTATION_CONFIG_INFO
  end
```

It requires Ruby 2.3+. If this generator is intended to support Ruby 2.2, we could leave a TODO to suggest moving to a squiggle heredoc once the option is available.

lib/tasks/instrumentation_generator/instrumentation.thor
fallwith: I like what you have now. It gives you a block of configuration, and I don't get that "block" meaning confused with a Ruby callable block or a block or code or whatnot.

fallwith (approved):

tannalynn (approved):

PR #1560, opened 2022-10-21
don't explicitly include jruby-openssl

fallwith:
Previously, jruby-openssl was being brought in for both CRuby where it is absolutely not needed and JRuby 9k+, where it is likely not needed as a separate Gemfile entry. The gem is now no longer brought in.


fallwith:
CI tests demonstrated as passing with #1561

tannalynn (approved):

PR #1562, opened 2022-10-24
8T compression: ensure that the value is a symbol

fallwith:
for the 8T compression setting, specify that it comes in as a string and then convert it to a symbol when using it with grpc


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2472,7 +2472,7 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
         :'infinite_tracing.compression_level' => {
           :default => :none,
```
kaylareopelle, line 2473:
```suggestion
          :default => 'none',
```

kaylareopelle (approved):

fallwith:
@hannahramadan et. al. - please do not merge this PR until v8.12.0 has gone out

PR #1566, opened 2022-10-27
Return params if undefined or nil

hannahramadan:
Return params if the RAILS_FILTER_CLASS is undefined or nil

### DIFF
--- a/lib/new_relic/agent/parameter_filtering.rb
+++ b/lib/new_relic/agent/parameter_filtering.rb
@@ -34,7 +34,7 @@ module NewRelic
       end
 
       def filter_using_rails(params, filters)
-        return params if !defined?(RAILS_FILTER_CLASS) || (RAILS_FILTER_CLASS.nil?)
+        return params if rails_filter_class?
 
         pre_filtered_params = filter_rails_request_parameters(params)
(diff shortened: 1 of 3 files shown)


lib/new_relic/agent/parameter_filtering.rb
```diff
@@ -34,7 +34,7 @@ def apply_filters(env, params)
       end
 
       def filter_using_rails(params, filters)
-        return params if RAILS_FILTER_CLASS.nil?
+        return params if !defined?(RAILS_FILTER_CLASS) || (RAILS_FILTER_CLASS.nil?)
```
kaylareopelle, line 37: You might already be working on this, but I wanted to drop in a reminder that the if condition might be a good candidate for a private method! Maybe named something like, `rails_filter_class?`

hannahramadan: Does [this](https://github.com/newrelic/newrelic-ruby-agent/pull/1566/commits/6edc1e66b4d29b94c2556755955f4606068e119c) work? :)

fallwith (approved):

kaylareopelle (approved):

PR #1577, opened 2022-10-31
Replace .zero? with == 0 in agent code

kaylareopelle:
In all cases except Float's `.zero?` method, using `== 0` for the comparison is more performant in microbenchmarks. [Source](https://andycroll.com/ruby/use-zero-for-integer-float-comparison/) 

In addition, replace the generation of a new string each time `'STDOUT'` is referenced with a constant for all values.


lib/new_relic/agent/agent_logger.rb
```diff
@@ -132,7 +132,7 @@ def create_null_logger
       end
 
       def wants_stdout?
-        ::NewRelic::Agent.config[:log_file_path].casecmp("STDOUT").zero?
+        ::NewRelic::Agent.config[:log_file_path].casecmp("STDOUT") == 0
```
fallwith, line 135: Given the micro level performance concern, we ought to stop initializing a new string `'STDOUT'` every time here and use a constant instead. NOTE: I recommend naming the constant something like `STANDARD_OUT` to avoid conflicting with Ruby's existing `STDOUT` constant.

```shell
$ ruby test.rb                                                                                                           
       user     system      total        real
zero?  0.008423   0.000190   0.008613 (  0.008612)
== 0  0.006353   0.000066   0.006419 (  0.006451)
zero? with a constant  0.004898   0.000004   0.004902 (  0.004904)
== 0 with a constant  0.004168   0.000001   0.004169 (  0.004171)
upcase and match with a constant  0.066901   0.001593   0.068494 (  0.068576)
```

lib/new_relic/agent/audit_logger.rb
```diff
@@ -95,7 +95,7 @@ def ensure_log_path
       end
 
       def wants_stdout?
-        ::NewRelic::Agent.config[:'audit_log.path'].casecmp("STDOUT").zero?
+        ::NewRelic::Agent.config[:'audit_log.path'].casecmp('STDOUT') == 0
```
fallwith, line 98: Same "use a `STANDARD_OUT` constant" recommendation here.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -157,7 +157,7 @@ def self.agent_enabled
         def self.audit_log_path
           proc {
             log_file_path = NewRelic::Agent.config[:log_file_path]
-            wants_stdout = (log_file_path.casecmp('STDOUT').zero?)
+            wants_stdout = (log_file_path.casecmp('STDOUT') == 0)
```
fallwith, line 160: Same "use a `STANDARD_OUT` constant" recommendation here.

lib/new_relic/agent/configuration/environment_source.rb
```diff
@@ -51,7 +51,7 @@ def set_dotted_alias(original_config_setting)
 
         def set_log_file
           if ENV['NEW_RELIC_LOG']
-            if ENV['NEW_RELIC_LOG'].casecmp('STDOUT').zero?
+            if ENV['NEW_RELIC_LOG'].casecmp('STDOUT') == 0
```
fallwith, line 54: Same "use a `STANDARD_OUT` constant" recommendation here.

fallwith (changes requested):

tannalynn (approved):
Looks great! 🎉

fallwith (approved):

PR #1585, opened 2022-11-02
Correctly identify Unicorn when used with Rack v3

### LINKED ISSUE #1394: CI: Determine the root cause of Rack v3.0.0 causing Unicorn errors
**Rack is a core technology. To meet our SLA, this work must be complete by December 5, 2022**

When Rack v3.0.0 is involved, 2 of the Unicorn CI tests fail. These failures are perhaps related to #1393 which sees errors reported for HTTP client gems, but the errors in question are different and specific to Unicorn.

We should determine the root cause of the failures and determine if the agent, Rack, and/or Unicorn needs any changes.

fallwith:
With version 3.0.0 of Rack in play, the agent can be fooled into thinking that web dispatching duties are being handled by Webrick. This is because the agent performs a less strict test for Webrick compared to other dispatchers; checking only for the existence of related namespaces to be defined.

With these changes, the agent will now treat Webrick as the last dispatcher checked after all other anticipated production use dispatchers have not been located. Specifically, this means that Webrick is checked after Unicorn, Rainbows, and FastCGI; all of which may have been incorrectly reporting as having Webrick as the dispatcher if Rack v3.0.0 was being used.

Resolves #1394

### DIFF
--- a/lib/new_relic/local_environment.rb
+++ b/lib/new_relic/local_environment.rb
@@ -71,10 +71,10 @@ module NewRelic
         thin
         mongrel
         litespeed
+        unicorn
         webrick
         fastcgi
         rainbows
-        unicorn
       ]
       # TODO: MAJOR VERSION - remove rainbows
       while dispatchers.any? && @discovered_dispatcher.nil?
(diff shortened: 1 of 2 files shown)


kaylareopelle (approved):
Thank you, James!

tannalynn (approved):
🎉

PR #1591, opened 2022-11-04
NR-1222/rack-test-v2

### LINKED ISSUE #1222: CI: Investigate test failure when using rack-test 2.0.0
## Description
We use rack-test in a lot of our test suites. Recently rack-test 2.0.0 was released, which caused one test in multiverse[rails] to consistently fail across all ruby and rails versions that use this version of rack-test. All other tests appear to be working as expected when using rack-test 2.0.0. For now, we are pinning rack-test < 2.0 to avoid this blocking our CI runs.

```
Failure:
ActionControllerLiveRumTest#test_excludes_rum_instrumentation_when_streaming_with_action_stream_true [/Users/tmcclure/projects/ruby-agent/test/multiverse/suites/rails/action_controller_live_rum_test.rb:45]:
Expected "<html><head>\n<script type=\"text/javascript\">window.NREUM||(NREUM={});NREUM.info={\"beacon\":\"beacon\",\"errorBeacon\":\"\",\"licenseKey\":\"key\",\"applicationID\":\"\",\"transactionName\":\"IQteEUMLWV4EFk4RCwYBUQEeBkdTCAo+FxEQAVEI\",\"queueTime\":0,\"applicationTime\":18,\"agent\":\"\"}</script>\n<script type=\"text/javascript\">JS LOADER IN DA HOUSE</script></head><body>Brains!</body></html>" 
to include "<html><head></head><body>Brains!</body></html>".
```

hannahramadan:
As of v2.0.0, rack-test defaults to HTTP/1.0 to work around the fact some apps return Transfer-Encoding: chunked responses, which rack-test doesn't have a decoder for. See [rack-test discussion](https://github.com/rack/rack-test/issues/300) on this decision.

Only HTTP/1.1 supports chunking, which our [`steaming?`](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/rack/browser_monitoring.rb#L106-L111) method checks for. For streaming tests to work with rack-test v2, we need to specify HTTP/1.1 when using streaming.

Closes #1222

### DIFF
--- a/lib/new_relic/rack/browser_monitoring.rb
+++ b/lib/new_relic/rack/browser_monitoring.rb
@@ -104,6 +104,7 @@ module NewRelic
       end
 
       def streaming?(env, headers)
+        # Chunked transfer encoding is a streaming data transfer mechanism available only in HTTP/1.1
         return true if headers && headers['Transfer-Encoding'] == 'chunked'
 
         defined?(ActionController::Live) &&
--- a/test/multiverse/suites/rails/Envfile
+++ b/test/multiverse/suites/rails/Envfile
@@ -44,7 +44,7 @@ def minitest_rails_version(rails_version = nil)
 end
 
 def pre2_rack_test
-  %Q(gem 'rack-test', '< 2')
+  %Q(gem 'rack-test')
 end
 
 def gem_list(rails_version = nil)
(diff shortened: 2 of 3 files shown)


kaylareopelle (dismissed):
YAY! I'm so glad we have an example with how to use `:env` in our `get` calls in our codebase now! Awesome work! I love the extended commit message too. 

I wonder if there's any other due diligence that can occur to help future people reading the `#streaming?` method.

fallwith (approved):

PR #1600, opened 2022-11-08
Rack instrumentation: support and test with v3+

### LINKED ISSUE #791: Update Core Tech: Rack Gem
This is a follow-up story from Ignite Sprint's goal.
- [ ] Understand if methods were deprecated by gem owners.
- [ ] Find new methods that are of value
- [ ] Instrument those new methods
- [ ] Update our multiverse to dynamically pull the latest version so that we do not have to manually update each gem version number manually.

fallwith:
Rack instrumentation updates:

* With Rack v3.0.0, `Rack::Builder#new` now supports a block. Update our instrumentation to handle a block and still apply tracing.
* For Ruby versions that support Rack v3+, have the Rack multiverse suite test with an unpinned version of Rack so that we're always grabbing and testing with the latest release.

resolves #791

### DIFF
--- a/test/multiverse/suites/rack/Envfile
+++ b/test/multiverse/suites/rack/Envfile
@@ -14,7 +14,9 @@ PUMA_VERSIONS = [
 ]
 
 def gem_list(puma_version = nil)
-  rack_version = ' ~> 2.1.4' unless RUBY_VERSION >= '2.3.0'
+  # try to test with the latest Rack version by not specifying a version
+  # constraint at all, but provide one for older Rubies not supporting Rack v3+
+  rack_version = ", '~> 2.1.4'" if RUBY_VERSION < '2.3.0'
   <<-RB
     gem 'puma'#{puma_version}
     gem 'rack'#{rack_version}


test/multiverse/suites/rack/Envfile
```diff
@@ -14,10 +14,10 @@ PUMA_VERSIONS = [
 ]
 
 def gem_list(puma_version = nil)
-  rack_version = RUBY_VERSION >= '2.3.0' ? '~> 2.2.3' : '~> 2.1.4'
+  rack_version = ' ~> 2.1.4' unless RUBY_VERSION >= '2.3.0'
   <<-RB
     gem 'puma'#{puma_version}
-    gem 'rack', '#{rack_version}'
```
kaylareopelle, line 20: The new implementation is missing an `, '` to correctly separate the version number from the gem name. Without this, a bundler syntax error is thrown (which is where the CI Rack failures are coming from).  Since we're down to just one version, perhaps adding the `, '` to the `rack_version` variable assignment would look cleaner than keeping it on line 20?

lib/new_relic/agent/instrumentation/rack/chain.rb
```diff
@@ -21,8 +21,16 @@ def to_app_with_newrelic_deferred_dependency_detection
...
-              run_with_tracing(app) { |wrapped_app| run_without_newrelic(wrapped_app, *args) }
+            def run_with_newrelic(app = nil, *args, &block)
+              app_or_block = app || block
+              run_with_tracing(app_or_block) do |wrapped|
+                # Rack::Builder#run for Rack v3+ supports a block, and does not
+                # support args. Whether a block or an app is provided, that
+                # callable object will be wrapped into a MiddlewareProxy
+                # instance. That proxy instance must then be passed to
+                # run_without_newrelic as the app argument.
+                block ? run_without_newrelic(wrapped, &nil) : run_without_newrelic(wrapped, *args)
```
kaylareopelle, line 32: I don't think I've ever seen `&nil` before. Can you tell me what it means?

test/multiverse/suites/rack/rack_builder_test.rb
```diff
@@ -112,6 +112,21 @@ def instance.middleware_instrumentation_enabled?; false; end
     end
   end
 
+  # Rack::Builder#run supports blocks as of v3.0.0
+  def test_run_with_tracing_with_the_original_run_method_is_given_a_block
+    skip 'Requires Rack v3+' unless defined?(Rack::RELEASE) && Rack::RELEASE >= '3.0.0'
```
kaylareopelle, line 117: Interesting! I didn't realize they weren't on the VERSION constant train 🚂

kaylareopelle (changes requested):

test/multiverse/suites/rack/Envfile
fallwith: Thank you! Addressed with [cfe4922](https://github.com/newrelic/newrelic-ruby-agent/pull/1600/commits/cfe4922ffcad9b540cd8bb14e827c14e32bc811f)

lib/new_relic/agent/instrumentation/rack/chain.rb
fallwith: Absolutely. A Ruby method will use `super` (no args specified) to pass all of its own arguments or `super()` (with args specified) to pass only the specified arguments, but in both cases the calling method's own block argument will be passed along. Because we end up with a middleware proxy object for tracing, passing it as a block won't work. So I just `nil` the original block out here and ensure that the proxy object is always passed as an argument.

The original `run` method we are tracing allows either an argument or a block (but not both) and simply treats whichever came in as the `Builder` instance's `@app` object. Given that blocks and arguments are treated identically and that the new support for a block is really just a syntax convenience thing, this approach of our always standardizing on the argument approach should work.

Beyond that, Rack v3's `run` doesn't seem to support `*args`. I'm keeping them around for the older versions, but if you've passed in a block then I know you're on v3+ and I don't use them.

Please let me know if I can rework the comment blocks to help capture any of this info better.

test/multiverse/suites/rack/rack_builder_test.rb
fallwith: There is a `Rack::VERSION` constant, but it references the [spec](https://github.com/rack/rack/blob/main/SPEC.rdoc). We need `Rack::RELEASE` to check the gem version itself.

PR #1604, opened 2022-11-08
resume testing Rack/Puma with Ruby 3.2.0

fallwith:
with the recent Rack v3 changes that have taken place, things seem to work with Ruby 3.2.0 now


kaylareopelle (approved):
Thanks for cleaning this up!

PR #1608, opened 2022-11-09
CI: pause testing of Rack/Puma with Ruby 3.2.0

fallwith:
revert PR 1604, and go back to skipping the Rack/Puma tests with Ruby 3.2.0-preview2 until we determine the root cause of the compatibility
  issues that still persist in our CI system


tannalynn:
closed in favor of https://github.com/newrelic/newrelic-ruby-agent/pull/1609

Issue #1361, opened 2022-08-30
CI: Introduce redis gem v5+ for testing


fallwith:
**Note: As a core technology, we have 90 days to adopt this change. This is due November 27, 2022**

The `redis/Envfile` has had its `nil` entry (signifying the use of the latest stable `redis` gem version) commented out for now due to some CI errors that happen when when `redis` v5 is used instead of v4 or v3.

We should investigate and address whatever it takes to get v5 working with our CI checks while retaining support for the older versions as well.

The Redis server version should also be inspected to make sure we're using a recent version.

kaylareopelle:
Some misc todos: 
* Need to make chain work with `call_v` Redis 5.x method
* Keep getting tests to pass 
* See if we need to set the host for our Redis tests at all (removing the host argument from Redis.new allowed us to connect to the Redis server running in Docker)

PR #1611, opened 2022-11-09
Update Redis instrumentation to support v5

kaylareopelle:
Co-authored-by: hramadan <hramadan@newrelic.com>

Redis gem 5.x comprises a major structural refactor for the library.

* Calls to methods like get and set are no longer routed through `call`, but through a method named `call_v`
* The methods we instrument have been moved into a new library, redis-client, a dependency of the redis gem.
* Prepending onto the new library's class (::RedisClient) for the `call_v` and connect methods maintains consistent reporting behavior with earlier versions of Redis.
* The location, behavior, and arguments of pipelined and multi methods changed.
* In previous versions, pipeline and multi calls were routed through a method named `call_pipeline`. There is a similar method in Redis 5.x named `call_pipelined`. However, this method is not defined in ::RedisClient, preventing our prepend instrumentation from working. ::RedisClient defines `multi` and `pipelined` methods, but they do not have access to the NoSQL statement as an argument or other instance variable/method on the caller.
* For `call_pipelined`, we now leverage Redis 5.x's new instrumentation API, `::RedisClient.register(module_name)`.
* This process introduces instrumentation too late to capture errors at the segment level. Despite this shortcoming, this strategy is used for `pipelined` and `multi` calls because it has access to the NoSQL statement

Closes: #1361

P.S. Test coverage is very high for Redis instrumentation, so we could rely on the existing test suite to inform us of the gaps with this new major version.

### DIFF
--- a/lib/new_relic/agent/instrumentation/redis/chain.rb
+++ b/lib/new_relic/agent/instrumentation/redis/chain.rb
@@ -9,16 +9,28 @@ module NewRelic::Agent::Instrumentation
         ::Redis::Client.class_eval do
           include NewRelic::Agent::Instrumentation::Redis
 
-          alias_method(:call_without_new_relic, :call)
+          if method_defined?(:call_v)
+            alias_method(:call_v_without_new_relic, :call_v)
 
-          def call(*args, &block)
-            call_with_tracing(args[0]) { call_without_new_relic(*args, &block) }
+            def call_v(*args, &block)
+              call_v_with_tracing(args[0]) { call_v_without_new_relic(*args, &block) }
+            end
           end
 
-          alias_method(:call_pipeline_without_new_relic, :call_pipeline)
+          if method_defined?(:call)
+            alias_method(:call_without_new_relic, :call)
 
-          def call_pipeline(*args, &block)
-            call_pipeline_with_tracing(args[0]) { call_pipeline_without_new_relic(*args, &block) }
+            def call(*args, &block)
+              call_with_tracing(args[0]) { call_without_new_relic(*args, &block) }
+            end
+          end
+
+          if method_defined?(:call_pipeline)
+            alias_method(:call_pipeline_without_new_relic, :call_pipeline)
+
+            def call_pipeline(*args, &block)
+              call_pipeline_with_tracing(args[0]) { call_pipeline_without_new_relic(*args, &block) }
+            end
           end
 
           alias_method(:connect_without_new_relic, :connect)
--- a/lib/new_relic/agent/instrumentation/redis/chain.rb
+++ b/lib/new_relic/agent/instrumentation/redis/chain.rb
@@ -13,7 +13,7 @@ module NewRelic::Agent::Instrumentation
             alias_method(:call_v_without_new_relic, :call_v)
 
             def call_v(*args, &block)
-              call_v_with_tracing(args[0]) { call_v_without_new_relic(*args, &block) }
+              call_with_tracing(args[0]) { call_v_without_new_relic(*args, &block) }
             end
           end
 
--- a/lib/new_relic/agent/instrumentation/redis/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/redis/instrumentation.rb
@@ -10,7 +10,7 @@ module NewRelic::Agent::Instrumentation
     LOCALHOST = 'localhost'
     MULTI_OPERATION = 'multi'
     PIPELINE_OPERATION = 'pipeline'
-    REDIS_5 = Gem::Version.new(::Redis::VERSION) >= Gem::Version.new('5.0.0') && !!defined?(::RedisClient)
+    HAS_REDIS_CLIENT = Gem::Version.new(::Redis::VERSION) >= Gem::Version.new('5.0.0') && !!defined?(::RedisClient)
 
     def connect_with_tracing
       with_tracing(CONNECT, database: db) { yield }
--- a/lib/new_relic/agent/instrumentation/redis/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/redis/instrumentation.rb
@@ -31,7 +31,7 @@ module NewRelic::Agent::Instrumentation
       with_tracing(operation, statement: statement, database: db) { yield }
     end
 
-    # Used for Redis 5.x
+    # Used for Redis 5.x+
     def call_pipelined_with_tracing(pipeline)
       operation = pipeline.flatten.include?('MULTI') ? MULTI_OPERATION : PIPELINE_OPERATION
       statement = ::NewRelic::Agent::Datastores::Redis.format_pipeline_commands(pipeline)
--- a/lib/new_relic/agent/instrumentation/redis/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/redis/instrumentation.rb
@@ -6,38 +6,58 @@ module NewRelic::Agent::Instrumentation
   module Redis
     PRODUCT_NAME = 'Redis'
     CONNECT = 'connect'
-    UNKNOWN = "unknown"
-    LOCALHOST = "localhost"
+    UNKNOWN = 'unknown'
+    LOCALHOST = 'localhost'
     MULTI_OPERATION = 'multi'
     PIPELINE_OPERATION = 'pipeline'
+    REDIS_5 = Gem::Version.new(::Redis::VERSION) >= Gem::Version.new('5.0.0') && !!defined?(::RedisClient)
 
+    def connect_with_tracing
+      with_tracing(CONNECT, database: db) { yield }
+    end
+
+    # Used for Redis 4.x and 3.x
     def call_with_tracing(command, &block)
       operation = command[0]
       statement = ::NewRelic::Agent::Datastores::Redis.format_command(command)
 
-      with_tracing(operation, statement) { yield }
+      with_tracing(operation, statement: statement, database: db) { yield }
     end
 
+    # Used for Redis 4.x and 3.x
     def call_pipeline_with_tracing(pipeline)
       operation = pipeline.is_a?(::Redis::Pipeline::Multi) ? MULTI_OPERATION : PIPELINE_OPERATION
       statement = ::NewRelic::Agent::Datastores::Redis.format_pipeline_commands(pipeline.commands)
 
-      with_tracing(operation, statement) { yield }
+      with_tracing(operation, statement: statement, database: db) { yield }
     end
 
-    def connect_with_tracing
-      with_tracing(CONNECT) { yield }
+    # Used for Redis 5.x
+    def call_pipelined_with_tracing(pipeline)
+      operation = pipeline.flatten.include?('MULTI') ? MULTI_OPERATION : PIPELINE_OPERATION
+      statement = ::NewRelic::Agent::Datastores::Redis.format_pipeline_commands(pipeline)
+
+      # call_pipelined isn't invoked on the client object, so use client.db to
+      # access the client instance var on self
+      with_tracing(operation, statement: statement, database: client.db) { yield }
+    end
+
+    # Used for Redis 5.x
+    def call_v_with_tracing(command, &block)
+      operation = command[0]
+      statement = ::NewRelic::Agent::Datastores::Redis.format_command(command)
+      with_tracing(operation, statement: statement, database: db) { yield }
     end
 
     private
 
-    def with_tracing(operation, statement = nil)
+    def with_tracing(operation, statement: nil, database: nil)
       segment = NewRelic::Agent::Tracer.start_datastore_segment(
         product: PRODUCT_NAME,
         operation: operation,
         host: _nr_hostname,
         port_path_or_id: _nr_port_path_or_id,
-        database_name: db
+        database_name: database
       )
       begin
         segment.notice_nosql_statement(statement) if statement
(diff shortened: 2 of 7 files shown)


lib/new_relic/agent/instrumentation/redis/instrumentation.rb
```diff
@@ -6,38 +6,58 @@ module NewRelic::Agent::Instrumentation
   module Redis
     PRODUCT_NAME = 'Redis'
     CONNECT = 'connect'
-    UNKNOWN = "unknown"
-    LOCALHOST = "localhost"
+    UNKNOWN = 'unknown'
+    LOCALHOST = 'localhost'
     MULTI_OPERATION = 'multi'
     PIPELINE_OPERATION = 'pipeline'
+    REDIS_5 = Gem::Version.new(::Redis::VERSION) >= Gem::Version.new('5.0.0') && !!defined?(::RedisClient)
```
fallwith, line 13: Given the `>=` nature of this indicator, I'd prefer a constant name that was more futureproof (for Redis 6) and specific in its description of what will happen if it is `true`. Possibly `USES_REDIS_CLIENT` or `HAS_REDIS_CLIENT` might work to signify that the separately namespaced redis client class.

lib/new_relic/agent/instrumentation/redis/instrumentation.rb
```diff
@@ -6,38 +6,58 @@ module NewRelic::Agent::Instrumentation
...
 
+    def connect_with_tracing
+      with_tracing(CONNECT, database: db) { yield }
```
fallwith, line 16: Does the monkeypatched class itself define `db`?

lib/new_relic/agent/instrumentation/redis/middleware.rb
```diff
@@ -0,0 +1,16 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic::Agent::Instrumentation
+  module RedisClient
+    module Middleware
+      # This module is used to instrument Redis 5.x+
```
fallwith, line 8: I like the `+` here. Could you please update the comments in `instrumentation.rb` and `prepend.rb` to use it as well for consistency?

test/multiverse/suites/redis/redis_instrumentation_test.rb
```diff
@@ -13,7 +13,7 @@ class NewRelic::Agent::Instrumentation::RedisInstrumentationTest < Minitest::Tes
     def after_setup
       super
       # Default timeout is 5 secs; a flushall takes longer on a busy box (i.e. CI)
-      @redis = Redis.new(:host => redis_host, :timeout => 25)
+      @redis ||= Redis.new(timeout: 25)
```
fallwith, line 16: Would you please explain the removals of the `redis_host` calls? Also, have ALL of the calls been removed, allowing us to remove the method definition itself now?

lib/new_relic/agent/instrumentation/redis/chain.rb
```diff
@@ -9,16 +9,28 @@ def self.instrument!
         ::Redis::Client.class_eval do
```
tannalynn, line 9: For prepend you're doing 2 different classes based on redis version, ::RedisClient  ::Redis::Client. How come that isn't necessary here? I would think that we'd be prepending/chaining on the same class always, so I'm curious about how this is working.

lib/new_relic/agent/instrumentation/redis/instrumentation.rb
```diff
@@ -6,38 +6,58 @@ module NewRelic::Agent::Instrumentation
...
+      operation = pipeline.flatten.include?('MULTI') ? MULTI_OPERATION : PIPELINE_OPERATION
+      statement = ::NewRelic::Agent::Datastores::Redis.format_pipeline_commands(pipeline)
+
+      # call_pipelined isn't invoked on the client object, so use client.db to
+      # access the client instance var on self
+      with_tracing(operation, statement: statement, database: client.db) { yield }
+    end
+
+    # Used for Redis 5.x
+    def call_v_with_tracing(command, &block)
```
tannalynn, line 46: It looks like the methods `call_v_with_tracing` and `call_with_tracing` are identical in everything other than method name. If i'm understanding correctly, maybe we don't need to have both methods. perhaps we could get rid of this method and just use call_with_tracing for both. so chain could be something like 
```ruby
if method_defined?(:call_v)
  alias_method(:call_v_without_new_relic, :call_v)

  def call_v(*args, &block)
    call_with_tracing(args[0]) { call_v_without_new_relic(*args, &block) }
  end
end

if method_defined?(:call)
  alias_method(:call_without_new_relic, :call)

  def call(*args, &block)
    call_with_tracing(args[0]) { call_without_new_relic(*args, &block) }
  end
end
```

test/multiverse/suites/redis/redis_instrumentation_test.rb
```diff
@@ -144,7 +144,7 @@ def test_does_not_record_statement_on_individual_command_node_by_default
       refute get_node[:statement]
     end
 
-    def test_records_metrics_for_set_in_web_transaction
+    def test_records_metrics_for_get_in_web_transaction
```
tannalynn, line 147: Wow, good eye, thats odd that these two test names seemed to be swapped based on what the actual test does!

lib/new_relic/agent/instrumentation/redis/chain.rb
kaylareopelle: Fantastic question! After taking a closer look, it seems like the Redis 5 instrumentation does not need to use the `redis-client` library (`::RedisClient`) as the entrypoint for chain/prepend after all! I thought it needed to be the source because I couldn't find a definition for `connect` in the `::Redis::Client` class, however, after looking at the caller locations for both chain and prepend using only `::Redis::Client`, it all still works! I'll go back to using ::Redis::Client for both since the change seems unnecessary.

Here's the relevant locations for chain if you're interested: 

```
 "~/dev/newrelic-ruby-agent/lib/new_relic/agent/instrumentation/redis/instrumentation.rb:16:in `connect_with_tracing'",
 "~/dev/newrelic-ruby-agent/lib/new_relic/agent/instrumentation/redis/chain.rb:39:in `connect'",
 "~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/redis-client-0.11.1/lib/redis_client.rb:647:in `raw_connection'",
 "~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/redis-client-0.11.1/lib/redis_client.rb:614:in `ensure_connected'",
 "~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/redis-client-0.11.1/lib/redis_client.rb:213:in `call_v'",
 "~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/redis-5.0.5/lib/redis/client.rb:73:in `call_v'",
 "~/dev/newrelic-ruby-agent/lib/new_relic/agent/instrumentation/redis/chain.rb:16:in `block in call_v'",
 "~/dev/newrelic-ruby-agent/lib/new_relic/agent/instrumentation/redis/instrumentation.rb:50:in `block in call_v_with_tracing'",
 "~/dev/newrelic-ruby-agent/lib/new_relic/agent/instrumentation/redis/instrumentation.rb:65:in `block in with_tracing'",
 "~/dev/newrelic-ruby-agent/lib/new_relic/agent/tracer.rb:355:in `capture_segment_error'",
 "~/dev/newrelic-ruby-agent/lib/new_relic/agent/instrumentation/redis/instrumentation.rb:65:in `with_tracing'",
 "~/dev/newrelic-ruby-agent/lib/new_relic/agent/instrumentation/redis/instrumentation.rb:50:in `call_v_with_tracing'",
 "~/dev/newrelic-ruby-agent/lib/new_relic/agent/instrumentation/redis/chain.rb:16:in `call_v'",
 "~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/redis-5.0.5/lib/redis.rb:167:in `block in send_command'",
 "~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/redis-5.0.5/lib/redis.rb:166:in `synchronize'",
 "~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/redis-5.0.5/lib/redis.rb:166:in `send_command'",
 "~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/redis-5.0.5/lib/redis/commands/strings.rb:191:in `get'",
 ```

Comment (kaylareopelle) [reply lib/new_relic/agent/instrumentation/redis/chain.rb:9]: Resolved in: 57320913dbd719ce65656e6db97dfe5f80d7a4c4

File: lib/new_relic/agent/instrumentation/redis/instrumentation.rb (hunk 1 above)
Comment (kaylareopelle) [reply lib/new_relic/agent/instrumentation/redis/instrumentation.rb]: Great point. Resolved in: a8198510ec8eaae23e51cb5271ef7eb250dd28a7

File: lib/new_relic/agent/instrumentation/redis/instrumentation.rb (hunk 2 above)
Comment (kaylareopelle) [reply lib/new_relic/agent/instrumentation/redis/instrumentation.rb:16]: Correct! It is available in all cases except the middleware, where `client.db` is defined. You can see an example of this call within #call_pipelined_with_tracing. If you'd prefer it be more specific, I could apply the `#_nr_client` method in all cases where `db` is called.

File: lib/new_relic/agent/instrumentation/redis/instrumentation.rb (hunk 6 above)
Comment (kaylareopelle) [reply lib/new_relic/agent/instrumentation/redis/instrumentation.rb]: You're absolutely right! They were better off different when the middleware was used because we needed to access the database and command differently, but using prepend, they're the same.

Fixed in: 5a66d239247527bc75ca7a5338fb46968387e29e

File: lib/new_relic/agent/instrumentation/redis/middleware.rb (hunk 3 above)
Comment (kaylareopelle) [reply lib/new_relic/agent/instrumentation/redis/middleware.rb:8]: Sure thing! Resolved in: d89e3b40145355970ba4c546df76cd12707e96c0

File: test/multiverse/suites/redis/redis_instrumentation_test.rb (hunk 4 above)
Comment (kaylareopelle) [reply test/multiverse/suites/redis/redis_instrumentation_test.rb:16]: While running the tests locally and across multiple versions with `redis_host`, both @hannahramadan and I ran into the following error: 

```
Redis::CannotConnectError: getaddrinfo: nodename nor servname provided, or not known
```

We had docker running Redis and this consistently occurred. Removing the `host: redis_host` argument resolved these problems and did not impact the CI, so we deemed this argument unnecessary. It has been removed from all locations, so the method definition can be removed now too. Thanks for noticing that! 

Resolved in: d49ef0fd39fc497c9bed9f0a9273c74e177e1d77

**Review by tannalynn — DISMISSED** [review]
🎉

File: test/multiverse/suites/redis/redis_instrumentation_test.rb (hunk 4 above)
Comment (kaylareopelle) [reply test/multiverse/suites/redis/redis_instrumentation_test.rb:16]: My bad, it looks like `redis_host` is still used so that metric names can be dynamically generated based on where the tests are running. 

Reverted in: 0d8f978da144f2f8ff5bde963b14117a5d18a5cb

**Review by tannalynn — APPROVED** [review]

File: test/multiverse/suites/redis/redis_instrumentation_test.rb [hunk 8]
```diff
@@ -336,8 +336,8 @@ def simulated_error_class
     end
 
     def simulate_read_error
-      redis = Redis.new(:host => redis_host)
-      redis.send(client).stubs("establish_connection").raises(simulated_error_class, "Error connecting to Redis")
```
kaylareopelle, line 340: The stubbed method was changed to `connect` because `establish_connection` does not exist in Redis 5.0

test/multiverse/suites/redis/redis_instrumentation_test.rb
fallwith: Cool. Currently we just do "are we in Docker?" very generically. I could see us needing to be more specific in future, such as "are we in Docker AND in an automated CI environment and not just a desktop based Docker system?"

fallwith (approved):

test/multiverse/suites/redis/redis_instrumentation_test.rb
kaylareopelle: So true! Good point. Do you think there's an appropriate action to take now to help future devs?

stevenou:
@kaylareopelle hi, I stumbled upon this PR because I am using `redis-client` with `newrelic_rpm` and am now encountering `undefined method 'db' for #<RedisClient>`

I am unclear on whether `redis-client` is even supported by `newrelic_rpm` or only `redis`...? I made the switch because `sidekiq` since version 7 now uses `redis-client` instead of `redis`. 

Wondering if you can provide any guidance and/or whether this is something that is on your radar to address?

kaylareopelle:
Hi @stevenou, thanks for bringing this to our attention. We're taking a look at the problem and hope to have more info for you soon.

angelatan2:
Hello @stevenou, thanks for sharing these findings so quickly.  You can follow our development in this ticket https://github.com/newrelic/newrelic-ruby-agent/issues/1639. I believe they are a similar issues.

PR #1613, opened 2022-11-10
Move Sidekiq classes into their own files

kaylareopelle:
In addition, the class names have been updated to reflect their locations


tannalynn (approved):
🎉

PR #1616, opened 2022-11-11
Sidekiq delayed extensions support TODO

fallwith:
Added TODO to remove instrumentation for Sidekiq delayed extensions now that they are no longer available in Sidekiq v7+


kaylareopelle (approved):
Schön gemacht!

Issue #794, opened 2021-09-23
Update Core Tech: Excon Gem


angelatan2:
This is a follow-up story from Ignite Sprint's goal.
- [x] Understand if methods were deprecated by gem owners.
None of the currently instrumented methods, [request_call](https://github.com/excon/excon/blob/master/lib/excon/connection.rb#L112), [response_call](https://github.com/excon/excon/blob/master/lib/excon/connection.rb#L206), or [error_call](https://github.com/excon/excon/blob/master/lib/excon/connection.rb#L106), have been deprecated. 
- [x]  Find new methods that are of value
There have been no new introduced methods of value—and hardly any new methods at that! Excon does version updates frequently, however most releases only include small refactors and certificate updates. They appear to follow semver versioning and are still on major version 0, pointing to no major updates to consider. [CHANGELOG](https://github.com/excon/excon/blob/master/changelog.txt)
- [x] Instrument those new methods
NA
- [x] Update our multiverse to dynamically pull the latest version so that we do not have to manually update each gem version number manually.
We are currently always pulling in the latest version and all tests are passing. We appear to be unnecessarily testing different versions, so a follow-up PR will be created to remove these and reduce testing overhead.

hannahramadan:
- [x] Understand if methods were deprecated by gem owners.
None of the currently instrumented methods, [request_call](https://github.com/excon/excon/blob/master/lib/excon/connection.rb#L112), [response_call](https://github.com/excon/excon/blob/master/lib/excon/connection.rb#L206), or [error_call](https://github.com/excon/excon/blob/master/lib/excon/connection.rb#L106), have been deprecated. 

- [x]  Find new methods that are of value
There have been no new introduced methods of value—and hardly any new methods at that! Excon does version updates frequently, however most releases only include small refactors and certificate updates. They appear to follow semver versioning and are still on major version 0, pointing to no major updates to consider. [CHANGELOG](https://github.com/excon/excon/blob/master/changelog.txt)

- [x] Instrument those new methods
NA

- [x] Update our multiverse to dynamically pull the latest version so that we do not have to manually update each gem version number manually.
We are currently always pulling in the latest version and all tests are passing. We appear to be uncessarily testing different versions, so a followup PR will be created to remove these and reduce testing overhead.

kaylareopelle:
@hannahramadan  - Are we all clear to close out this issue?

kaylareopelle:
Chatted with Hannah and we're good to close! 😎

PR #1620, opened 2022-11-11
Test minimum supported version

hannahramadan:
- Test the minimum supported version of Excon
- Add deprecation message
- Don't test JRuby on old versions of Excon

Related to #794


kaylareopelle (approved):
This looks great as-is, if anything were to change, since we just pulled out the supportability metrics for the other deprecations (like a few hours ago 😛), I would recommend removing line 49.

PR #1621, opened 2022-11-11
Create Ruby 2.3 logged deprecation message

kaylareopelle:
This change will log a message to our users if they're on Ruby 2.3, warning them about the eventual removal of support.

### DIFF
--- a/lib/new_relic/control/instrumentation.rb
+++ b/lib/new_relic/control/instrumentation.rb
@@ -94,8 +94,6 @@ module NewRelic
         :deprecated_ruby_version,
         deprecation_msg
       )
-
-      ::NewRelic::Agent.record_metric("Supportability/Deprecated/RubyVersion", 1)
     end
 
     include Instrumentation
--- a/lib/new_relic/control/instrumentation.rb
+++ b/lib/new_relic/control/instrumentation.rb
@@ -62,7 +62,7 @@ module NewRelic
           File.join(instrumentation_path, app.to_s, '*.rb')
         @instrumentation_files.each { |pattern| load_instrumentation_files(pattern) }
         DependencyDetection.detect!
-        ruby_22_deprecation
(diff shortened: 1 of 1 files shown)


lib/new_relic/control/instrumentation.rb
```diff
@@ -98,6 +99,21 @@ def ruby_22_deprecation
       ::NewRelic::Agent.record_metric("Supportability/Deprecated/Ruby22", 1)
     end
 
+    def ruby_23_deprecation
```
fallwith, line 102: I'd prefer to see a combined "support for Ruby versions below 2.4" message instead of these separate 2.2 and 2.3 ones.

As currently implemented, if I was on 2.2 and saw the 2.2 specific message, I might think upgrading to 2.3 would get me into the clear. But then I'd run into the 2.3 specific message as well.

fallwith (changes requested):

lib/new_relic/control/instrumentation.rb
kaylareopelle: Great point! Resolved in: 26be8d344d1bb3dc28fdfbc8400ad3eff628d9da

lib/new_relic/control/instrumentation.rb
```diff
@@ -95,7 +95,7 @@ def ruby_22_deprecation
         deprecation_msg
       )
 
-      ::NewRelic::Agent.record_metric("Supportability/Deprecated/Ruby22", 1)
+      ::NewRelic::Agent.record_metric("Supportability/Deprecated/RubyVersion", 1)
```
fallwith, line 98: What do you think of using something like `Ruby/#{RUBY_VERSION}` here?

fallwith (approved):

fallwith (approved):

PR #1623, opened 2022-11-12
Ruby 3.2.0-preview3 compatibiility

fallwith:
Updates for Ruby 3.2.0-preview3 compatibility.

NOTE: CI checks are blocked by https://github.com/ruby/setup-ruby/pull/397

### DIFF
--- a/lib/new_relic/agent/instrumentation/redis/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/redis/instrumentation.rb
@@ -2,20 +2,12 @@
 # See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
 # frozen_string_literal: true
 
+require_relative 'constants'
+
 module NewRelic::Agent::Instrumentation
   module Redis
-    PRODUCT_NAME = 'Redis'
-    CONNECT = 'connect'
-    UNKNOWN = 'unknown'
-    LOCALHOST = 'localhost'
-    MULTI_OPERATION = 'multi'
-    PIPELINE_OPERATION = 'pipeline'
-    HAS_REDIS_CLIENT = defined?(::Redis) &&
-      Gem::Version.new(::Redis::VERSION) >= Gem::Version.new('5.0.0') &&
-      !!defined?(::RedisClient)
-
     def connect_with_tracing
-      with_tracing(CONNECT, database: db) { yield }
+      with_tracing(Constants::CONNECT, database: db) { yield }
     end
 
     def call_with_tracing(command, &block)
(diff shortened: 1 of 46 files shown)


lib/new_relic/agent/instrumentation/redis/instrumentation.rb
```diff
@@ -10,7 +10,9 @@ module Redis
     LOCALHOST = 'localhost'
     MULTI_OPERATION = 'multi'
     PIPELINE_OPERATION = 'pipeline'
-    HAS_REDIS_CLIENT = Gem::Version.new(::Redis::VERSION) >= Gem::Version.new('5.0.0') && !!defined?(::RedisClient)
+    HAS_REDIS_CLIENT = defined?(::Redis) &&
+      Gem::Version.new(::Redis::VERSION) >= Gem::Version.new('5.0.0') &&
+      !!defined?(::RedisClient)
```
fallwith, line 15: This constant is loaded by a simple `require` of the class performed by `test/new_relic/agent/instrumentation/instrumentation_test.rb`, which runs without the `redis` gem. We need the `defined?(::Redis)` guard to prevent errors.

lib/new_relic/agent/instrumentation/sidekiq/extensions/delayed_class.rb
```diff
@@ -8,7 +8,7 @@
 #
 #       see https://github.com/mperham/sidekiq/issues/5076 for the discussion
 #       of the removal, which includes mentions of alternatives
-if Sidekiq::VERSION < '7.0.0'
+if defined?(Sidekiq::VERSION) && Sidekiq::VERSION < '7.0.0'
```
fallwith, line 11: This class is loaded by a simple require performed by `test/new_relic/agent/instrumentation/instrumentation_test.rb`, which operates without Sidekiq. As a result, we need to guard with an extra `defined?` check here.

test/multiverse/lib/multiverse/suite.rb
```diff
@@ -12,7 +12,6 @@
 require 'base64'
 require 'fileutils'
 require 'digest'
-require_relative '../multiverse'
```
fallwith, line 15: `../multiverse` itself requires this class. Removing this line fixes a circular dependency.

test/new_relic/load_test.rb
```diff
@@ -4,7 +4,7 @@
 
 require_relative '../test_helper'
 require 'resolv'
-require 'mocha/setup'
+require 'mocha/minitest'
```
fallwith, line 7: `require 'mocha'` and `require 'mocha/setup'` are deprecated. Users are encouraged to perform a more specific `require`, such as `mocha/minitest` in our case.

test/multiverse/lib/multiverse/suite.rb
```diff
@@ -110,12 +114,12 @@ def load_dependencies(gemfile_text, env_index, should_print = true)
 
     def with_potentially_mismatched_bundler
       yield
-    rescue Bundler::LockfileError => error
+    rescue ::Bundler::LockfileError => error
```
kaylareopelle, line 117: Does 3.2.0-preview3 require the `::` prefix? Or was this misc cleanup?

fallwith: After I addressed the circular dependency issue, multiverse ended up not knowing what Bundler was anymore. I added the :: to try to help it along. Then I realized I needed to require_relative the bundler_patch.rb file which delivers the needed "require 'bundler'" line. After I did that and it fixed everything, I felt the extra specificity of the colons could stay as well even though it wasn't really needed for anything.

kaylareopelle: Thanks for clarifying!

PR #1625, opened 2022-11-12
move Redis instrumentation constants to a class

fallwith:
the nightly CI system is reporting tens of thousands of 'already defined constant' warnings from the new Redis instrumentation changes.

moving the constants out into a class to help with this


kaylareopelle:
Thank you, James!

Issue #1639, opened 2022-11-16
NoMethod error raised from redis instrumentation when queueing ActiveJob sidekiq job


fcheung:
## Description

When enqueueing a job I get an error with this backtrace

```
NoMethodError
undefined method `db' for #<RedisClient redis://127.0.0.1:6379/0>

      with_tracing(operation, statement: statement, database: client.db) { yield }

newrelic_rpm (8.13.0) lib/new_relic/agent/instrumentation/redis/instrumentation.rb in call_pipelined_with_tracing at line 35
newrelic_rpm (8.13.0) lib/new_relic/agent/instrumentation/redis/middleware.rb in call_pipelined at line 12
redis-client (0.11.1) lib/redis_client.rb in connect at line 677
redis-client (0.11.1) lib/redis_client.rb in raw_connection at line 647
redis-client (0.11.1) lib/redis_client.rb in ensure_connected at line 614
redis-client (0.11.1) lib/redis_client.rb in pipelined at line 349
redis-client (0.11.1) lib/redis_client/decorator.rb in pipelined at line 51
sidekiq (7.0.1) lib/sidekiq/client.rb in block in raw_push at line 214
connection_pool (2.3.0) lib/connection_pool.rb in block (2 levels) in with at line 65
connection_pool (2.3.0) lib/connection_pool.rb in handle_interrupt at line 64
connection_pool (2.3.0) lib/connection_pool.rb in block in with at line 64
connection_pool (2.3.0) lib/connection_pool.rb in handle_interrupt at line 61
connection_pool (2.3.0) lib/connection_pool.rb in with at line 61
sidekiq (7.0.1) lib/sidekiq/client.rb in raw_push at line 211
sidekiq (7.0.1) lib/sidekiq/client.rb in push at line 92
sidekiq (7.0.1) lib/sidekiq/client.rb in push at line 160
activejob (7.0.4) lib/active_job/queue_adapters/sidekiq_adapter.rb in enqueue at line 22
activejob (7.0.4) lib/active_job/enqueuing.rb in block in enqueue at line 67
activesupport (7.0.4) lib/active_support/callbacks.rb in block in run_callbacks at line 118
newrelic_rpm (8.13.0) lib/new_relic/agent/instrumentation/active_job.rb in block in run_in_trace at line 63
newrelic_rpm (8.13.0) lib/new_relic/agent/method_tracer.rb in block in trace_execution_scoped at line 74
[...]

PR #1640, opened 2022-11-16
Call client.config instead of client

kaylareopelle:
# Overview

In some cases, the RedisClient object cannot directly access methods like db, port, or path. These methods are always available on the client.config object.

Resolves: #1639 

Submitter Checklist:
-  [X] Include a link to the related GitHub issue, if applicable
- [ ] ~~Include a security review link, if applicable~~

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


fallwith (approved):

sidekiq_with_redis_test.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    assert_equal value, result
+  end
+
+  private
+
+  def run_these_tests?
+    # these tests aren't impacted by Ruby version
+    return false unless RUBY_VERSION >= '3.1.2'
```
tannalynn, line 67: Im confused by the comment above this. How come we're checking the ruby version if the tests aren't impacted by the ruby version?

sidekiq_with_redis_test.rb
```diff
@@ -0,0 +1,72 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'bundler/inline'
+
```
tannalynn, line 6: Is this test in the root directory of the agent instead of in test/?

sidekiq_with_redis_test.rb
fallwith: Without this check, every single Ruby being tested by the CI will perform the test. Given that we're doing an inline bundling of `redis` and that the test won't manifest itself differently with a different Ruby version, this check exists to ensure that we perform as few runs of this test as possible.

I'm very open to changing or removing this check, and/or the comment above it. Give it some thought and let me know what you'd like this bit of the test to look like.

sidekiq_with_redis_test.rb
fallwith: Odd. Thanks for catching that, will address!

fallwith: Moved with [eb8f2ef](https://github.com/newrelic/newrelic-ruby-agent/pull/1640/commits/eb8f2efb850c7a30935945e6b8cedd71768d6e4a), thanks!

sidekiq_with_redis_test.rb
tannalynn: ahh I see, so it's not really about the ruby version so much as it's just about running it just once on the CI to save time. And yeah, I would appreciate it if the comment was a little more specific about why it's doing this, since this isn't something we typically do. thank you for explaining!

fallwith: I gave it a rework with [516dd6f](https://github.com/newrelic/newrelic-ruby-agent/pull/1640/commits/516dd6f4603da7609cc2b476f95b9892ce92eac4)

tannalynn (approved):
Thanks for getting more detailed about when this runs!

tannalynn (approved):

fallwith:
reminder: this bugfix branch has been mentioned publicly and should not be deleted until after v8.13.1 is released

PR #1649, opened 2022-11-22
stop using CGI

fallwith:
rather than upgrade to `cgi` v0.3.5 to mitigate
[CVE-2021-33621](https://www.ruby-lang.org/en/news/2022/11/22/http-response-splitting-in-cgi-cve-2021-33621/), let's just stop using CGI


kaylareopelle (approved):
Thanks, James! I like this solution.

PR #1652, opened 2022-11-28
Test darwin hw.logicalcpu > 0

### LINKED ISSUE #1431: CI: Add unit test to verify that hw.logicalcpu is > 0
The `NewRelic::Agent::SystemInfo.processor_info_darwin ` method performs `sysctl_value('hw.logicalcpu')`. According to this TODO in the tests, we don't have a unit test to verify that the response from that call is a positive value. We should add such a test and remove the TODO comment.

```ruby
test/new_relic/agent/system_info_test.rb
   179:	 # TODO: in processor_info_darwin, sysctl_value('hw.logicalcpu') > 0 is untested
```

hannahramadan:
There are 3 ways to set the value of `num_logical_processors`: 1 primary and 2 fallback options. Previously, testing covered the second fallback option, but not the first. This commit adds a test for the first fallback value.

Closes #1431

### DIFF
--- a/test/new_relic/agent/system_info_test.rb
+++ b/test/new_relic/agent/system_info_test.rb
@@ -179,7 +179,6 @@ class NewRelic::Agent::SystemInfoTest < Minitest::Test
     end
   end
 
-  # TODO: in processor_info_darwin, sysctl_value('hw.logicalcpu') > 0 is untested
   def test_processor_info_darwin_fallback
     counts = {'hw.packages' => 2,
               'hw.physicalcpu_max' => 0,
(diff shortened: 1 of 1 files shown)


fallwith:
The class is at 84% line coverage, but the method in question now has 100% line and branch coverage.

Screen Shot 2022-11-28 at 2 16 25 PM

fallwith (approved):

kaylareopelle (approved):
I think this looks great, Hannah! 

If you wanted to take it a step further, there is some duplication between the existing test and the new test that could be separated into constants or a method.

PR #1659, opened 2022-11-29
instrument Rails custom events

fallwith:
add instrumentation - just basic stopwatching for now - for custom events that a monitored application has defined and configured notifications for

ideally in future, this support would provide a viable vendor agnostic approach for manually instrumenting anything that the agent doesn't automatically detect and instrument on its own


test/new_relic/agent/instrumentation/rails/custom_events_subscriber_test.rb
```diff
@@ -0,0 +1,95 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require_relative '../../../../test_helper'
+require 'new_relic/agent/instrumentation/custom_events_subscriber'
+
+module NewRelic::Agent::Instrumentation
+  class CustomEventsSubscriberTest < Minitest::Test
+    TOPIC = 'abre los ojos'
```
kaylareopelle, line 10: Does ActiveSupport allow topics with whitespace?

kaylareopelle (commented):
This is very exciting stuff. Would you add a CHANGELOG entry?

fallwith:
> This is very exciting stuff. Would you add a CHANGELOG entry?

Yep, given that we now reference the PR in the CHANGELOG, I needed to have it created first. Here's a follow-up commit with a CHANGELOG entry: [2eccfba](https://github.com/newrelic/newrelic-ruby-agent/pull/1659/commits/2eccfba810537554f54a940b84198f4cdd463d79)

test/new_relic/agent/instrumentation/rails/custom_events_subscriber_test.rb
fallwith: Yep!

```ruby
#!/usr/bin/env ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'rails', require: false
end

require 'active_support/notifications'
require 'active_support/isolated_execution_state'

EVENT_NAME = 'abre los ojos'

ActiveSupport::Notifications.subscribe EVENT_NAME do |name, started, finished, unique_id, data|
  puts "name: >>#{name}<<"
  puts "started: >>#{started}<<"
  puts "finished: >>#{finished}<<"
  puts "unique_id: >>#{unique_id}<<"
  puts "data: >>#{data.inspect}<<"
end

ActiveSupport::Notifications.instrument EVENT_NAME, this: :data do
  rand(1148)
end
```

CHANGELOG.md
```diff
@@ -2,12 +2,16 @@
...
 
+  * **Support for Rails ActiveSupport::Notifications for custom events**
+
+    When the new `custom_events_instrumentation_topics` configuration parameter is be set equal to a comma delimited list of custom event topics to subscribe to, the agent will now subscribe to each of the topics specified and report instrumentation for the events when they take place. [Creating custom events](https://guides.rubyonrails.org/active_support_instrumentation.html#creating-custom-events) is simple and now reporting instrumentation for them to New Relic is simple as well. [PR#1659](https://github.com/newrelic/newrelic-ruby-agent/pull/1659)
```
kaylareopelle, line 13:
```suggestion
    When the new `custom_events_instrumentation_topics` configuration parameter is set equal to a comma delimited list of custom event topics to subscribe to, the agent will now subscribe to each of the topics specified and report instrumentation for the events when they take place. [Creating custom events](https://guides.rubyonrails.org/active_support_instrumentation.html#creating-custom-events) is simple and now reporting instrumentation for them to New Relic is simple as well. [PR#1659](https://github.com/newrelic/newrelic-ruby-agent/pull/1659)
```

lib/new_relic/agent/instrumentation/custom_events_subscriber.rb
```diff
@@ -0,0 +1,37 @@
...
+      finishable = pop_segment(id)
+      finishable.finish if finishable
+    rescue => e
+      log_notification_error(e, name, 'finish')
+    end
+
+    private
+
+    def transaction_name(name)
+      "Controller/ActiveSupport/CustomEvents/#{name}"
```
kaylareopelle, line 34: It's such a bummer we have to start these with `Controller/`. Does fixing that seem like a 9.0-worthy change?

test/new_relic/agent/instrumentation/rails/custom_events_subscriber_test.rb
kaylareopelle: Awesome, thank you!

kaylareopelle (dismissed):

lib/new_relic/agent/instrumentation/custom_events_subscriber.rb
fallwith: We should definitely gather some feedback on naming conventions on the whole to better serve the (current) Ruby community, and then gate anything significant with a major version bump.

For this feature here, it's following the `Controller/` pattern used by all other notification instrumentation. I suppose we could use `Custom/` instead. Do you want to have us think more on naming for this feature specifically or hold off for the bigger version-bump-requiring conversation?

kaylareopelle: I thought `Controller/` was a magic prefix that we needed to use in order for the action to be permitted as the root node of a transaction? I'll poke around the codebase to check that assertion.

If `Controller/*` isn't a magic required prefix, I think it'd be worth discussing what we want to name the transactions since we're creating a new type of transaction. I'd prefer to get it right before release rather than change it on customers in a major version.

test/helpers/misc.rb
```diff
@@ -118,3 +118,9 @@ def wait_until_not_nil(give_up_after = 3, &block)
     current_tries += 1
   end
 end
+
+def skip_unless_minitest5_or_above
```
hannahramadan, line 122: The method name has me thinking Minitest5 wouldn't be skipped, but with `>` version 5, would Minitest 5 also be skipped?

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -422,6 +422,17 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'Path to <b>newrelic.yml</b>. If undefined, the agent checks the following directories (in order): <b>config/newrelic.yml</b>, <b>newrelic.yml</b>, <b>$HOME/.newrelic/newrelic.yml</b> and <b>$HOME/newrelic.yml</b>.'
         },
+        :'custom_events_instrumentation_topics' => {
+          :default => [],
+          :public => true,
+          :type => Array,
+          :allowed_from_server => true,
```
tannalynn, line 429: This should probably be set to false.
I don't think the server would be sending anything for it anyways since that isn't something the server knows about. 
Plus, from what I understand, this config is only used during startup when we are installing instrumentation. Configs that are only used right at startup can't be allowed_from_server because the agent would have already finished setup by the time the server response is received.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -422,6 +422,17 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :dynamic_name => true,
```
tannalynn, line 430: I don't think having this marked as a dynamic name is necessary since you are calling this config directly in `lib/new_relic/agent/instrumentation/rails_notifications/custom_events.rb:29`. It should be safe to just remove this line.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -422,6 +422,17 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :description => <<-DESCRIPTION
+A comma separated list of ActiveSupport custom event topics to subscribe to and
```
tannalynn, line 432: I think the wording here could be changed since the type is set to Array. In yml files you can specify an array in a non comma separated list way. Like
```
custom_events_instrumentation_topics:
  - first_thing
  - second_thing
```
We could make it less specific and just say something like `an array of activesupport customer event topics`?

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1187,6 +1198,14 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => deprecated_description(:'instrumentation.curb', 'If `true`, disables instrumentation for the curb gem.')
         },
+        :disable_custom_events_instrumentation => {
+          :default => false,
+          :public => true,
+          :type => Boolean,
+          :dynamic_name => true,
```
tannalynn, line 1205: Same situation as the other config with dynamic_name, we should be able to remove this line since you're calling the config for this directly in the custom_events file

lib/new_relic/agent/instrumentation/custom_events_subscriber.rb
tannalynn: Hmmm. I'm not 100% sure. The magic prefix thing DOES seem familiar to me. The agent might need it, or maybe the UI needs it? Or both or neither, I'm not sure. Maybe we could test out changing it to something else and seeing how it ends up looking in the UI?

test/helpers/misc.rb
tannalynn: +1

lib/new_relic/agent/instrumentation/custom_events_subscriber.rb
hannahramadan: Assuming we aren't tied to `Controller/`, I do like the `Custom/`prefix better. I think it gives a nice indicator of what a user is looking at/ that someone on their team found a particular metric important to capture, and leads down a clearer path that those metrics are customizable.

test/helpers/misc.rb
fallwith: Ah, I can see that. We're doing an alphabetical sort comparison here (as we do elsewhere in the agent).

The constant will hold a value like `"5.16.3"` or even `"5.0.0"` and when we do a string comparison with `>` to `"5"`, all v5 versions will satisfy the check.

I'll give some thought to a helper method in future that makes our agent's use of string comparisons for something that seems numeric to be consistent and clearer.

tannalynn: Ooh right, I forgot string comparison feels unintuitive sometimes lol. Thanks for explaining!

hannahramadan: Ooo right right you've mentioned this before. Thank you!

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: Agreed! Good point!

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -422,6 +422,17 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
         },
+        :'custom_events_instrumentation_topics' => {
+          :default => [],
+          :public => true,
+          :type => Array,
+          :allowed_from_server => true,
+          :dynamic_name => true,
+          :description => <<-DESCRIPTION
+A comma separated list of ActiveSupport custom event topics to subscribe to and
+provide instrumentation for. Ex: "my.custom.event,another.event,a.third.event"
```
kaylareopelle, line 433: We usually write "For example," in the descriptions if you'd like to make this entry consistent.

lib/new_relic/agent/instrumentation/custom_events_subscriber.rb
kaylareopelle: I like the idea of testing it in the UI. I'm hesitant about using the `Custom/` prefix because that is used by customers creating custom metrics: 
https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent.rb#L183 
https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent.rb#L209

I think my preference would be to start it with `ActiveSupport/`.

kaylareopelle: Re: `Controller/*`, I think this came up while we were adding gRPC instrumentation: https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/grpc/server/instrumentation.rb#L115 
I can't tell from the commit why it's named this way, and the agent spec doesn't have this requirement.

kaylareopelle: It could be due to this method: https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/rules_engine.rb#L26-L41

fallwith: It looks like the existing Rails notifications transactions don't all begin with `Controller/`  (especially given that we have a view one that correctly starts with `View/`) so there's no need to hold on to `Controller/` for grouping all of a customer's notifications transactions together as I had originally thought.

I like Kayla's take on disambiguating a Rails custom event from our own agent's custom events and not using `Custom/` which is reserved for the latter.

I'll put together Kayla's recommended change:

```diff
- "Controller/ActiveSupport/CustomEvents/#{name}"
+ "ActiveSupport/CustomEvents/#{name}"
```

and get a UI test going to see what we think.

kaylareopelle: Thank you, James!

lib/new_relic/agent/configuration/default_source.rb
fallwith: @tannalynn do I then also need to include this disclaimer?

```yaml
  <Callout variant="caution">
    This option can't be set via environment variable.
  </Callout>
```

tannalynn: hmm i thought that was only on our hash configs, but it looks like it is on the ignore/expected classes configs even though thats array. i don't think it says that on any other arrays? I looked at our env source class and it looks like it where does type stuff it doesn't have anything for arrays https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/configuration/environment_source.rb#L82
So I guess I'm not sure! I always thought arrays could be passed in as a string on the env source and then parsed, but maybe that doesn't work at all.

fallwith: Gotcha. I've got to integration test the transaction name change anyhow, so I'll do some env var driven testing around this too. Thank you!

lib/new_relic/agent/instrumentation/custom_events_subscriber.rb
fallwith: The recommended change looks great to me in the UI, with the transactions appearing as non-web when recorded on their own and not segments belonging to a web transaction.

Screen Shot 2022-11-30 at 9 41 44 PM

lib/new_relic/agent/configuration/default_source.rb
fallwith: Updated with https://github.com/newrelic/newrelic-ruby-agent/pull/1659/commits/951476b03bd2beafa3f9804a04b79b0b8a88ba6f

lib/new_relic/agent/configuration/default_source.rb
fallwith: Updated with https://github.com/newrelic/newrelic-ruby-agent/pull/1659/commits/951476b03bd2beafa3f9804a04b79b0b8a88ba6f

lib/new_relic/agent/configuration/default_source.rb
fallwith: Updated with https://github.com/newrelic/newrelic-ruby-agent/pull/1659/commits/951476b03bd2beafa3f9804a04b79b0b8a88ba6f

lib/new_relic/agent/configuration/default_source.rb
fallwith: Updated with https://github.com/newrelic/newrelic-ruby-agent/pull/1659/commits/951476b03bd2beafa3f9804a04b79b0b8a88ba6f

lib/new_relic/agent/configuration/default_source.rb
fallwith: Updated with https://github.com/newrelic/newrelic-ruby-agent/pull/1659/commits/951476b03bd2beafa3f9804a04b79b0b8a88ba6f

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -358,6 +358,19 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :description => 'Sets the level of detail of log messages. Possible log levels, in increasing verbosity, are: `error`, `warn`, `info` or `debug`.'
         },
         # General
+        :active_support_custom_events_topics => {
```
kaylareopelle, line 361: During our conversation in Slack yesterday, we talked about changing this to `active_support_custom_events_instrumentation_topics`. After taking another look at the Rails guide, I would like to propose changing "topics" to "name". 

I couldn't find a reference to the word "topics" in the documentation. Please let me know if I'm mistaken. It seems like the word "name" is used. If we use the same word, it'll be more searchable.

`active_support_custom_events_names`

lib/new_relic/agent/instrumentation/active_support.rb
```diff
@@ -0,0 +1,12 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+# This is a helper file that will allow apps using ActiveSupport without Rails
+# to still leverage all ActiveSupport based instrumentation functionality
+# offered by the agent that would otherwise be gated by the detection of Rails.
+
+# ActiveSupport custom events notifications
```
kaylareopelle, line 9:
```suggestion
# ActiveSupport notifications custom events
```

kaylareopelle (dismissed):

kaylareopelle (approved):

lib/new_relic/agent/configuration/default_source.rb
fallwith: I concur. I was going after a generic pub/sub architecture with "topic" but "name" is definitely all I see in the guides. Good call. Addressed with [25a2b3b](https://github.com/newrelic/newrelic-ruby-agent/pull/1659/commits/25a2b3b04c041d371b93e721dd9e96a5c38bf525)

fallwith:
Will address the (unrelated) JRuby CI issues separately.

PR #1666, opened 2022-11-30
Add workflow_dispatch to CI workflow for PRs

kaylareopelle:
# Overview
Sometimes only a few of our actions get run on the PR and we need to make a new commit to un-stick it. This will allow us to get confirmation about the changes in those cases.


fallwith:
Closing in favor of #1667 (which adds only the desired change without any additional autoformatting taking place)

PR #1673, opened 2022-12-02
Redis instrumentation: support older redis-client

### LINKED ISSUE #1650: Upgrading from 8.12 to 8.13.1 produced 500 error due to Redis monitoring throwing NameError
^^ Provide a general summary of the issue in the title above. ^^ 

After upgrading from version 8.12 to version 8.13.1 of newrelic_rpm we began getting a 500 error due to the Redis monitoring component of the New Relic agent not finding "client".

## Description
newrelic_rpm breaking Rails app on Heroku using Redis because of NameError

This shows up in the Rails Console when trying to access the Rails cache in an app using the Redis cache. 

`NameError (undefined local variable or method `client' for RedisClient::Middlewares:Module`

## Expected Behavior
It is expected that it will working normally and not break the app.

## Steps to Reproduce
Using newrelic_rpm 8.13.1, open a Rails console where the New Relic agent is enabled and the 8.13.1 version of the newrelic_rpm is being used.
In the Rails console, try to access the keys in the Rails cache with Rails.cache.redis.keys
What should happen is that a listing of keys returns.  What happens instead is that there is a NameError.
This can be fixed by disabling the agent or reverting to the 8.12 versions.

## Your Environment
Ruby 3.12, newrelic_rpm 8.13.1

## Additional context
N/A

### LINKED ISSUE #1671: Redis gem instrumentation: add better error handling around v5+ specific instrumentation
v8.13.0 of the agent introduced support for v5 of the `redis` gem, and v8.13.1 followed to address a Sidekiq v7 specific use case.

Now with v8.13.1, we have had reports of additional issues that the agent has when attempting to get at the lowest underlying Redis client object within the `redis` v5 instrumentation.

While we continue to research these reports and attempt to produce reproductions, in parallel or beforehand we should take steps to harden the impacted `redis` v5 instrumentation to prevent it from crashing. From the customer's perspective, any issue encountered with the Redis v5 instrumentation should result in either parial or no Redis instrumentation with the rest of the agent's functionality continuing to operate normally. Effectively, the error handling should automatically provide at least as good of an experience as rolling back to an agent version that does not have the v5 instrumentation or disabling the instrumentation support in configuration.

fallwith:
Redis instrumentation:

* update `call_pipelined_with_tracing` to yield early (instead of crashing) if a db value cannot be obtained
* rename `_nr_client` to `_nr_redis_client_config` to make it more clear that it holds a config object and not a client object, even though config getters can be called directly on the `redis` gem's client object
* update `_nr_redis_client_config` to add support for `redis-client` gem versions older than v0.11
* update `_nr_redis_client_config` to raise upon failing to get at the underlying config object
* minor exception handling updates to include class names
* update the unit tests accordingly

resolves #1650 
resolves #1671

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,18 +12,18 @@
 
     When the new `active_support_custom_events_names` configuration parameter is set equal to an array of custom event names to subscribe to, the agent will now subscribe to each of the names specified and report instrumentation for the events when they take place. [Creating custom events](https://guides.rubyonrails.org/active_support_instrumentation.html#creating-custom-events) is simple and now reporting instrumentation for them to New Relic is simple as well. [PR#1659](https://github.com/newrelic/newrelic-ruby-agent/pull/1659)
 
-  * **Bugfix: Support older versions of the redis-client gem, handle unknown Redis database index
+  * **Bugfix: Support older versions of the RedisClient gem, handle unknown Redis database index
 
-    With version 8.13.0 of the agent, support was added for `redis` gem v5+ and the new `redis-client` gem. With versions of `redis-client` older than v0.11, the agent could cause the monitored application to crash when attempting to determine the Redis database index. Version 8.14.0 adds two related improvements. Firstly, support for `redis-client` versions older than v0.11 has been added to get at the database index value. Secondly, the agent will no longer crash or impact the monitored application in the event that the database index cannot be obtained. [Issue#1650](https://github.com/newrelic/newrelic-ruby-agent/issues/1650) [PR#1673](https://github.com/newrelic/newrelic-ruby-agent/pull/1673)
+    With version 8.13.0 of the agent, support was added for the [redis-rb](https://github.com/redis/redis-rb) gem v5+ and the new [RedisClient](https://rubygems.org/gems/redis-client) gem. With versions of RedisClient older than v0.11, the agent could cause the monitored application to crash when attempting to determine the Redis database index. Version 8.14.0 adds two related improvements. Firstly, support for RedisClient versions older than v0.11 has been added to get at the database index value. Secondly, the agent will no longer crash or impact the monitored application in the event that the database index cannot be obtained. Thank you very much to our community members [@mbsmartee](https://github.com/mbsmartee) and [@patatepartie](https://github.com/patatepartie) for bringing this issue to our attention, for helping us determine how to best reproduce it, and for testing out the update. We appreciate your help! [Issue#1650](https://github.com/newrelic/newrelic-ruby-agent/issues/1650) [PR#1673](https://github.com/newrelic/newrelic-ruby-agent/pull/1673)
 
 
   ## v8.13.1
 
   Version 8.13.1 of the agent provides a bugfix for Redis v5.0 instrumentation.
 
-  * **Fix NoMethodError when using Sidekiq v7.0 with Redis Client v0.11**
+  * **Fix NoMethodError when using Sidekiq v7.0 with RedisClient v0.11**
 
-    In some cases, the `RedisClient` object cannot directly access methods like db, port, or path. These methods are always available on the `client.config` object. This raised a `NoMethodError` in environments that used Sidekiq v7.0 and [Redis Client](https://rubygems.org/gems/redis-client) v0.11. Thank you to [fcheung](https://github.com/fcheung) and [@stevenou](https://github.com/stevenou) for bringing this to our attention! [Issue#1639](https://github.com/newrelic/newrelic-ruby-agent/issues/1639)
+    In some cases, the `RedisClient` object cannot directly access methods like db, port, or path. These methods are always available on the `client.config` object. This raised a `NoMethodError` in environments that used Sidekiq v7.0 and [RedisClient](https://rubygems.org/gems/redis-client) v0.11. Thank you to [fcheung](https://github.com/fcheung) and [@stevenou](https://github.com/stevenou) for bringing this to our attention! [Issue#1639](https://github.com/newrelic/newrelic-ruby-agent/issues/1639)
 
 
   ## v8.13.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
   ## v8.14.0
 
-  Version 8.14.0 of the agent restores desired Capistrano-based changelog lookup functionalty when a deployment is performed and delivers support for instrumenting Rails custom event notifications.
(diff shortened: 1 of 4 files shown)


fallwith:
One way to reproduce the issue:

1. Create a new Rails 7 app. NOTE: do not use `--minimal` as it does not prep ActionCable to use Redis for development.
2. Optionally confirm that `config/cable.yml` is referencing Redis for `development`.
3. Add `newrelic_rpm` v8.13.0 or v8.13.1 and `redis-client` < v0.10 in `Gemfile` and bundle.
4. Start a redis server.
5. Run `bin/rails console` and perform the following in the console:

```
require 'action_cable/subscription_adapter/redis'
server = ActionCable.server
keys = server.config.cable.symbolize_keys 
conn = ActionCable::SubscriptionAdapter::Redis.redis_connector.call(keys)
listener = ActionCable::SubscriptionAdapter::Redis::Listener.new(:redis, nil)
listener.listen(conn)
```

A reproduce will produce a crash with a stack trace. A success after applying this PR will produce a constantly polling listening process without any errors (hit ctrl-c to stop it from listening).

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
     When the new `active_support_custom_events_names` configuration parameter is set equal to an array of custom event names to subscribe to, the agent will now subscribe to each of the names specified and report instrumentation for the events when they take place. [Creating custom events](https://guides.rubyonrails.org/active_support_instrumentation.html#creating-custom-events) is simple and now reporting instrumentation for them to New Relic is simple as well. [PR#1659](https://github.com/newrelic/newrelic-ruby-agent/pull/1659)
 
+  * **Bugfix: Support older versions of the redis-client gem, handle unknown Redis database index
```
kaylareopelle, line 15: I think we generally capitalize the gem name, but I'd be open to changing this. In 8.13.1 `redis-client` was Redis Client. 

```suggestion
  * **Bugfix: Support older versions of the Redis Client gem, handle unknown Redis database index
```

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
...
+
+    With version 8.13.0 of the agent, support was added for `redis` gem v5+ and the new `redis-client` gem. With versions of `redis-client` older than v0.11, the agent could cause the monitored application to crash when attempting to determine the Redis database index. Version 8.14.0 adds two related improvements. Firstly, support for `redis-client` versions older than v0.11 has been added to get at the database index value. Secondly, the agent will no longer crash or impact the monitored application in the event that the database index cannot be obtained. [Issue#1650](https://github.com/newrelic/newrelic-ruby-agent/issues/1650) [PR#1673](https://github.com/newrelic/newrelic-ruby-agent/pull/1673)
```
kaylareopelle, line 17:
```suggestion
    With version 8.13.0 of the agent, support was added for the `redis` gem v5+ and the new `redis-client` gem. With versions of `redis-client` older than v0.11, the agent could cause the monitored application to crash when attempting to determine the Redis database index. Version 8.14.0 adds two related improvements. Firstly, support for `redis-client` versions older than v0.11 has been added to get at the database index value. Secondly, the agent will no longer crash or impact the monitored application in the event that the database index cannot be obtained. [Issue#1650](https://github.com/newrelic/newrelic-ruby-agent/issues/1650) [PR#1673](https://github.com/newrelic/newrelic-ruby-agent/pull/1673)
```

kaylareopelle: Should we mention the folks who chimed in on the issue?

kaylareopelle (dismissed):
Thanks for the `nil` value adjustment! The comments are all suggestions/questions that don't need to be addressed. Feel free to disregard and merge.

lib/new_relic/agent/instrumentation/redis/instrumentation.rb
```diff
@@ -52,21 +59,35 @@ def with_tracing(operation, statement: nil, database: nil)
...
     rescue => e
-      NewRelic::Agent.logger.debug("Failed to retrieve Redis port_path_or_id: #{e}")
+      NewRelic::Agent.logger.debug("Failed to retrieve Redis port_path_or_id: #{e.class} - #{e.message}")
       Constants::UNKNOWN
     end
 
-    def _nr_client
-      @nr_client ||= self.is_a?(::Redis::Client) ? self : client.config
+    def _nr_redis_client_config
+      @nr_config ||= begin
```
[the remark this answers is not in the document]
fallwith: The `begin` and `end` here create a block of code for the memoization with `@nr_config`. That instance variable gets set equal to the result of the entire block.

Alternatively, we could use 2 separate methods, like so to call a method instead of an anonymous block.

```ruby
def important_value
  @important_value ||= initialize_important_value
end

def initialize_important_value
  # multiple line logic goes here
end
```

CHANGELOG.md
fallwith: In this context we have fully 5 names to keep track of:

- "Redis", the name of the datastore software itself
- "redis-rb", the project name for the "redis" gem
- "redis", a long running Ruby gem for Redis
- "RedisClient", the project name for the "redis-client" gem
- "redis-client", a relatively new Ruby gem for Redis

My intention was to disambiguate the gem names from the "Redis" datastore software name.

Frustratingly, we seem to have never referred the "redis-rb" project correctly as either "redis-rb" or "redis", opting instead for the confusing "Redis". As for "RedisClient", we've tossed a space into that one and gone with "Redis Client" for v8.13.1.

I'll go ahead and use "redis-rb" and "RedisClient" for v8.13.1 and v8.14.0 and leave the older changelog entries alone. I'm okay with using project names instead of gem names, but not capitalizing them unless the project namer has done so.

fallwith: Updated with [772de0a](https://github.com/newrelic/newrelic-ruby-agent/pull/1673/commits/772de0ad48a8e280953684ff81dba2ff1c344ea0)

CHANGELOG.md
fallwith: Updated with [772de0a](https://github.com/newrelic/newrelic-ruby-agent/pull/1673/commits/772de0ad48a8e280953684ff81dba2ff1c344ea0) - great call!

hannahramadan (approved):
makes sense!

lib/new_relic/agent/instrumentation/redis/instrumentation.rb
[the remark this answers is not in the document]
kaylareopelle: Thanks for explaining!

kaylareopelle (approved):

PR #1688, opened 2022-12-08
CI: fix redis-rb instrumentation test

fallwith:
A redis-rb instrumentation update for v8.14.0 was initially going to perform an early yield but was later changed to not have to. The corresponding test needs to be updated to no longer refer to a now missing yield related variable.


kaylareopelle (approved):
Thank you! Database tests look like they're passing now.

PR #1691, opened 2022-12-09
Add test for vendors

### LINKED ISSUE #1434: CI: verify the vendors in the trace context cross agent tests for distributed tracing
The trace context cross agent tests for distributed tracing have the following TODO:

```ruby
test/new_relic/agent/distributed_tracing/trace_context_cross_agent_test.rb
   255:	 # TODO: check vendors in test_case_attributes
```

that TODO should be satisfied or removed as not providing any value.

hannahramadan:
Add test to ensure expected and actual `vendors` match. Resolve TODO.

closes #1434

### DIFF
--- a/test/new_relic/agent/distributed_tracing/trace_context_cross_agent_test.rb
+++ b/test/new_relic/agent/distributed_tracing/trace_context_cross_agent_test.rb
@@ -253,7 +253,11 @@ module NewRelic
               %Q(Unexpected #{event_type} attribute "#{key}")
           end
 
-          # TODO: check vendors in test_case_attributes
+          (test_case_attributes['vendors'] || []).each do |k, v|
+            assert_equal v,
+              actual_attributes[k.to_s],
+              %Q(Wrong "#{k}" #{event_type} attribute; expected #{v.inspect}, was #{actual_attributes[k.to_s].inspect})
+          end
         end
 
         def verify_transaction_intrinsics(test_case)


kaylareopelle (approved):
Kudos for addressing this old TODO! Thanks @tannalynn  for your help!

test/new_relic/agent/distributed_tracing/trace_context_cross_agent_test.rb
```diff
@@ -253,7 +253,11 @@ def verify_attributes(test_case_attributes, actual_attributes, event_type)
               %Q(Unexpected #{event_type} attribute "#{key}")
           end
 
-          # TODO: check vendors in test_case_attributes
+          (test_case_attributes['vendors'] || []).each do |k, v|
```
fallwith, line 256: Would you please explain the need for `|| []`? I want to make sure the TODO is fully satisfied by there always being vendor checks performed.

test/new_relic/agent/distributed_tracing/trace_context_cross_agent_test.rb
```diff
@@ -253,7 +253,11 @@ def verify_attributes(test_case_attributes, actual_attributes, event_type)
...
+            assert_equal v,
+              actual_attributes[k.to_s],
+              %Q(Wrong "#{k}" #{event_type} attribute; expected #{v.inspect}, was #{actual_attributes[k.to_s].inspect})
```
fallwith, line 259: It is my understanding that `k` and `v` are always strings for this new test. Is that correct? If so, all of the `to_s` and `#inspect` calls can be removed. Also NOTE that by virtue of using interpolation (`#{}`), Ruby will automatically call `#to_s` on the interpolated object.

test/new_relic/agent/distributed_tracing/trace_context_cross_agent_test.rb
```diff
@@ -253,7 +253,11 @@ def verify_attributes(test_case_attributes, actual_attributes, event_type)
...
+          end
```
fallwith, line 260: When I run the new test, I see that it there is a `'vendors'` key present only twice.

In those 2 cases, for me `k` is equal to either `'44@nr'`, `'foo'`, or `'bar'`. `v` is always `nil`. The corresponding `actual_attributes[k.to_s]` value is always `nil`. So at least on my machine with a single Ruby version being tested, we're really only ever comparing `nil` with `nil` and getting a false sense of security from it. Am I doing something incorrectly?

My wild guess is that given the `test_case_attributes['vendors']` value of `['foo', 'bar']`, you wanted `k` to end up holding `'foo'` and `v` to end up holding `'bar'`. If that's the case, we'll need to change the iteration syntax.

fallwith (changes requested):

test/new_relic/agent/distributed_tracing/trace_context_cross_agent_test.rb
fallwith: addressed with https://github.com/newrelic/newrelic-ruby-agent/pull/1692

test/new_relic/agent/distributed_tracing/trace_context_cross_agent_test.rb
fallwith: addressed with https://github.com/newrelic/newrelic-ruby-agent/pull/1692

test/new_relic/agent/distributed_tracing/trace_context_cross_agent_test.rb
fallwith: addressed with https://github.com/newrelic/newrelic-ruby-agent/pull/1692

PR #1699, opened 2022-12-14
Add record_metric_once method to Agent module

### LINKED ISSUE #1687: Create a way to record a metric only once
We need to know when our Concurrent Ruby instrumentation is lit up by our customers's codebases, not just when it is installed because it's a dependency of Rails. 

The working idea to solve this problem right now is to record a metric the first time one of the methods we instrument gets called. 

This prevents a bunch of metrics from being recorded unnecessarily and gives us the visibility we're after.
We have a strategy to `log_once`, let's make something similar for metrics.

Acceptance Criteria:
* record a supportability metric only the first time an instrumented method is invoked

tannalynn:
# Overview
Added a method for the agent to keep track of metrics that it wants to only record once during the life of the agent. 

closes #1687

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


lib/new_relic/agent.rb
```diff
@@ -203,6 +204,13 @@ def record_metric(metric_name, value) # THREAD_LOCAL_ACCESS
       agent.stats_engine.tl_record_unscoped_metrics(metric_name, value)
     end
 
+    def record_metric_once(metric_name, value = 0.0)
+      return if @metrics_already_recorded.include?(metric_name)
```
fallwith, line 208: For large datasets, taking the memory consumption hit to use a hash instead of an array can really speed up existence checks. Here's an example benchmark script:

```ruby
require 'securerandom'
require 'benchmark'

NAME_COUNT = 1000
TIMES = 10_000

METRIC_NAMES = NAME_COUNT.times.each_with_object([]) { |_i, a| a << SecureRandom.hex }
ARRAY = METRIC_NAMES.dup.shuffle
HASH = METRIC_NAMES.each_with_object({}) { |n, h| h[n] = 1 }
TESTERS = TIMES.times.each_with_object([]) { |_i, a| a << METRIC_NAMES.sample }

Benchmark.bm do |x|
  x.report('Array') { TESTERS.each { |t| ARRAY.include?(t) } }
  x.report('Hash') { TESTERS.each { |t| HASH.include?(t) } }
end
```

My `1000` number much be much larger than the total number of unique metric names we anticipate storing. If so, please lower it in the script and see if there's any performance benefit between the array and hash approaches.

tannalynn: Hmm, that's a good point. I'm not really expecting this to be a commonly used thing, especially since the agent has gone this long already without needing something like this. 
I did run the lookup_variations.rb in the microbenchmark repo ( i also added an array.any? one too bc i was curious), and i went with array because it was slightly the fastest, with the hash being 'sameish'. 
```
        array lookup:  7074614.3 i/s
         hash lookup:  6527173.8 i/s - same-ish: difference falls within error
          set lookup:  6429436.8 i/s - same-ish: difference falls within error
    array lookup any:  4251835.8 i/s - 1.66x slower
 array detect lookup:  2151400.8 i/s - 3.29x slower
```
BUT there only 9 things in the array for that, so I haven't run the same thing with a larger number. I can update it to a hash instead if we think this has the potential to grow to be a large array. What do you think?

fallwith: Ah, I had spaced on `Set` before. That's a great option for this use case for the lookup speed of a hash without wasted space with values we don't want.

Here are  2 options for using Set for you to test with:

### Option 1: drop-in replacement for Array

```diff
- @metrics_already_recorded = []
+ @metrics_already_recorded = Set.new
```

### Option 2: Leverage Set#add?

It's possible that `Set#add?` is more efficient than combining `Set#include?` and `Set#<<`:

```diff
- @metrics_already_recorded = []
+ @metrics_already_recorded = Set.new

  def record_metric_once(metric_name, value = 0.0)
-   return if @metrics_already_recorded.include?(metric_name)
- 
    record_metric(metric_name, value)
-   @metrics_already_recorded << metric_name
+   @metrics_already_recorded.add?(metric_name)
  end
```

NOTE: Both your original `record_metric_once` version and this one do not return the result of `record_metric`. If that's desired, we'll have to refactor.

lib/new_relic/agent.rb
```diff
@@ -109,6 +109,7 @@ class SerializationError < StandardError; end
     @logger = nil
     @tracer_lock = Mutex.new
     @tracer_queue = []
+    @metrics_already_recorded = []
```
kaylareopelle, line 112: We only add metrics to this array when `#record_metric_once` is called, correct? So the number of items in the object would max out with the number of times we call this method?

lib/new_relic/agent.rb
kaylareopelle: Love this discussion! I would change part of your recommendation for `Set#add`, @fallwith: 

```diff
-  def record_metric_once(metric_name, value = 0.0)
+  def record_metric_once(metric_name)
-    return if @metrics_already_recorded.include?(metric_name)
+    return unless @metrics_already_recorded.add?(metric_name)
      record_metric(metric_name, value)
  end 
```

This will exit if `add?` returns nil (meaning the item is already in the Set), and record the metric if it's being added to the Set for the first time.

(All provided, of course, this is the best/most performant option according to @tannalynn)

kaylareopelle (changes requested):
Thank you, Tanna! This is looking great. Could you add a performance test to this PR as well? Since we're talking about performance for this method already, it seems like a good candidate.

lib/new_relic/agent.rb
fallwith: Thanks for this correction, @kaylareopelle!

tannalynn: updated in 777da8451c05d7a244170fcd9e41725d09f82ea4

lib/new_relic/agent.rb
tannalynn: It would actually max out with the number of times we call this method with a unique metric name. If we call this method 100 times all with the same metric, it would still only be a size of 1.  If we called it with a unique metric name each time though, that would increase it, yes.

tannalynn:
> Thank you, Tanna! This is looking great. Could you add a performance test to this PR as well? Since we're talking about performance for this method already, it seems like a good candidate.

added in cd0c15117caa4dcbbccdd5ad0fcc62739defe350

lib/new_relic/agent.rb
```diff
@@ -203,6 +204,12 @@ def record_metric(metric_name, value) # THREAD_LOCAL_ACCESS
       agent.stats_engine.tl_record_unscoped_metrics(metric_name, value)
     end
 
+    def record_metric_once(metric_name, value = 0.0)
+      return if @metrics_already_recorded.add?(metric_name)
```
fallwith, line 208: `Set#add?` will return the set itself if the addition took place and `nil` otherwise. Given that you only want to record the metric once, I think you should use `return unless` here instead of `return if`.

From looking at your new test, I think the reason it passes is because the 2nd call actually causes the metric to be recorded and then there's no 3rd call which would have caused a 2nd recording. I think you should update the test to perform fully 3 separate `record_metric_once` calls and then have it confirm that only 1 call to `record_metric` is made after the `if -> unless` change is made.

fallwith (changes requested):

lib/new_relic/agent.rb
tannalynn: thanks for catching that! the way it's working currently is anti-helpful lol. 
updated in 026baa594e6b0e17c408f6cb6cda7c56259f5df1

fallwith (approved):

Issue #662, opened 2021-04-29
RPM forces Rails frameworks to load during initialization


jdelStrother:
## Description

Loading the newrelic gem causes ActionView::Base, ActiveRecord::Base & ActiveJob::Base to load [before config-initialization](https://github.com/newrelic/newrelic-ruby-agent/blob/1b807886b7fd462dad16f9d77ddc30f5f1fa174e/lib/newrelic_rpm.rb#L40). Normally this should be deferred to after_initialize.

One particular example: Rails generates a `config/initializers/new_framework_defaults_6_1.rb` file which configures ActiveRecord settings (eg `Rails.application.config.active_record.has_many_inversing = true`).  However, if ActiveRecord::Base has already loaded, this config has no effect, and any changes to it are silently ignored.

## Expected Behavior
ActionView::Base, ActiveRecord::Base & ActiveJob::Base aren't loaded until after config/initializers are processed.

Here's the stack where ActiveRecord is loaded:
```

/nix/store/666hwyf9pmlq3vw6z9p9lmshbinvwpri-ruby2.7.2-activerecord-6.1.3.1/lib/ruby/gems/2.7.0/gems/activerecord-6.1.3.1/lib/active_record/base.rb:315:in `<module:ActiveRecord>'
/nix/store/666hwyf9pmlq3vw6z9p9lmshbinvwpri-ruby2.7.2-activerecord-6.1.3.1/lib/ruby/gems/2.7.0/gems/activerecord-6.1.3.1/lib/active_record/base.rb:15:in `<main>'
/nix/store/i3zqcq6pzd6bdgvgrf2xrmbb1vnrj8vm-ruby2.7.2-zeitwerk-2.4.2/lib/ruby/gems/2.7.0/gems/zeitwerk-2.4.2/lib/zeitwerk/kernel.rb:34:in `require'
/nix/store/jbxyvfbx31cj67cnin6qfb8969h23qni-ruby2.7.2-activesupport-6.1.3.1/lib/ruby/gems/2.7.0/gems/activesupport-6.1.3.1/lib/active_support/dependencies.rb:332:in `block in require'
/nix/store/jbxyvfbx31cj67cnin6qfb8969h23qni-ruby2.7.2-activesupport-6.1.3.1/lib/ruby/gems/2.7.0/gems/activesupport-6.1.3.1/lib/active_support/dependencies.rb:299:in `load_dependency'
[...]

PR #1701, opened 2022-12-15
Remove defer_rails_initialization config

kaylareopelle:
# Overview

Configs are loaded during initialization. This fix was placed behind a feature flag referencing the config value, but that value will always be falsey at initialization, so the fix was unreachable.

This change will always use the new, split initialization process.

Closes: #662 

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


kaylareopelle:
@jdelStrother - This branch is up to date with the upcoming 8.15.0 release. We aren't including it in this release because we weren't able to schedule the fixes this cycle. We're hoping to use an environment variable to allow folks to toggle the setting instead of the config value. If that doesn't work, this will need to wait until a major release. We'll leave the PR open and update it before releases to make sure you continue to have access to the feature.

kaylareopelle:
Picking this up at a later date, closing the PR for now.

PR #1712, opened 2022-12-24
CI: test Padrino with both Sinatra 2 and 3

fallwith:
Now that Padrino 0.15.2 is out with support for Sinatra 3, let's test Padrino with both Sinatra 2 and 3.

thanks, @nesquena


nesquena:
Sure I hope that 0.15.2 resolves the issues! cc @jkowens

kaylareopelle (approved):

PR #1716, opened 2022-12-29
Use Node 16 GitHub Actions

kaylareopelle:
Node 12 is deprecated for GitHub Actions and Node 16+ is required. This PR upgrades actions from the marketplace that have been identified by GitHub as running on Node 12 to newer versions. It also updates our self-written actions to use Node 16.

Actions run with deprecation warnings: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/3795916450 

One issue I'm seeing in the output with this change is a warning related to caching mysql: "Failed to restore: Tar failed with error: The process '/usr/bin/tar' failed with exit code 2" 

I believe this is because it's running for the first time, so there isn't a cache to restore. We should keep an eye on this when the PR is merged.

### DIFF
--- a/.github/actions/annotate/action.yml
+++ b/.github/actions/annotate/action.yml
@@ -2,5 +2,5 @@ name: 'Annotate Errors'
 description: 'Annotates errors if an errors.txt file is present'
 author: New Relic
 runs:
-  using: 'node12'
+  using: 'node16'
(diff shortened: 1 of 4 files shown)


hannahramadan (approved):
Nice! Interested to see if that warning goes away with caching.

PR #1733, opened 2023-01-05
Add workflow_dispatch to performance tests

kaylareopelle:
This will allow perf tests to run on demand


hannahramadan (approved):
I've seen workflow dispatch but am just now finally looking it up. Super cool!

fallwith (approved):

PR #1736, opened 2023-01-06
CI Updates 

tannalynn:
# Overview
- Moves all jruby ci into its own workflow that will be run on schedule (but won't notify slack) and can be triggered manually
- Chowns the directory mysql needs to be installed in, allowing the cache action to sucessfully pull down files for mysql from the cache without having to use a forked cache action
- Removes the rubocop option to enable new cops automatically. Since new cops being enabled automatically has cause a lot of toil, this will prevent unexpected rubocop updates from blocking PRs and releases.


.github/workflows/ci.yml
```diff
@@ -16,8 +16,8 @@ jobs:
       - run: bundle
       - run: rubocop
 
+
   unit_tests:
-    needs: run_rubocop
```
fallwith, line 20: This change wasn't mentioned in the PR description.

hannahramadan, line 20: Does this mean none of the tests need rubocop to pass as a prerequisite?

.github/workflows/ci.yml
```diff
@@ -245,8 +252,8 @@ jobs:
           path: lib/coverage_*/.resultset.json
           retention-days: 2
 
+
   infinite_tracing:
-    needs: run_rubocop
```
fallwith, line 249: This change wasn't mentioned in the PR description.

fallwith (approved):

Issue #1513, opened 2022-10-01
Rails notifications: Add subscriptions to ActionMailer and ActionMailbox topics


fallwith:
Rails will publish content to the following 2 ActionMailer topics:

- `deliver.action_mailer`
- `process.action_mailer`

We should subscribe to these in a similar manner used for `*.action_controller` and `*.action_view` topics.

FOR TEAM DISCUSSION: we do not yet instrument ActionMailer itself. This issue does not depend on us doing so, but it might be a good idea to group all ActionMailer work together.

FOR TEAM DISCUSSION: we should also consider subscribing to `process.action_mailbox` which is mail related.

**Update 2022-11-07** the team discussed this one and concluded:

* There should be 2 new subscriber configs/classes; one for ActionMailer and one for ActionMailbox. These should subscribe to the 3 topics listed in the description above.
* There should be 2 new subscriber agent config settings to a) disable the on-by-default subscribing to ActionMailer and b) disable the on-by-default subscribing to ActionMailbox
* For a good example class to make copies of for ActionMailer and ActionMailbox, see the [ActiveStorage instrumentation class](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/active_storage.rb) and the [ActiveStorage subscriber class](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/active_storage_subscriber.rb).

kaylareopelle:
Closed by #1740

PR #1740, opened 2023-01-13
ActionMailer and ActionMailbox instrumentation

fallwith:
add Rails notifications based instrumentation support for ActionMailer and ActionMailbox

addresses #1513

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1146,14 +1146,14 @@ A map of error classes to a list of messages. When an error of one of the classe
           :public => true,
           :type => Boolean,
           :allowed_from_server => false,
-          :description => 'If `true`, disables ActionMailbox instrumentation.'
+          :description => 'If `true`, disables Action Mailbox instrumentation.'
         },
         :disable_action_mailer => {
           :default => false,
           :public => true,
           :type => Boolean,
           :allowed_from_server => false,
-          :description => 'If `true`, disables ActionMailer instrumentation.'
+          :description => 'If `true`, disables Action Mailer instrumentation.'
         },
         :disable_activejob => {
           :default => false,
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1189,7 +1189,7 @@ A map of error classes to a list of messages. When an error of one of the classe
           :type => Boolean,
           :dynamic_name => true,
           :allowed_from_server => false,
-          :description => 'If `true`, disables instrumentation for ActiveRecord 4, 5, and 6.'
+          :description => 'If `true`, disables instrumentation for Active Record 4+'
         },
         :disable_bunny => {
           :default => false,
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1145,7 +1145,6 @@ A map of error classes to a list of messages. When an error of one of the classe
           :default => false,
           :public => true,
           :type => Boolean,
-          :dynamic_name => true,
           :allowed_from_server => false,
           :description => 'If `true`, disables ActionMailbox instrumentation.'
         },
@@ -1153,7 +1152,6 @@ A map of error classes to a list of messages. When an error of one of the classe
           :default => false,
           :public => true,
           :type => Boolean,
-          :dynamic_name => true,
           :allowed_from_server => false,
           :description => 'If `true`, disables ActionMailer instrumentation.'
         },
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1141,6 +1141,22 @@ A map of error classes to a list of messages. When an error of one of the classe
           :allowed_from_server => false,
           :description => 'If `true`, disables Action Cable instrumentation.'
         },
+        :disable_action_mailbox => {
+          :default => false,
+          :public => true,
+          :type => Boolean,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => 'If `true`, disables ActionMailbox instrumentation.'
+        },
+        :disable_action_mailer => {
+          :default => false,
+          :public => true,
+          :type => Boolean,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => 'If `true`, disables ActionMailer instrumentation.'
+        },
         :disable_activejob => {
           :default => false,
           :public => true,
--- a/lib/new_relic/agent/instrumentation/action_mailbox.rb
+++ b/lib/new_relic/agent/instrumentation/action_mailbox.rb
@@ -20,7 +20,7 @@ DependencyDetection.defer do
   end
 
   executes do
-    ActiveSupport::Notifications.subscribe(/^process\.action_mailbox$/,
+    ActiveSupport::Notifications.subscribe(/\A[^\.]+\.action_mailbox\z/,
       NewRelic::Agent::Instrumentation::ActionMailboxSubscriber.new)
   end
 end
--- a/lib/new_relic/agent/instrumentation/action_mailbox.rb
+++ b/lib/new_relic/agent/instrumentation/action_mailbox.rb
@@ -11,9 +11,8 @@ DependencyDetection.defer do
     defined?(ActiveSupport) &&
       defined?(ActionMailbox) &&
       ActionMailbox.respond_to?(:gem_version) && # 'require "action_mailbox"' doesn't require version...
-      ActionMailbox.gem_version >= Gem::Version.new('7.1.0.alpha') # notifications added in Rails 7.1
-
-    !NewRelic::Agent::Instrumentation::ActionMailboxSubscriber.subscribed?
+      ActionMailbox.gem_version >= Gem::Version.new('7.1.0.alpha') && # notifications added in Rails 7.1
+      !NewRelic::Agent::Instrumentation::ActionMailboxSubscriber.subscribed?
   end
 
   executes do
--- a/lib/new_relic/agent/instrumentation/action_mailbox.rb
+++ b/lib/new_relic/agent/instrumentation/action_mailbox.rb
@@ -1,0 +1,27 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'new_relic/agent/instrumentation/action_mailbox_subscriber'
+
+DependencyDetection.defer do
+  named :action_mailbox
+
+  depends_on do
+    defined?(ActiveSupport) &&
+      defined?(ActionMailbox) &&
+      ActionMailbox.respond_to?(:gem_version) && # 'require "actionmalbox"' doesn't require version...
+      ActionMailbox.gem_version >= Gem::Version.new('7.1.0.alpha') # notifications added in Rails 7.1
+
+    !NewRelic::Agent::Instrumentation::ActionMailboxSubscriber.subscribed?
+  end
+
+  executes do
+    NewRelic::Agent.logger.info('Installing ActionMailbox instrumentation')
+  end
+
+  executes do
+    ActiveSupport::Notifications.subscribe(/^process\.action_mailbox$/,
+      NewRelic::Agent::Instrumentation::ActionMailboxSubscriber.new)
+  end
+end
--- a/lib/new_relic/agent/instrumentation/action_mailbox.rb
+++ b/lib/new_relic/agent/instrumentation/action_mailbox.rb
@@ -10,7 +10,7 @@ DependencyDetection.defer do
   depends_on do
     defined?(ActiveSupport) &&
       defined?(ActionMailbox) &&
-      ActionMailbox.respond_to?(:gem_version) && # 'require "actionmalbox"' doesn't require version...
+      ActionMailbox.respond_to?(:gem_version) && # 'require "action_mailbox"' doesn't require version...
       ActionMailbox.gem_version >= Gem::Version.new('7.1.0.alpha') # notifications added in Rails 7.1
 
     !NewRelic::Agent::Instrumentation::ActionMailboxSubscriber.subscribed?
--- a/lib/new_relic/agent/instrumentation/action_mailbox_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/action_mailbox_subscriber.rb
@@ -26,8 +26,6 @@ module NewRelic
 
         def start_segment(name, id, payload)
           segment = Tracer.start_segment(name: metric_name(name, payload))
-          segment.params[:key] = payload[:key]
-          segment.params[:exist] = payload[:exist] if payload.key?(:exist)
           push_segment(id, segment)
         end
 
--- a/lib/new_relic/agent/instrumentation/action_mailbox_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/action_mailbox_subscriber.rb
@@ -41,12 +41,12 @@ module NewRelic
         end
 
         def metric_name(name, payload)
-          service = payload[:service]
+          mailbox = payload[:mailbox].class.name
           method = method_from_name(name)
-          "Ruby/ActionMailbox/#{service}Service/#{method}"
+          "Ruby/ActionMailbox/#{mailbox}/#{method}"
         end
 
-        PATTERN = /\Aservice_([^\.]*)\.action_mailbox\z/
+        PATTERN = /\A([^\.]*)\.action_mailbox\z/
         UNKNOWN = "unknown".freeze
 
         METHOD_NAME_MAPPING = Hash.new do |h, k|
--- a/lib/new_relic/agent/instrumentation/action_mailbox_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/action_mailbox_subscriber.rb
@@ -1,0 +1,66 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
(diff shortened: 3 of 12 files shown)


lib/new_relic/agent/instrumentation/action_mailbox.rb
```diff
@@ -0,0 +1,27 @@
...
+
+require 'new_relic/agent/instrumentation/action_mailbox_subscriber'
+
+DependencyDetection.defer do
+  named :action_mailbox
+
+  depends_on do
+    defined?(ActiveSupport) &&
+      defined?(ActionMailbox) &&
+      ActionMailbox.respond_to?(:gem_version) && # 'require "actionmalbox"' doesn't require version...
```
hannahramadan, line 13: The comment has a misspelling of 'actionmailbox'!

fallwith: Rats! There's also supposed to be an underscore in there, "action_mailbox"! Fixed with 0aa67aed20990eba253ae8dd1dd3661a2406929d

lib/new_relic/agent/instrumentation/action_mailbox_subscriber.rb
```diff
@@ -0,0 +1,66 @@
...
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'new_relic/agent/instrumentation/notifications_subscriber'
+
+module NewRelic
+  module Agent
+    module Instrumentation
+      class ActionMailboxSubscriber < NotificationsSubscriber
+        def start(name, id, payload)
```
kaylareopelle, line 11: The barebones `start`, `finish`, etc. methods in this class are very encouraging. It looks like we're close to adding them to `NotificationsSubscriber` and simplifying our `Subscriber` sub-classes.

lib/new_relic/agent/instrumentation/action_mailer.rb
```diff
@@ -0,0 +1,26 @@
...
+require 'new_relic/agent/instrumentation/action_mailer_subscriber'
+
+DependencyDetection.defer do
+  named :action_mailer
+
+  depends_on do
+    defined?(ActiveSupport) &&
+      defined?(ActionMailer) &&
+      ActionMailer.respond_to?(:gem_version) &&
+      ActionMailer.gem_version >= Gem::Version.new('5.0') &&
```
kaylareopelle, line 14: I didn't realize ActionMailer existed in 5.0. I thought it was a 6.0 release.

test/new_relic/agent/instrumentation/rails/action_mailbox_subscriber.rb
```diff
@@ -0,0 +1,123 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require_relative '../../../../test_helper'
+require 'new_relic/agent/instrumentation/action_mailer_subscriber'
+
```
kaylareopelle, line 7: Do you know if we're at 100% test coverage for ActionMailbox and ActionMailer?

kaylareopelle (approved):

lib/new_relic/agent/instrumentation/action_mailbox.rb
```diff
@@ -0,0 +1,26 @@
...
+      ActionMailbox.gem_version >= Gem::Version.new('7.1.0.alpha') && # notifications added in Rails 7.1
+      !NewRelic::Agent::Instrumentation::ActionMailboxSubscriber.subscribed?
+  end
+
+  executes do
+    NewRelic::Agent.logger.info('Installing ActionMailbox instrumentation')
+  end
+
+  executes do
+    ActiveSupport::Notifications.subscribe(/^process\.action_mailbox$/,
```
kaylareopelle, line 23: What do you think about using a wildcard match instead of `process`? I know it's different than the original ticket, but it would be a bit more future-proof.

lib/new_relic/agent/instrumentation/action_mailer_subscriber.rb
```diff
@@ -0,0 +1,66 @@
...
+          if segment = pop_segment(id)
+            if exception = exception_object(payload)
+              segment.notice_error(exception)
+            end
+            segment.finish
+          end
+        end
+
+        def metric_name(name, payload)
+          service = payload[:service]
```
kaylareopelle, line 44: Is `:service` part of the payload for ActionMailer? I'm only seeing the keys listed on this guide: https://guides.rubyonrails.org/active_support_instrumentation.html#action-mailer 

It seems like `mailer` might be a better fit.

lib/new_relic/agent/instrumentation/action_mailer_subscriber.rb
```diff
@@ -0,0 +1,66 @@
...
+          method = method_from_name(name)
+          "Ruby/ActionMailer/#{service}Service/#{method}"
+        end
+
+        PATTERN = /\Aservice_([^\.]*)\.action_mailer\z/
```
kaylareopelle, line 49: This may need to be changed too. I don't believe all the `action_mailer` events are prefixed with `service`.

lib/new_relic/agent/instrumentation/action_mailbox.rb
kaylareopelle: Also, could you add a link to the code where the events are defined for ActionMailbox? I'd like to see them/understand the keys in the payload.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1141,6 +1141,22 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
         },
+        :disable_action_mailbox => {
+          :default => false,
+          :public => true,
+          :type => Boolean,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => 'If `true`, disables ActionMailbox instrumentation.'
+        },
+        :disable_action_mailer => {
```
tannalynn, line 1152: Does dependency detection automatically check these configs? I see that they are marked as dynamic and you aren't referencing them directly, but I can't remember how that works.

lib/new_relic/agent/instrumentation/action_mailbox_subscriber.rb
fallwith: Yep, these are mostly if not entirely identical to the ones from the ActiveStorage subscriber and pretty generic. It would be great to DRY some of these common patterns up after the conclusion of the milestone.

test/new_relic/agent/instrumentation/rails/action_mailbox_subscriber.rb
fallwith: I modeled the testing off of the recent ActiveSupport custom events stuff which delivered 100% coverage, but I haven't checked the coverage for this PR specifically. This gives me an idea to add an optional "branch" scope to Stoat's `coverage` function. Granted, in this case with ActionMailbox we're only see full coverage on the CI run that includes Rails Edge, so that's perhaps a one off scenario that Stoat won't be expected to reach.

lib/new_relic/agent/instrumentation/action_mailbox.rb
fallwith: My take would be to hold off on wildcarding for now, and when we circle back after having some minimal coverage for every event type we can discuss perhaps standardizing on a wildcard approach for all subscribers. Given that this ActionMailbox class only has a single event that's still in alpha, I'm also leery of updating it in particular for future proofing, but I can certainly see how `./^*\.action_mailbox$/` would work. Maybe on the circle back we can just do `./^.*_#{subscriber_class}$/` dynamically or whatnot.

fallwith: The Rails Edge code that delivers the lone ActionMailbox event is here: https://github.com/rails/rails/blob/246c13f8df833cd8d8ae423af253bf06ceca0fc0/actionmailbox/lib/action_mailbox/base.rb#L81

lib/new_relic/agent/instrumentation/action_mailer_subscriber.rb
fallwith: Great call. This seems to be an example of borrowing from ActiveStorage where I shouldn't have. Will test with `mailer`.

lib/new_relic/agent/instrumentation/action_mailer_subscriber.rb
fallwith: Great call

lib/new_relic/agent/instrumentation/action_mailbox.rb
kaylareopelle: Thank you! Let's hold off then.

lib/new_relic/agent/instrumentation/action_mailbox_subscriber.rb
```diff
@@ -0,0 +1,66 @@
...
+          end
+        end
+
+        def metric_name(name, payload)
+          service = payload[:service]
+          method = method_from_name(name)
+          "Ruby/ActionMailbox/#{service}Service/#{method}"
+        end
+
+        PATTERN = /\Aservice_([^\.]*)\.action_mailbox\z/
```
kaylareopelle, line 49: @fallwith - I'm sure you're already on this, but for completeness's sake the `service`-related feedback for ActionMailer applies here too. From the link you shared, it looks like `:mailbox` and `:inbound_email` are the keys in the payload.

https://github.com/rails/rails/blob/246c13f8df833cd8d8ae423af253bf06ceca0fc0/actionmailbox/lib/action_mailbox/base.rb#L109-L114

lib/new_relic/agent/configuration/default_source.rb
fallwith: Removed the "dynamic" key/val pair for the new configs with [354f0de](https://github.com/newrelic/newrelic-ruby-agent/pull/1740/commits/354f0de483791f52daec5bae941b6e513015e127)

lib/new_relic/agent/instrumentation/action_mailer_subscriber.rb
fallwith: Reworked payload parsing with [354f0de](https://github.com/newrelic/newrelic-ruby-agent/pull/1740/commits/354f0de483791f52daec5bae941b6e513015e127)

lib/new_relic/agent/instrumentation/action_mailer_subscriber.rb
fallwith: Reworked with [354f0de](https://github.com/newrelic/newrelic-ruby-agent/pull/1740/commits/354f0de483791f52daec5bae941b6e513015e127)

lib/new_relic/agent/instrumentation/action_mailbox_subscriber.rb
fallwith: Reworked with [354f0de](https://github.com/newrelic/newrelic-ruby-agent/pull/1740/commits/354f0de483791f52daec5bae941b6e513015e127)

lib/new_relic/agent/instrumentation/action_mailbox.rb
fallwith: Switched to a wildcard with [354f0de](https://github.com/newrelic/newrelic-ruby-agent/pull/1740/commits/354f0de483791f52daec5bae941b6e513015e127)

lib/new_relic/agent/instrumentation/action_mailbox_subscriber.rb
```diff
@@ -0,0 +1,66 @@
...
+          method = method_from_name(name)
+          "Ruby/ActionMailbox/#{mailbox}/#{method}"
+        end
+
+        PATTERN = /\A([^\.]*)\.action_mailbox\z/
+        UNKNOWN = "unknown".freeze
+
+        METHOD_NAME_MAPPING = Hash.new do |h, k|
+          if PATTERN =~ k
+            h[k] = $1
```
hannahramadan, line 54: What is the `$1` doing?

fallwith: The `$1` here is referring to the first regex capture.

```ruby
'hannah' =~ /(han|nah)/
puts $1
  # => 'han'
```
To prevent confusion, Rubocop recommends using `Regexp.last_match(1)` instead of `$1`. Given that `$1` is already used by the other subscriber classes, I'm keeping it for now until we take a pass at DRYing and refactoring all subscriber classes post milestone.

newrelic.yml
```diff
@@ -169,7 +169,13 @@ common: &default_settings
...
   # disable_action_cable_instrumentation: false
 
-  # If true, disables instrumentation for Active Record 4, 5, and 6.
+  # If true, disables Action Mailbox instrumentation.
+  # disable_action_mailbox: false
+
+  # If true, disables Action Mailer instrumentation.
+  # disable_action_mailer: false
+
+  # If true, disables instrumentation for Active Record 4+
```
kaylareopelle, line 178: Since this reference is being updated, could you update the [default_source.rb reference](https://github.com/newrelic/newrelic-ruby-agent/pull/1740/files#diff-d116d3c8c7bebe204a18ee3983ac59c5d15bd70da7123f2a2c4a12f144bca181R1192) as well?

lib/new_relic/agent/instrumentation/action_mailbox_subscriber.rb
```diff
@@ -0,0 +1,66 @@
...
+
+          finish_segment(id, payload)
+        rescue => e
+          log_notification_error(e, name, 'finish')
+        end
+
+        def start_segment(name, id, payload)
+          segment = Tracer.start_segment(name: metric_name(name, payload))
+          segment.params[:key] = payload[:key]
+          segment.params[:exist] = payload[:exist] if payload.key?(:exist)
```
kaylareopelle, line 30: I think this might've been a copy/paste mistake. I think you just want to include `:mailbox`? Or maybe no params since it's in the name?

lib/new_relic/agent/instrumentation/action_mailer_subscriber.rb
```diff
@@ -0,0 +1,86 @@
...
+
+require 'new_relic/agent/instrumentation/notifications_subscriber'
+
+module NewRelic
+  module Agent
+    module Instrumentation
+      # NOTE: as of v7.1.0.0.alpha, deliver.action_mailer will provide
+      # an empty payload hash ({}) to #start, so in this subscriber class
+      # we defer params population until #finish and start the segment with
+      # a temporary name that is later replaced
```
kaylareopelle, line 13: 👏

test/new_relic/agent/instrumentation/action_mailbox_subscriber_test.rb
```diff
@@ -0,0 +1,14 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require_relative '../../../test_helper'
+require 'new_relic/agent/instrumentation/action_mailbox_subscriber'
+
+if defined?(ActionMailbox) &&
+    ActionMailbox.respond_to?(:gem_version) && # 'require "actionmalbox"' doesn't require version...
```
kaylareopelle, line 9:
```suggestion
    ActionMailbox.respond_to?(:gem_version) && # 'require "actionmailbox"' doesn't require version...
```

test/new_relic/agent/instrumentation/rails/action_mailbox_subscriber.rb
```diff
@@ -0,0 +1,151 @@
...
+    end
+
+    def test_start_when_not_traced
+      SUBSCRIBER.state.stub :is_execution_traced?, false do
+        in_transaction do |txn|
+          SUBSCRIBER.start(NAME, ID, {})
+
+          assert_empty txn.segments
+        end
+      end
```
kaylareopelle, line 46: This test was very helpful to me for the Active Support ticket! Thanks for writing it!

test/new_relic/agent/instrumentation/rails/action_mailer_subscriber.rb
```diff
@@ -0,0 +1,189 @@
...
+    #     in_transaction do |txn|
+    #       SUBSCRIBER.stub :pop_segment, -> { raise 'kaboom' } do
+    #         SUBSCRIBER.finish(NAME, ID, {})
+    #       end
+
+    #       assert_equal 1, txn.segments.size
+    #     end
+    #   end
+    #   logger.verify
+    # end
```
kaylareopelle, line 96: Did you mean to leave this commented?

kaylareopelle (changes requested):

lib/new_relic/agent/instrumentation/action_mailbox_subscriber.rb
fallwith: Ah, I started using the `:mailbox` payload key/val pair in `#metric_name` but then forgot to come back here and remove these 2 copy/paste keys. Thanks for catching that! I've gone ahead and gone with the "no params" approach. The only other param besides the mailbox class name that we're using in the segment name is an email address which I'm deliberately not going to store.

Addressed with [e71e37b](https://github.com/newrelic/newrelic-ruby-agent/pull/1740/commits/e71e37bde75ea6622e789faeecf8c8976a66dd53)

test/new_relic/agent/instrumentation/rails/action_mailer_subscriber.rb
fallwith: I could have sworn I deleted it multiple times already. Weird. It is a slightly different variation on a test of the same name that is not commented out. Hopefully finally removed now with [e71e37b](https://github.com/newrelic/newrelic-ruby-agent/pull/1740/commits/e71e37bde75ea6622e789faeecf8c8976a66dd53)

kaylareopelle (approved):
Thanks for the changes, James!

PR #1742, opened 2023-01-18
Add Active Support instrumentation

### LINKED ISSUE #1514: Rails notifications: subscribe to ActiveSupport topics
Rails (and potentially non-Rails ActiveSupport based apps) will publish content to the following 3 ActiveSupport topics:

- `cache_read.active_support`
- `cache_generate.active_support`
- `cache_write.active_support`

We should subscribe to these in a similar manner used for `*.action_controller` and `*.action_view topics`.

* There should be 1 new subscriber configs/classes for ActiveSupport. It should subscribe to the 3 topics listed in the description above.
* There should be 1 new subscriber agent config settings to disable the on-by-default subscribing to ActiveSupport
* For a good example class to make copies of for ActiveSupport, see the [ActiveStorage instrumentation class](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/active_storage.rb) and the [ActiveStorage subscriber class](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/active_storage_subscriber.rb)

kaylareopelle:
Active Support has notifications built in to instrument caching-related activities, including:
* `cache_read`
* `cache_generate`
* `cache_fetch_hit`
* `cache_write`
* `cache_delete`
* `cache_exist?`

The payload always includes the `:key` and the name of the `:store` class. `cache_read` also includes whether the read is a `:hit`, and `:super_operation` when a read is used with `#fetch`.

https://guides.rubyonrails.org/active_support_instrumentation.html#active-support

When these operations are executed, a new segment will be started that includes the payload within the params attribute.

Closes #1514

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ This upcoming release of the agent adds instrumentation for Active Support cachi
 
   | Configuration name | Default | Behavior |
   | ----- | ----- | ----- |
-  | `disable_active_support` | `false` | If `true`, disables ActiveSupport instrumentation. |
+  | `disable_active_support` | `false` | If `true`, disables Active Support instrumentation. |
 
 ## 8.15.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # New Relic Ruby Agent Release Notes
 
+## Upcoming Release
+
+This upcoming release of the agent adds instrumentation for Active Support caching operations.
+
+- **Add Active Support Instrumentation**
+
+  Active Support instruments caching-related operations. This change includes the instrumentation in the agent. A new segment is created when one of the operations is called. The key, store, and other attributes will be included on the segment as params. [PR#1742](https://github.com/newrelic/newrelic-ruby-agent/pull/1742)
+
+  | Configuration name | Default | Behavior |
+  | ----- | ----- | ----- |
+  | `disable_active_support` | `false` | If `true`, disables ActiveSupport instrumentation. |
+
 ## 8.15.0
 
 Version 8.15.0 of the agent confirms compatibility with Ruby 3.2.0, adds instrumentation for concurrent-ruby, and confirms Sinatra 3 compatibility with Padrino 0.15.2. It also enables batching and compression for Infinite Tracing.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This upcoming release of the agent adds instrumentation for Active Support cachi
 
 - **Add Active Support Instrumentation**
 
-  Active Support instruments caching-related operations. This change includes the instrumentation in the agent. A new segment is created when one of the operations is called. The key, store, and other attributes will be included on the segment as params. [PR#1742](https://github.com/newrelic/newrelic-ruby-agent/pull/1742)
+  Instrumentation is now automatically provided for all [Active Support caching](https://guides.rubyonrails.org/caching_with_rails.html) operations. Whenever a caching operation is performed, a New Relic segment is created that contains timing information as well as parameters for the cache key, store, and other relevant data. [PR#1742](https://github.com/newrelic/newrelic-ruby-agent/pull/1742)
 
   | Configuration name | Default | Behavior |
   | ----- | ----- | ----- |
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1137,7 +1137,6 @@ A map of error classes to a list of messages. When an error of one of the classe
           :default => false,
           :public => true,
           :type => Boolean,
-          :dynamic_name => true,
           :allowed_from_server => false,
           :description => 'If `true`, disables Action Cable instrumentation.'
         },
@@ -1145,25 +1144,22 @@ A map of error classes to a list of messages. When an error of one of the classe
           :default => false,
           :public => true,
           :type => Boolean,
-          :dynamic_name => true,
           :allowed_from_server => false,
-          :description => 'If `true`, disables ActiveJob instrumentation.'
+          :description => 'If `true`, disables Active Job instrumentation.'
         },
         :disable_active_storage => {
           :default => false,
           :public => true,
           :type => Boolean,
-          :dynamic_name => true,
           :allowed_from_server => false,
-          :description => 'If `true`, disables ActiveStorage instrumentation.'
+          :description => 'If `true`, disables Active Storage instrumentation.'
         },
         :disable_active_support => {
           :default => false,
           :public => true,
           :type => Boolean,
-          :dynamic_name => true,
           :allowed_from_server => false,
-          :description => 'If `true`, disables ActiveSupport instrumentation.'
+          :description => 'If `true`, disables Active Support instrumentation.'
         },
         :disable_activerecord_instrumentation => {
           :default => value_of(:skip_ar_instrumentation),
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+
+This upcoming release of the agent adds instrumentation for Active Support caching operations.
+
+- **Add Active Support Instrumentation**
+
+  Active Support instruments caching-related operations. This change includes the instrumentation in the agent. A new segment is created when one of the operations is called. The key, store, and other attributes will be included on the segment as params. [PR#1742](https://github.com/newrelic/newrelic-ruby-agent/pull/1742)
+
+  | Configuration name | Default | Behavior |
+  | ----- | ----- | ----- |
+  | `disable_active_support` | `false` | If `true`, disables ActiveSupport instrumentation. |
```
fallwith, line 13: "Active Support" and "ActiveSupport" are both correct depending on context, but seeing them side by side in the same file like this makes me wonder if we should standardize on one or the other.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1157,6 +1157,14 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'If `true`, disables ActiveStorage instrumentation.'
         },
+        :disable_active_support => {
+          :default => false,
+          :public => true,
+          :type => Boolean,
+          :dynamic_name => true,
```
fallwith, line 1164: I need to fix this in my PR as well, but I don't believe this new parameter name is dynamic, given that it exists in `newrelic.yml` and is statically known before agent startup.

lib/new_relic/agent/instrumentation/active_support.rb
```diff
@@ -2,11 +2,22 @@
...
+    defined?(ActiveSupport) &&
+      !NewRelic::Agent::Instrumentation::ActiveSupportSubscriber.subscribed?
+  end
+
+  executes do
+    NewRelic::Agent.logger.info('Installing ActiveSupport instrumentation')
+  end
+
+  executes do
+    ActiveSupport::Notifications.subscribe(/\.active_support$/,
```
fallwith, line 20: I'm still not clear on why we use `^` and `$` in some places and `\A` and `\z` in others for these patterns.

newrelic.yml
```diff
@@ -175,6 +175,9 @@ common: &default_settings
   # If true, disables Active Storage instrumentation.
   # disable_active_storage: false
 
+  # If true, disables Active Support instrumentation.
```
fallwith, line 178: `default_source.rb` uses `ActiveSupport` for the description but `newrelic.yml` here is using `Active Support`.

fallwith (changes requested):

CHANGELOG.md
kaylareopelle: Oof, great point. I usually see "Active Support" in the Rails guides, so I was trying to follow that. I missed the config spelling though. Are you comfortable with the spaced version?

lib/new_relic/agent/instrumentation/active_support.rb
kaylareopelle: Great question. I think it has more to do with the person's preference at the time of writing. For me, it was copy-pasta. I wonder if one of them has a performance benefit. Do you know off the top of your head? If not, perhaps we can investigate as part of the post-milestone refactor.

CHANGELOG.md
kaylareopelle: The spaced version has been incorporated in this commit: 278635542852a978fcb57364752500c445f237c6

Happy to revert.

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: Fixed in: 278635542852a978fcb57364752500c445f237c6 

I also tackled a few others. Let me know if you'd like me to move those to a separate PR.

newrelic.yml
kaylareopelle: Made consistent in: 278635542852a978fcb57364752500c445f237c6

CHANGELOG.md
kaylareopelle: Oof. This reference was fixed in this commit: 4c10f14f713ef094a86497628a9b864a2e1d1a92

lib/new_relic/agent/instrumentation/active_support.rb
fallwith: Good call on post-milestone. I think `\A` and `\z` are best. They represent the absolute beginning and end of the entire text, whereas `^` and `$` represent the start and stop of any given line.

For example, this text blob will be matched with your current regex, but arguably shouldn't.

```text
This is a simple
.active_support
example
```

Given that the input is expected to a) not be user crafted and b) always exactly one line, this regex and others like it should be fine for now.

fallwith (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # New Relic Ruby Agent Release Notes
 
+## Upcoming Release
+
+This upcoming release of the agent adds instrumentation for Active Support caching operations.
+
+- **Add Active Support Instrumentation**
+
+  Active Support instruments caching-related operations. This change includes the instrumentation in the agent. A new segment is created when one of the operations is called. The key, store, and other attributes will be included on the segment as params. [PR#1742](https://github.com/newrelic/newrelic-ruby-agent/pull/1742)
```
tannalynn, line 9: > This change includes the instrumentation in the agent.

I'm not sure what this sentence means, really. Could we reword or remove it?

test/new_relic/agent/instrumentation/rails/active_support_subscriber.rb
```diff
@@ -0,0 +1,213 @@
...
+          in_transaction('test') do
+            generate_event('cache_new_method.active_support')
+          end
+
+          assert_metrics_recorded "#{METRIC_PREFIX}#{DEFAULT_STORE}/new_method"
+        end
+
+        def test_failsafe_if_event_does_not_match_expected_pattern
+          in_transaction('test') do
+            generate_event('charcuterie_build_a_board_workshop')
```
tannalynn, line 79: lmao

CHANGELOG.md
fallwith: There's some ambiguity here in the whole paragraph given that Active Support is what's providing a high level caching interface for a Rails or other type of Ruby app AND what's providing pub/sub notifications for events related to the caching. Here's an attempt at rewrite:

```
Instrumentation is now automatically provided for all [Active Support caching](https://guides.rubyonrails.org/caching_with_rails.html) operations. Whenever a caching operation is performed, a New Relic segment will be created that contains timing information as well as parameters for the cache key, store, and other relevant data. [PR#1742](https://github.com/newrelic/newrelic-ruby-agent/pull/1742)
```

> Instrumentation is now automatically provided for all [Active Support caching](https://guides.rubyonrails.org/caching_with_rails.html) operations. Whenever a caching operation is performed, a New Relic segment will be created that contains timing information as well as parameters for the cache key, store, and other relevant data. [PR#1742](https://github.com/newrelic/newrelic-ruby-agent/pull/1742)

kaylareopelle: Great points! Thanks for the rewrite, James. Incorporated in: c5db1c26942912fcbe69ecc284567b95003ebf1c

PR #1743, opened 2023-01-19
Active Support instrumentation: add test

fallwith:
Add a unit test for the Active Support instrumentation that does not directly invoke the subscriber class's `#start` and `#finish` methods manually, but instead performs a real Active Storage cache write and confirms that the instrumentation works as expected.

@kreopelle I have found a potential problem with the `NotificationsSubscriber` class being used for `ActionMailer` and `ActionMailbox` notifications. I wanted to be sure that your new `ActiveSupport` instrumentation did not have the same problem. I authored this new test and am happy to report that it confirms the new `ActiveSupport` instrumentation is not impacted by the same problem. I figured we might as well commit the test, so here's a PR.

### DIFF
--- a/test/new_relic/agent/instrumentation/rails/active_support_subscriber.rb
+++ b/test/new_relic/agent/instrumentation/rails/active_support_subscriber.rb
@@ -200,6 +200,10 @@ module NewRelic
         end
 
         def test_an_actual_active_storage_cache_write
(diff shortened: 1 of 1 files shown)


kaylareopelle (approved):
Thank you, @fallwith! This makes the testing for our new instrumentation much more robust.

PR #1744, opened 2023-01-20
Subscribe to additional Action controller topics

### LINKED ISSUE #1510: Rails notifications: subscribe to additional ActionController topics
We currently subscribe to a single ActionController related topic, `process_action.action_controller` via `lib/new_relic/agent/instrumentation/rails_notifications/action_controller.rb`. We should be subscribing to these others as well:

- `start_processing.action_controller`
- `send_file.action_controller`
- `send_data.action_controller`
- `redirect_to.action_controller`
- `halted_callback.action_controller`
- `unpermitted_parameters.action_controller`

Our subscription method is regex based, so we could potentially use `/\.action_controller$/`.

So the regex [here](https://github.com/newrelic/newrelic-ruby-agent/blob/bfb70572fec478f42c4b0ff68ea235c89bd66fd5/lib/new_relic/agent/instrumentation/rails_notifications/action_controller.rb#L30) would simply be altered so that it could match the existing `process_action.action_controller` topic AND all of the other topic listed above.

Resources:
- [Rails API for ActiveSupport Notifications](https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html)
- [ActionController topics/common event hooks](https://edgeguides.rubyonrails.org/active_support_instrumentation.html#action-controller)

### LINKED ISSUE #1512: Rails notifications: subscribe to additional ActionView topics
Via `lib/new_relic/agent/instrumentation/action_view_subscriber.rb`, we currently subscribe to 3 ActionView topics. These topics are stored as constants `RENDER_TEMPLATE_EVENT_NAME`, `RENDER_PARTIAL_EVENT_NAME`, and `RENDER_COLLECTION_EVENT_NAME` in that class.

Rails will publish to another topic, `render_layout.action_view`. We should subscribe to this one as well.

tannalynn:
- Added a start/finish methods to the parent class to share with other classes that are not doing anything complicated. I also removed the methods from subclasses that no longer need it. 
- I also moved the tests related to the start and finish methods into a notifications_subscriber test
- Since the new action controller topics were so different from process_action, it made more sense to create a new subscriber class for it
- I did not subscribe to `start_processing` because it was just creating duplicates of process_action
- 
[passing on all versions with latest commit](https://github.com/newrelic/newrelic-ruby-agent/actions/runs/3999676236)
closes #1510 
closes #1512


test/new_relic/agent/instrumentation/rails/active_support_subscriber.rb
```diff
@@ -163,42 +159,6 @@ def test_pop_segment_returns_false
           end
         end
 
-        def test_start_logs_notification_error
```
tannalynn, line 166: Moved these tests into the notification_subscriber_test since these are testing what is now shared code

lib/new_relic/agent/instrumentation/action_controller_other_subscriber.rb
```diff
@@ -0,0 +1,46 @@
...
+        end
+
+        def add_segment_params(segment, payload, name)
+          segment.params[:filter] = payload[:filter] if payload[:filter]
+          segment.params[:keys] = payload[:keys] if payload[:keys]
+          segment.params[:original_path] = payload[:request].original_fullpath if payload[:request]
+
+          if payload[:context]
+            segment.params[:action] = payload[:context][:action]
+            segment.params[:controller] = payload[:context][:controller].to_s
```
fallwith, line 27: Is there a way to get at the string value that is being provided for the `:controller` payload key [here](https://github.com/rails/rails/blob/fc950324bdbca393b29f1c26d2bf4e7bb2aa8c42/actionpack/lib/action_controller/metal/instrumentation.rb#L59)?

If not, and we're left working with an actual controller class here, I suggest using `name` instead of `to_s` to better convey that the payload value is a class.

lib/new_relic/agent/instrumentation/action_controller_other_subscriber.rb
```diff
@@ -0,0 +1,46 @@
...
+          end
+        end
+
+        def metric_name(name, payload)
+          controller_path = controller_name_for_metric(payload)
+          "Ruby/ActionController#{"/#{controller_path}" if controller_path}/#{name.gsub(/\.action_controller/, '')}"
```
fallwith, line 33: Conditionally including the extra `/` delimiter is very slick! One consideration here - there might be some confusion between the `controller_name_for_metric` method name saying "name" and the variable holding its result saying "path".

lib/new_relic/agent/instrumentation/notifications_subscriber.rb
```diff
@@ -47,6 +47,30 @@ def self.subscribe(pattern)
...
+          log_notification_error(e, name, 'finish')
+        end
+
+        def start_segment(name, id, payload)
+          raise NotImplementedError
+        end
+
+        def finish_segment(name, id, payload)
+          raise NotImplementedError
+        end
```
fallwith, line 72: This is a great cleanup, thank you.

lib/new_relic/agent/instrumentation/rails_notifications/action_controller.rb
```diff
@@ -29,5 +30,14 @@
...
       .subscribe(/^process_action.action_controller$/)
+
+    subs = %w[send_file
+      send_data
+      redirect_to
+      halted_callback
+      unpermitted_parameters]
+
+    NewRelic::Agent::Instrumentation::ActionControllerOtherSubscriber \
+      .subscribe(Regexp.new("^(#{subs.join('|')}).action_controller$"))
```
fallwith, line 41: The dynamic "OR"ing of the event names is neat to see! The literal period here should be escaped (`.` -> `\.`) so that the pattern will only allow a literal period and not any character.

fallwith (changes requested):

lib/new_relic/agent/instrumentation/rails_notifications/action_controller.rb
tannalynn: oh yeah, great point! I ended up having to double escape it though because it's starting as a string first, and then gets converted into a regex. 
53639679fc7b9da07442ed96b68449c2dbe983a7

lib/new_relic/agent/instrumentation/action_controller_other_subscriber.rb
tannalynn: Updated in 53639679fc7b9da07442ed96b68449c2dbe983a7

lib/new_relic/agent/instrumentation/action_controller_other_subscriber.rb
tannalynn: What you linked to is the payload you get for process_action, which is handled by the original subscriber class already. Unfortunately, these payloads don't have the same stuff in them. 
Also, it should already be a string at this point, so .name wouldn't work, but calling to_s is redundant, so i just removed it instead. 
fbccb6e6d302c3e42a88e15f3a556399944a795b

lib/new_relic/agent/instrumentation/notifications_subscriber.rb
kaylareopelle, line 72: Could you add tests for the "NotImplementedErrors"? Simplecov reports them as untested.

lib/new_relic/agent/instrumentation/rails_notifications/action_controller.rb
fallwith: oh interesting - it's specifically the interpolation of the double quoted string that chomps up the single backslash if not prefixed by an additional backslash:

```ruby
irb(main):001:0> Regexp.new('a\.test')
=> /a\.test/
irb(main):002:0> Regexp.new("a\.test")
=> /a.test/
irb(main):003:0> Regexp.new("a\\.test")
=> /a\.test/
irb(main):004:0> /a\.test/
=> /a\.test/
```

Also on the topic of regexes for notification subscriptons, Kayla and I have been talking about standardizing on `\A` and `\z` instead of `^` and `$` and I'd also like to see us standardize on informing Ruby's regex engine that we don't actually want to capture the ORed list of words (so `(#{subs.join('|')})` would become `(?:#{subs.join('|')})`. But these 2 refactors can go in later post milestone.

fallwith (dismissed):

kaylareopelle (changes requested):

kaylareopelle (approved):
Thank you, Tanna! Great work!

lib/new_relic/agent/instrumentation/notifications_subscriber.rb
```diff
@@ -47,6 +47,30 @@ def self.subscribe(pattern)
...
+          finish_segment(id, payload)
+        rescue => e
+          log_notification_error(e, name, 'finish')
+        end
+
+        def start_segment(name, id, payload)
+          raise NotImplementedError
+        end
+
+        def finish_segment(name, id, payload)
```
hannahramadan, line 70: Looks like `finish_segment` doesn't ever call `name`, so it could remove as an arg!

kaylareopelle (dismissed):

fallwith (approved):

fallwith (dismissed):

fallwith (approved):

test/new_relic/agent/instrumentation/rails/active_support_subscriber.rb
```diff
@@ -116,8 +114,6 @@ def test_super_operation_recorded_as_attribute_on_traces
           trace = last_transaction_trace
           tt_node = find_node_with_name(trace, "#{METRIC_PREFIX}#{DEFAULT_STORE}/read")
 
-          puts txn.segments.last.inspect
```
kaylareopelle, line 119: Thank you for cleaning up these puts statements! Sorry I forgot to remove them!

kaylareopelle (approved):
Great work, Tanna!

PR #1748, opened 2023-01-20
Subscribe to additional Action View topic

### LINKED ISSUE #1512: Rails notifications: subscribe to additional ActionView topics

hannahramadan:
- subscribe to `render_layout.action_view`
- add test for `render_layout.action_view`
- refactor ActionViewSubscriber's `start` and `finish` to accommodate for new `start` and `finish` methods inherited from NotificationsSubscriber

closes #1512

### DIFF
--- a/lib/new_relic/agent/instrumentation/action_view_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/action_view_subscriber.rb
@@ -61,12 +61,14 @@ module NewRelic
         RENDER_TEMPLATE_EVENT_NAME = 'render_template.action_view'.freeze
         RENDER_PARTIAL_EVENT_NAME = 'render_partial.action_view'.freeze
         RENDER_COLLECTION_EVENT_NAME = 'render_collection.action_view'.freeze
+        RENDER_LAYOUT_EVENT_NAME = 'render_layout.action_view'.freeze
 
         def metric_action(name)
           case name
           when /#{RENDER_TEMPLATE_EVENT_NAME}$/o then 'Rendering'
           when RENDER_PARTIAL_EVENT_NAME then 'Partial'
           when RENDER_COLLECTION_EVENT_NAME then 'Partial'
+          when RENDER_LAYOUT_EVENT_NAME then 'Layout'
           end
         end
 
--- a/lib/new_relic/agent/instrumentation/action_view_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/action_view_subscriber.rb
@@ -66,6 +66,7 @@ module NewRelic
           when RENDER_PARTIAL_EVENT_NAME then 'Partial'
           when RENDER_COLLECTION_EVENT_NAME then 'Partial'
           when RENDER_LAYOUT_EVENT_NAME then 'Layout'
+          else 'Unknown'
           end
         end
 
--- a/lib/new_relic/agent/instrumentation/action_view_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/action_view_subscriber.rb
@@ -10,28 +10,25 @@ module NewRelic
   module Agent
     module Instrumentation
       class ActionViewSubscriber < NotificationsSubscriber
-        def start(name, id, payload) # THREAD_LOCAL_ACCESS
+        def start_segment(name, id, payload)
           parent = segment_stack[id].last
           metric_name = format_metric_name(name, payload, parent)
 
           event = ActionViewEvent.new(metric_name, payload[:identifier])
-          if state.is_execution_traced? && recordable?(name, metric_name)
+
+          if recordable?(name, metric_name)
             event.finishable = Tracer.start_segment(name: metric_name)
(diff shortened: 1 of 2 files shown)


test/new_relic/agent/instrumentation/rails/action_view_subscriber.rb
```diff
@@ -110,6 +110,23 @@ def test_records_metrics_for_simple_collection
...
+        :virtual_path => 'model/_layout')
+      advance_process_time(2.0)
+      @subscriber.finish('!render_layout.action_view', :id,
+        :virtual_path => 'model/_layout')
+      @subscriber.finish('render_layout.action_view', :id, params)
+    end
+    expected = {:call_count => 1, :total_call_time => 2.0}
+
+    assert_metrics_recorded('View/model/_layout.html.erb/Layout' => expected)
+  end
```
kaylareopelle, line 128: Could you walk me through this test? What I think I'm seeing is:
* pause the clock
* start two similar events, but one with a false name (because it leads with a `!`), for the same params
* advance the clock
* finish both events
* make sure only one of the events is recorded

Is that right, or is there something else going on too? Are only metrics supposed to be recorded for `render_layout`, not segments?

(It looks like a great test, I just want to make sure I'm reading it correctly!)

lib/new_relic/agent/instrumentation/action_view_subscriber.rb
```diff
@@ -10,28 +10,25 @@ module NewRelic
...
+          if recordable?(name, metric_name)
             event.finishable = Tracer.start_segment(name: metric_name)
           end
           push_segment(id, event)
-        rescue => e
-          log_notification_error(e, name, 'start')
         end
 
-        def finish(name, id, payload)
+        def finish_segment(id, payload)
```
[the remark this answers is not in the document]
kaylareopelle: We chatted about this and decided to make a recommendation on the parent branch to remove `name` from the parent class's version of `#finish_segment`.

test/new_relic/agent/instrumentation/rails/action_view_subscriber.rb
kaylareopelle: We chatted about this and the description above matches my understanding, except the events that start with `!` are the ones we want recorded! Descriptive comment available in [NewRelic::Agent::ActionViewSubscriber#recordable?](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/action_view_subscriber.rb#L51-L59)

kaylareopelle (approved):
Great work, Hannah!

lib/new_relic/agent/instrumentation/action_view_subscriber.rb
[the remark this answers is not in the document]
fallwith: So... `NotificationsSubcriber#finish_segment` does indeed have 3 argument method definition, but what really matters is how many arguments the caller passes. The calling method, `NotificationsSubscriber#finish` only passes 2 arguments.

I think this is an issue that @hannahramadan has identified in #1744 (cc @tannalynn).

We should do one of two things to `NotificationsSubscriber`:

- Pass `name` to `finish_segment`
- Change the base definition of `finish_segment` to indicate that only 2 arguments are passed in.

lib/new_relic/agent/instrumentation/action_view_subscriber.rb
```diff
@@ -61,12 +58,15 @@ def recordable?(event_name, metric_name)
...
         RENDER_COLLECTION_EVENT_NAME = 'render_collection.action_view'.freeze
+        RENDER_LAYOUT_EVENT_NAME = 'render_layout.action_view'.freeze
 
         def metric_action(name)
           case name
           when /#{RENDER_TEMPLATE_EVENT_NAME}$/o then 'Rendering'
           when RENDER_PARTIAL_EVENT_NAME then 'Partial'
           when RENDER_COLLECTION_EVENT_NAME then 'Partial'
+          when RENDER_LAYOUT_EVENT_NAME then 'Layout'
+          else 'Unknown'
```
fallwith, line 69: There are places where "Unknown" (capital U) are used, but for the subscriber stuff we've been consistently using "unknown" (lower case u). Perhaps this should be changed?

kaylareopelle: Great catch. I wonder if we should define the `UNKNOWN` constant in NotificationsSubscriber so we have the same UNKNOWN everywhere?

PR #1752, opened 2023-01-23
ActionMailer testing: always use a fresh mailer

fallwith:
The ActionMailer subscriber tests that perform work with ActionMailer::Base were occasionally flaking due to what appears to be problems with the reuse of the same mailer instance.

Now each test requiring a mailer will instantiate its own unique one, complete with its own unique instance variables. This appears to have completely eliminated the flaking, but just in case there is still an issue the underlying flake condition of attachments not being ready is now tested for and handled gracefully.


tannalynn (approved):
🎉

PR #1760, opened 2023-01-25
Make test output nicer

tannalynn:
**Hide warnings that we can't do anything about in tests**
This adds the warning gem to all of our test runs (unit, env, and multiverse) and will ignore all warnings that come from gem code that we can't control. This makes it way nicer and we will only see warnings from OUR code, that we can fix!

**Make file name output more compact for skipped tests**
Also, i changed the "Skipping tests in...." statements to output just the filename instead of the whole huge long path name.  it being super long was really annoying to me, personally, and made it harder to read/see what was going on. now it'll just do the file name without the whole path. 

**Added env variable VERBOSE_TEST_OUTPUT and hide a lot of test statements**
I also added an env variable check for a bunch of stuff in the test output. If VERBOSE_TEST_OUTPUT is not set, then the tests will not output a variety of statements including skipped files, test time info, full bundle info etc. If that env variable is set, then the test output will be exactly like it was before. This does NOT set the verbose flag for minitest. I think this will be nice for quick test runs where you're just trying to see if something is passing or not. 

**Removed verbose flag from CI test runs**
We way verbose was being passed to our CI runs wasn't working anyways, so I just removed them since we are used to it without verbosity. 

**Removed railsedge for now**
railsedge tests are failing after some recent rails commits for the parameter filtering because our `Rails.application.config.to_prepare` block doesn't appear to be getting called. I'll make an issue for this for us to check on this soon and see whats up with it.

### DIFF
--- a/test/multiverse/lib/multiverse/suite.rb
+++ b/test/multiverse/lib/multiverse/suite.rb
@@ -184,7 +184,7 @@ module Multiverse
       end
 
       result = red(result) unless $?.success?
-      puts result if ENV["VERBOSE_TEST_OUPUT"]
+      puts result if ENV["VERBOSE_TEST_OUTPUT"]
       $?
     end
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,7 @@ jobs:
           command: bundle exec rake test:env[${{ env.rails }}]
         env:
           DB_PORT: ${{ job.services.mysql.ports[3306] }}
-          VERBOSE_TEST_OUPUT: true
+          VERBOSE_TEST_OUTPUT: true
           COVERAGE: true
 
       - name: Save coverage results
@@ -233,7 +233,7 @@ jobs:
           max_attempts: 2
           command: bundle exec rake test:multiverse[group="${{ matrix.multiverse }}"]
         env:
-          VERBOSE_TEST_OUPUT: true
+          VERBOSE_TEST_OUTPUT: true
           MYSQL_PASSWORD: root
           DB_PASSWORD: root
           DB_PORT: ${{ job.services.mysql.ports[3306] }}
@@ -283,7 +283,7 @@ jobs:
           max_attempts: 2
           command: bundle exec rake test:multiverse[group=infinite_tracing]
         env:
-          VERBOSE_TEST_OUPUT: true
+          VERBOSE_TEST_OUTPUT: true
           SERIALIZE: 1
           COVERAGE: true
 
--- a/.github/workflows/ci_cron.yml
+++ b/.github/workflows/ci_cron.yml
@@ -119,7 +119,7 @@ jobs:
           max_attempts: 2
           command:  bundle exec rake test:env[${{ env.rails }}]
         env:
-          VERBOSE_TEST_OUPUT: true
+          VERBOSE_TEST_OUTPUT: true
           DB_PORT: ${{ job.services.mysql.ports[3306] }}
 
 
@@ -251,7 +251,7 @@ jobs:
           max_attempts: 2
           command:  bundle exec rake test:multiverse[group="${{ matrix.multiverse }}"]
         env:
-          VERBOSE_TEST_OUPUT: true
+          VERBOSE_TEST_OUTPUT: true
           DB_PASSWORD: root
           MYSQL_PASSWORD: root
           DB_PORT: ${{ job.services.mysql.ports[3306] }}
@@ -292,7 +292,7 @@ jobs:
           max_attempts: 2
           command:  bundle exec rake test:multiverse[group=infinite_tracing]
         env:
-          VERBOSE_TEST_OUPUT: true
+          VERBOSE_TEST_OUTPUT: true
           SERIALIZE: 1
 
       - name: Annotate errors
--- a/.github/workflows/ci_jruby.yml
+++ b/.github/workflows/ci_jruby.yml
@@ -44,7 +44,7 @@ jobs:
           max_attempts: 2
           command:  bundle exec rake test:env[norails,rails61,rails60]
         env:
-          VERBOSE_TEST_OUPUT: true
+          VERBOSE_TEST_OUTPUT: true
           DB_PORT: ${{ job.services.mysql.ports[3306] }}
           JRUBY_OPTS: --dev --debug
 
@@ -150,7 +150,7 @@ jobs:
         run: ./.github/workflows/scripts/retry_command 
         env: 
           TEST_CMD: "bundle exec rake test:multiverse[group=${{ matrix.multiverse }}]"
-          VERBOSE_TEST_OUPUT: true
+          VERBOSE_TEST_OUTPUT: true
           RETRY_ATTEMPS: 2
           SERIALIZE: 1
           DB_PORT: ${{ job.services.mysql.ports[3306] }}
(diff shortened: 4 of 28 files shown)


fallwith:
Given that our CI system currently has all verbosity ramped up, I think we should take one of these approaches:

- Only discard vendor error messages locally, but still include them in CI
- Stop being so verbose in CI. Layer in these changes and various others that decrease verbosity, perhaps with a toggle that can be used only when we actively wish to debug something.

tannalynn:
> Given that our CI system currently has all verbosity ramped up, I think we should take one of these approaches:
> 
> * Only discard vendor error messages locally, but still include them in CI
> * Stop being so verbose in CI. Layer in these changes and various others that decrease verbosity, perhaps with a toggle that can be used only when we actively wish to debug something.

I don't think we should keep warnings from gems in our CI. Since there is nothing we can do about these warnings and they aren't related to our code, I don't see what we gain by leaving them enabled in the CI. The main point of hiding these warnings it to decrease the amount of noise so that it's easier to see the output of tests and easier to notice the warnings from our codebase that we can actually fix. 

Also, what verbosity are you considering decreasing for the CI? I'd be down to change some of that, but I'm not sure what you mean. We don't run our any of our tests with verbosity enabled for minitest and the output of the tests in the CI is very similar to how looks when I run the tests locally. Could you clarify what you're referring to?

fallwith:
> Also, what verbosity are you considering decreasing for the CI? I'd be down to change some of that, but I'm not sure what you mean. We don't run our any of our tests with verbosity enabled for minitest and the output of the tests in the CI is very similar to how looks when I run the tests locally. Could you clarify what you're referring to?

The CI leverages both the `--verbose` MiniTest flag set via the `TESTOPTS` environment variable and it also leverages our own multiverse suite's `verbose` argument. Both of these represent overriding of the default behavior which is not as verbose. While I'm not familiar with the history of the CI verbosity being added, I can see how using transient resources like we do with CI would leave us unable to manually re-run things with the exact same setup and desire as much logging output as possible. That's my take on why the verbosity settings are present in the CI now, so I'm thinking we should either stick with that and always be verbose in CI or question it across the board.

tannalynn:
> The CI leverages both the `--verbose` MiniTest flag set via the `TESTOPTS` environment variable and it also leverages our own multiverse suite's `verbose` argument. Both of these represent overriding of the default behavior which is not as verbose. While I'm not familiar with the history of the CI verbosity being added, I can see how using transient resources like we do with CI would leave us unable to manually re-run things with the exact same setup and desire as much logging output as possible. That's my take on why the verbosity settings are present in the CI now, so I'm thinking we should either stick with that and always be verbose in CI or question it across the board.

Oooh, i see why I was confused. I never noticed it was trying to do that, but it sure looks like its trying to set verbose, but it actually doesn't work lmao. its because env variable has to be before the command. 
So it's 
```
bundle exec rake test:env[${{ env.rails }}] TESTOPTS="--verbose"
```
 right now, which doesn't actually set anything for the command. If we want it to be actually running verbose, we would need to change it to 
```
TESTOPTS="--verbose" bundle exec rake test:env[${{ env.rails }}]
```
I'd vote for not doing that though, because it's kind of annoying IMO to see all the time. the output looks like 
```
NewRelic::Agent::Agent::MethodTracer::ClassMethods::AddMethodTracerTest#test_validate_options_nonhash = 0.00 s = .
NewRelic::Agent::Agent::MethodTracer::ClassMethods::AddMethodTracerTest#test_check_for_illegal_keys_positive = 0.00 s = .
NewRelic::Agent::Agent::MethodTracer::ClassMethods::AddMethodTracerTest#test_newrelic_method_exists_negative = 0.00 s = .
NewRelic::Agent::Agent::MethodTracer::ClassMethods::AddMethodTracerTest#test_newrelic_method_exists_positive = 0.00 s = .
NewRelic::Agent::Agent::MethodTracer::ClassMethods::AddMethodTracerTest#test_check_for_push_scope_and_metric_negative = 0.00 s = .
NewRelic::Agent::MethodTracer::TraceExecutionScopedTest#test_trace_execution_scope_runs_passed_block_and_returns_its_value = 0.00 s = .
NewRelic::Agent::MethodTracer::TraceExecutionScopedTest#test_metric_recording_in_nested_transactions = 0.00 s = .
NewRelic::Agent::MethodTracer::TraceExecutionScopedTest#test_trace_execution_scoped_does_not_swallow_errors = 0.00 s = .
```
which it doesn't now which is why i was confused about it lol. 
As for multiverse, i'm not sure what passing verbose in is supposed to do? I ran it locally and the output looks the same with and without it. So either i'm not noticing what it's doing, or it isn't doing what it should be? Not sure. If it's supposed to be setting the minitest verbose stuff, it isn't also.

So I'd be down to remove both of those, since as far as I can tell, they aren't doing anything currently anyway lol. If that sounds good I can go ahead and remove those!

fallwith (approved):
Thanks for explaining and removing the verbosity stuff as well, @tannalynn. I seem to remember the multiverse's `verbose` setting triggering extra unrelated stuff such as bringing in the Pry gem - also not needed for CI.

fallwith (dismissed):

test/test_helper.rb
```diff
@@ -2,6 +2,17 @@
...
+if RUBY_VERSION >= '2.4.0'
+  require 'warning'
+  Gem.path.each do |path|
+    Warning.ignore(//, path)
+  end
+
+  # this is to ignore warnings that happen on the CI only
+  # the site_ruby part of the path needs to be removed if it exists otherwise the CI keeps doing the warnings
+  Warning.ignore(//, Gem::RUBYGEMS_DIR.gsub(/site_ruby\//, ''))
+end
```
kaylareopelle, line 14: You might be able to put this into a file similar to test/simplecov_test_helper.rb and require that file here and in `multiverse/suite.rb` to reduce duplication.

kaylareopelle (approved):
I like this Tanna, thanks for these changes. I have one suggestion to reduce duplication that you don't need to incorporate if you'd prefer not to.

test/test_helper.rb
tannalynn: good idea! I added that in 062b1c77308e6d16bc8c6353fefe4a2520ca4276

test/multiverse/lib/multiverse/suite.rb
```diff
@@ -387,10 +391,13 @@ def execute_child_environment(env_index, instrumentation_method)
         log_test_running_process
         configure_before_bundling
 
+        # require_relative('../../../warning_test_helper')
```
kaylareopelle, line 394: whoops, I think this got left behind!

tannalynn: oh good catch, my bad lol. fixed! 33b68e474abe813c3091a4ab9d703d64b197e5ad

kaylareopelle (approved):

fallwith (commented):
Looking at `MIN_TEST_OUTPUT`, 2 thoughts come to mind:

- It reads really similar to `MINI_TEST` 😄 
- I wonder if this behavior should be the default, and only those contexts (possibly CI) would see the output.

We could address both of these with something like `VERBOSE_TEST_OUTPUT` that does the opposite of what's being done now.

fallwith (dismissed):
Left a suggestion to consider flipping the MIN_TEST_OUTPUT logic around and only be verbose when an env var is set instead of only being quiet when one is set.

But that's not a request for change. If you've given it thought and like the way it is now, that's fine with me. Approved.

fallwith (approved):

PR #1761, opened 2023-01-25
Active Job notifications subscriber

### LINKED ISSUE #1515: Rails notifications: subscribe to ActiveJob topics
Rails will publish content to the following 4 ActiveJob topics:

- `enqueue_at.active_job`
- `enqueue.active_job`
- `enqueue_retry.active_job`
- `perform.active_job`

We should subscribe to these in a similar manner used for `*.action_controller` and `*.action_view topics`.

**NOTE:** we currently instrument ActiveJob without taking advantage of subscriptions. Make sure to not collide / overlap with that existing instrumentation.

fallwith:
Subscribe to Active Job notifications and process the 7 (at the time of Rails v7.1.0.0.alpha) related events to produce additional instrumentation for the library.

resolves #1515

### DIFF
--- a/lib/new_relic/agent/instrumentation/active_job.rb
+++ b/lib/new_relic/agent/instrumentation/active_job.rb
@@ -12,7 +12,7 @@ DependencyDetection.defer do
   end
 
   executes do
-    NewRelic::Agent.logger.info('Installing ActiveJob instrumentation')
+    NewRelic::Agent.logger.info('Installing base ActiveJob instrumentation')
 
     ActiveSupport.on_load(:active_job) do
       ActiveJob::Base.around_enqueue do |job, block|
@@ -24,6 +24,19 @@ DependencyDetection.defer do
       end
 
       NewRelic::Agent::PrependSupportability.record_metrics_for(ActiveJob::Base)
+    end
+  end
+
+  executes do
+    if defined?(ActiveSupport) &&
+        ActiveJob.respond_to?(:gem_version) &&
+        ActiveJob.gem_version >= Gem::Version.new('6.0.0') &&
+        !NewRelic::Agent.config[:disable_activejob] &&
+        !NewRelic::Agent::Instrumentation::ActiveJobSubscriber.subscribed?
+      NewRelic::Agent.logger.info('Installing notifications based ActiveJob instrumentation')
+
+      ActiveSupport::Notifications.subscribe(/\A[^\.]+\.active_job\z/,
+        NewRelic::Agent::Instrumentation::ActiveJobSubscriber.new)
     end
   end
 end
--- a/lib/new_relic/agent/instrumentation/active_job_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/active_job_subscriber.rb
@@ -10,22 +10,9 @@ module NewRelic
       class ActiveJobSubscriber < NotificationsSubscriber
         PAYLOAD_KEYS = %i[adapter db_runtime error job wait]
 
-        def start_segment(name, id, payload)
-          segment = Tracer.start_segment(name: metric_name(name, payload))
-
+        def add_segment_params(segment, payload)
           PAYLOAD_KEYS.each do |key|
             segment.params[key] = payload[key] if payload.key?(key)
-          end
-
-          push_segment(id, segment)
-        end
-
-        def finish_segment(id, payload)
-          if segment = pop_segment(id)
-            if exception = exception_object(payload)
-              segment.notice_error(exception)
-            end
-            segment.finish
           end
         end
 
@@ -36,13 +23,12 @@ module NewRelic
         end
 
         PATTERN = /\A([^\.]+)\.active_job\z/
-        UNKNOWN = 'unknown'.freeze
 
         METHOD_NAME_MAPPING = Hash.new do |h, k|
           if PATTERN =~ k
             h[k] = $1
           else
-            h[k] = UNKNOWN
+            h[k] = NewRelic::UNKNOWN
           end
         end
 
--- a/lib/new_relic/agent/instrumentation/active_job_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/active_job_subscriber.rb
@@ -1,0 +1,55 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'new_relic/agent/instrumentation/notifications_subscriber'
+
+module NewRelic
+  module Agent
+    module Instrumentation
+      class ActiveJobSubscriber < NotificationsSubscriber
+        PAYLOAD_KEYS = %i[adapter db_runtime error job wait]
+
+        def start_segment(name, id, payload)
+          segment = Tracer.start_segment(name: metric_name(name, payload))
+
+          PAYLOAD_KEYS.each do |key|
+            segment.params[key] = payload[key] if payload.key?(key)
+          end
+
+          push_segment(id, segment)
+        end
+
+        def finish_segment(id, payload)
+          if segment = pop_segment(id)
+            if exception = exception_object(payload)
+              segment.notice_error(exception)
+            end
+            segment.finish
+          end
+        end
+
+        def metric_name(name, payload)
+          queue = payload[:job].queue_name
+          method = method_from_name(name)
+          "Ruby/ActiveJob/#{queue}/#{method}"
+        end
+
+        PATTERN = /\A([^\.]+)\.active_job\z/
+        UNKNOWN = 'unknown'.freeze
+
+        METHOD_NAME_MAPPING = Hash.new do |h, k|
+          if PATTERN =~ k
+            h[k] = $1
+          else
+            h[k] = UNKNOWN
+          end
+        end
+
+        def method_from_name(name)
+          METHOD_NAME_MAPPING[name]
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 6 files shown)


test/simplecov_test_helper.rb
```diff
@@ -7,4 +7,11 @@
 # required Ruby >= 2.5.0 and Ruby 2.6.0 was marked for EOL
 SIMPLECOV_MIN_RUBY_VERSION = '2.7.0'
 
-require 'simplecov' if RUBY_VERSION >= SIMPLECOV_MIN_RUBY_VERSION
```
fallwith, line 10: Currently for some tests with some Rubies, we hit a "file not found" error, which can be puzzling because it's not obvious that there's a LoadError involved (no backtrace is provided) and passing things such as `--trace` won't reach the subshell used for the env tests.

To repro, simply use Ruby 2.7.5 to perform `bundle exec rake test:env[rails52]`. It's perfectly reasonable that Ruby 2.7.5 might be used in conjunction with Rails 5.2. But while Ruby 2.7+ will satisfy the SimpleCov helper check, the Rails 5.2 test env `Gemfile` does not specify `simplecov`.

Now we simply rescue and report load errors but don't let them stop the tests from running.

lib/new_relic/agent/instrumentation/active_job.rb
```diff
@@ -26,6 +26,19 @@
...
     end
   end
+
+  executes do
+    if defined?(ActiveSupport) &&
+        ActiveJob.respond_to?(:gem_version) &&
+        ActiveJob.gem_version >= Gem::Version.new('6.0.0') &&
+        !NewRelic::Agent.config[:disable_activejob] &&
+        !NewRelic::Agent::Instrumentation::ActiveJobSubscriber.subscribed?
+      NewRelic::Agent.logger.info('Installing notifications based ActiveJob instrumentation')
```
kaylareopelle, line 36: I apologize if this pattern has been around for a few PRs and I'm just noticing it. Why did you choose to put an `if` statement inside the `executes` block, instead of using a `depends_on` block? Since we have two different types of instrumentation for Active Job, should the registration of the instrumentation be in different files?

lib/new_relic/agent/instrumentation/active_job_subscriber.rb
```diff
@@ -0,0 +1,55 @@
...
+        end
+
+        def finish_segment(id, payload)
+          if segment = pop_segment(id)
+            if exception = exception_object(payload)
+              segment.notice_error(exception)
+            end
+            segment.finish
+          end
+        end
```
kaylareopelle, line 30: This method looks very similar to the one in `NotificationsSubscriber`. Do we need it here too?

lib/new_relic/agent/instrumentation/active_job_subscriber.rb
```diff
@@ -0,0 +1,55 @@
...
+
+        def metric_name(name, payload)
+          queue = payload[:job].queue_name
+          method = method_from_name(name)
+          "Ruby/ActiveJob/#{queue}/#{method}"
+        end
+
+        PATTERN = /\A([^\.]+)\.active_job\z/
+        UNKNOWN = 'unknown'.freeze
```
kaylareopelle, line 39: Should we use the new constant here? `NewRelic::UNKNOWN`?

test/new_relic/agent/instrumentation/rails/active_job_subscriber.rb
```diff
@@ -0,0 +1,201 @@
...
+
+    def test_start_when_not_traced
+      SUBSCRIBER.state.stub :is_execution_traced?, false do
+        in_transaction do |txn|
+          SUBSCRIBER.start(NAME, ID, {})
+
+          assert_empty txn.segments
+        end
+      end
+    end
```
kaylareopelle, line 48: I think this might be covered with Tanna's new notifications subscriber tests

test/new_relic/agent/instrumentation/rails/active_job_subscriber.rb
```diff
@@ -0,0 +1,201 @@
...
+
+    # TODO: test for retry_stopped
+    # retry_stopped.active_job
+    # def test_an_actual_job_event_retry
+    #   in_transaction do |txn|
+    #     # ???
+    #     validate_transaction(txn, 'retry_stopped')
+    #   end
+    # end
```
kaylareopelle, line 183: Do we still need this?

test/new_relic/agent/instrumentation/rails/active_job_subscriber.rb
```diff
@@ -0,0 +1,201 @@
...
+        in_transaction do |txn|
+          SUBSCRIBER.stub :start_segment, -> { raise 'kaboom' } do
+            SUBSCRIBER.start(NAME, ID, {})
+          end
+
+          assert_equal 1, txn.segments.size
+        end
+      end
+      logger.verify
+    end
```
kaylareopelle, line 66: I think this one may also be a dupe of the notifications subscriber tests.

lib/new_relic/agent/instrumentation/active_job.rb
fallwith: Thanks for taking a look. You didn't miss this pattern anywhere else, and it's definitely representative of a compromise that bears further discussion. What we have here is:

- Existing ActiveJob instrumentation with existing `depends_on` checks
- New functionality that requires new `depends_on` type checks
- The new functionality still requires all of the existing `depends_on` checks
- The new functionality should use the existing config param
- The existing functionality is usable for older Rails apps, but the new stuff requires v6+

I had a choice to create a second instrumentation file. The drawback there is that I would have to either have the new file reference the existing one or create a third shared file or module that both instrumentation halves reference to be able to share the same checks and config setting.

Given that I wanted all of this functionality to collectively represent ActiveJob instrumentation that is all on or all off, and how small the new instrumentation's checks are, I made the call to put a conditional inside of an `executes` block.

I'm open to feedback on using a separate file, using a separate config option, and anything else.

lib/new_relic/agent/instrumentation/active_job_subscriber.rb
fallwith: The work on this PR began before `NotificationsSubscriber` delivered this method and I missed that. I knew about `start` and `finish`, but not `start_segment` and `finish_segment`. Will update.

test/new_relic/agent/instrumentation/rails/active_job_subscriber.rb
fallwith: It's the 1 out of 7 that we don't have a complete integration style test for. From reviewing the ActiveJob code to see how it works, I'm comfortable that our existing instrumentation will process the events just fine. But given that the test file has 6 other individual tests with event name comment labels, I figured I'd leave the seventh behind to communicate that we know about it. Happy to remove and ignore it and/or create a backlogged story for it if you'd prefer one of those other approaches.

test/new_relic/agent/instrumentation/rails/active_job_subscriber.rb
fallwith: Addressed with [6e2aaf5](https://github.com/newrelic/newrelic-ruby-agent/pull/1761/commits/6e2aaf5f00b3447670f68fe508a2d7c52bd6263d)

test/new_relic/agent/instrumentation/rails/active_job_subscriber.rb
fallwith: Updated the TODO comment so that it no longer looks like a forgotten bit of coding.

Addressed with [6e2aaf5](https://github.com/newrelic/newrelic-ruby-agent/pull/1761/commits/6e2aaf5f00b3447670f68fe508a2d7c52bd6263d)

test/new_relic/agent/instrumentation/rails/active_job_subscriber.rb
fallwith: Addressed with [6e2aaf5](https://github.com/newrelic/newrelic-ruby-agent/pull/1761/commits/6e2aaf5f00b3447670f68fe508a2d7c52bd6263d)

lib/new_relic/agent/instrumentation/active_job_subscriber.rb
fallwith: Addressed with [6e2aaf5](https://github.com/newrelic/newrelic-ruby-agent/pull/1761/commits/6e2aaf5f00b3447670f68fe508a2d7c52bd6263d)

lib/new_relic/agent/instrumentation/active_job_subscriber.rb
fallwith: Addressed with [6e2aaf5](https://github.com/newrelic/newrelic-ruby-agent/pull/1761/commits/6e2aaf5f00b3447670f68fe508a2d7c52bd6263d)

lib/new_relic/agent/instrumentation/active_job.rb
kaylareopelle: Thanks for breaking down the compromise. I think this works here and can be the topic of discussion in some other forum. We've talked before about standardizing the number of depends_on/executes blocks and I think this is a good case to consider when we have that discussion.

lib/new_relic/agent/instrumentation/active_job_subscriber.rb
kaylareopelle: Thank you

test/new_relic/agent/instrumentation/rails/active_job_subscriber.rb
kaylareopelle: Thanks, James. Could you create a story and add it to the Improve Rails Instrumentation 2.0 milestone?

kaylareopelle (approved):
Thanks for the changes, James!

test/new_relic/agent/instrumentation/rails/active_job_subscriber.rb
fallwith: Sure thing. Created #1764

kaylareopelle: Thank you!

PR #1766, opened 2023-01-26
wire up disable_ config params for notifications

fallwith:
Make sure that each ActiveSupport notifications based instrumentation class respects its `disable_` config parameter


kaylareopelle (dismissed):
Thanks for your proactivity, James.

tannalynn (approved):
🎉

PR #1770, opened 2023-01-26
Fix active job 6.0 failures when retry job

tannalynn:
# Overview
rails 6.0 seems to have issues retrying a job if the arg passed in is a class. it raises the error `ActiveJob::SerializationError: Unsupported argument type: Class`. However it's totally fine and works well if you aren't attempting to retry the job. It also works fine on 6.1+.
This works around the issue by using a string instead if on rails 6.0.

I also noticed some warnings when running the env tests because the warning helper was being called later when starting the env tests. I added it to the env test rake files so it will start working immediately instead. 

passing all: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/4019560610

### DIFF
--- a/test/environments/norails/Rakefile
+++ b/test/environments/norails/Rakefile
@@ -5,6 +5,8 @@
 # Add your own tasks in files placed in lib/tasks ending in .rake,
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
+require_relative '../../warning_test_helper'
+
 require 'rake'
 require 'tasks/all'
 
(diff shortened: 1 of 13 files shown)


fallwith (approved):
Looks great, thank you for getting to the bottom of this and implementing such a clean fix!

I see there's some warning suppression additions that went into this PR. That's totally fine, but please (in future) drop a quick byline in the PR description calling attention to it. Sometimes devs working across multiple changes will accidentally commit the wrong content, and it's helpful being able to see in the PR description that a change was intentional.

tannalynn:
> Looks great, thank you for getting to the bottom of this and implementing such a clean fix!
> 
> I see there's some warning suppression additions that went into this PR. That's totally fine, but please (in future) drop a quick byline in the PR description calling attention to it. Sometimes devs working across multiple changes will accidentally commit the wrong content, and it's helpful being able to see in the PR description that a change was intentional.

For sure, sounds good! I actually did mention it in the description, in the second section. I'll try to bring more attention to that kind of thing a little better in the future though!

PR #1780, opened 2023-02-01
Add vale to PR workflow

kaylareopelle:
The docs-website uses this linter in its CI.
Since its introduction, we've needed to update change log entries after the fact.
Let's add it to our PR workflow, in hopes that we can preemptively address its feedback.

https://vale.sh/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,14 @@ jobs:
       - run: bundle
       - run: rubocop
 
+  vale:
+    name: vale linter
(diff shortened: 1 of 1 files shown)


kaylareopelle:
There's a lot of repo-specific config files that would need to stay in sync with the docs-website. I don't think this is ready to be added at this time. I opened an issue on the docs-website requesting a custom GitHub action with the styles, ini file, etc. for distribution to NR teams.

https://github.com/newrelic/docs-website/issues/11655

PR #1781, opened 2023-02-01
Add test running script and update test readme

tannalynn:
- I turned my bash functions into a script! It should make running tests simple and consistent across all the test types. Using the script, now we can run the unit, env, and multiverse tests basically the same way. This will help prevent the issue of forgetting which tests require which way to specify a test name or file.
- Added instructions to the test readme on using the test script and setting up aliases in your .bash_profile or .zshenv files. 
- Also updated the test readme to include a better way to specify test file and test name for the unit and env tests that works with both, instead of the previously documented way that doesn't. I did leave the old way at the bottom of the file in case we wanted it in the future, although I can remove it if that is preferred.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -28,6 +28,8 @@ add these aliases in ~/.bash_profile (or ~/.zshenv, depending on what shell you 
     alias bermq="./test/script/run_tests -q"
     alias ber="./test/script/run_tests"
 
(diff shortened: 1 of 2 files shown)


fallwith (approved):

kaylareopelle (approved):
🦄

Issue #957, opened 2022-02-14
Remove all previously deprecated configuration options


tannalynn:
In the next major release 9.0.0 of the agent, we should remove all configuration options marked with `deprecated => true` in `default_source.rb`. The replacement configuration options that currently reference these deprecated options should have their default values updated to no longer reference those options as well.

tannalynn:
For our next major release, leave the deprecated configuration options referencing analytic events, as those have been recently deprecated. All other deprecated configuration options should be able to be removed.

kaylareopelle:
Some of these deprecated attributes are called out as such in [this document](https://docs.newrelic.com/docs/apm/agents/ruby-agent/attributes/enable-disable-attributes-ruby/). Let's update it when we work on this ticket.

PR #1782, opened 2023-02-01
Remove all previously deprecated configuration options

tannalynn:
Removes all deprecated config options except `cross_application_tracing.enabled`, as CAT is still deprecated but not being removed yet.

I had to update several tests to use the new configs, and also deleted a bunch of tests that were testing to make sure the deprecated (and now removed) configs were still working. 

[all ci run](https://github.com/newrelic/newrelic-ruby-agent/actions/runs/4087048207)
closes #https://github.com/newrelic/newrelic-ruby-agent/issues/957


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -370,12 +370,12 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
   - a.third.event
           DESCRIPTION
         },
+        # this is only set via server side config
         :apdex_t => {
           :default => 0.5,
-          :public => true,
+          :public => false,
```
tannalynn, line 376: So it turns out this config is only deprecated for use in the newrelic.yml and env variables. This should only be set via server side config.
The agent however, still uses this value for things so we need to keep the config. So I've removed the deprecated tag and instead have made this a non public config option.

test/new_relic/agent/javascript_instrumentor_test.rb
```diff
@@ -158,7 +158,7 @@ def test_browser_timing_header_safe_when_json_dump_fails
 
   def test_config_data_for_js_agent
     nr_freeze_process_time
-    with_config(CAPTURE_ATTRIBUTES => true) do
+    with_config(ATTRIBUTES_ENABLED => true) do
```
tannalynn, line 161: these tests were all using the old config. Other tests in this file were already using the new config as `ATTRIBUTES_ENABLED`, so these tests were just updated to use that now

lib/new_relic/agent/attribute_filter.rb
```diff
@@ -124,8 +124,6 @@ def prep_attributes_exclude_rules(config)
 
       def prep_capture_params_rules(config)
         build_rule(['request.parameters.*'], include_destinations_for_capture_params(config[:capture_params]), true)
-        build_rule(['job.resque.args.*'], include_destinations_for_capture_params(config[:'resque.capture_params']), true)
```
tannalynn, line 127: These attributes are no longer treated special and separate, they now work exactly like any other attribute given to the `attribute.include` config.

lib/new_relic/agent/configuration/high_security_source.rb
```diff
@@ -11,8 +11,6 @@ class HighSecuritySource < DottedHash
         def initialize(local_settings)
           super({
             :capture_params => false,
-            :'resque.capture_params' => false,
-            :'sidekiq.capture_params' => false,
```
tannalynn, line 15: These are both now covered by the below line
```
:'attributes.include' => []
```

lib/new_relic/agent/instrumentation/grape.rb
```diff
@@ -9,7 +9,7 @@
 DependencyDetection.defer do
   # Why not just :grape? newrelic-grape used that name already, and while we're
   # not shipping yet, overloading the name interferes with the plugin.
-  named :grape_instrumentation
+  @name = :grape_instrumentation
```
tannalynn, line 12: All of the instrumentation files that had `named` changed to `@name=` is due to a test we have that automatically checks to make sure all files using `named` have a config that matches the provided name. Since the the config for grape is `instrumentation.grape`, the test was failing because it was checking for `instrumentation.grape_instrumentation`.

test/new_relic/agent/attribute_filter_test.rb
```diff
@@ -94,16 +94,24 @@ def test_resque_capture_params_false_adds_exclude_rule_for_request_parameters
     end
 
     def test_resque_capture_params_true_allows_request_params_for_traces_and_errors
-      with_config(:'resque.capture_params' => true) do
+      with_config(:'attributes.include' => ['job.resque.args.*']) do
         filter = AttributeFilter.new(NewRelic::Agent.config)
         result = filter.apply('job.resque.args.*', AttributeFilter::DST_NONE)
 
-        assert_destinations %w[transaction_tracer error_collector], result
+        expected_destinations = %w[
```
tannalynn, line 101: Expected destinations changed because the deprecated config was super old and didn't work with events

test/new_relic/agent/configuration/security_policy_source_test.rb
```diff
@@ -177,33 +177,10 @@ def test_message_parameters_disabled
           end
         end
 
-        def test_job_arguments_enabled
```
tannalynn, line 180: Deleted because these tests are only testing to make sure the new removed configs worked correctly. Since it is now treated like any other attribute, the tests for `attribute.include` now cover these situations.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -713,21 +690,7 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
-          :deprecated => true,
-          :allowed_from_server => true,
-          :dynamic_name => true,
-          :description => <<~DESCRIPTION
-            Use `error_collector.ignore_classes` instead. Specify a comma-delimited list of error classes that the agent should ignore.
 
-                <Callout variant="caution">
-                  Server side configuration takes precedence for this setting over all environment configurations. This differs from all other configuration settings where environment variable take precedence over server side configuration.
-                </Callout>
-          DESCRIPTION
```
kaylareopelle, line 729: Does `error_collector.ignore_classes` have a similar SSC priority that we should call out in the description of `error_collector.ignore_classes?

kaylareopelle (approved):
Your comments within the PR were very helpful, @tannalynn. I have one small question, but it can be disregarded.

fallwith (approved):

lib/new_relic/agent/configuration/default_source.rb
tannalynn: No, that shouldn't affect ignore_classes. The only config source higher than server side config is env vars, but ignore_classes is not accepted via env vars, so server side config is highest precedence anyways

kaylareopelle: Awesome, thanks!

kaylareopelle (approved):
Approved! Feel free to put change log stuff in a separate PR :)

tannalynn:
> Approved! Feel free to put change log stuff in a separate PR :)

Great idea, I'll do that!

PR #1784, opened 2023-02-02
disable, make private ActionDispatch notifications

fallwith:
the ActiveSupport event notifications based instrumentation for ActionDispatch overlaps too greatly with our existing Rails middleware instrumentation, leading to confusing and potentially costly overlap for Rails users.

ideally, we'd like to standardize on leveraging Rails notifications for all Rails instrumentation and remove any custom Rails instrumentation that overlaps. But in this case with ActionDispatch the notifications aren't available until Rails v6, and we aren't prepared at this time to stop supporting earlier Rails versions.

so for now, retain the notifications based instrumentation for ActionDispatch but have it configured - privately - to be disabled by default.


tannalynn (approved):
ty! 🎉 
we'll want to make sure we update the docs PR as well

kaylareopelle:
Slack thread for reference: https://newrelic.slack.com/archives/GGXADQ9F1/p1675288101710279

kaylareopelle (approved):
Thank you, James!

PR #1790, opened 2023-02-03
Enable defer_rails_initialization config by environment variable only

kaylareopelle:
# Overview 
Configs are loaded during initialization. This fix was placed behind a feature flag referencing the config value, but that value will always be false at initialization, so the bug fix was unreachable.

Environment variables are accessible before initialization. The environment variable `NEW_RELIC_DEFER_RAILS_INITIALIZATION` may be set to true to enable this option.

TODO:
- [ ] Fix unit tests 

Resolves: #662


kaylareopelle:
Opened on this branch by mistake

PR #1796, opened 2023-02-06
Removed configuration changelog

tannalynn:
This adds the changelog entry listing all removed public configuration options and provides the replacement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,8 @@
   | `sidekiq.capture_params`                  | `attributes.include`                      | `attributes.include: ['job.sidekiq.args.*']`                                        |
   | `transaction_tracer.capture_attributes`   | `transaction_tracer.attributes.enabled`   | `transaction_tracer.attributes.enabled: false`                                      |
   | `error_collector.ignore_errors`           | `error_collector.ignore_classes`          | `error_collector.ignore_classes: ['ActionController::RoutingError', 'CustomError']` |
-  | `analytics_events.enabled`                 | `transaction_events.enabled`              | `transaction_events.enabled: false`                                                 |
-  | `analytics_events.max_samples_stored`      | `transaction_events.max_samples_stored`   | `transaction_events.max_samples_stored: 1200`                                       |
+  | `analytics_events.enabled`                | `transaction_events.enabled`              | `transaction_events.enabled: false`                                                 |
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,55 @@
...
+  | `analytics_events.capture_attributes`     | `transaction_events.attributes.enabled`   | `transaction_events.attributes.enabled: false`                                      |
+  | `resque.capture_params`                   | `attributes.include`                      | `attributes.include: ['job.resque.args.*']`                                         |
+  | `sidekiq.capture_params`                  | `attributes.include`                      | `attributes.include: ['job.sidekiq.args.*']`                                        |
+  | `error_collector.ignore_errors`           | `error_collector.ignore_classes`          | `error_collector.ignore_classes: ['ActionController::RoutingError', 'CustomError']` |
+  | `analytics_events.enabled                 | `transaction_events.enabled`              | `transaction_events.enabled: false`                                                 |
+  | `analytics_events.max_samples_stored      | `transaction_events.max_samples_stored`   | `transaction_events.max_samples_stored: 1200`                                       |
```
fallwith, line 22:
```suggestion
  | `analytics_events.max_samples_stored`      | `transaction_events.max_samples_stored`   | `transaction_events.max_samples_stored: 1200`                                       |
```

fallwith (commented):
Perhaps we should sort the table based on an alphabetical listing of the "removed" values.

tannalynn:
> Perhaps we should sort the table based on an alphabetical listing of the "removed" values.

I had that at first, but I ended up grouping them by similar function instead since it seemed to make more sense to me. So now it's attributes > ignore > events > instrumentation, where before the attribute related ones were kind of sprinkled through the whole thing which I thought wasn't great. I did just update it to reorganize them so they are alphabetized better in each section though. How does that look?

fallwith:
> How does that look?

Looks great! The only ones that stand out now are the `dalli` ones. Are those supposed to come before `dj` or are they where they are now for deliberate grouping with `memcached`?

tannalynn:
> > How does that look?
> 
> Looks great! The only ones that stand out now are the `dalli` ones. Are those supposed to come before `dj` or are they where they are now for deliberate grouping with `memcached`?

Oh yeah, I forgot I did that because it was last week, but I did specifically group them with the memcache ones since it's all the same instrumentation, but I'm open to changing that if you think it would make more sense.

fallwith (approved):

fallwith:
> I did specifically group them

Sounds great! Approved.

PR #1797, opened 2023-02-06
Allow `set_transaction_name` to change name and category of current transaction.

### LINKED ISSUE #1064: set_transaction_name does not apply :task as a category
## Description 
The public API method `set_transaction_name` allows the user to optionally pass in a `:category` argument with the values of either `:controller`, `:task`, `:middleware`, or `:uri`. These values map to overarching categories for transactions.

We have a customer who tried to use this method with the `:task` category, but did not have their transactions successfully assigned as Non-Web/Background tasks. 

A refactor to the method `NewRelic::Agent::Transaction#set_overriding_transaction_name` made it so it would not be possible for any transaction pre-categorized as a web transaction to be overridden to be a non-web transaction.

## Expected Behavior 
Users of the public API should be able to pass in the `category: :task` argument to `set_transaction_name` and have the transaction be re-categorized as a Non-Web or Background transaction. This change should not re-categorize the parent transaction (if there is one).

[Internal Reference: GTSE-14114](https://newrelic.atlassian.net/browse/GTSE-14114)

## T-Shirt size

tannalynn:
This updates how `set_transaction_name` works so it functions more as expected and accomplishes what customers have expected it to do. Now, when set_transaction_name is called with a different category that is already assigned, the category and name will be changed to what is passed in. Previously, nothing would happen if the category was different. 

The `set_transaction_name` api docs did not need to be updated because the way it is worded already implies that this is how it was supposed to work in the first place. 

closes #1064

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,7 +73,7 @@
 
 - **Changes how the API method `set_transaction_name` works**
 
-  When the method `NewRelic::Agent.set_transaction_name` is called, it will now always change the name and category of the currently running transaction to what is passed in to the method. This is a change from how it functioned in previous agent versions. Previously, if set_transaction_name was called with a new transaction name and a new category that did not match the category that was already assignd to a transaction, neither the new name nor category would be saved to the transaction. 
+  When the method `NewRelic::Agent.set_transaction_name` is called, it will now always change the name and category of the currently running transaction to what is passed in to the method. This is a change from how it functioned in previous agent versions. Previously, if `set_transaction_name` was called with a new transaction name and a new category that did not match the category that was already assigned to a transaction, neither the new name nor category would be saved to the transaction. 
 
 
 - **Dropped method: `NewRelic::Agent.disable_transaction_tracing`**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v9.0.0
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -70,6 +70,12 @@
     - Rainbows
     - Sunspot
 
+
+- **Changes how the API method `set_transaction_name` works**
+
+  When the method `NewRelic::Agent.set_transaction_name` is called, it will now always change the name and category of the currently running transaction to what is passed in to the method. This is a change from how it functioned in previous agent versions. Previously, if set_transaction_name was called with a new transaction name and a new category that did not match the category that was already assignd to a transaction, neither the new name nor category would be saved to the transaction.
```
fallwith, line 76:
```suggestion
  When the method `NewRelic::Agent.set_transaction_name` is called, it will now always change the name and category of the currently running transaction to what is passed in to the method. This is a change from how it functioned in previous agent versions. Previously, if `set_transaction_name` was called with a new transaction name and a new category that did not match the category that was already assigned to a transaction, neither the new name nor category would be saved to the transaction. 
```

fallwith (changes requested):

fallwith (approved):

CHANGELOG.md
tannalynn: oh awesome, thanks for catching that!

Issue #1756, opened 2023-01-24
Add Fiber instrumentation


tannalynn:
Update the agent to work better within the context of fibers. This work will be very similar to the thread instrumentation, but added to the Fiber class. We will need to copy the current transaction over, and update how we're keeping track of the span parents to account for fibers, as currently it is only taking threads into account.

fallwith:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/APM-Agent-Overview-614b054ff9a50b000f9039d7/issues/newrelic/newrelic-ruby-agent/1756?planning-poker) with Zenhub @kaylareopelle @tannalynn @hannahramadan

PR #1802, opened 2023-02-07
Add fiber instrumentation

tannalynn:
Adds instrumentation to the Fiber class.
This will allow the agent to treat fibers the same way it is currently treating threads, making it possible to automatically record spans created inside of fibers and will create a fiber spans.
This will share the `instrumentation.thread.tracing` config to control whether the current transaction is carried over to the new fiber. It made more sense to use the same config as threads for this behavior, since there isn't any reason only one would need to be disabled. 

closes https://github.com/newrelic/newrelic-ruby-agent/issues/1756

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,11 @@
 
   The configuration option `instrumentation.thread.tracing` is now enabled by default. This will allow the agent to properly monitor code that occurs inside of threads. If you are currently using custom instrumentation to start a new transaction inside of threads, this may be a breaking change, as it will no longer start a new transaction if one already exists.
 
+- **Add Fiber Instrumentation**
+
+  Fibers are now automatically instrumented similarly to threads. This can be configured using `instrumentation.fiber`.
+
+
 - **Ruby 2.2 and 2.3 Dropped**
 
   Support for Ruby 2.2 and 2.3 dropped with this release. They are no longer included in our test matrices and are not supported
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,7 @@
 
 - **Add Fiber Instrumentation**
 
-  Fibers are now automatically instrumented similarly to threads. This can be configured using `instrumentation.fiber`.
+  `Fiber` instances are now automatically instrumented similarly to `Thread` instances. This can be configured using `instrumentation.fiber`.
 
 
 - **Ruby 2.2 and 2.3 Dropped**
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1318,7 +1318,7 @@ module NewRelic
           :type => String,
           :dynamic_name => true,
           :allowed_from_server => false,
-          :description => 'Controls auto-instrumentation of the Fiber library at start up. May be one of [auto|prepend|chain|disabled].'
+          :description => 'Controls auto-instrumentation of the Fiber class at start up. May be one of [auto|prepend|chain|disabled].'
         },
         :'instrumentation.concurrent_ruby' => {
           :default => 'auto',
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1312,6 +1312,14 @@ module NewRelic
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of bunny at start up.  May be one of [auto|prepend|chain|disabled].'
         },
+        :'instrumentation.fiber' => {
+          :default => 'auto',
+          :public => true,
+          :type => String,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => 'Controls auto-instrumentation of the Fiber library at start up. May be one of [auto|prepend|chain|disabled].'
+        },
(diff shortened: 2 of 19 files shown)


CHANGELOG.md
```diff
@@ -56,6 +56,11 @@
 
   The configuration option `instrumentation.thread.tracing` is now enabled by default. This will allow the agent to properly monitor code that occurs inside of threads. If you are currently using custom instrumentation to start a new transaction inside of threads, this may be a breaking change, as it will no longer start a new transaction if one already exists.
 
+- **Add Fiber Instrumentation**
+
+  Fibers are now automatically instrumented similarly to threads. This can be configured using `instrumentation.fiber`.
```
fallwith, line 61:
```suggestion
  `Fiber` instances are now automatically instrumented similarly to `Thread` instances. This can be configured using `instrumentation.fiber`.
```

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1312,6 +1312,14 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of bunny at start up.  May be one of [auto|prepend|chain|disabled].'
         },
+        :'instrumentation.fiber' => {
+          :default => 'auto',
+          :public => true,
+          :type => String,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => 'Controls auto-instrumentation of the Fiber library at start up. May be one of [auto|prepend|chain|disabled].'
```
fallwith, line 1321:
```suggestion
          :description => 'Controls auto-instrumentation of the Fiber class at start up. May be one of [auto|prepend|chain|disabled].'
```

lib/new_relic/agent/instrumentation/fiber.rb
```diff
@@ -0,0 +1,25 @@
...
+
+DependencyDetection.defer do
+  named :fiber
+
+  depends_on do
+    defined?(Fiber)
+  end
+
+  executes do
+    NewRelic::Agent.logger.info('Installing fiber instrumentation')
```
fallwith, line 17:
```suggestion
    NewRelic::Agent.logger.info('Installing Fiber instrumentation')
```

newrelic.yml
```diff
@@ -334,6 +334,10 @@ common: &default_settings
   # May be one of [auto|prepend|chain|disabled].
   # instrumentation.bunny: auto
 
+# Controls auto-instrumentation of Fiber at start up.
+# May be one of [auto|prepend|chain|disabled]
+# instrumentation.fiber: auto
```
fallwith, line 339:
```suggestion
  # Controls auto-instrumentation of Fiber at start up.
  # May be one of [auto|prepend|chain|disabled]
  # instrumentation.fiber: auto
```

test/multiverse/suites/thread/Envfile
```diff
@@ -0,0 +1,8 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+instrumentation_methods :chain, :prepend
+
+gemfile <<~RB
+RB
```
fallwith, line 8:
```suggestion
gemfile nil
```

fallwith (approved):

lib/new_relic/agent/tracer.rb
```diff
@@ -410,9 +410,17 @@ def clear_state
...
+        end
+
+        def thread_tracing_enabled?
+          NewRelic::Agent.config[:'instrumentation.thread.tracing']
+        end
+
         def thread_block_with_current_transaction(*args, segment_name:, parent: nil, &block)
           parent ||= current_segment
-          current_txn = ::Thread.current[:newrelic_tracer_state].current_transaction if ::Thread.current[:newrelic_tracer_state] && ::Thread.current[:newrelic_tracer_state].is_execution_traced?
+          current_txn = ::Thread.current[:newrelic_tracer_state]&.current_transaction if ::Thread.current[:newrelic_tracer_state]&.is_execution_traced?
```
hannahramadan, line 423: nice use of >2.3 syntax 😎

lib/new_relic/agent/tracer.rb
```diff
@@ -410,9 +410,17 @@ def clear_state
 
         alias_method :tl_clear, :clear_state
 
+        def current_segment_key
+          ::Fiber.current.object_id
```
tannalynn, line 414: Using the fiber id instead of the thread id (which is what we did before), still totally works with threads, because every new thread is a new fiber as well, but it also allows us to accurately track multiple fibers that exist inside of a single thread.

lib/new_relic/agent/transaction.rb
```diff
@@ -260,20 +257,26 @@ def initialize(category, options)
         end
       end
 
-      def parent_thread_id
-        ::Thread.current.nr_parent_thread_id if ::Thread.current.respond_to?(:nr_parent_thread_id)
+      def state
```
tannalynn, line 260: This just makes sense the transaction is always using the correct state object. we were storing the state on the transaction originally, which wasn't causing any issues, but this is just a preemptive safety change for just in case situations.

PR #1828, opened 2023-02-17
Update grpc gemlist

hannahramadan:
Lock grpc dependency `google-protobuf` to 3.21.12 for Rubies less than 2.5.9. 

google-protobuf 3.22 depends on ruby >=2.6, so we need to pin the version.


tannalynn (approved):
🎉

PR #1837, opened 2023-02-24
CAT: do not insert headers for gRPC requests

### LINKED ISSUE #1730: gRPC call causes "ArgumentError: 'X-NewRelic-ID' is an invalid header key, must match [a-z0-9-_.]+" with Google Ads API
## Description
When we call Google Ads API like that example, it causes `"ArgumentError: 'X-NewRelic-ID' is an invalid header key, must match [a-z0-9-_.]+"`.
https://github.com/googleads/google-ads-ruby/blob/d29519bc6c9ec23cad348b9b0ff5375c08a9fdf8/examples/account_management/list_accessible_customers.rb#L28-L40

## Expected Behavior (Acceptance Criteria)

no `ArgumentError` is thrown.

A stack trace from Rollbar is below:
```
File /usr/local/lib/ruby/gems/2.7.0/gems/grpc-1.50.0-x86_64-linux/src/ruby/lib/grpc/generic/active_call.rb:373 in run_batch
File /usr/local/lib/ruby/gems/2.7.0/gems/grpc-1.50.0-x86_64-linux/src/ruby/lib/grpc/generic/active_call.rb:373 in request_response
File /usr/local/lib/ruby/gems/2.7.0/gems/grpc-1.50.0-x86_64-linux/src/ruby/lib/grpc/generic/client_stub.rb:174 in block (2 levels) in request_response
File /usr/local/lib/ruby/gems/2.7.0/gems/grpc-1.50.0-x86_64-linux/src/ruby/lib/grpc/generic/interceptors.rb:181 in block in intercept!
File /usr/local/lib/ruby/gems/2.7.0/gems/google-ads-googleads-20.0.0/lib/google/ads/google_ads/interceptors/error_interceptor.rb:65 in request_response
File /usr/local/lib/ruby/gems/2.7.0/gems/grpc-1.50.0-x86_64-linux/src/ruby/lib/grpc/generic/interceptors.rb:175 in intercept!
File /usr/local/lib/ruby/gems/2.7.0/gems/grpc-1.50.0-x86_64-linux/src/ruby/lib/grpc/generic/client_stub.rb:173 in block in request_response
File /usr/local/lib/ruby/gems/2.7.0/gems/gapic-common-0.9.0/lib/gapic/grpc/service_stub/rpc_call.rb:121 in call
File /usr/local/lib/ruby/gems/2.7.0/gems/gapic-common-0.9.0/lib/gapic/grpc/service_stub.rb:156 in call_rpc
File /usr/local/lib/ruby/gems/2.7.0/gems/google-ads-googleads-20.0.0/lib/google/ads/google_ads/v12/services/customer_service/client.rb:322 in list_accessible_customers
File /usr/local/lib/ruby/gems/2.7.0/gems/google-ads-googleads-20.0.0/lib/google/ads/google_ads/service_wrapper.rb:53 in public_send
File /usr/local/lib/ruby/gems/2.7.0/gems/google-ads-googleads-20.0.0/lib/google/ads/google_ads/service_wrapper.rb:53 in method_missing
```

https://github.com/grpc/grpc/blob/91091e3668144de9c6aa392f496bb7639f7025a7/src/ruby/lib/grpc/generic/active_call.rb#L373

## Steps to Reproduce

Our logic is similar with the example code: https://github.com/googleads/google-ads-ruby/blob/d29519bc6c9ec23cad348b9b0ff5375c08a9fdf8/examples/account_management/list_accessible_customers.rb#L28-L40
We are also investigating reproduce condition.

## Your Environment

- Ruby 2.7.6
- newrelic_rpm (8.13.1)
- grpc (1.50.0)
- gapic-common (0.9.0)
- google-ads-googleads (20.0.0)

## Additional context
Add any other context about the problem here. For example, relevant community posts or support tickets.

- https://github.com/newrelic/newrelic-ruby-agent/issues/1337

Additional AC:
In the event that CAT is enabled, do not raise an error and handle the use case gracefully.

fallwith:
CAT: do not insert headers for outbound gRPC requests, as gRPC libraries will validate all headers against a regex pattern that the CAT headers do not currently satisfy (CAT uses capital letters).

resolves #1730

### DIFF
--- a/lib/new_relic/agent/distributed_tracing/cross_app_tracing.rb
+++ b/lib/new_relic/agent/distributed_tracing/cross_app_tracing.rb
@@ -1,4 +1,3 @@
-# -*- ruby -*-
 # This file is distributed under New Relic's license terms.
 # See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
 # frozen_string_literal: true
@@ -53,6 +52,9 @@ module NewRelic
 
       def insert_cross_app_header(headers)
         return unless CrossAppTracing.cross_app_enabled?
+        # do not insert CAT headers for gRPC requests
+        # https://github.com/newrelic/newrelic-ruby-agent/issues/1730
+        return if grpc_headers?(headers)
 
         @is_cross_app_caller = true
         txn_guid = transaction.guid
@@ -248,6 +250,10 @@ module NewRelic
         if (referring_guid = payload.referring_guid)
           txn.attributes.add_intrinsic_attribute(:referring_transaction_guid, referring_guid)
         end
+      end
+
+      def grpc_headers?(headers)
+        headers&.class&.name&.include?('::GRPC::')
       end
     end
   end
--- a/lib/new_relic/agent/distributed_tracing/cross_app_tracing.rb
+++ b/lib/new_relic/agent/distributed_tracing/cross_app_tracing.rb
@@ -52,9 +52,6 @@ module NewRelic
 
       def insert_cross_app_header(headers)
         return unless CrossAppTracing.cross_app_enabled?
-        # do not insert CAT headers for gRPC requests
-        # https://github.com/newrelic/newrelic-ruby-agent/issues/1730
-        return if grpc_headers?(headers)
 
         @is_cross_app_caller = true
         txn_guid = transaction.guid
@@ -250,10 +247,6 @@ module NewRelic
         if (referring_guid = payload.referring_guid)
           txn.attributes.add_intrinsic_attribute(:referring_transaction_guid, referring_guid)
         end
-      end
-
-      def grpc_headers?(headers)
-        headers&.class&.name&.include?('::GRPC::')
       end
     end
   end
--- a/test/multiverse/suites/agent_only/cross_application_tracing_test.rb
+++ b/test/multiverse/suites/agent_only/cross_application_tracing_test.rb
@@ -61,6 +61,57 @@ class CrossApplicationTracingTest < Minitest::Test
     assert_includes attributes_for(last_traced_error, :intrinsic), :client_cross_process_id
   end
 
+  def test_grpc_headers_returns_nil_if_the_headers_object_is_nil
+    assert_nil tracer.send(:grpc_headers?, nil)
+  end
+
+  def test_grpc_headrs_returns_nil_if_the_headers_object_class_is_nil
+    headers = MiniTest::Mock.new
+    headers.expect :class, nil
+
+    assert_nil tracer.send(:grpc_headers?, headers)
+    headers.verify
+  end
+
+  def test_grpc_headers_returns_nil_if_the_headers_object_class_name_is_nil
+    klass = MiniTest::Mock.new
+    klass.expect :name, nil
+    headers = MiniTest::Mock.new
+    headers.expect :class, klass
+
+    assert_nil tracer.send(:grpc_headers?, headers)
+    klass.verify
+    headers.verify
+  end
+
+  def test_grpc_headers_returns_false_if_the_headers_object_class_name_does_not_include_grpc
+    klass = MiniTest::Mock.new
+    klass.expect :name, 'The::Old::Man::and::the::Sea'
+    headers = MiniTest::Mock.new
+    headers.expect :class, klass
+
+    refute tracer.send(:grpc_headers?, headers)
+    klass.verify
+    headers.verify
+  end
+
+  def test_grpc_headers_returns_true_if_the_headers_object_class_name_does_include_grpc
+    klass = MiniTest::Mock.new
+    klass.expect :name, 'NewRelic::Agent::Instrumentation::GRPC::Client::RequestWrapper'
+    headers = MiniTest::Mock.new
+    headers.expect :class, klass
+
+    assert tracer.send(:grpc_headers?, headers)
+    klass.verify
+    headers.verify
+  end
+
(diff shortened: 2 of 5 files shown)


lib/new_relic/agent/distributed_tracing/cross_app_tracing.rb
```diff
@@ -249,6 +251,10 @@ def assign_intrinsic_transaction_attributes(state)
           txn.attributes.add_intrinsic_attribute(:referring_transaction_guid, referring_guid)
         end
       end
+
+      def grpc_headers?(headers)
+        headers&.class&.name&.include?('::GRPC::')
```
kaylareopelle, line 256: I love seeing these safe navigation operators!

test/multiverse/suites/agent_only/cross_application_tracing_test.rb
```diff
@@ -61,6 +61,57 @@ def test_cross_app_error_attaches_process_id_to_intrinsics
     assert_includes attributes_for(last_traced_error, :intrinsic), :client_cross_process_id
   end
 
+  def test_grpc_headers_returns_nil_if_the_headers_object_is_nil
+    assert_nil tracer.send(:grpc_headers?, nil)
+  end
+
+  def test_grpc_headrs_returns_nil_if_the_headers_object_class_is_nil
```
kaylareopelle, line 68:
```suggestion
  def test_grpc_headers_returns_nil_if_the_headers_object_class_is_nil
```

lib/new_relic/agent/distributed_tracing/cross_app_tracing.rb
```diff
@@ -53,6 +52,9 @@ def cat_path_hash
 
       def insert_cross_app_header(headers)
         return unless CrossAppTracing.cross_app_enabled?
+        # do not insert CAT headers for gRPC requests
+        # https://github.com/newrelic/newrelic-ruby-agent/issues/1730
+        return if grpc_headers?(headers)
```
kaylareopelle, line 57: It looks like "then" case coverage is missing for this line.

tannalynn (commented):
Looks good! 
I am curious though why you went with this approach instead of adding a check in the grpc instrumentation to not call the insert headers method is cat was enabled?

fallwith:
> Looks good! I am curious though why you went with this approach instead of adding a check in the grpc instrumentation to not call the insert headers method is cat was enabled?

@tannalynn looking at it again, I think we should have the check occur in the grpc instrumentation itself. Originally I wanted the scope to be as small as possible (just CAT headers) to preserve any potential good headers functionality, but now that I see that the scope is too small.

The gRPC instrumentation invokes `NewRelic::Agent::Transaction::ExternalRequestSegment#add_request_headers`.

That method does the following:

- Updates the segment name when a host can be obtained from the request
- Inserts the synthetics header (bad, won't satisfy gRPC header regex)
- Inserts the DT header (bad, won't satisfy gRPC header regex)

Currently I'm only addressing the 3rd item on the list. I need to get after the 2nd one as well. I could of course just prevent the call from being made entirely and skip all 3 items. What are your thoughts on that and the loss of item 1 for grpc?

lib/new_relic/agent/distributed_tracing/cross_app_tracing.rb
fallwith: Odd. There was an existing test for the presence of the headers that should continue to test the `return` not being reached, and there has been a new test added to test the `return` being reached. Both sides of the `if` conditional should be getting tested. I'll revisit after I move the check as Tanna is suggesting.

tannalynn:
> > Looks good! I am curious though why you went with this approach instead of adding a check in the grpc instrumentation to not call the insert headers method is cat was enabled?
> 
> @tannalynn looking at it again, I think we should have the check occur in the grpc instrumentation itself. Originally I wanted the scope to be as small as possible (just CAT headers) to preserve any potential good headers functionality, but now that I see that the scope is too small.
> 
> The gRPC instrumentation invokes `NewRelic::Agent::Transaction::ExternalRequestSegment#add_request_headers`.
> 
> That method does the following:
> 
> * Updates the segment name when a host can be obtained from the request
> * Inserts the synthetics header (bad, won't satisfy gRPC header regex)
> * Inserts the DT header (bad, won't satisfy gRPC header regex)
> 
> Currently I'm only addressing the 3rd item on the list. I need to get after the 2nd one as well. I could of course just prevent the call from being made entirely and skip all 3 items. What are your thoughts on that and the loss of item 1 for grpc?

IMO I think that's fine. It should only be skipped if someone actually is using CAT, which is super old, deprecated, and not many people use it, so for the vast majority of people this change won't effect anything.

lib/new_relic/agent/distributed_tracing/cross_app_tracing.rb
fallwith: Removed with the [26e4415](https://github.com/newrelic/newrelic-ruby-agent/pull/1837/commits/26e4415437bb1e86f455b45bedc947507a7b0b93) rework.

lib/new_relic/agent/distributed_tracing/cross_app_tracing.rb
fallwith: Moot with the [26e4415](https://github.com/newrelic/newrelic-ruby-agent/pull/1837/commits/26e4415437bb1e86f455b45bedc947507a7b0b93) rework.

kaylareopelle (approved):
Nice refactor!

PR #1838, opened 2023-02-24
Update typhoeus multiverse suite_condition

kaylareopelle:
This test suite wasn't running on CRuby because there wasn't a check for the presence of Java before the /etc files, which exist on my machine


tannalynn (approved):
Interesting! I hadn't seen this come up before, glad you figured out what was going on 🎉

PR #1839, opened 2023-02-24
Log request headers

### LINKED ISSUE #1657: Show request headers instead of object
When a customer uses the HTTPClient library for their external requests, they do not see the headers for the requests in the logs, but instead the human-readable representation of the object.

Example: 
`OUTGOING REQUEST HEADERS: #<NewRelic::Agent::HTTPClients::HTTPClientRequest:0x00007fd0dda983e0>` 

The content following the colon should be a list of headers.

The `DistributedTracer#log_request_headers` method is responsible for logging this line of code. It expects the first argument, `headers` will always be an interpolatable value that includes the header values. 

This method is invoked by `ExternalRequestSegment#add_request_headers`, which calls `DistributedTracer#insert_headers` with the request object as an argument, followed by `DistributedTracer#log_request_headers`.

`DistributedTracer#insert_headers` uses the argument it receives from `ExternalRequestSegment#add_request_headers` for multiple method calls:
* insert_trace_context_headers
* insert_distributed_trace_header
* insert_cross_app_header
* log_request_headers

It's unclear if these method calls are silently failing when the external request is an HTTPClientRequest.

## Acceptance Criteria
* Logged headers for HTTPClientRequest objects are the headers and not the human-readable representation of the HTTPClientRequest object
* All HTTP client libraries have their headers correctly logged
* Methods called in `DistributedTracer#insert_headers` are able to do what they need to with the header argument for each external request library (make sure there aren't silent failures happening when other DT headers are getting added)

kaylareopelle:
# Overview

* Adds a `#headers` method to all `NewRelic::Agent::HTTPClients::AbstractRequest` subclasses that returns a hash of the headers for the request 
* Updates `NewRelic::Agent::Transaction::DistributedTracer#log_request_headers` to check the `headers` argument's class. If it is an instance of `NewRelic::Agent::HTTPClients::AbstractRequest`, the `#headers` method will be called and logged. If not, the argument as-passed will be printed

Resolves #1657

### DIFF
--- a/lib/new_relic/agent/http_clients/abstract.rb
+++ b/lib/new_relic/agent/http_clients/abstract.rb
@@ -14,9 +14,9 @@ module NewRelic
       #
       # @api public
       class AbstractRequest
-        [:[], :[]=, :type, :host_from_header, :host, :method, :headers].each do |name|
+        [:[], :[]=, :type, :host_from_header, :host, :method, :headers, :uri].each do |name|
           define_method "#{name}" do |i|
-            not_implemented
+            not_implemented(__method__)
           end
         end
 
--- a/lib/new_relic/agent/http_clients/abstract.rb
+++ b/lib/new_relic/agent/http_clients/abstract.rb
@@ -15,8 +15,8 @@ module NewRelic
       # @api public
       class AbstractRequest
         [:[], :[]=, :type, :host_from_header, :host, :method, :headers, :uri].each do |name|
-          define_method "#{name}" do |i|
-            not_implemented(__method__)
+          define_method("#{name}") do
+            not_implemented(name)
           end
         end
 
--- a/lib/new_relic/agent/http_clients/abstract.rb
+++ b/lib/new_relic/agent/http_clients/abstract.rb
@@ -15,7 +15,7 @@ module NewRelic
       # @api public
       class AbstractRequest
         [:[], :[]=, :type, :host_from_header, :host, :method, :headers, :uri].each do |name|
-          define_method("#{name}") do
+          define_method(name) do
             not_implemented(name)
           end
         end
--- a/lib/new_relic/agent/http_clients/abstract.rb
+++ b/lib/new_relic/agent/http_clients/abstract.rb
@@ -14,7 +14,7 @@ module NewRelic
       #
       # @api public
       class AbstractRequest
-        [:[], :[]=, :type, :host_from_header, :host, :method, :headers, :uri].each do |name|
+        %i[[] []= type host_from_header host method headers uri].each do |name|
           define_method(name) do
             not_implemented(name)
           end
--- a/lib/new_relic/agent/http_clients/abstract.rb
+++ b/lib/new_relic/agent/http_clients/abstract.rb
@@ -37,6 +37,10 @@ module NewRelic
         def method
           raise NotImplementedError, MUST_IMPLEMENT_ERROR % [self.class, __method__]
         end
+
+        def headers
+          raise NotImplementedError, MUST_IMPLEMENT_ERROR % [self.class, __method__]
+        end
       end
 
       # This class implements the adaptor pattern and is used internally provide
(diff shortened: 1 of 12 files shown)


tannalynn (approved):
🎉

lib/new_relic/agent/http_clients/abstract.rb
```diff
@@ -37,6 +37,10 @@ def host
         def method
           raise NotImplementedError, MUST_IMPLEMENT_ERROR % [self.class, __method__]
         end
+
+        def headers
+          raise NotImplementedError, MUST_IMPLEMENT_ERROR % [self.class, __method__]
+        end
```
fallwith, line 43: I suppose we could technically DRY this up with something that reuses everything except the method name but that might be overkill...

```ruby
        def method
          not_implemented(__method__)
        end

        def headers
          not_implemented(__method__)
        end

        private

        def not_implemented(method_name)
          raise NotImplementedError, MUST_IMPLEMENT_ERROR % [self.class, method_name]
        end
```

lib/new_relic/agent/http_clients/httpclient_wrappers.rb
```diff
@@ -57,12 +57,16 @@ def uri
         end
 
         def [](key)
-          request.headers[key]
+          headers[key]
         end
 
         def []=(key, value)
           request.http_header[key] = value
```
fallwith, line 64: I'm tempted to move `[]` and `[]=` into the base class and just require every subclass to define a `headers` method. This seems to be the only method that wouldn't work with that, explicitly doing something other than `headers[key] = value`.

fallwith (approved):

lib/new_relic/agent/http_clients/abstract.rb
kaylareopelle: Good idea! I'll put that in a separate technical debt ticket/PR to avoid overloading the fix.

kaylareopelle: Changed my mind! Fixed in 0b28bf78cf699ce57f346cf26fda78585d76aa96

fallwith: Using `define_method` is a slick approach. The new `not_implemented` method requires an argument though, and the callers don't appear to be providing one.

lib/new_relic/agent/http_clients/httpclient_wrappers.rb
kaylareopelle: I think there's a lot of similarities between the `[]` and `[]=` methods between the subclasses, but there's some uniqueness in typheous, httpclient, and net_http versions of `[]=` that make me hesitant to switch them over. 

There should now be a `headers` method in each subclass that accesses a hash of headers. In some libraries the hash of headers is not the same object where things are supposed to be added (ex. http.rb, net_http). 

Thoughts?

lib/new_relic/agent/http_clients/abstract.rb
kaylareopelle: Great catch! Fixed. I was also missing the URI method.

kaylareopelle: edd33782b5a87d86a9ec671da3b3afd66b5f5f45

lib/new_relic/agent/http_clients/httpclient_wrappers.rb
fallwith: Fully 3 Ruby HTTP clients being incompatible with the approach is definitely a strong reason to not go with it. Sounds good.

Having the common `headers` method for getting values is great at least. Nice!

lib/new_relic/agent/http_clients/abstract.rb
kaylareopelle: Fix commit: 236788a0bef6043eb2c8caf7658f9576aa11ade4

fallwith (dismissed):

fallwith (approved):

fallwith:
Released in v9.1.0 of the `newrelic_rpm` gem.

Issue #1787, opened 2023-02-02
Review rubocops that became available with 9.0.0


hannahramadan:
The discontinued support of Ruby 2.2 and 2.3 in agent 9.0.0 freed up a few new rubocop cops for use. We should review each of these new cops and decide to either implement and cleanup our codebase accordingly, or decide that we won't follow that cop, move it to `rubocop.yml` and mark it as `Enabled: false`

- Performance/RegexpMatch
- Performance/Sum
- Performance/UnfreezeString
- Style/FetchEnvVar
- Style/RedundantStringEscape
- Style/SafeNavigation

A note about Style/RedundantStringEscape:
We may not be able to fix this since we're specifically testing the behavior for when there IS an extra `\` at the end of the string. We should probably just remove this from `rubocop_todo` and add a disable cop comment to that line in the test file.

kaylareopelle:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/APM-Agent-Overview-614b054ff9a50b000f9039d7/issues/newrelic/newrelic-ruby-agent/1787?planning-poker) with Zenhub @fallwith @hannahramadan @tannalynn

Issue #1246, opened 2022-07-01
Replace String.new() with +string


kaylareopelle:
Ruby 2.3 introduces the [+string](https://ruby-doc.org/core-3.1.2/String.html#method-i-2B-40) method. After conducting some benchmarking tests, `+string` has a performance benefit over `String.new()` in both object allocations and iterations per second. 

When we remove support for Ruby 2.2 in major release 9.0.0, we can enhance our performance by replacing calls to String.new() with +string.

tannalynn:
Removed the the 9.0 milestone
Since we will no longer support 2.2 as of 9.0 , this will not be a breaking change after 9.0 is released

kaylareopelle:
Closed by #1787

PR #1840, opened 2023-02-24
Rubocops compatible with Ruby 2.4+

kaylareopelle:
# Overview

Our agent's previous support of Ruby 2.2 and 2.3 prevented some Rubocop linters from being enabled. This PR enables the following:
* [Performance/RegexpMatch](https://docs.rubocop.org/rubocop-performance/cops_performance.html#performanceregexpmatch) - safe
* [Performance/Sum](https://docs.rubocop.org/rubocop-performance/cops_performance.html#performancesum) - unsafe autocorrection; covered by tests 
* [Performance/UnfreezeString](https://docs.rubocop.org/rubocop-performance/cops_performance.html#performanceunfreezestring) - unsafe autocorrection; covered by tests
* [Style/RedundantStringEscape](https://docs.rubocop.org/rubocop/1.46/cops_style.html#styleredundantstringescape) - safe 
* [Style/SafeNavigation](https://docs.rubocop.org/rubocop/1.46/cops_style.html#stylesafenavigation) - unsafe autocorrection

It disables Style/FetchEnvVar. When `ENV.fetch` fails, a `KeyError` or the explicitly specified default
value is returned. We don't want to raise an unexpected error in a customer's environment. Also, fetch is slightly less performant than `ENV[]` when there is not a non-nil replacement value.

Resolves: https://github.com/newrelic/newrelic-ruby-agent/issues/1787 
Resolves: https://github.com/newrelic/newrelic-ruby-agent/issues/1246

### DIFF
--- a/infinite_tracing/lib/infinite_tracing/config.rb
+++ b/infinite_tracing/lib/infinite_tracing/config.rb
@@ -64,7 +64,7 @@ module NewRelic::Agent
       # is overridden by the presence of the port on the host entry.
       def port_from_host_entry
         port_str = NewRelic::Agent.config[:'infinite_tracing.trace_observer.host'].scan(%r{:(\d+)$}).flatten
-        if port = (port_str[0] and port_str[0].to_i)
+        if port = (port_str[0]&.to_i)
           NewRelic::Agent.logger.warn(":'infinite_tracing.trace_observer.port' is ignored if present because :'infinite_tracing.trace_observer.host' specifies the port")
           return port
         end
--- a/infinite_tracing/lib/infinite_tracing/config.rb
+++ b/infinite_tracing/lib/infinite_tracing/config.rb
@@ -64,7 +64,7 @@ module NewRelic::Agent
       # is overridden by the presence of the port on the host entry.
       def port_from_host_entry
         port_str = NewRelic::Agent.config[:'infinite_tracing.trace_observer.host'].scan(%r{:(\d+)$}).flatten
-        if port = (port_str[0]&.to_i)
+        if port = port_str[0]&.to_i
           NewRelic::Agent.logger.warn(":'infinite_tracing.trace_observer.port' is ignored if present because :'infinite_tracing.trace_observer.host' specifies the port")
           return port
         end
--- a/lib/new_relic/agent/commands/agent_command_router.rb
+++ b/lib/new_relic/agent/commands/agent_command_router.rb
@@ -29,9 +29,7 @@ module NewRelic
           @handlers['start_profiler'] = proc { |cmd| thread_profiler_session.handle_start_command(cmd) }
           @handlers['stop_profiler'] = proc { |cmd| thread_profiler_session.handle_stop_command(cmd) }
 
-          if event_listener
-            event_listener.subscribe(:before_shutdown, &method(:on_before_shutdown))
-          end
+          event_listener&.subscribe(:before_shutdown, &method(:on_before_shutdown))
         end
 
         def new_relic_service
--- a/lib/new_relic/agent/commands/agent_command_router.rb
+++ b/lib/new_relic/agent/commands/agent_command_router.rb
@@ -30,6 +30,10 @@ module NewRelic
           @handlers['stop_profiler'] = proc { |cmd| thread_profiler_session.handle_stop_command(cmd) }
 
           event_listener&.subscribe(:before_shutdown, &method(:on_before_shutdown))
+          # the following statement needs else branch coverage
+          if event_listener # rubocop:disable Style/SafeNavigation
+            event_listener.subscribe(:before_shutdown, &method(:on_before_shutdown))
+          end
         end
 
         def new_relic_service
--- a/lib/new_relic/agent/commands/agent_command_router.rb
+++ b/lib/new_relic/agent/commands/agent_command_router.rb
@@ -29,7 +29,6 @@ module NewRelic
           @handlers['start_profiler'] = proc { |cmd| thread_profiler_session.handle_start_command(cmd) }
           @handlers['stop_profiler'] = proc { |cmd| thread_profiler_session.handle_stop_command(cmd) }
 
-          event_listener&.subscribe(:before_shutdown, &method(:on_before_shutdown))
           # the following statement needs else branch coverage
           if event_listener # rubocop:disable Style/SafeNavigation
             event_listener.subscribe(:before_shutdown, &method(:on_before_shutdown))
--- a/lib/new_relic/agent/custom_event_aggregator.rb
+++ b/lib/new_relic/agent/custom_event_aggregator.rb
@@ -27,7 +27,7 @@ module NewRelic
         return unless enabled?
 
         type = @type_strings[type]
-        unless type =~ EVENT_TYPE_REGEX
+        unless EVENT_TYPE_REGEX.match?(type)
           note_dropped_event(type)
           return false
         end
--- a/lib/new_relic/agent/instrumentation/action_controller_other_subscriber.rb
+++ b/lib/new_relic/agent/instrumentation/action_controller_other_subscriber.rb
@@ -28,7 +28,7 @@ module NewRelic
 
         def controller_name_for_metric(payload)
           # redirect_to
-          return payload[:request].controller_class.controller_path if payload[:request] && payload[:request].controller_class
+          return payload[:request].controller_class.controller_path if payload[:request]&.controller_class
 
           # unpermitted_parameters
           ::NewRelic::LanguageSupport.constantize(payload[:context][:controller]).controller_path if payload[:context] && payload[:context][:controller]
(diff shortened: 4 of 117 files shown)


fallwith:
@kaylareopelle perhaps it's an issue best left for a separate PR, but my preference would be that for everywhere we now plan to align with the defaults, we would simply remove the `Enabled: false` based override from `.rubocop_todo.yml` without actually redundantly adding a new `Enabled: true` section to `.rubocop.yml`.

kaylareopelle:
> @kaylareopelle perhaps it's an issue best left for a separate PR, but my preference would be that for everywhere we now plan to align with the defaults, we would simply remove the `Enabled: false` based override from `.rubocop_todo.yml` without actually redundantly adding a new `Enabled: true` section to `.rubocop.yml`.

I'm comfortable with that change. I'm not sure if all of these are on by default. I'll check Rubocop and adjust accordingly. I think makes sense to add another issue to adjust the other cops, since some of the enabled cops may be disabled by default.

kaylareopelle:
@fallwith  - RedundantStringEscape and Sum are pending, so I left those references in to make sure they get applied. The others were removed in 03d82e65a3a5b281165b0d7fe2b45c7875693d5c

lib/new_relic/agent/system_info.rb
```diff
@@ -128,7 +128,7 @@ def self.parse_cpuinfo(cpuinfo)
 
         num_physical_packages = cores.keys.map(&:first).uniq.size
         num_physical_cores = cores.size
-        num_logical_processors = cores.values.reduce(0, :+)
```
kaylareopelle, line 131: This cop is unsafe if sum is called on an object that doesn't have a sum method, such as a String. `cores.values` should always be an array of integers based on the default values of the Hash defined on [line 110](https://github.com/newrelic/newrelic-ruby-agent/pull/1840/commits/02b7aabde79f256521674c65c1287e3d585950b0#diff-f6f8dd00b6df0ce9cdcb61bc2ff7ce94f894a821e859d9c1e0e991c57d2608acL110)

lib/new_relic/agent/transaction.rb
```diff
@@ -181,7 +181,7 @@ def self.add_agent_attribute(key, value, default_destinations)
 
       def add_agent_attribute(key, value, default_destinations)
         @attributes.add_agent_attribute(key, value, default_destinations)
-        current_segment.add_agent_attribute(key, value) if current_segment
```
kaylareopelle, line 184: no else

lib/new_relic/agent/transaction/trace_node.rb
```diff
@@ -149,10 +147,8 @@ def each_node_with_nest_tracking(&block)
           summary = yield(self)
           summary.current_nest_count += 1 if summary
 
-          if @children
-            @children.each do |node|
-              node.each_node_with_nest_tracking(&block)
-            end
+          @children&.each do |node|
+            node.each_node_with_nest_tracking(&block)
```
kaylareopelle, line 151: no then

lib/new_relic/agent/transaction_time_aggregator.rb
```diff
@@ -101,7 +101,7 @@ def current_thread
 
         def thread_is_alive?(thread_id)
           thread = thread_by_id(thread_id)
-          thread && thread.alive?
```
kaylareopelle, line 104: no else

lib/new_relic/agent/transaction_time_aggregator.rb
```diff
@@ -145,8 +145,7 @@ def transaction_time_in_thread(timestamp, thread_id, entry)
         end
 
         def log_missing_elapsed_transaction_time
-          transaction_name = Tracer.current_transaction &&
```
kaylareopelle, line 148: no coverage

lib/new_relic/cli/commands/deployments.rb
```diff
@@ -33,7 +33,7 @@ def initialize(command_line_args)
     @changelog = nil
     @user = nil
     super(command_line_args)
-    @description ||= @leftover && @leftover.join(" ")
```
kaylareopelle, line 36: no else

lib/new_relic/dependency_detection.rb
```diff
@@ -37,7 +37,7 @@ def dependency_by_name(name)
 
   def installed?(name)
     item = dependency_by_name(name)
-    item && item.executed
```
kaylareopelle, line 40: no coverage

lib/new_relic/helper.rb
```diff
@@ -63,7 +63,7 @@ def run_command(command)
         raise NewRelic::CommandRunFailedError.new("Failed to run command '#{command}': #{message}")
       end
 
-      output.chomp if output
```
kaylareopelle, line 66: no else

lib/new_relic/rack/browser_monitoring.rb
```diff
@@ -96,11 +96,11 @@ def autoinstrument_source(response, headers, js_to_inject)
       end
 
       def html?(headers)
-        headers[CONTENT_TYPE] && headers[CONTENT_TYPE].include?(TEXT_HTML)
```
kaylareopelle, line 99: no else

infinite_tracing/lib/infinite_tracing/config.rb
```diff
@@ -64,7 +64,7 @@ def trace_observer_host
       # is overridden by the presence of the port on the host entry.
       def port_from_host_entry
         port_str = NewRelic::Agent.config[:'infinite_tracing.trace_observer.host'].scan(%r{:(\d+)$}).flatten
-        if port = (port_str[0] and port_str[0].to_i)
+        if port = (port_str[0]&.to_i)
```
fallwith, line 67: These parens can go now that we don't have the `and` anymore.

fallwith (approved):

lib/new_relic/agent/commands/agent_command_router.rb
```diff
@@ -29,7 +29,9 @@ def initialize(event_listener = nil)
           @handlers['start_profiler'] = proc { |cmd| thread_profiler_session.handle_start_command(cmd) }
           @handlers['stop_profiler'] = proc { |cmd| thread_profiler_session.handle_stop_command(cmd) }
 
-          if event_listener
+          event_listener&.subscribe(:before_shutdown, &method(:on_before_shutdown))
+          # the following statement needs else branch coverage
+          if event_listener # rubocop:disable Style/SafeNavigation
             event_listener.subscribe(:before_shutdown, &method(:on_before_shutdown))
```
tannalynn, line 35: I think it's doing it twice now (line 32 and also 35) instead of once like before

lib/new_relic/agent/custom_event_aggregator.rb
```diff
@@ -27,7 +27,7 @@ def record(type, attributes)
         return unless enabled?
 
         type = @type_strings[type]
-        unless type =~ EVENT_TYPE_REGEX
+        unless EVENT_TYPE_REGEX.match?(type)
```
tannalynn, line 30: Is it supposed to be `regex.match?(string)`? It seems like in the file below this it's doing `string.match?(regex)`. Although maybe they both do the exact same thing? I'm not sure, just something I noticed

```
EVENT_TYPE_REGEX.match?(type)
# VS
sql&.match?(COMPONENTS_REGEX_MAP[:single_quotes])
```

lib/new_relic/agent/instrumentation/action_controller_other_subscriber.rb
```diff
@@ -28,7 +28,7 @@ def metric_name(name, payload)
 
         def controller_name_for_metric(payload)
           # redirect_to
-          return payload[:request].controller_class.controller_path if payload[:request] && payload[:request].controller_class
+          return payload[:request].controller_class.controller_path if payload[:request]&.controller_class
```
tannalynn, line 31: Unfortunately this line merge conflicts with the PR I merged this morning, i made some changes to this method.

kaylareopelle: I thought that might happen. Fixed in: 054c25bb827ddd3a55e560bf843074e2c1553fc1

lib/new_relic/agent/custom_event_aggregator.rb
kaylareopelle: Great question! Both are valid. The `String`, `Regex`, and `Symbol` classes have a `match?` method. The caller/argument change based on the class `match?` is called on.

infinite_tracing/lib/infinite_tracing/config.rb
kaylareopelle: Thank you! Resolved in: 45e17184fb95e9d79d458bd33fd5486eefdc392c

fallwith (approved):

tannalynn (approved):
🎉

PR #1841, opened 2023-02-25
Multiverse: always show options, fix 'not found'

fallwith:
Multiverse CI tests:

- Always report the options used, pass or fail
- Don't report "No tests found with those options." just because a test failed or errored out.
- Rename "run_test" to "passed" to better convey that it contains the boolean result of the tests having been performed

Unfortunately given that the v5.x `run` method instantiates its own reporter that falls out of scope as soon as the method completes, all we are left with is a boolean result of the run. We can't make a determination as to whether or not 0 tests were found and force that condition to be an error condition with a non-zero exit code. Perhaps in future we could hook into the reporter chain with our own plugin to relay the test count back to us at the end.


kaylareopelle (approved):
Thanks for your proactivity!

Issue #1520, opened 2022-10-03
CI: Address all RuboCop Metrics/AbcSize violations


fallwith:
With #1426 and the [Sunrise Ruby milestone](https://github.com/newrelic/newrelic-ruby-agent/milestone/23) we addressed many AbcSize violations and referenced all remaining work in `.rubocop_todo.yml`. We should circle back to the outstanding items to address them and get their references removed from the TODO file as time permits.

PR #1849, opened 2023-03-01
RuboCop: address some AbcSize issues

fallwith:
Address some Metrics/AbcSize issues to get our maximum ABC size down closer towards a reasonable value

helps with #1520

### DIFF
--- a/lib/new_relic/agent/agent_helpers/start_worker_thread.rb
+++ b/lib/new_relic/agent/agent_helpers/start_worker_thread.rb
@@ -60,36 +60,13 @@ module NewRelic
         end
 
         def create_and_run_event_loop
-          data_harvest = :"#{Agent.config[:data_report_period]}_second_harvest"
-          event_harvest = :"#{Agent.config[:event_report_period]}_second_harvest"
-
           @event_loop = create_event_loop
+          data_harvest = :"#{Agent.config[:data_report_period]}_second_harvest"
           @event_loop.on(data_harvest) do
             transmit_data
           end
-
-          @event_loop.on(interval_for(TRANSACTION_EVENT_DATA)) do
-            transmit_analytic_event_data
-          end
-          @event_loop.on(interval_for(CUSTOM_EVENT_DATA)) do
-            transmit_custom_event_data
-          end
-          @event_loop.on(interval_for(ERROR_EVENT_DATA)) do
-            transmit_error_event_data
-          end
-          @event_loop.on(interval_for(SPAN_EVENT_DATA)) do
-            transmit_span_event_data
-          end
-          @event_loop.on(interval_for(LOG_EVENT_DATA)) do
-            transmit_log_event_data
-          end
-
-          @event_loop.on(:reset_log_once_keys) do
-            ::NewRelic::Agent.logger.clear_already_logged
-          end
-          @event_loop.fire_every(Agent.config[:data_report_period], data_harvest)
-          @event_loop.fire_every(Agent.config[:event_report_period], event_harvest)
-          @event_loop.fire_every(LOG_ONCE_KEYS_RESET_PERIOD, :reset_log_once_keys)
+          establish_interval_transmissions
+          establish_fire_everies(data_harvest)
 
           @event_loop.run
         end
@@ -160,6 +137,37 @@ module NewRelic
               end
             end
           end
+        end
+
+        private
+
+        def establish_interval_transmissions
+          @event_loop.on(interval_for(TRANSACTION_EVENT_DATA)) do
+            transmit_analytic_event_data
+          end
+          @event_loop.on(interval_for(CUSTOM_EVENT_DATA)) do
+            transmit_custom_event_data
+          end
+          @event_loop.on(interval_for(ERROR_EVENT_DATA)) do
+            transmit_error_event_data
+          end
+          @event_loop.on(interval_for(SPAN_EVENT_DATA)) do
+            transmit_span_event_data
+          end
+          @event_loop.on(interval_for(LOG_EVENT_DATA)) do
+            transmit_log_event_data
+          end
+        end
+
+        def establish_fire_everies(data_harvest)
+          @event_loop.on(:reset_log_once_keys) do
+            ::NewRelic::Agent.logger.clear_already_logged
+          end
+
+          event_harvest = :"#{Agent.config[:event_report_period]}_second_harvest"
+          @event_loop.fire_every(Agent.config[:data_report_period], data_harvest)
+          @event_loop.fire_every(Agent.config[:event_report_period], event_harvest)
+          @event_loop.fire_every(LOG_ONCE_KEYS_RESET_PERIOD, :reset_log_once_keys)
         end
       end
     end
--- a/lib/new_relic/cli/commands/deployments.rb
+++ b/lib/new_relic/cli/commands/deployments.rb
@@ -162,24 +162,24 @@ class NewRelic::Cli::Deployments < NewRelic::Cli::Command
   def options
     OptionParser.new(%Q(Usage: #{$0} #{self.class.command} [OPTIONS] ["description"] ), 40) do |o|
       o.separator("OPTIONS:")
-      o.on("-a", "--appname=NAME", String,
-        "Set the application name.",
-        "Default is app_name setting in newrelic.yml. Available only when using API v1.") { |e| @appname = e }
-      o.on("-i", "--appid=ID", String,
-        "Set the application ID",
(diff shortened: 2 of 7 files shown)


lib/new_relic/agent/agent_helpers/start_worker_thread.rb
```diff
@@ -161,6 +138,37 @@ def deferred_work!(connection_options)
...
+          end
+          @event_loop.on(interval_for(SPAN_EVENT_DATA)) do
+            transmit_span_event_data
+          end
+          @event_loop.on(interval_for(LOG_EVENT_DATA)) do
+            transmit_log_event_data
+          end
+        end
+
+        def establish_fire_everies(data_harvest)
```
hannahramadan, line 162: fun method name lol `everies`

lib/new_relic/control/frameworks/rails.rb
```diff
@@ -44,26 +44,38 @@ def rails_config
...
+          install_dependency_detection
+          install_browser_monitoring_and_agent_hooks
+        rescue => e
+          ::NewRelic::Agent.logger.error('Failure during init_config for Rails. Is Rails required in a non-Rails ' \
+                                         'app? Set NEW_RELIC_FRAMEWORK=ruby to avoid this message. The Ruby agent ' \
+                                         'will continue running, but Rails-specific features may be missing. ' \
+                                         "#{e.class} - #{e.message}")
+        end
+
+        def install_dependency_detection
```
hannahramadan, line 56: Love turning that comment into a method!

lib/new_relic/rack/browser_monitoring.rb
```diff
@@ -69,30 +69,33 @@ def should_instrument?(env, status, headers)
...
         # Only scan the first 50k (roughly) then give up.
         beginning_of_source = source[0..SCAN_LIMIT]
         meta_tag_positions = find_meta_tag_positions(beginning_of_source)
         if body_start = find_body_start(beginning_of_source)
           if insertion_index = find_insertion_index(meta_tag_positions, beginning_of_source, body_start)
             source = source_injection(source, insertion_index, js_to_inject)
           else
-            NewRelic::Agent.logger.debug("Skipping RUM instrumentation. Could not properly determine location to inject script.")
+            NewRelic::Agent.logger.debug('Skipping RUM instrumentation. Could not properly determine location to ' \
+                                         'inject script.')
```
hannahramadan, line 91: Why add` \ ` in the debug message?

hannahramadan (approved):

lib/new_relic/rack/browser_monitoring.rb
fallwith: Ah, so while I was fixing the `AbcSize` issue I decided to also address another RuboCop issue with this code block to do with line length. By using backslashes, I just broke the super long debug text up into 120 column chunks.

hannahramadan: Oh just a length thing! Cool thank you

lib/new_relic/agent/agent_helpers/start_worker_thread.rb
```diff
@@ -161,6 +138,37 @@ def deferred_work!(connection_options)
...
+          end
+          @event_loop.on(interval_for(ERROR_EVENT_DATA)) do
+            transmit_error_event_data
+          end
+          @event_loop.on(interval_for(SPAN_EVENT_DATA)) do
+            transmit_span_event_data
+          end
+          @event_loop.on(interval_for(LOG_EVENT_DATA)) do
+            transmit_log_event_data
+          end
```
kaylareopelle, line 159: This still feels like a decent amount of duplication. Would it work to put tbe *_EVENT_DATA constants into an array and loop through them?

lib/new_relic/cli/commands/deployments.rb
```diff
@@ -163,24 +163,24 @@ def set_params_v2(request)
...
-      o.on("-l", "--license-key=KEY", String,
-        "Specify the license key of the account for the app being deployed") { |l| @license_key = l }
-      o.on("-c", "--changes",
-        "Read in a change log from the standard input") { @changelog = STDIN.read }
+      o.on('-r', '--revision=REV', String,
+        'Specify the revision being deployed. Required when using New Relic REST API v2') { |r| @revision = r }
+      o.on('-l', '--license-key=KEY', String,
+        'Specify the license key of the account for the app being deployed') { |l| @license_key = l }
+      o.on('-c', '--changes',
+        'Read in a change log from the standard input') { @changelog = STDIN.read }
```
kaylareopelle, line 183: I love seeing these single quotes. 

Isn't there a built-in linter for changing double quotes to singles for areas without interpolation? I wonder if that might be worth turning on?

lib/new_relic/rack/browser_monitoring.rb
```diff
@@ -69,30 +69,33 @@ def should_instrument?(env, status, headers)
 
       private
 
-      def autoinstrument_source(response, headers, js_to_inject)
+      def autoinstrument_source(response, _headers, js_to_inject)
```
kaylareopelle, line 72: What do you think about cleaning up the callers and removing the headers arg altogether?

kaylareopelle (approved):
Thanks for taking care of this! I have a few suggestions/comments. 

Did you happen to check SimpleCov for the updated files? 

I ask because I think I may have forgotten last time I worked on a ticket like this and accidentally created a bug.

lib/new_relic/agent/agent_helpers/start_worker_thread.rb
fallwith: The short answer for why I didn't do so is that I'm trying to very tightly scope this PR's changes to RuboCop related things. I brought in changes in line length and quotes because there are cops for those things, but I didn't want to do any refactoring to the code beyond that.

For a summary of a longer answer, that thought did come to mind in this section of the code but each constant is mapped to a different method call and the names don't all follow a common convention. For example: `TRANSACTION_EVENT_DATA` maps to `transmit_analytic_event_data` which breaks the pattern of `transmit_<CONSTANT.downcase>. I also did not want to dive into a deep understanding of the requirements of what needs to take place in the context of an event loops `do` block. It's possible that something like `send(methodize(CONSTANT)` wouldn't work the same way. So I stuck to just satisfying the cop.

lib/new_relic/cli/commands/deployments.rb
fallwith: You can tell RuboCop to insist on the use of single quotes for strings without interpolation or you can tell RuboCop to insist on double quotes everywhere. I believe RuboCop currently defaults to the first one of those, but there has been well reasoned arguments for the second approach as well.

I believe we are currently disabling that cop and allowing any kind of quotes at any time regardless of interpolation.

If the devs decide on a standard, we can update the RuboCop config accordingly. Until then, I've been operating with my personal preference of RuboCop defaults.

lib/new_relic/rack/browser_monitoring.rb
fallwith: This was another case of me sticking to just RuboCop related stuff and not wanting to introduce any other refactoring that was out of scope and potentially bug-ridden. The reason for changing `headers` to `_headers` is that there's actually a cop that will insist on you doing so for unused variables. I'm not sure why it's not on for this project, but I figured I'd bring us in line with it while I was here.

For this specific method, it looks like it only has a single lone caller, so we should be able to do the refactor without too much risk.

lib/new_relic/agent/agent_helpers/start_worker_thread.rb
tannalynn: Yeah that makes sense to me. One thought I had is that in the future if we want to further refactor this we could put the constants into a hash that has the methods in it, that way we don't have to try to guess or interpolate what the method name is supposed to be.

tannalynn (approved):

lib/new_relic/cli/commands/deployments.rb
hannahramadan: For the record! I prefer single quotes. Double quotes alert me to interpolation, so I'd love to see this single quote cop enabled.

lib/new_relic/agent/agent_helpers/start_worker_thread.rb
kaylareopelle: Nice! I totally agree with making tightly scoped changes. This is great as-is.

lib/new_relic/cli/commands/deployments.rb
kaylareopelle: My preference is also single quotes unless in interpolation, though it can get hairy if we have apostrophes in our strings.

PR #1854, opened 2023-03-03
RuboCop AbcSize driven updates:

fallwith:
- refactor the logger's `call` method to adhere to AbcSize guidelines
- exempt `Transaction.new`
- exempt curb's chain based instrumentation

helps with #1520


tannalynn (approved):
This looks good, I like how it seems much neater now

PR #1856, opened 2023-03-06
Use !value.nil? instead of nil != value

kaylareopelle:
This change supports microbenchmarking tests on Ruby 3.0 and Ruby 2.7

### DIFF
--- a/lib/new_relic/agent/distributed_tracing/cross_app_tracing.rb
+++ b/lib/new_relic/agent/distributed_tracing/cross_app_tracing.rb
@@ -27,7 +27,7 @@ module NewRelic
       end
 
       def is_cross_app_callee?
-        cross_app_payload != nil
+        !cross_app_payload.nil?
       end
 
       def is_cross_app?
--- a/lib/new_relic/agent/transaction/synthetics_sample_buffer.rb
+++ b/lib/new_relic/agent/transaction/synthetics_sample_buffer.rb
@@ -13,7 +13,7 @@ module NewRelic
         end
 
         def allow_sample?(sample)
-          sample.synthetics_resource_id != nil
+          !sample.synthetics_resource_id.nil?
         end
 
         def truncate_samples


lib/new_relic/agent/distributed_tracing/cross_app_tracing.rb
```diff
@@ -27,7 +27,7 @@ def is_cross_app_caller?
       end
 
       def is_cross_app_callee?
-        cross_app_payload != nil
+        !cross_app_payload.nil?
```
fallwith, line 30: Or perhaps the object itself can be evaluated for truthiness and we could use `!!value` instead of `nil != value`:

```ruby
def is_cross_app_callee?
   !!cross_app_payload
end
```

kaylareopelle: That works in both locations, since it seems very unlikely either evaluated object would be `false`. 

However, running microbenchmarks, the `!value.nil?` strategy is more performant than `!!value`–sometimes slightly, sometimes significantly.

```ruby
require 'benchmark'
require 'benchmark/ips'
require 'benchmark/ipsa'

Benchmark.ipsa do |b|
  n = 1_000_000

  value1 = 1
  value2 = nil

  b.report('value != nil') do
    n.times do
      value1 != nil
      value2 != nil
    end
  end

  b.report('!value.nil?') do
    n.times do
      !value1.nil?
      !value2.nil?
    end
  end

  b.report('!!value') do
    n.times do
      !!value1
      !!value2
    end
  end
end
```

**Iteration 1**
```
Allocations -------------------------------------
        value != nil       0/0  alloc/ret        0/0  strings/ret
         !value.nil?       0/0  alloc/ret        0/0  strings/ret
             !!value       0/0  alloc/ret        0/0  strings/ret
Warming up --------------------------------------
        value != nil     1.000  i/100ms
         !value.nil?     1.000  i/100ms
             !!value     1.000  i/100ms
Calculating -------------------------------------
        value != nil      3.890  (± 0.0%) i/s -     20.000 
         !value.nil?     18.795  (± 0.0%) i/s -     94.000 
             !!value     18.267  (± 5.5%) i/s -     92.000 
```

**Iteration 2**
```
Allocations -------------------------------------
        value != nil       0/0  alloc/ret        0/0  strings/ret
         !value.nil?       0/0  alloc/ret        0/0  strings/ret
             !!value       0/0  alloc/ret        0/0  strings/ret
Warming up --------------------------------------
        value != nil     1.000  i/100ms
         !value.nil?     1.000  i/100ms
             !!value     1.000  i/100ms
Calculating -------------------------------------
        value != nil      4.014  (± 0.0%) i/s -     21.000 
         !value.nil?     18.540  (± 5.4%) i/s -     93.000 
             !!value     18.115  (± 0.0%) i/s -     91.000 
```

**Iteration 3**
```
Allocations -------------------------------------
        value != nil       0/0  alloc/ret        0/0  strings/ret
         !value.nil?       0/0  alloc/ret        0/0  strings/ret
             !!value       0/0  alloc/ret        0/0  strings/ret
Warming up --------------------------------------
        value != nil     1.000  i/100ms
         !value.nil?     1.000  i/100ms
             !!value     1.000  i/100ms
Calculating -------------------------------------
        value != nil      4.005  (± 0.0%) i/s -     21.000 
         !value.nil?     18.843  (± 0.0%) i/s -     95.000 
             !!value     17.803  (± 0.0%) i/s -     89.000 
```

kaylareopelle: This was ran using: `ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-darwin21]`

fallwith: If I replace `Benchmark.ipsa` with `Benchmark.bm`, the numbers seem to trend the other direction.

Here's the use of a simple stopwatch instead of `Benchmark`:

```ruby
value1 = 1
value2 = nil
n = 1_000_000

approaches = {'value != nil' => -> { value1 != nil; value2 != nil },
              '!value.nil?' => -> { !value1.nil?; !value2.nil? },
              '!!value' => -> { !!value1; !!value2 }}

approaches.each do |name, lambda|
  start = Time.now.to_f
  n.times { lambda.call }
  stop = Time.now.to_f
  puts "#{name} took #{stop - start} secs"
end
```

fallwith: > If I replace Benchmark.ipsa with Benchmark.bm, the numbers seem to trend the other direction.

Whoops! These are supposed to trend the other direction. `bm` = lower is better, and `ipsa` = higher is better. My mistake.

fallwith (approved):

PR #1857, opened 2023-03-07
Fix warnings in unit tests

tannalynn:
Fixes a couple of warnings related to the way our tests were written

There was a test that was the same name as another test (test_provides_accurate_info_for_an_instance_method)
```
/Users/tmcclure/projects/ruby-agent/test/new_relic/agent/method_tracer_helpers_test.rb:94: warning: method redefined; discarding old test_provides_accurate_info_for_an_instance_method
/Users/tmcclure/projects/ruby-agent/test/new_relic/agent/method_tracer_helpers_test.rb:82: warning: previous definition of test_provides_accurate_info_for_an_instance_method was here
```

There were 2 regexs used in tests that wanted it surrounded by parens 
```
/Users/tmcclure/projects/ruby-agent/test/new_relic/agent/tracer_test.rb:403: warning: ambiguity between regexp and two divisions: wrap regexp in parentheses or add a space after `/' operator
/Users/tmcclure/projects/ruby-agent/test/new_relic/agent/tracer_test.rb:416: warning: ambiguity between regexp and two divisions: wrap regexp in parentheses or add a space after `/' operator
```

### DIFF
--- a/test/new_relic/agent/tracer_test.rb
+++ b/test/new_relic/agent/tracer_test.rb
@@ -400,7 +400,7 @@ module NewRelic
             Thread.new { 'woof' }.join
           end
 
-          assert_match (/Ruby\/Thread\/Thread\d+\/Fiber\d+/), txn.segments.last.name
+          assert_match %r{Ruby/Thread/Thread\d+/Fiber\d+}, txn.segments.last.name
         end
       end
 
@@ -413,7 +413,7 @@ module NewRelic
             Thread.new { 'woof' }.join
           end
 
-          assert_match (/Ruby\/Thread$/), txn.segments.last.name
+          assert_match %r{Ruby/Thread$}, txn.segments.last.name
         end
       end
 
--- a/test/new_relic/agent/tracer_test.rb
+++ b/test/new_relic/agent/tracer_test.rb
@@ -400,7 +400,7 @@ module NewRelic
             Thread.new { 'woof' }.join
           end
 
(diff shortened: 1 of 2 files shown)


kaylareopelle (approved):
Thanks for cleaning this up!

test/new_relic/agent/tracer_test.rb
```diff
@@ -400,7 +400,7 @@ def test_thread_ids_included_when_enabled
             Thread.new { 'woof' }.join
           end
 
-          assert_match /Ruby\/Thread\/Thread\d+\/Fiber\d+/, txn.segments.last.name
+          assert_match (/Ruby\/Thread\/Thread\d+\/Fiber\d+/), txn.segments.last.name
```
fallwith, line 403: In both of these cases, due to the fact that the patterns themselves need to reference literal forward slashes, I think we should have used `%r{}` in the first place to not have to use the backslashes for escaping. For example:

```ruby
assert_match %r{Ruby/Thread/Thread\d+/Fiber\d+}, txn.segments.last.name
```

fallwith (approved):

test/new_relic/agent/tracer_test.rb
tannalynn: Great suggestion! I updated it to that in 4f4b507be7f803fcd10b2de3773e81ebe03b9e04

fallwith (approved):

PR #1858, opened 2023-03-07
errors inbox fingerprinting

### LINKED ISSUE #1845: Create a new set_error_group_callback public API and set a new error_group agent attribute
A new public `set_error_group_callback` API endpoint must be created in adherence with the `APIs/error_group.md` internal specifications doc.

The API must accept as its only argument an instance of a Ruby `Proc` which can be either a proc or lambda but will be referred to generically here as a proc.

The proc itself must accept as argument(s) - the exception class instance for a noticed error and as much other information the agent has at notice time such as the transaction name, the transaction user attributes, any web app info such as request URI, response code, and status, etc. It must return exactly one string value: the desired error group to associate the error with. See the `APIs/error_group.md` internal specifications doc for a full list of ideal inputs.

When the Ruby agent notices an error, it will check to see if it has an error group callback proc that was supplied by the customer. If so, it will invoke that callback with the error instance and expect to get back either a non empty string, a `nil`, or an empty string. If the agent receives a `nil` or an empty string, it discards these and takes no new action compared to what it currently does to process the noticed error. If the agent receives a non empty string, it will use the string as the value for a new `error_group` agent attribute that is sent up within the agent attributes hash for the noticed error.

fallwith:
Allow the customer to register a proc to be invoked on each noticed error in order to associate each noticed error with an error group

Changes made:

- A new public `set_error_group_callback` API method exists. It must be given a Ruby proc or lambda (must be of type `Proc`)
- When a noticed error is created by the error collector, it will invoke the callback with a hash of information about the error, and apply the resulting error group string (if returned) as the error group.
- Any noticed error with an error group property set will have that error group name present in its agent attributes with the key `error.group.name`.
- Internal errors (of which we only have a single stats hash related type currently) will now be routed through the error collector's method for creating a noticed error, thus ensuring the error group callback is invoked for the error.

resolves #1845

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,39 @@
 # New Relic Ruby Agent Release Notes
 
+## error_fingerprinting_and_user_tracking
+  
+  Version x.x.x of the agent delivers support for two new [errors inbox](https://docs.newrelic.com/docs/errors-inbox/errors-inbox/) features: error fingerprinting and user tracking.
+
+- **Error fingerprinting**
+
+  A new `set_error_group_callback` public API method now exists that will accept a user defined proc. The proc will be invoked for each noticed error and whenever it returns a string, that string will be used as the error group name for the error and will take precedence over any server-side grouping that takes place with the New Relic errors inbox. This gives users much greater control over the grouping of their errors.
+
+  The customer defined proc will be expected to receive exactly one input argument, a hash. The hash contains the following:
+
+  |  Key                 | Value                                                                        |
+  | ---------------------| ---------------------------------------------------------------------------- |
+  | `:error`             | The Ruby error class instance. Offers `#class`, `#message`, and `#backtrace` |
+  | `:customAttributes`  | Any customer defined customed attributed for the current transaction         |
+  | `:'request.uri'`     | The current requeset URI if available                                        |
+  | `:'http.statusCode'` | The HTTP status code (200, 404, etc.) if available                           |
+  | `:'http.method'`     | The HTTP method (GET, PUT, etc.) if available                                |
+  | `:'error.expected'`  | Whether (true) or not (false) the error was expected                         |
+  | `:'options'`         | The options hash passed to `NewRelic::Agent.notice_error                     |
+
+  The callback only needs to be set once per initialization of the New Relic agent.
+
+  Example usage:
+
+  ```
+  proc = proc { |hash| "Access" if hash[:'http.statusCode'].to_s == '401' }
+  NewRelic::Agent.set_error_group_callback(proc)
+  ```
+
+- **User tracking**
+
+  A new `set_user_id` public API method now exists that will accept a string representation of a user id and associate that user id with the current transaction. Transactions and errors will then have a new `enduser.id` agent attribute associated with them. This will allow agent users to tag transactions and errors as belonging to given user ids in support of greater filtering and alerting capabilities.
+
+
 ## dev
 
   Upcoming version removes Distributed Tracing warnings from agent logs when using Sidekiq, fixes a bug regarding logged request headers, and is tested against the recently released JRuby 9.4.2.0.
--- a/lib/new_relic/agent.rb
+++ b/lib/new_relic/agent.rb
@@ -333,6 +333,7 @@ module NewRelic
         return
       end
 
+      record_api_supportability_metric(:set_error_group_callback)
       @error_group_callback = callback_proc
     end
 
--- a/lib/new_relic/agent.rb
+++ b/lib/new_relic/agent.rb
@@ -105,6 +105,8 @@ module NewRelic
     # placeholder name used when we cannot determine a transaction's name
     UNKNOWN_METRIC = '(unknown)'.freeze
 
+    attr_reader :error_group_callback
+
     @agent = nil
     @error_group_callback = nil
     @logger = nil
--- a/lib/new_relic/agent.rb
+++ b/lib/new_relic/agent.rb
@@ -106,6 +106,7 @@ module NewRelic
     UNKNOWN_METRIC = '(unknown)'.freeze
 
     @agent = nil
+    @error_group_callback = nil
     @logger = nil
     @tracer_lock = Mutex.new
     @tracer_queue = []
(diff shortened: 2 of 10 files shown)


lib/new_relic/agent.rb
```diff
@@ -293,6 +294,46 @@ def notice_error(exception, options = {})
       nil # don't return a noticed error data structure. it can only hurt.
     end
 
+    # Set a callback proc for determining an error's error group name
```
kaylareopelle, line 297: Potential refactoring opportunity, though idk how it might impact Rdoc: 

This file is already very long. I think we've been adding things so that the APIs can all be accessed in the same way, but we could still do this if each API was in its own module that was included by NewRelic::Agent.

```ruby
module Child
	def child_method
		puts 'hi data!'
	end
end

module Parent
	extend self 
	include Child
end 

Parent.child_method 
# hi data!
# => nil                                                  
```

lib/new_relic/agent.rb
```diff
@@ -293,6 +294,46 @@ def notice_error(exception, options = {})
...
+    #
+    # @param [Proc] the callback proc
+    #
+    # Typically this method should be called only once to set a callback for
```
kaylareopelle, line 301: Should you mention that it should be called when the app initializes, but after the agent has loaded? Or does call time not matter?

lib/new_relic/agent.rb
```diff
@@ -293,6 +294,46 @@ def notice_error(exception, options = {})
...
+    # :'http.method' => The HTTP method (GET, PUT, etc.) if available
+    # :'error.expected' => Whether (true) or not (false) the error was expected
+    # :options => The options hash passed to `NewRelic::Agent.notice_error`
+    #
+    # @api public
+    #
+    def set_error_group_callback(callback_proc)
+      unless callback_proc.is_a?(Proc)
+        NewRelic::Agent.logger.error('The error group callback proc must be an instance of Proc, ' \
+                                     "got #{callback_proc.class}")
```
kaylareopelle, line 330: What about including the name of the method in the log message, so it might be easier to identify? 

ex: `The NewRelic::Agent.set_error_group_callback proc...`

lib/new_relic/noticed_error.rb
```diff
@@ -27,6 +27,9 @@ class NewRelic::NoticedError
 
   DESTINATION = NewRelic::Agent::AttributeFilter::DST_ERROR_COLLECTOR
 
+  AGENT_ATTRIBUTE_REQUEST_URI = :'request.uri'
+  AGENT_ATTRIBUTE_ERROR_GRUP = :'error.group.name'
```
kaylareopelle, line 31: lil typo "GRUP" => "GROUP"

lib/new_relic/agent.rb
fallwith: Call time does not matter. Essentially, as long as Ruby doesn't error out saying that it doesn't know what `NewRelic::Agent` is, you're ready to call it. We're just populating an instance var with the result of the call, so we're ready immediately.

lib/new_relic/agent.rb
fallwith: We have at least these concerns currently with `lib/new_relic/agent.rb`:

- The customer facing RDoc incorrectly lists class methods as instance ones. For example, `NewRelic::Agent.notice_error` gets listed as `NewRelic::Agent#notice_error`.
- There are multiple classes being defined in the file
- The file has grown a bit unwieldy for devs to maintain, and could potentially benefit from being broken up a bit

There is, however, a big benefit that I can see for keeping all of the public methods inside the file, and that's that it will make them easier to find for customers. If you are told about a public API method, `NewRelic::Agent.set_error_group_callback`, you will start your code search at `lib/new_relic/agent.rb` looking for a `set_error_group_callback` method.

In any event, I'd like to see any refactoring of this class discussed and then tackled with a dedicated separate story.

lib/new_relic/noticed_error.rb
fallwith: Thank you! Addressed with e4657bfb143a8199401740a0cd2259d28c79be23

lib/new_relic/agent.rb
fallwith: Thank you! Addressed (while pairing) with e4657bfb143a8199401740a0cd2259d28c79be23

kaylareopelle:
> The logic to update the transaction's agent attributes with the noticed error's request.uri was modified to also update those attributes with the error.group.name value.

@fallwith  - Is this still part of the PR?

lib/new_relic/supportability_helper.rb
```diff
@@ -44,6 +44,7 @@ module SupportabilityHelper
       :record_metric,
       :recording_web_transaction?,
       :require_test_helper,
+      :set_error_group_callabck,
```
kaylareopelle, line 47: Small typo:

```suggestion
      :set_error_group_callback,
```

kaylareopelle (commented):
I haven't had a chance to think through the tests yet, but will get back to those on Monday!

CHANGELOG.md
```diff
@@ -1,5 +1,39 @@
 # New Relic Ruby Agent Release Notes
 
+## error_fingerprinting_and_user_tracking
```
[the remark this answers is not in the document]
fallwith: Erik and I met with the docs and errors inbox teams and it was agreed that bits like this would be handled by the agents.

lib/new_relic/agent/error_collector.rb
```diff
@@ -306,6 +308,32 @@ def drop_buffered_data
         @error_event_aggregator.reset!
         nil
       end
+
+      private
+
+      def update_error_group_name(noticed_error, exception, options)
+        return unless error_group_callback
```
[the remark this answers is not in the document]
fallwith: The way it works is that we can be assured that this check will return either `nil` or a proc. `NewRelic::Agent.error_group_callback` method won't accept anything but a proc.

lib/new_relic/agent/error_trace_aggregator.rb
```diff
@@ -81,9 +81,10 @@ def notice_agent_error(exception)
           # Already seen this class once? Bail!
           return if @errors.any? { |err| err.exception_class_name == exception.class.name }
 
-          trace = exception.backtrace || caller.dup
-          noticed_error = NewRelic::NoticedError.new('NewRelic/AgentError', exception)
-          noticed_error.stack_trace = trace
+          noticed_error = NewRelic::Agent.instance.error_collector.create_noticed_error(exception,
+            {metric: 'NewRelic/AgentError'})
+          noticed_error.stack_trace = caller.dup unless exception.backtrace
```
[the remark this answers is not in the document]
fallwith: The original objective was to use `caller.dup` in lieu of a backtrace if the given exception object didn't have one. That objective is still upheld. The noticed error object is now created via `NewRelic::Agent.instance.error_collector.create_noticed_error` so that we can rely on the error group being applied as desired. That method handles the setting of the noticed error's `stack_trace` equal to the exception's backtrace. Then once we've got a return value back from the method call, we still want to apply the `caller.dup` value if the backtrace is `nil` for the exception. NOTE that we unfortunately cannot do this:

```ruby
noticed_error.stack_trace ||= caller.dup
```

because `noticed_error` will always return a string placeholder when the backtrace is missing.

fallwith:
> > The logic to update the transaction's agent attributes with the noticed error's request.uri was modified to also update those attributes with the error.group.name value.
> 
> @fallwith - Is this still part of the PR?

Good call. We landed on a decision around that bit of code in separate PR #1871. Removed the entry from this PR's description, thank you.

kaylareopelle (approved):

Issue #1847, opened 2023-02-28
Create a new set_user_id public API and set a new enduser.id agent attribute


fallwith:
A new public `set_user_id` API endpoint must be created in adherence with the `APIs/user_tracking.md` internal specifications doc.

The API must accept as its only argument a string representing a unique identifier for one of the customer's users. This string can be a UUID, a string representation of a unique database id, or other unique identifier. Upon receiving this value, the agent will associate the given user id with the current transaction.

When the agent notices an error, an `enduser.id` agent attribute will be present in the agent attributes hash. Its value will be the string value passed to the `set_user_id` API method. If the API method has not been invoked for the current transaction, or if a `nil` or empty string (`''`) value has been supplied to the method, then the `enduser.id` attribute will be left out of the agent attributes hash entirely.

fallwith:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/APM-Agent-Overview-614b054ff9a50b000f9039d7/issues/newrelic/newrelic-ruby-agent/1847?planning-poker) with Zenhub @hannahramadan @kaylareopelle @tannalynn

fallwith:
Addressed by #1868

PR #1868, opened 2023-03-07
Error inbox add set_user_id API method

tannalynn:
Adds the api method `NewRelic::Agent.set_user_id`.

closes https://github.com/newrelic/newrelic-ruby-agent/issues/1847

### DIFF
--- a/lib/new_relic/agent.rb
+++ b/lib/new_relic/agent.rb
@@ -605,7 +605,7 @@ module NewRelic
       end
     end
 
-    # Set the user id for the current transaction
+    # Set the user id for the current transaction. When present, this value will be included in the agent attributes for transaction and error events as 'enduser.id'.
     #
     # @param [String] user_id    The user id to add to the current transaction attributes
     #
--- a/lib/new_relic/agent.rb
+++ b/lib/new_relic/agent.rb
@@ -605,6 +605,26 @@ module NewRelic
       end
     end
 
+    # Set the user id for the current transaction
+    #
+    # @param [String] user_id    The user id to add to the current transaction attributes
+    #
+    # @api public
+    def set_user_id(user_id)
+      record_api_supportability_metric(:set_user_id)
+
+      if user_id.nil? || user_id.empty?
+        ::NewRelic::Agent.logger.warn('NewRelic::Agent.set_user_id called with a nil or empty user id.')
+        return
+      end
+
+      default_destinations = NewRelic::Agent::AttributeFilter::DST_TRANSACTION_TRACER |
+        NewRelic::Agent::AttributeFilter::DST_TRANSACTION_EVENTS |
+        NewRelic::Agent::AttributeFilter::DST_ERROR_COLLECTOR
+
+      NewRelic::Agent::Transaction.add_agent_attribute('enduser.id', user_id, default_destinations)
+    end
+
     # @!endgroup
 
     # @!group Transaction naming
--- a/lib/new_relic/agent.rb
+++ b/lib/new_relic/agent.rb
@@ -622,7 +622,7 @@ module NewRelic
         NewRelic::Agent::AttributeFilter::DST_TRANSACTION_EVENTS |
         NewRelic::Agent::AttributeFilter::DST_ERROR_COLLECTOR
 
-      NewRelic::Agent::Transaction.add_agent_attribute('enduser.id', user_id, default_destinations)
+      NewRelic::Agent::Transaction.add_agent_attribute(:'enduser.id', user_id, default_destinations)
     end
 
     # @!endgroup
--- a/lib/new_relic/supportability_helper.rb
+++ b/lib/new_relic/supportability_helper.rb
@@ -46,6 +46,7 @@ module NewRelic
       :require_test_helper,
       :set_sql_obfuscator,
       :set_transaction_name,
+      :set_user_id,
       :shutdown,
       :start_segment,
       :trace,
--- a/test/new_relic/agent_test.rb
+++ b/test/new_relic/agent_test.rb
@@ -73,6 +73,9 @@ module NewRelic
         refute NewRelic::Agent.config[:'transaction_tracer.enabled']
         refute NewRelic::Agent.config[:'error_collector.enabled']
       end
+
+      # resets the SSC to empty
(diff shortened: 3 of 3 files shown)


lib/new_relic/agent.rb
```diff
@@ -605,6 +605,26 @@ def add_custom_span_attributes(params)
       end
     end
 
+    # Set the user id for the current transaction
```
kaylareopelle, line 608:
```suggestion
    # Set the user id for the current transaction. When present, this value will be included in the agent attributes for transaction and error events as 'enduser.id'.
```

kaylareopelle (commented):
Great work! I thought of something small to add to the API description.

test/new_relic/agent_test.rb
```diff
@@ -529,6 +532,32 @@ def test_modules_and_classes_return_name_properly
...
+        NewRelic::Agent.set_user_id(test_user)
+        NewRelic::Agent.notice_error(NewRelic::TestHelpers::Exceptions::TestError.new)
+      end
+
+      [harvest_transaction_events!, harvest_error_events!].each do |events|
+        assert_equal events[1][0][2]['enduser.id'], test_user
+      end
+    end
+
+    def test_set_user_id_nil_or_empty_error
```
hannahramadan, line 548: Never seen a combo test like this! Pretty cool

kaylareopelle (dismissed):
🎉

lib/new_relic/agent.rb
```diff
@@ -605,6 +605,26 @@ def add_custom_span_attributes(params)
       end
     end
 
+    # Set the user id for the current transaction. When present, this value will be included in the agent attributes for transaction and error events as 'enduser.id'.
+    #
+    # @param [String] user_id    The user id to add to the current transaction attributes
+    #
+    # @api public
+    def set_user_id(user_id)
+      record_api_supportability_metric(:set_user_id)
```
fallwith, line 614: Given that we expect that `set_user_id` will be called very frequently (at least once per transaction), I think incrementing can end up creating a lot of egress. Perhaps we should simply record once, so that we know the feature is being used without concern for how many invocations.

```diff
+ SET_USER_ID_SUPPORTABILITY_METRIC = "#{NewRelic::SupportabilityHelper::API_SUPPORTABILITY_PREFIX}set_user_id/Invoked"


- record_api_supportability_metric(:set_user_id)
+ record_metric_once(SET_USER_ID_SUPPORTABILITY_METRIC)
```

lib/new_relic/supportability_helper.rb
```diff
@@ -46,6 +46,7 @@ module SupportabilityHelper
       :require_test_helper,
       :set_sql_obfuscator,
       :set_transaction_name,
+      :set_user_id,
```
fallwith, line 49: If we're only going to record once, this line can go away entirely.

lib/new_relic/agent.rb
```diff
@@ -605,6 +605,26 @@ def add_custom_span_attributes(params)
...
+      if user_id.nil? || user_id.empty?
+        ::NewRelic::Agent.logger.warn('NewRelic::Agent.set_user_id called with a nil or empty user id.')
+        return
+      end
+
+      default_destinations = NewRelic::Agent::AttributeFilter::DST_TRANSACTION_TRACER |
+        NewRelic::Agent::AttributeFilter::DST_TRANSACTION_EVENTS |
+        NewRelic::Agent::AttributeFilter::DST_ERROR_COLLECTOR
+
+      NewRelic::Agent::Transaction.add_agent_attribute('enduser.id', user_id, default_destinations)
```
fallwith, line 625:
```suggestion
      NewRelic::Agent::Transaction.add_agent_attribute(:'enduser.id', user_id, default_destinations)
```

This is not something explicitly spelled out in the spec because it's Ruby specific, but the attribute should be a symbol. Once we've merged this, we can go back into the agent spec (which has a TBD section for agent specific implementations) and make a note there.

test/new_relic/agent_test.rb
```diff
@@ -529,6 +532,32 @@ def test_modules_and_classes_return_name_properly
...
+    def test_set_user_id_attribute
+      test_user = 'test_user_id'
+
+      in_transaction do |txn|
+        NewRelic::Agent.set_user_id(test_user)
+        NewRelic::Agent.notice_error(NewRelic::TestHelpers::Exceptions::TestError.new)
+      end
+
+      [harvest_transaction_events!, harvest_error_events!].each do |events|
+        assert_equal events[1][0][2]['enduser.id'], test_user
```
fallwith, line 544:
```suggestion
        assert_equal events[1][0][2][:'enduser.id'], test_user
```

test/new_relic/agent_test.rb
```diff
@@ -529,6 +532,32 @@ def test_modules_and_classes_return_name_properly
...
+
+    def test_set_user_id_nil_or_empty_error
+      [nil, ''].each do |test_user_id|
+        in_transaction do |txn|
+          NewRelic::Agent.set_user_id(test_user_id)
+          NewRelic::Agent.notice_error(NewRelic::TestHelpers::Exceptions::TestError.new)
+        end
+
+        [harvest_transaction_events!, harvest_error_events!].each do |events|
+          refute events[1][0][2].key?('enduser.id')
```
fallwith, line 556:
```suggestion
          refute events[1][0][2].key?(:'enduser.id')
```

fallwith (changes requested):

lib/new_relic/agent.rb
tannalynn: That makes sense, my concern with that would be it would cause inconsistencies with the way we record this metric compared to all other api method metrics. We have a lot of other API methods that can also be called once or more per transaction, so this method wouldn't be doing anything unusually time consuming as is.
If this is a change we want to make, I think we should make this change to all API methods to keep things consistent.

fallwith: Great call on consistency, @tannalynn. We discussed this one more offline and will keep the original approach here as-is.

lib/new_relic/supportability_helper.rb
fallwith: We've decided to record every time, as we do with all other API calls.

hannahramadan (approved):

kaylareopelle (approved):

PR #1871, opened 2023-03-09
NoticedError: remove defunct request.uri feature

fallwith:
[1fcfbdd2f7970b68182895c5c792a8c288a01b53](https://github.com/newrelic/newrelic-ruby-agent/commit/1fcfbdd2f7970b68182895c5c792a8c288a01b53) introduced some custom logic to `NoticedError` to treat the request URI in a special manner.

[9f7636e5072db0c5d63257be59ec4701e16348de](https://github.com/newrelic/newrelic-ruby-agent/commit/9f7636e5072db0c5d63257be59ec4701e16348de) removed this custom logic, but did not do enough clean up. A comment and a code block remained that will no longer effect the noticed error's agent attributes hash.

Removed the defunct code.

### DIFF
--- a/lib/new_relic/noticed_error.rb
+++ b/lib/new_relic/noticed_error.rb
@@ -141,21 +141,9 @@ class NewRelic::NoticedError
   end
 
   def build_agent_attributes(merged_attributes)
-    agent_attributes = if @attributes
-      @attributes.agent_attributes_for(DESTINATION)
-    else
-      NewRelic::EMPTY_HASH
-    end
+    return NewRelic::EMPTY_HASH unless @attributes
 
-    # It's possible to override the request_uri from the transaction attributes
-    # with a uri passed to notice_error. Add it to merged_attributes filter and
-    # merge with the transaction attributes, possibly overriding the request_uri
-    if request_uri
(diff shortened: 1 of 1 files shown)


kaylareopelle (approved):
Thanks, James! Nice to see code cleanup was all that needed to happen. What are your thoughts about a CHANGELOG entry for this work?

fallwith:
> Thanks, James! Nice to see code cleanup was all that needed to happen. What are your thoughts about a CHANGELOG entry for this work?

I'm currently thinking to not create an entry for it. I recognize this as yet more cleanup (hopefully the last bit of cleanup!) for request URI handling functionality that began with https://github.com/newrelic/newrelic-ruby-agent/commit/e8b5240772be46460a5517464239c9c21dfbd424 and was later reverted - all without CHANGELOG entries previously.

I could also see value in having a CHANGELOG entry present in the event of a bug being introduced and giving customers something to associate the bug with, but I'm not worried about that in this case. A conditional and 2 lines of code inside it were removed. Those 2 lines a) added content to an internal hash that was not being read from again (because the method that reads from it was previously reverted out and b) spawning a 3rd hash from using `#merge` with 2 others, only to never store that 3rd hash anywhere.

PR #1882, opened 2023-03-13
NoticedError: update attr cache for error group

fallwith:
`NewRelic::Agent::ErrorCollector#create_noticed_error`, which is the sole consumer of `NoticedError#error_group=`, ends up calling `NoticedError#processed_attributes` fully 3 times in order to build the hash that is passed to the customer error group callback.

As a result, the agent attributes for the noticed error end up being memoized. Because of the memoization, the agent attributes hash is not rebuilt later on when it comes time to submit those attributes to the collector.

Updated `NoticedError#error_group=` to effectively bust the memoization cache and ensure that the error group name makes it into the agent attributes hash.

### DIFF
--- a/lib/new_relic/noticed_error.rb
+++ b/lib/new_relic/noticed_error.rb
@@ -143,15 +143,9 @@ class NewRelic::NoticedError
   end
 
   def build_agent_attributes(merged_attributes)
-    return error_group_hash unless @attributes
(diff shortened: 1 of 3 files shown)


tannalynn (approved):
🎉

kaylareopelle (approved):
🍵

PR #1895, opened 2023-03-17
Decrease number of middle versions of tested gems

tannalynn:
this PR removes some middle versions of gems that have the same behavior from our multiverse testing. Since our instrumentation and tests don't change for these versions, we don't need to test all of them since the code we're instrumenting doesn't change between these versions.
I checked to make sure we didn't have any code in our instrumentation or tests that would change for different versions, so these should be good to remove.


fallwith (dismissed):

test/multiverse/suites/curb/Envfile
```diff
@@ -40,13 +40,4 @@ gemfile <<~RB
   gem 'curb', '#{curb_gem_version}'
   gem 'rack'
   gem 'json', :platforms => [:rbx, :mri_18]
-
-  
-RB
-
-gemfile <<~RB
```
tannalynn, line 47: This one was just a exact duplicate of the one above, so we were running all the curb tests twice for the same version

test/multiverse/suites/activemerchant/Envfile
```diff
@@ -4,8 +4,6 @@
 
 ACTIVEMERCHANT_VERSIONS = [
   [nil, 3.0],
-  ['1.121.0', 2.5, 3.0],
-  ['1.106.0', 2.4, 3.0],
   ['1.65.0', 2.4, 2.4]
```
kaylareopelle, line 7: This'd probably be a different ticket, but I wonder if we can take off the 2.4 constraints when we see them now since we don't have any versions less than 2.4 in the mix? I forget how exactly this works.

test/multiverse/suites/curb/Envfile
kaylareopelle: :sweat_smile: thanks for catching this!

hannahramadan (approved):

test/multiverse/suites/activemerchant/Envfile
tannalynn: I think if the max version we want it to run is 2.4, we still have to include it like `2.4, 2.4` so it sees both the min and max, since passing it just one ruby version uses it as the minimum version. I think that's the same for any version we are specifying a max ruby version for.

test/multiverse/suites/activemerchant/Envfile
```diff
@@ -4,8 +4,6 @@
 
 ACTIVEMERCHANT_VERSIONS = [
   [nil, 3.0],
```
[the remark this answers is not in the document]
tannalynn: Oh good catch, I had missed noticing that the min version here was 3.0. I'm not sure why it would be 3.0 though? I just tested running it on 2.6 locally and it worked fine, so i'll update that to 2.5 and see how that does on the CI. 
172d6b3fc3e34ed46154722da0c3b5ea097923bb

kaylareopelle (approved):

PR #1896, opened 2023-03-20
JavaScriptInstrumentor: log about custom attrs

fallwith:
Add debug level logging statements to the JavaScriptInstrumentor's handling of custom attributes.

### DIFF
--- a/lib/new_relic/agent/javascript_instrumentor.rb
+++ b/lib/new_relic/agent/javascript_instrumentor.rb
@@ -185,9 +185,14 @@ module NewRelic
 
       def append_custom_attributes!(txn, atts)
         custom_attributes = txn.attributes.custom_attributes_for(NewRelic::Agent::AttributeFilter::DST_BROWSER_MONITORING)
-        unless custom_attributes.empty?
-          atts[ATTS_USER_SUBKEY] = custom_attributes
+        if custom_attributes.empty?
+          NewRelic::Agent.logger.debug("#{self.class}: No custom attributes found to append.")
+          return
         end
+
+        NewRelic::Agent.logger.debug("#{self.class}: Appending the following custom attribute keys for browser " \
+          "monitoring: #{custom_attributes.keys}")
+        atts[ATTS_USER_SUBKEY] = custom_attributes
       end
 
       def append_agent_attributes!(txn, atts)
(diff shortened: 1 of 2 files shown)


lib/new_relic/agent/javascript_instrumentor.rb
```diff
@@ -185,9 +185,14 @@ def add_attributes(data, txn)
...
         custom_attributes = txn.attributes.custom_attributes_for(NewRelic::Agent::AttributeFilter::DST_BROWSER_MONITORING)
-        unless custom_attributes.empty?
-          atts[ATTS_USER_SUBKEY] = custom_attributes
+        if custom_attributes.empty?
+          NewRelic::Agent.logger.debug("#{self.class}: No custom attributes found to append.")
+          return
         end
+
+        NewRelic::Agent.logger.debug("#{self.class}: Appending the following custom attribute keys for browser " \
+          "monitoring: #{custom_attributes.keys}")
```
kaylareopelle, line 194: Do you think there's a chance the arg might not be a hash, but a string? I'm wondering if we want a safe navigation operator here.

fallwith: Thanks for checking on this - we definitely don't want to introduce bugs in efforts to add debug logs.

My current take is that we're good to go as-is. Here's my reasoning:

- `&.` will guard against `nil`, but not an unexpected type. `non_hash&.keys` would still error out. If we're concerned about type, we'll need more than `&.`
- Zooming out to regard the `append_custom_attributes!` method on the whole, I see that we already perform `custom_attributes.empty?` with code that predates this PR. That `#empty?` check would fail if called on either `nil` or an object that doesn't offer an `#empty?` method.
-  The new debug statement won't be reached until the `#empty?` check is satisfied.

If we're still concerned about `custom_attributes` being `nil` or non-hash, those concerns will involve code that predates this PR given the existing `#empty?` call. Let me know if you'd like me to address that with a separate PR, but I think this one won't make things any worse and is good to go.

kaylareopelle: I think this is fine to merge without further checks. Thank you for your thoughtful and well-researched response!

kaylareopelle (approved):

PR #1916, opened 2023-03-31
Rubocop updates

kaylareopelle:
Innovation meeting project: 
* Audit of some linters turned off in the initial run and addressing them. These changes are all from safe autocorrect.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -605,11 +605,6 @@ Lint/EmptyClass:
 Lint/EmptyConditionalBody:
   Enabled: false
 
-# Disabling for now
-# Lint/EmptyEnsure:
-#   Enabled: false
-#   # AutoCorrect: true
-
 Lint/EmptyExpression:
   Enabled: true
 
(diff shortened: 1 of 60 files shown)


.rubocop.yml
```diff
@@ -661,9 +606,9 @@ Lint/EmptyConditionalBody:
   Enabled: false
 
 # Disabling for now
-Lint/EmptyEnsure:
-  Enabled: false
-  # AutoCorrect: true
+# Lint/EmptyEnsure:
```
hannahramadan, line 609: Do we want to remove this entirely vs comment out?

test/performance/lib/performance/result.rb
```diff
@@ -6,8 +6,8 @@
 
 module Performance
   class Result
-    attr_reader :test_name, :measurements, :tags, :timer, :artifacts
-    attr_accessor :exception, :iterations
+    attr_reader :test_name, :measurements, :tags, :timer, :artifacts, :exception
```
fallwith, line 9: I'm surprised there isn't a "these should be alphabetized" cop in play.

fallwith (approved):

test/performance/lib/performance/result.rb
kaylareopelle: It might be on the list of disabled cops that we didn't get to yesterday!

tannalynn (approved):

.rubocop.yml
kaylareopelle: Great catch! Fixed in: 03134c1433aa21e2ffef9cc67cc47ef1ce3f92f4

fallwith (approved):

PR #1922, opened 2023-04-04
Add ::Chain to end of instrumentation class names

kaylareopelle:
`DependencyDetection#log_and_instrument` takes the second to last namespace and uses it in the
`Supportability/instrumention/#{supportability_name}/#{chain|prepend}` metrics

This was common in new instrumentation because the instrumentation generator templates didn't include `::Chain` (and the class I based the generator off of didn't have this either!)


tannalynn (approved):
Thank you!

fallwith (approved):

PR #1925, opened 2023-04-06
Slack notification update

hannahramadan:
Update Ruby setup.

From [setup-ruby docs](https://github.com/ruby/setup-ruby):
> Important: Prefer ruby/setup-ruby@v1. If you pin to a commit or release, only the Ruby versions available at the time of the commit will be available, and you will need to update it to use newer Ruby versions, see [Versioning](https://github.com/ruby/setup-ruby#versioning).

To keep up-to-date with new ruby versions, we stick to using us setup-ruby@v1

### DIFF
--- a/.github/workflows/slack_notifications.yml
+++ b/.github/workflows/slack_notifications.yml
@@ -8,7 +8,7 @@ jobs:
   gem_notifications:
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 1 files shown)


kaylareopelle (approved):

hannahramadan:
Closing in favor of #1926

PR #1926, opened 2023-04-06
Update all GHA to use SHAs

### LINKED ISSUE #1901: CI: Update all YAML files to reference 3rd party actions by sha
I believe we already have every 3rd party GitHub Actions action referenced by version. Let's go one step further and pin them to specific shas.

hannahramadan:
As part of security updates, all Actions versions now use a full-length SHA. 

Each one is using the latest stable version.

closes #1901

### DIFF
--- a/.github/workflows/repolinter.yml
+++ b/.github/workflows/repolinter.yml
@@ -15,17 +15,17 @@ jobs:
     steps:
       - name: Test Default Branch
         id: default-branch
-        uses: actions/github-script@v2
+        uses: actions/github-script@d7906e4ad0b1822421a7e6a35d5ca353c962f410 # tag v6.4.1
         with:
           script: |
-            const data = await github.repos.get(context.repo)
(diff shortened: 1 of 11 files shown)


.github/workflows/repolinter.yml
```diff
@@ -15,17 +15,17 @@ jobs:
     steps:
       - name: Test Default Branch
         id: default-branch
-        uses: actions/github-script@v2
+        uses: actions/github-script@d7906e4ad0b1822421a7e6a35d5ca353c962f410 # tag v6.4.1
         with:
           script: |
-            const data = await github.repos.get(context.repo)
+            const data = await github.rest.repos.get(context.repo)
```
hannahramadan, line 21: A [breaking change in v5](https://github.com/actions/github-script#breaking-changes-in-v5) moves where methods live.

fallwith (approved):
It's great to have these locked down and I really appreciate the tag version comments. Nicely done!

PR #1927, opened 2023-04-07
Efficiency improvements for segment completion

### LINKED ISSUE #1345: newrelic_rpm slows down requests *4x* when used with graphql dataloader
## Description

Following on to https://github.com/newrelic/newrelic-ruby-agent/issues/1264, I deployed v8.10.0 to production. Then I deployed, in isolation, `use GraphQL::Tracing::NewRelicTracing, set_transaction_name: true`. I found that, for key queries in our app, performance degraded from 1-2 seconds to 6-7s, about a 4x slowdown. This is better than 10x, but still unacceptable for production use.

*Suggested Priority (P1,P2,P3,P4,P5): P5 (until more info is received)*
*Suggested T-Shirt size (S, M, L, XL, Unknown): Unknown*

### LINKED ISSUE #1910: Thread tracing causes process to hang
## Description
We recently attempted to upgrade the version of the agent we use from 8.16 to 9.0 and ran into an issue where, after our code had completed a Sidekiq job, the process would be spinning at 100% CPU usage seemingly indefinitely.

We were able to prevent this issue by disabling the thread tracing (setting `instrumentation.thread.tracing` to `false`), so it seems like this issue may have been present for a while but only manifested with the automatic enabling of thread tracing that was included in 9.0.

Notably, this only seems to occur for a particular job that we have. We use Sidekiq for job processing, and this issue would manifest for a particularly long-running Sidekiq job that made many database requests. Reducing the number of DB requests made by that job prevented this issue from manifesting.

## Expected Behavior
We expect the New Relic agent to publish the traces quickly and without using 100% of a CPU.

## Steps to Reproduce
I wish I had more specific steps for you here — the job we're looking at takes over ten minutes to run so it's a long debug cycle. We were not able to reproduce the issue when running the job outside of Sidekiq, so it seems like maybe some interaction of the threading used by Sidekiq with the thread tracing done in NewRelic.

If this seems worth digging into then I can look further into creating a reproducible example, let me know!

## Your Environment
Running on Heroku 18 (Ubuntu 18.04).
New Relic Ruby Agent: 9.0.0
Ruby: 3.1.2
Sidekiq: 6.5.6

fallwith:
When lots of instrumented activity takes place nested within an outer instrumented action, the number of child segments belonging to a parent segment may reach a quantity (likely 10,000 or higher) that causes the agent to consume considerable CPU resources and either take a long time or hang indefinitely while processing the child segments at parent segment finish time.

Examples of nested activity are:

- Instrumented data store operations that occur within an instrumented background job process
- Bulk data load operations that take place during a public API method call to trace the execution of a block of code

Version 9.0 of the agent began enabling the
`instrumentation.thread.tracing` config parameter by default. The agent's `Thread` based tracing controlled by this parameter allows the agent to record metrics for activity taking place within Ruby threads that would otherwise be hidden from view of the agent.

After studying exactly what the agent does with child segments in quantities of 20,000 to 100,000, the agent's maintainers noticed that the parent segment spends a lot of time looking for start/finish time overlaps between all child segments and itself. This is critically important for the creation of accurate timing information when child segments operate concurrently, so as to not double report the time spent in the overlap periods. But when large quantities of child segments are present to be processed - even if they don't feature any overlap at all in timing - the inefficiencies in the overlap calculations can be lengthy and resource intensive.

With the changes included here, the process of calculating child segment overlap is now thought to be dramatically more efficient without sacrificing any accuracy.

Relates to #1910 
Relates to #1345


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## dev
 
+  Version <tbd> of the agent introduces some performance improvements for working with high numbers of nested actions, and deprecates instrumentation for the `memcached` and `memcache-client` gems (with `dalli` still being supported).
+
+- **Feature: Performance enhancements for handling high numbers of nested actions
```
```suggestion
- **Feature: Performance enhancements for handling high numbers of nested actions**
```

kaylareopelle (approved):
🍵

PR #1930, opened 2023-04-07
auto convert :auto to :prepend/:chain in config

### LINKED ISSUE #1924: Update all instrumentation related config values after dependency detection to convert "auto" to "prepend" or "chain"
Our `instrumentation.*` params, such as `instrumentation.net_http` currently support four possible values: `disabled`, `auto`, `prepend`, and `chain`.

The `auto` value (the default) sees the agent attempt to dynamically determine whether method prepending or chaining should be used. After this automated determination has been made and the relevant method(s) have been updated, it would be great to be able to query the agent configuration afterwards and introspect whether "chain" or "prepend" was selected.

Given that the "auto" value serves no purpose post dependency detection time, is the default, and can't be set via a server side configuration change, it might be best to simply convert the "auto" value into either a "prepend" or "chain" value so that querying the config for any given instrumentation will expose the method updating approach that was taken.

fallwith:
For `:'instrumentation.*'` configuration parameters that are set to :auto (the default), the agent will automatically determine whether to use method prepending or chaining. The agent will now update its in-memory configuration to give each relevant parameter a value of either :prepend or :chain so that the result of the determination can be introspected. This is intended to help 3rd party libraries that wish to further enhance the agent's instrumentation capabilities by prepending or chaining additional logic. Environment variable, YAML file, and server-side configuration based values are not impacted.

resolves #1924

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## dev
 
-- **Feature: The agent configuration will now reflect whether method prepending or chaining was used per instrumentation
+- **Feature: The agent configuration will now reflect whether method prepending or chaining was used per instrumentation**
 
   For `:'instrumentation.*'` configuration parameters that are set to :auto (the default), the agent will automatically determine whether to use method prepending or chaining. The agent will now update its in-memory configuration to give each relevant parameter a value of either :prepend or :chain so that the result of the determination can be introspected. This is intended to help 3rd party libraries that wish to further enhance the agent's instrumentation capabilities by prepending or chaining additional logic. Environment variable, YAML file, and server-side configuration based values are not impacted. [PR#1930](https://github.com/newrelic/newrelic-ruby-agent/pull/1930)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## dev
 
+- **Feature: The agent configuration will now reflect whether method prepending or chaining was used per instrumentation
+
+  For `:'instrumentation.*'` configuration parameters that are set to :auto (the default), the agent will automatically determine whether to use method prepending or chaining. The agent will now update its in-memory configuration to give each relevant parameter a value of either :prepend or :chain so that the result of the determination can be introspected. This is intended to help 3rd party libraries that wish to further enhance the agent's instrumentation capabilities by prepending or chaining additional logic. Environment variable, YAML file, and server-side configuration based values are not impacted. [PR#1930](https://github.com/newrelic/newrelic-ruby-agent/pull/1930)
+
 - **Feature: Deprecate memcached and memcache-client instrumentation**
 
   Instrumentation for the memcached and memcache-client libraries is deprecated and will be removed during the next major release.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## dev
 
-- **Feature: The agent configuration will now reflect whether method prepending or chaining was used per instrumentation**
+- **Feature: The agent configuration will now reflect whether module prepending or method chaining was used for instrumentation**
 
-  For `:'instrumentation.*'` configuration parameters that are set to :auto (the default), the agent will automatically determine whether to use method prepending or chaining. The agent will now update its in-memory configuration to give each relevant parameter a value of either :prepend or :chain so that the result of the determination can be introspected. This is intended to help 3rd party libraries that wish to further enhance the agent's instrumentation capabilities by prepending or chaining additional logic. Environment variable, YAML file, and server-side configuration based values are not impacted. [PR#1930](https://github.com/newrelic/newrelic-ruby-agent/pull/1930)
+  For `:'instrumentation.*'` configuration parameters that are set to :auto (the default), the agent will automatically determine whether to use module prepending or method chaining. The agent will now update its in-memory configuration to give each relevant parameter a value of either :prepend or :chain so that the result of the determination can be introspected. This is intended to help 3rd party libraries that wish to further enhance the agent's instrumentation capabilities by prepending or chaining additional logic. Environment variable, YAML file, and server-side configuration based values are not impacted. [PR#1930](https://github.com/newrelic/newrelic-ruby-agent/pull/1930)
 
 - **Feature: Deprecate memcached and memcache-client instrumentation**
 
--- a/lib/new_relic/dependency_detection.rb
+++ b/lib/new_relic/dependency_detection.rb
@@ -179,6 +179,15 @@ module DependencyDetection
       return valid_value
     end
 
+    # update any :auto config value to be either :prepend or :chain after auto
+    # determination has selected one of those to use
+    def update_config_value(use_prepend)
+      if config_key && auto_configured?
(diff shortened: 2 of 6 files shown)


prateek-ap:
@fallwith There is a bug in the [replace_or_add_config](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/configuration/manager.rb#L78) method which updates the config when there is any new config is received from server source or any other source. 
[evaluate_and_apply_transformations](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/configuration/manager.rb#L142) method always updates from Default source and hence the changes made in between(in this case the instrumentation config) are lost once the server source is applied.
For the instrumentation config either we will need to create a new Map or need to fix this issue.
Probable fix can be, update only those parameters in config which are supposed to be updated by server source configuration.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## dev
 
+- **Feature: The agent configuration will now reflect whether method prepending or chaining was used per instrumentation
```
```suggestion
- **Feature: The agent configuration will now reflect whether method prepending or chaining was used per instrumentation**
```

fallwith:
> @fallwith There is a bug in the [replace_or_add_config](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/configuration/manager.rb#L78) method which updates the config when there is any new config is received from server source or any other source. [evaluate_and_apply_transformations](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/configuration/manager.rb#L142) method always updates from Default source and hence the changes made in between(in this case the instrumentation config) are lost once the server source is applied. For the instrumentation config either we will need to create a new Map or need to fix this issue. Probable fix can be, update only those parameters in config which are supposed to be updated by server source configuration.

Hi @prateek-ap. The dependency detection based parameters can't be configured server-side so we should be okay to simply preserve them after auto-detection. I have updated the cache resetting logic with [7735f92](https://github.com/newrelic/newrelic-ruby-agent/pull/1930/commits/7735f928121d32768ddf9ded4927ee4472aaa1b6).

prateek-ap:
Hi @fallwith The cache data is now preserving, Thanks. 
But, I have observation that few `instrumentation.<module>` parameter values are enabled, while the VALID_CONFIG_VALUES are only [:auto, :disabled, :prepend, :chain].
This makes discrepancies for few module instrumentation, example:
`:"instrumentation.mongo"=>"enabled"`
`:"instrumentation.excon"=>"enabled"`
How to resolve instrumentation mechanism used(alias or prepend) by Dependency detection for these cases?

fallwith:
> Hi @fallwith The cache data is now preserving, Thanks. But, I have observation that few `instrumentation.<module>` parameter values are enabled, while the VALID_CONFIG_VALUES are only [:auto, :disabled, :prepend, :chain]. This makes discrepancies for few module instrumentation, example: `:"instrumentation.mongo"=>"enabled"` `:"instrumentation.excon"=>"enabled"` How to resolve instrumentation mechanism used(alias or prepend) by Dependency detection for these cases?

@prateek-ap this is a great point! The Mongo and Excon instrumentation config parameters are definitely odd. As you point out, they are not set to `auto|disabled|prepend|chain` like the others, but `enabled|disabled` instead. And I think there might also be a bug preventing them from respecting `disabled`. I will use a separate PR to address that issue.

The good news is that you don't not have to worry about the `newrelic_rpm` Mongo and Excon instrumentations because they do not touch any module code. The Mongo instrumentation calls `Mongo::Monitoring::Global.subscribe` and the Excon instrumentation calls `<<` on `Excon.defaults[:middlewares]`. Because the `newrelic_rpm` agent is not performing any module prepending or chaining for the Mongo and Excon libraries, you won't have to worry about matching the same approach. You can either follow the same patterns used with `subscribe` and `<<` or perform module prepending/chaining if you prefer. And optionally you can skip the instrumentation entirely if you find the agent's config value is set to `'disabled'`.

fallwith:
See #1956 for addressing the Mongo/Excon config param issue, but that issue should not be a blocker here.

prateek-ap:
Hi @fallwith I understand that APM doesn't doing alias or prepend on mongo and excon library and that's why, by default have enabled, but In case of Security agent I will need a deterministic decision, otherwise, I will instrument using default approach (i.e. prepend) and that can conflict with other library instrumenting mongo or excon may be using chaining.

lib/new_relic/dependency_detection.rb
```diff
@@ -179,6 +179,15 @@ def fetch_config_value(key)
       return valid_value
     end
 
+    # update any :auto config value to be either :prepend or :chain after auto
+    # determination has selected one of those to use
+    def update_config_value(use_prepend)
+      if config_key && auto_configured?
+        NewRelic::Agent.config.instance_variable_get(:@cache)[config_key] = use_prepend ? :prepend : :chain
+      end
+      use_prepend
```
hannahramadan, line 188: Is this saying we are always returning `true` or `false`?

fallwith: Precisely. We hook into this existing method that returns a boolean, so here we make sure we still always return a boolean.

kaylareopelle (approved):
This is great as-is, take or leave the CHANGELOG rec.

lib/new_relic/dependency_detection.rb
hannahramadan: Got it I see what is happening now. Thank you :)

hannahramadan (dismissed):

fallwith:
> Hi @fallwith I understand that APM doesn't doing alias or prepend on mongo and excon library and that's why, by default have enabled, but In case of Security agent I will need a deterministic decision, otherwise, I will instrument using default approach (i.e. prepend) and that can conflict with other library instrumenting mongo or excon may be using chaining.

Hi @prateek-ap. With this PR/branch, for all config parameters that start with `instrumentation.`, you can rely on the following for their values:

- **prepend:** Agent used the prepend approach to provide instrumentation
- **chain:** Agent used the chain approach to provide instrumentation
- **disabled:** Agent did not provide instrumentation
- **enabled:** Agent provided instrumentation, but did not perform monkeypatching (no chain, no prepend)

So if you see `'enabled'`, feel free to use a default desired approach (prepend) without any worry of conflicting. For all `instrumentation.` parameters that have a value of `'enabled'`, the agent used neither prepend nor chain, so there won't be a conflict.

For Mongo and Excon in particular, you are welcome to follow the same pattern that we use for working with those libraries (for Mongo we subscribe, and for Excon we add to the middleware array) or if you prefer you can use prepend. Either way, you should not conflict with the agent for those two.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## dev
 
+- **Feature: The agent configuration will now reflect whether method prepending or chaining was used per instrumentation**
```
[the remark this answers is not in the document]
fallwith: Certainly! Addressed with [bf3b5ee](https://github.com/newrelic/newrelic-ruby-agent/pull/1930/commits/bf3b5ee288564b1c72f44cc139a5b742463a3a0a)

hannahramadan (approved):

fallwith:
@prateek-ap these changes are in the `dev` branch now.

PR #1932, opened 2023-04-10
Add CODEOWNERS file

### LINKED ISSUE #1906: CI: Check-in CODEOWNERS files
Each repo should have a GitHub formatted `CODEOWNERS` file in the root or in the `.github` subdirectory, despite any potential redundancies with the repo's access configuration in the web UI.

At a minimum, be sure to mark the `.github/workflows` directory as modifiable only by the appropriate set of teams and individuals. For example:

```text
github/**   @username @newrelic/teamname
```

hannahramadan:
Add CODEOWNERS file. The presence of this file means that any changes to the newrelic-ruby-agent repository need to be reviewed by at least one code owner. This feature automatically requests reviews from the code owners when a PR changes any owned file (in this case, all of them).

Closes #1906

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# Ruby agent code owners:
+*       @fallwith @hannahramadan @kaylareopelle @tannalynn


kaylareopelle:
Thanks, @hannahramadan! One question -- does the `*` syntax before the name means there's requirements on every file and not just the `.github` directory mentioned in the original ticket?

hannahramadan:
> #1906

@kaylareopelle yes! That `*` means that any file modified under our repo sends one of us a review request. The original ticket mentioned "At a minimum, make sure all .github/workflow files get a review", but seemed like it would be more secure to be a part of every change. We can definitely edit this file to only get notified about workflow changes if it becomes annoying or doesn't seem valuable.

PR #1939, opened 2023-04-12
Add permissions to GITHUB_TOKEN

### LINKED ISSUE #1904: CI: Set GHA permissions to read only
For all team GitHub repos, go to Settings > Actions > General and make sure that the Workflow Permissions section is configured to "Read" and not "Read and write".

Audit the impact of this change and create additional issues as necessary.

hannahramadan:
Actions that use `GITHUB_TOKEN` will be affected when GHA permissions are set to read-only (#1904) as a part of security updates. To allow these actions to continue working properly (write), we need to add `permissions` to the job. 

`GITHUB_TOKEN` exists in the following places: 

- `release.yml` - updated
- `ci.yml` - updated
- `assignproj.yml` - updated
- `ci_cron.yml` - NOT updated. The job that uses this is only reading the results of the last run.

Closes #1904

### DIFF
--- a/.github/workflows/assignproj.yml
+++ b/.github/workflows/assignproj.yml
@@ -4,8 +4,11 @@ on:
   issues:
     types: [opened]
 
+permissions:
+  issues: write
+
 env:
-  MY_GITHUB_TOKEN: ${{ secrets.RUBY_GITHUB_TOKEN }}
+  MY_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
   assign_one_project:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,6 +41,8 @@ jobs:
     - name: Create github release
       uses: softprops/action-gh-release@de2c0eb89ae2a093876385947365aca7b0e5f844 # tag v0.1.15
       if: $(git tag -l ${{ env.VERSION }}) == false
+      permissions:
+        contents: write
       with:
         tag_name: ${{ env.VERSION }}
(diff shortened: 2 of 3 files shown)


.github/workflows/assignproj.yml
```diff
@@ -4,8 +4,10 @@ on:
   issues:
     types: [opened]
 
+permissions: write-all
+
 env:
-  MY_GITHUB_TOKEN: ${{ secrets.RUBY_GITHUB_TOKEN }}
```
hannahramadan, line 8: `RUBY_GITHUB_TOKEN` is only used here. `GITHUB_TOKEN` takes care of this so we don't need both.

.github/workflows/release.yml
```diff
@@ -41,6 +41,8 @@ jobs:
     - name: Create github release
       uses: softprops/action-gh-release@de2c0eb89ae2a093876385947365aca7b0e5f844 # tag v0.1.15
       if: $(git tag -l ${{ env.VERSION }}) == false
+      permissions:
+        contents: write
```
hannahramadan, line 45: This actions requires `content: write` permissions
https://github.com/softprops/action-gh-release#permissions

.github/workflows/assignproj.yml
```diff
@@ -4,8 +4,11 @@ on:
   issues:
     types: [opened]
 
+permissions:
+  issues: write
```
hannahramadan, line 8: I believe we just need issue writing permissions for assigning projects

tannalynn (approved):

PR #1957, opened 2023-04-21
Mongo/Excon instrumentation: use 'enabled'

### LINKED ISSUE #1956: The Mongo and Excon instrumentation config params should be set to "enabled" in the config template
The `:'instrumentation.mongo'` and `:'instrumentation.excon'` configuration parameters have ~at least these 2 issues~ one issue:

 - The `newrelic.yml` ERB template in the root of the repo provides commented out uses of an "auto" value, while the documentation states that the both configuration parameters should be set to either "enabled" or "disabled".
     - Note that setting either parameter to "auto" will behave like "enabled" does, so the problem here is confined to potential user confusion from seeing "auto" in use while also seeing a comment that says only "enabled" and "disabled" can be used.
- ~Setting either configuration parameter to a value of "disabled" does not seem to work. Both parameters remain enabled and instrumentation will be performed if either library is present at agent startup time.~

Definition of done:

- The `newrelic.yml` ERB template will use "enabled" values for these two parameters instead of "auto". The parameters will remain commented out, with this being the only change.
- ~Setting `:'instrumentation.mongo'` to "disabled" will prevent the agent from creating a Mongo subscription when the `Mongo` library is present.~
- ~Setting `:'instrumentation.excon'` to "disabled" will prevent the agent from adding the agent's Excon middleware to the Excon middleware array when the `Excon` library is present.~

**UPDATE** the disabling behavior is working fine. Inadvertently I was pulling values from the inner `@defaults` instance var instead of the `@cache` instance var, so of course I was always seeing `enabled` because that's the default. Setting it `'disabled'` in a config file works. We should still update the `newrelic.yml` ERB template.

fallwith:
update the `newrelic.yml` ERB template in the root of the repo to align with the comments for `:'instrumentation.mongo'` and `:'instrumentation.excon'` that state that the only allowed values are `'enabled'` and `'disabled'` by using `'enabled'` instead of `'auto'` in the commented out config param lines.

resolves #1956

### DIFF
--- a/newrelic.yml
+++ b/newrelic.yml
@@ -352,7 +352,7 @@ common: &default_settings
 
   # Controls auto-instrumentation of Excon at start up.
   # May be one of [enabled|disabled].
-  # instrumentation.excon: auto
+  # instrumentation.excon: enabled
 
   # Controls auto-instrumentation of Grape at start up.
   # May be one of [auto|prepend|chain|disabled].
@@ -388,7 +388,7 @@ common: &default_settings
 
   # Controls auto-instrumentation of Mongo at start up.
   # May be one of [enabled|disabled].
-  # instrumentation.mongo: auto
+  # instrumentation.mongo: enabled
 
   # Controls auto-instrumentation of Net::HTTP at start up.
   # May be one of [auto|prepend|chain|disabled].


hannahramadan (approved):
Nice catch.

kaylareopelle (approved):

Issue #1936, opened 2023-04-11
Possible memory leak since 9.0


fcheung:
## Description
After updating from 8.16 to 9.0 we  noticed our ruby  processes had steady memory growth. This occurs only in one of our applications, despite all of them having similar environments in terms of ruby version, rails version etc.

This screenshot show ruby process usage overtime on one of our dev servers over a weekend (ie no traffic other than loadbalancer health checks, maybe actioncable pings if anyone had left a window open)

<img width="1804" alt="Screenshot 2023-04-11 at 13 06 57" src="https://user-images.githubusercontent.com/5927/231158404-07fdf63c-6e07-4e1e-a5bf-e54743d68aac.png">

after we rolled back to 8.16 this issue went away.

A thing that is different about this application that springs to mind is that is does use actioncable, ie. has extra threads running in the background and 9.0 defaulted thread instrumentation to on, so maybe that is relevant?

## Steps to Reproduce

working on it ...

## Your Environment
- alpine linux 3.16
- ruby 3.2.1
- rails 7.0.4.2
- agent 9.0.0
- server: puma 6.1.1

## Additional context

I did find one other person mentioning this: https://github.com/newrelic/newrelic-ruby-agent/issues/1842#issuecomment-1447191439

## For Maintainers Only or Hero Triaging this bug
*Suggested Priority (P1,P2,P3,P4,P5):*
*Suggested T-Shirt size (S, M, L, XL, Unknown):*

hannahramadan:
Hi @fcheung 👋 Thank you for sharing this issue! You mentioned 9.0.0 enabling thread instrumentation by default and we're wondering if this is related to what you're seeing. To help narrow it down, would you be willing to try disabling thread instrumentation while using Ruby agent 9.0.0? 

To do this, add the following to your `newrelic.yml`
```ruby
# newrelic.yml
instrumentation.thread.tracing: false
```
Alternatively, you can disable thread instrumentation with the following environment variable:

```ruby
NEW_RELIC_INSTRUMENTATION_THREAD_TRACING=false
```
We're also wondering if you have long-running/indefinite transactions? Thank you for mentioning that you're working on a reproduction. That should help us get closer to a resolution!

cirdes:
@fcheung I have experienced the same.

fcheung:
Hi,

[...]

PR #1969, opened 2023-04-25
Bugfix Thread instrumentation continuing to use finished transactions on other threads

tannalynn:
This fixes a bug where the agent would continue to record segments on transactions that were already finished in another thread. This would lead to increased memory usage over time because the agent would continue to copy over the finished transaction as the still current transaction in the thread instrumentation, causing the agent to think it was still supposed to continue recording data on the actually finished transaction. Now the agent will check the finished state of a transaction before using it when a new thread is spawned. 
Additionally, if a transaction reaches the segment limit but the transaction is actually already finished on some other thread, the state will now reset, preventing future segments from being created and a supportability metric will be recorded.

#1936

### DIFF
--- a/lib/new_relic/agent/transaction/tracing.rb
+++ b/lib/new_relic/agent/transaction/tracing.rb
@@ -29,6 +29,11 @@ module NewRelic
           else
             segment.record_on_finish = true
             ::NewRelic::Agent.logger.debug("Segment limit of #{segment_limit} reached, ceasing collection.")
+
+            if initial_segment&.finished?
+              ::NewRelic::Agent.logger.debug("Transaction #{best_name} has finished but segments still being created, resetting state.")
+              NewRelic::Agent::Tracer.state.reset
+            end
           end
           segment.transaction_assigned
         end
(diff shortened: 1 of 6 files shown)


lib/new_relic/agent/transaction/tracing.rb
```diff
@@ -29,6 +29,12 @@ def add_segment(segment, parent = nil)
           else
             segment.record_on_finish = true
             ::NewRelic::Agent.logger.debug("Segment limit of #{segment_limit} reached, ceasing collection.")
+
+            if finished?
+              ::NewRelic::Agent.logger.debug("Transaction #{best_name} has finished but segments still being created, resetting state.")
```
fallwith, line 34: This new module code make use of a method, `best_name`, that doesn't exist in the module, so the code can only be used by an instance of the `Transaction` class. Seeing as the `Transaction` class is the only consumer (includer) of this `Tracing` module, this is probably ok.

fallwith (approved):
Memory copying for new threads can be tricky. I'm very glad to see you were able to get this addressed in such a clear way that doesn't risk adverse side effects.

hannahramadan (approved):

PR #1975, opened 2023-04-27
Fix CI failures on 2.5.9 and rails 61

tannalynn:
I have no idea why simply adding a failure message to the assert makes it not fail anymore, but it does.
https://github.com/newrelic/newrelic-ruby-agent/actions/runs/4822722532
I couldn't get it to happen locally, seems to only happen on the CI.


kaylareopelle (approved):
🎩 💥 🐰

hannahramadan (approved):
So weird. But hey! Nice to have error messages 🤷🏽‍♀️

PR #1977, opened 2023-04-27
Remove Ruby < 2.4 constraints

kaylareopelle:
# Overview
Clean up a few areas with conditionals related to Ruby 2.4

### DIFF
--- a/lib/new_relic/control/instrumentation.rb
+++ b/lib/new_relic/control/instrumentation.rb
@@ -62,7 +62,6 @@ module NewRelic
           File.join(instrumentation_path, app.to_s, '*.rb')
         @instrumentation_files.each { |pattern| load_instrumentation_files(pattern) }
         DependencyDetection.detect!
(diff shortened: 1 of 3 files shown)


lib/new_relic/control/instrumentation.rb
```diff
@@ -62,7 +62,6 @@ def _install_instrumentation
           File.join(instrumentation_path, app.to_s, '*.rb')
         @instrumentation_files.each { |pattern| load_instrumentation_files(pattern) }
         DependencyDetection.detect!
-        ruby_deprecation
         rails_32_deprecation
```
kaylareopelle, line 65: We still have a deprecation message going out when the user is running Rails 3.2. I think this makes sense, as we haven't ripped out the code to support Rails 3.2 yet.

fallwith (approved):

hannahramadan (approved):
Nice thanks for finding these ⚡️

PR #1987, opened 2023-05-01
Skip intermittently failing tests

kaylareopelle:
The NewRelic::Agent::PipeChannelManagerTest has two cases failing on Ruby 2.5.9 and Rails 6.0/6.1:
#test_listener_merges_error_events, #test_listener_merges_error_traces

These tests started failing after PR#1969 was merged: https://github.com/newrelic/newrelic-ruby-agent/pull/1969

The tests are now skipped in the specific conditions where they intermittently fail

This was tested locally using Ruby 2.5.9 and rails52 (where they should not be skipped), rails60, and rails61.

### DIFF
--- a/test/new_relic/agent/pipe_channel_manager_test.rb
+++ b/test/new_relic/agent/pipe_channel_manager_test.rb
@@ -79,6 +79,8 @@ class NewRelic::Agent::PipeChannelManagerTest < Minitest::Test
     end
 
     def test_listener_merges_error_traces
+      skip_for_ruby_2_5_9_and_rails_6_x
+
       sampler = NewRelic::Agent.agent.error_collector
       sampler.notice_error(Exception.new('message'), :uri => '/myurl/',
         :metric => 'path', :referer => 'test_referer',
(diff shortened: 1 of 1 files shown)


test/new_relic/agent/pipe_channel_manager_test.rb
```diff
@@ -360,4 +364,12 @@ def reset_lifetime_counts!(container)
     buffer.instance_variable_set(:@captured_lifetime, 0)
     buffer.instance_variable_set(:@seen_lifetime, 0)
   end
+
+  def skip_for_ruby_2_5_9_and_rails_6_x
+    # TODO: MAJOR VERSION - remove this method and its calls when Ruby 2.5 is no longer supported
+    skip if RUBY_VERSION == '2.5.9' &&
+      defined?(Rails::VERSION::STRING) &&
+      # version string starts with 6.0 or 6.1
+      Rails::VERSION::STRING =~ /\A6\.(0|1)/
```
fallwith, line 373: Given that there won't be a Rails 6.2, I think checking for `Rails::VERSION::MAJOR` and then doing `Rails::VERSION::MAJOR == 6` would suffice. Given the context (test code only) there's no need to change it though.

fallwith (approved):

test/new_relic/agent/pipe_channel_manager_test.rb
kaylareopelle: Thanks for this suggestion. I thought about that too, even though it seems unlikely/unprecedented 6.2 would ever come out, I figured it might be better to specify.

PR #1990, opened 2023-05-04
orphaned config test: ignore security configs

fallwith:
update the orphaned config regex with a negative lookbehind

### DIFF
--- a/test/new_relic/agent/configuration/orphan_configuration_test.rb
+++ b/test/new_relic/agent/configuration/orphan_configuration_test.rb
@@ -15,7 +15,7 @@ class OrphanedConfigTest < Minitest::Test
   def test_all_agent_config_keys_are_declared_in_default_source
     non_test_files.each do |file|
       lines_in(file).each_with_index do |line, index|
-        config_match = line.match(/Agent\.config\[:['"]?([a-z\._]+)['"]?\]/)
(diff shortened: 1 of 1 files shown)


test/new_relic/agent/configuration/orphan_configuration_test.rb
```diff
@@ -15,7 +15,7 @@ def setup
   def test_all_agent_config_keys_are_declared_in_default_source
     non_test_files.each do |file|
       lines_in(file).each_with_index do |line, index|
-        config_match = line.match(/Agent\.config\[:['"]?([a-z\._]+)['"]?\]/)
+        config_match = line.match(/(?<!Security::)Agent\.config\[:['"]?([a-z\._]+)['"]?\]/)
```
kaylareopelle, line 18: What does the < sign do in the RegEx? I don't think I've come across it before.

fallwith: It's supposed to give the human reviewer a hint that it's going the left / backwards direction. Here a "negative lookbehind" / "negative readbehind" is used to only match 'Agent' if it is not immediately preceded by 'Security::'.

kaylareopelle (approved):
Thanks, James!

PR #1996, opened 2023-05-09
Assign permission at job-level

hannahramadan:
Move permissions to job level. We should be able to set this at the workflow level (as previously), but that doesn't seem to work here. Seeing if job-level will fix this.

### DIFF
--- a/.github/workflows/assignproj.yml
+++ b/.github/workflows/assignproj.yml
@@ -4,10 +4,6 @@ on:
   issues:
     types: [opened]
 
-permissions:
-  issues: write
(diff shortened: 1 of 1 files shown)


.github/workflows/assignproj.yml
```diff
@@ -4,10 +4,6 @@ on:
   issues:
     types: [opened]
 
-permissions:
-  issues: write
```
hannahramadan, line 8: I don't think we actually need this permission set. Was leftover from a previous attempt.

fallwith (approved):

PR #2023, opened 2023-05-19
Active Record notifications: permit disabling

### LINKED ISSUE #2019: Active Record Notifications: config option to disable notifications does not work
To prevent the agent from attempting to subscribe to Active Support notifications for the Active Record library's activity, the `newrelic.yml` config file should be updated to contain `disable_active_record_notifications: true`. Our `newrelic.yml` template in the root of our repo and our documentation both correctly explain this.

Unfortunately, there is a typo in the agent code that prevents the configuration option from working, because the code is looking for "activerecord" instead of "active_record" for the configuration option. This is likely due to the "activerecord" name of the Active Record gem itself causing confusion.

We should fix the typo so that the `disable_active_record_notifications: true` configuration setting works as desired to disable the notifications.

fallwith:
To prevent the agent from subscribing to Active Record notifications and otherwise instrumenting Active Record, the configuration parameter `:disable_active_record_instrumentation` should be set to a value of `true` in the `newrelic.yml` config file.

This change permits that parameter to function as expected by fixing a typo in which the code expected `:disable_activerecord_instrumentation` without an underscore. Note that prior to this change, putting `:disable_activerecord_instrumentation` in the config file would also not work, because the config parameter is spelled correctly in `default_source.rb`. This change is needed for the parameter to function at all.

resolves #2019

### DIFF
--- a/lib/new_relic/agent/instrumentation/active_record_notifications.rb
+++ b/lib/new_relic/agent/instrumentation/active_record_notifications.rb
@@ -90,7 +90,7 @@ DependencyDetection.defer do
   end
 
   depends_on do
-    !NewRelic::Agent.config[:disable_activerecord_instrumentation] &&
+    !NewRelic::Agent.config[:disable_active_record_instrumentation] &&
       !NewRelic::Agent::Instrumentation::ActiveRecordSubscriber.subscribed?
   end
 
--- a/test/new_relic/agent/instrumentation/rails/active_record_subscriber.rb
+++ b/test/new_relic/agent/instrumentation/rails/active_record_subscriber.rb
@@ -238,6 +238,21 @@ class NewRelic::Agent::Instrumentation::ActiveRecordSubscriberTest < Minitest::T
     end
   end
 
+  def test_instrumentation_can_be_disabled
+    parameter = :disable_active_record_instrumentation
+
+    item = DependencyDetection.instance_variable_get(:@items).detect { |i| i.name == :active_record_notifications }
+
+    assert item, 'Could not locate the dependency detection item for Active Record notifications'
+    dependency_check = item.dependencies.detect { |d| d.source.match?(/#{parameter}/) }
+
+    assert dependency_check, "Could not locate the dependency check related to the #{parameter} configuration parameter"
+
+    with_config(parameter => false) do
+      refute dependency_check.call
+    end
+  end
+
   private
 
   def simulate_query(duration = nil)


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1104,11 +1104,12 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'If `true`, disables Active Support instrumentation.'
         },
-        :disable_activerecord_instrumentation => {
+        :disable_active_record_instrumentation => {
           :default => value_of(:skip_ar_instrumentation),
           :documentation_default => false,
           :public => true,
           :type => Boolean,
+          :aliases => %i[disable_activerecord_instrumentation],
```
hannahramadan, line 1112: nice i didn't know we had this `aliases` key

lib/new_relic/agent/instrumentation/active_record_notifications.rb
```diff
@@ -90,7 +90,8 @@ def log(sql, name = 'SQL', binds = [], type_casted_binds = [], statement_name =
   end
 
   depends_on do
-    !NewRelic::Agent.config[:disable_activerecord_instrumentation] &&
+    !NewRelic::Agent.config[:disable_active_record_instrumentation] &&
+      !NewRelic::Agent.config[:disable_active_record_notifications] &&
```
hannahramadan, line 94: If there a world where someone could have instrumentation without notifications or vice versa?

hannahramadan (approved):

lib/new_relic/agent/instrumentation/active_record_notifications.rb
fallwith: The answer for that world being possible is "definitely". The answer for that possibility potentially causing problems for this change, the answer is "maybe".

Because the agent predates Rails notifications, for some Rails functionality it instruments core Rails libraries and uses the "instrumentation" parameters to toggle that behavior. But for the notification based approaches that we take, we're supposed to use a "notifications" based parameter.

For this particular class it was previously toggled to off only if you used an "instrumentation" parameter, despite it taking a "notifications" approach. I don't want to break that existing functionality, so I left it in place and will now ALSO disable the notifications if the "notifications" parameter is present (because if you want to disable notifications, my thinking is that we should in fact disable notifications).

So there's a chance a situation exists that this change causes a problem for. By retaining backwards compatibility with existing checks and permutations of the related config parameters, I'm hoping I've protected us against potential problems.

PR #2035, opened 2023-05-25
Include the current transaction id with errors

fallwith:
If a transaction can be found in the current context while noticing an error, include that transaction's guid in the array that is JSON-ified for reporting later.


hannahramadan (approved):

fallwith:
resolves #2036

tannalynn (approved):
🎉

Issue #2038, opened 2023-05-26
CI: Address compatibility with Active Record 7.0.5


fallwith:
The recently released Active Record v7.0.5 works great from an agent instrumentation perspective and even passes all CI checks locally. But for some reason with GHA, we encounter the following new error:

```shell
/opt/hostedtoolcache/Ruby/3.3.0-preview1/x64/lib/ruby/gems/3.3.0+0/gems/activerecord-7.0.5/lib/active_record/tasks/database_tasks.rb:73:in `check_protected_environments!': You are attempting to modify a database that was last run in `test` environment. (ActiveRecord::EnvironmentMismatchError)
You are running in `default_env` environment. If you are sure you want to continue, first set the environment using:

        bin/rails db:environment:set


	from /opt/hostedtoolcache/Ruby/3.3.0-preview1/x64/lib/ruby/gems/3.3.0+0/gems/activerecord-7.0.5/lib/active_record/railties/databases.rake:18:in `block (2 levels) in <top (required)>'
	from /opt/hostedtoolcache/Ruby/3.3.0-preview1/x64/lib/ruby/gems/3.3.0+0/gems/rake-13.0.6/lib/rake/task.rb:281:in `block in execute'
```

fallwith:
resolved with #2039

PR #2039, opened 2023-05-27
CI: ignore Rails environment switches for db tasks

fallwith:
When switching between the default_env and test Rails environments, the Rails database tasks will raise
`ActiveRecord::EnvironmentMismatchError`.

Rails can either be explicitly told that the environment being switched to is expected via a task that sets the target environment, or the check itself can be disabled via an environment variable.

For now we are opting for the latter with
`ENV['DISABLE_DATABASE_ENVIRONMENT_CHECK']`.


fallwith:
resolves #2038

PR #2044, opened 2023-06-01
Configuration options and interface singleton for the security agent

fallwith:
- Define new `security.agent.*` and `security.*` configuration parameters
- Introduce a SecurityInterface singleton class to guarantee only one instance of the security agent is ever loaded
- Perform a `require` on the security agent if configured to do so
- For existing singleton classes, have them `require 'singleton'` to make it more obvious that they are singleton classes instead of relying on the Control class being loaded and delivering the `require`.


.gitignore
```diff
@@ -41,3 +41,4 @@ Brewfile.lock.json
 .github/actions/simplecov-report/lib/
 test/minitest/minitest_time_report
 gem_manifest_*.json
+nr-security-home
```
kaylareopelle, line 44: Do we still need this? I thought [NEWRELIC-8432](https://issues.newrelic.com/browse/NEWRELIC-8432) moved the directory into "log"?

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2212,6 +2212,73 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :allowed_from_server => false,
+          :description => 'If `true`, the security agent is loaded'
+        },
+        :'security.enabled' => {
+          :default => false,
+          :external => true,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => 'If `true`, the security agent is activated'
```
kaylareopelle, line 2231:
```suggestion
          :description => 'If `true`, the security agent is started'
```

We have a concept of the language agent "starting" rather than "activating". I think we should use a familiar term here.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2212,6 +2212,73 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :description => 'Defines the mode for the security agent to operate in. Currently only `IAST` is supported',
+          :dynamic_name => true
+        },
+        :'security.validator_service_url' => {
+          :default => 'wss://csec.nr-data.net',
+          :external => true,
+          :public => true,
+          :type => String,
+          :allowed_from_server => true,
+          :description => 'Defines the end point URL for posting security related data',
```
kaylareopelle, line 2248:
```suggestion
          :description => 'Defines the endpoint URL for posting security-related data',
```

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2212,6 +2212,73 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :public => true,
+          :type => String,
+          :allowed_from_server => true,
+          :description => 'Defines the mode for the security agent to operate in. Currently only `IAST` is supported',
+          :dynamic_name => true
+        },
+        :'security.validator_service_url' => {
+          :default => 'wss://csec.nr-data.net',
+          :external => true,
+          :public => true,
```
kaylareopelle, line 2245: Does this need to be a public setting? What's the use case for the value needing to be changed by a customer?

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2212,6 +2212,73 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :description => 'Defines the end point URL for posting security related data',
+          :dynamic_name => true
+        },
+        :'security.detection.rci.enabled' => {
+          :default => true,
+          :external => true,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => 'If `true`, enables rci detection'
```
kaylareopelle, line 2257: A quick Google for "rci detection" / "rci detection security" / "rci detection app security" doesn't give me much to understand what this means. Could we include a definition? I'd also like to capitalize "RCI".

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2212,6 +2212,73 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+        },
+        :'security.detection.rxss.enabled' => {
+          :default => true,
+          :external => true,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => 'If `true`, enables rxss detection'
```
kaylareopelle, line 2265:
```suggestion
          :description => 'If `true`, enables RXSS detection'
```

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2212,6 +2212,73 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+        },
+        :'security.detection.deserialization.enabled' => {
+          :default => true,
+          :external => true,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => 'If `true`, enables deserialization detection'
+        },
+        :'security.applicationinfo.port' => {
```
kaylareopelle, line 2275:
```suggestion
        :'security.application_info.port' => {
```

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2212,6 +2212,73 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :default => -1,
```
kaylareopelle, line 2276: Could we replace this value with the default port for Passenger? I believe it's [3000](https://www.phusionpassenger.com/library/config/standalone/intro.html#configuring-through-command-line-options-environment-variables-and-passengerfile-json)

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2212,6 +2212,73 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :public => true,
+          :type => Integer,
+          :external => true,
+          :allowed_from_server => false,
+          :description => 'The port the application listens on (Mandatory for Passenger, for other servers detected by default)'
```
kaylareopelle, line 2281:
```suggestion
          :description => 'The port the application is listening on. This setting is mandatory for Passenger users. Other servers should be detected by default.'
```

lib/new_relic/control/security_interface.rb
```diff
@@ -0,0 +1,40 @@
...
+      end
+
+      def waiting?
+        (@wait ||= false) == true
+      end
+
+      def init_agent
+        return if agent_started? || waiting?
+
+        if Agent.config[:'security.agent.enabled']
```
kaylareopelle, line 25: Don't we need both `security.agent.enabled` and `security.enabled` to be true to init the security agent?

newrelic.yml
```diff
@@ -692,6 +692,34 @@ common: &default_settings
   # If true, the agent automatically detects that it is running in a Pivotal Cloud Foundry environment.
   # utilization.detect_pcf: true
 
+  # BEGIN security agent
```
kaylareopelle, line 695: This might be a good place to put instructions about how to use the security agent (ex. only in a dedicated test environment, it will delete your data, etc.)

lib/new_relic/control.rb
```diff
@@ -8,7 +8,6 @@
 require 'new_relic/language_support'
 require 'new_relic/helper'
 
-require 'singleton'
```
kaylareopelle, line 11: Why did you decide to remove 'singleton' from this class? Is it because it was required but not included?

lib/new_relic/control/security_interface.rb
```diff
@@ -0,0 +1,40 @@
...
+          Agent.logger.info('Invoking New Relic security module')
```
kaylareopelle, line 26: Are the security module and the security agent different things? We use both terms in the logger messages in this file.

lib/new_relic/control.rb
kaylareopelle: Never mind -- Your great commit message clarified this for me: [NewRelic::Control::SecurityInterface](https://github.com/newrelic/newrelic-ruby-agent/pull/2044/commits/2e8db5d6e66584530e461d6f79a3ba032aa39d03)

lib/new_relic/agent/instrumentation/rack/instrumentation.rb
```diff
@@ -11,6 +11,7 @@ class << builder_class
             attr_accessor :_nr_deferred_detection_ran
           end
           builder_class._nr_deferred_detection_ran = false
+          NewRelic::Control::SecurityInterface.instance.wait = true
```
kaylareopelle, line 14: Are these changes to make sure the Rack instrumentation continues to load in a deferred way, even when the security agent is present?

lib/new_relic/agent/configuration/default_source.rb
fallwith: Committed, but these params will all change once the agent spec is finalized.

lib/new_relic/agent/configuration/default_source.rb
fallwith: Relayed the question to the agent spec PR.

lib/new_relic/agent/configuration/default_source.rb
fallwith: I'd like to hash these items out on the cross agent spec and then circle back here.

lib/new_relic/agent/configuration/default_source.rb
fallwith: I can't change this one without a matching merged PR on the security agent itself.

lib/new_relic/agent/configuration/default_source.rb
fallwith: Even though it's needed for Passenger, it's also used if populated by the other web servers. The `-1` seems to be a trick that allows the use of a numeric field to convey "don't use the actual number I'm giving you, use the default". I wonder if `nil` would work in the same way. No matter what, we'd need a change made on the security agent side before we can change this value here.

lib/new_relic/agent/configuration/default_source.rb
fallwith: I like where you're heading with this one. Would you please make one additional change? Currently "mandatory for ... users" and "Other servers" makes me mentally think we're conflating users and servers.

lib/new_relic/control/security_interface.rb
fallwith: Yes. The APM agent checks for one to guard the `require` and the security agent checks for the other one to guard the actually starting of the security agent and running its event loop.

newrelic.yml
fallwith: Good call. I think I can bring this to the cross agent spec discussion as well.

lib/new_relic/control/security_interface.rb
fallwith: Kinda. I think the intention by the original author we copied and pasted from was to convey a separation between the gem / library / module which is just a chunk of Ruby code at that point, and the security agent which is the label for the running, looping application.

lib/new_relic/agent/instrumentation/rack/instrumentation.rb
fallwith: Yes. I'd love to know more about how well our deferred behavior works with modern Ruby apps to know if it itself has any shortcomings or if there's anything we're doing that we don't need to do anymore in 2023.

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: Done!

lib/new_relic/agent/instrumentation/rack/instrumentation.rb
kaylareopelle: Great idea. Captured in https://github.com/newrelic/newrelic-ruby-agent/issues/2046

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: Would it be appropriate for me to create a PR for the security agent requesting these changes be made? Or should I start with a Jira ticket?

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: We have precedence in our code to have `nil` defaults for `Integer` configurations with `utilization.logical_processors` and `utilization.total_ram_mib`. I think I would prefer that approach over `-1`.

lib/new_relic/agent/configuration/default_source.rb
fallwith: I think Jira would be better as this option is universal and not Ruby specific, so it'll likely need to get circulated amongst those who look at Jira. I'm curious as to how frozen the values are at this point, but I support the change.

lib/new_relic/agent/configuration/default_source.rb
fallwith: This option is Ruby specific, so we have some leeway. We can submit a PR to replace -1 in both code bases. I can take that or you can.

lib/new_relic/agent/configuration/default_source.rb
fallwith: Whoops - I apologize - the port param IS Ruby specific as it was only needed to be user facing for Ruby. I'll leave it up to you on which approach to take to get it changed, but we can definitely do so without anything impacting the other language agents.

lib/new_relic/agent/configuration/default_source.rb
fallwith: I like how down here I realized correctly that it's Ruby specific.

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: Internal Jira ticket requesting changes: https://issues.newrelic.com/browse/NEWRELIC-8926

lib/new_relic/agent/configuration/default_source.rb
fallwith: The spec reviewers have decided all existing options will remain and be kept public. This particular option leaves the door open for an on-prem endpoint.

lib/new_relic/agent/configuration/default_source.rb
fallwith: Capitalizations have been made, and the request for additional info has been raised to the agent spec PR.

.gitignore
fallwith: The cross-agent issue has been raised. As of this time, we're in compliance with how things work for everyone, so no changes to be made just now. We'll handle Ruby specific changes should the spec discussion lead to a change.

PR #2051, opened 2023-06-06
Introduce log-level filtering

### LINKED ISSUE #1831: Implement Log filtering by log level
Add a new configuration `application_logging.forwarding.log_level` that will set the minimum level the ruby agent will record log events for. We will want to check the log level against this config any time we want to [record a log event](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/logger/instrumentation.rb#L54).
We will be able to compare the configuration to the [logging severity](https://ruby-doc.org/stdlib-2.7.0/libdoc/logger/rdoc/Logger/Severity.html) defined by the Logger class. 
We will also want to record a supportability metric one time, `Supportability/Logging/Forwarding/LogLevel`, any time this configuration option is being used.

kaylareopelle:
This creates a new configuration, `application_logging.forwarding.log_level`. This configuration defines the
minimum log level a log event should have to forward to New Relic.

Filtering is done in the `LogEventAggregator#record` method.

The default is "debug" which will forward all logs.

Setting the configuration to another built-in Ruby log level, such as "error", will forward only logs with the severity of error or above. This would mean logs with "error", "fatal" or "unknown" would be forwarded to New Relic.

This setting uses [Ruby's Logger::Severity constants integer values](https://github.com/ruby/ruby/blob/master/lib/logger/severity.rb#L6-L17) to determine precedence.

If an invalid log level is given to the configuration, the agent will default to forwarding all logs/having a "debug" level.

Log levels not defined as constants within Ruby's `Logger::Severity` class will always have their events forwarded.

This PR also: 
* replaces string references with constants in a few areas on the LogEventAggregator
* adds a pattern to ConfigScanning to permit config assignment to constants as a valid use case

Resolves #1831

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -749,6 +749,29 @@ module NewRelic
           :allowed_from_server => false,
           :description => 'If `true`, the agent captures log records emitted by your application.'
         },
+        :'application_logging.forwarding.log_level' => {
+          :default => 'debug',
+          :public => true,
+          :type => String,
+          :allowed_from_server => false,
+          :description => <<~DESCRIPTION
+            Sets the minimum level a log event must have to be forwarded to New Relic.
+
+            This is based on the integer values of Ruby's `Logger::Severity` constants: https://github.com/ruby/ruby/blob/master/lib/logger/severity.rb
+
+            The intention is to forward logs with the level given to the configuration, as well as any logs with a higher level of severity.
+
+            For example, setting this value to "debug" will forward all log events to New Relic. Setting this value to "error" will only forward log events with the levels "error", "fatal", and "unknown".
+
+            Valid values (ordered lowest to highest):
+              * "debug"
+              * "info"
+              * "warn"
+              * "error"
+              * "fatal"
+              * "unknown"
+          DESCRIPTION
+        },
         :'application_logging.forwarding.max_samples_stored' => {
           :default => 10000,
           :public => true,
--- a/lib/new_relic/agent/log_event_aggregator.rb
+++ b/lib/new_relic/agent/log_event_aggregator.rb
@@ -36,6 +36,7 @@ module NewRelic
       METRICS_ENABLED_KEY = :'application_logging.metrics.enabled'
       FORWARDING_ENABLED_KEY = :'application_logging.forwarding.enabled'
       DECORATING_ENABLED_KEY = :'application_logging.local_decorating.enabled'
+      LOG_LEVEL_KEY = :'application_logging.forwarding.log_level'
 
       def initialize(events)
         super(events)
@@ -63,8 +64,9 @@ module NewRelic
           end
         end
 
+        return if severity_too_low?(severity)
         return if formatted_message.nil? || formatted_message.empty?
-        return unless NewRelic::Agent.config[:'application_logging.forwarding.enabled']
+        return unless NewRelic::Agent.config[FORWARDING_ENABLED_KEY]
         return if @high_security
 
         txn = NewRelic::Agent::Transaction.tl_current
@@ -191,7 +193,7 @@ module NewRelic
       # these until harvest before recording them
       def record_customer_metrics
         return unless enabled?
-        return unless NewRelic::Agent.config[:'application_logging.metrics.enabled']
+        return unless NewRelic::Agent.config[METRICS_ENABLED_KEY]
 
         @counter_lock.synchronize do
           return unless @seen > 0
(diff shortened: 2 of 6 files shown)


lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -230,6 +232,32 @@ def truncate_message(message)
 
         message.byteslice(0...MAX_BYTES)
       end
+
```
kaylareopelle, line 235: Rather than continuing to increase the size of this file, I considered pushing the new methods into a `LogLevelFiltering` module and including that in the `LogEventAggregator`. We don't use that pattern a lot in the agent, so I wanted to hold back on that until we could discuss it as a team.

fallwith: I think holding off is a good call. The project wasn't built around a specific pattern that would help guide additional content and it predates RuboCop for that matter. Given the fact that the separation would not promote any reusability throughout the project or other gems and we'd be doing only to keep class sizes down, I think the benefit of compartmentalizing things all in one place for the human reviewer provides some good benefits.

Two things that I like to do to improve things with a large class:

- Make sure there's good `public` and `private` separation
- Have the methods appear in alphabetical order

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -230,6 +232,32 @@ def truncate_message(message)
...
+        log_level.upcase.to_sym
+      end
+
+      def severity_too_low?(severity)
+        severity_constant = format_log_level_constant(severity)
+        # always record custom log levels
+        return false unless Logger::Severity.constants.include?(severity_constant)
+
+        Logger::Severity.const_get(severity_constant) < Logger::Severity.const_get(minimum_log_level)
+      rescue NameError => e
```
fallwith, line 258: Can we remove this `rescue` now that both methods that perform a constant lookup use a `constants.include?` guard? If not, we should have a test that explains a pathway to reaching the `rescue`.

test/helpers/config_scanning.rb
```diff
@@ -13,6 +13,7 @@ module ConfigScanning
       REGISTER_CALLBACK_PATTERN = /register_callback\(:['"]?([a-z\._]+)['"]?\)/
       NAMED_DEPENDENCY_PATTERN = /^\s*named[ (]+\:?([a-z0-9\._]+).*$/
       EVENT_BUFFER_MACRO_PATTERN = /(capacity_key|enabled_key)\s+:['"]?([a-z\._]+)['"]?/
+      ASSIGNED_CONSTANT_PATTERN = /[A-Z]+\s*=\s*:['"]?([a-z\._]+)['"]?\s*/
```
fallwith, line 16: A minor issue (predating this PR) for these patterns is that we're using `['"]` twice instead of a backreference for the first match. This technically permits `key = "value'` with unmatched quotes. I think with it only being used in the tests, we're ok for now.

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -63,8 +64,9 @@ def record(formatted_message, severity)
           end
         end
 
+        return if severity_too_low?(severity)
```
fallwith, line 67: My mind really wanted the inverse with something like `return unless severity_level_permitted?` but after reading through the whole class it caught on.

lib/new_relic/agent/log_event_aggregator.rb
kaylareopelle: Great point! Fixed in 93eae88624a860ac88fe6420dff8e4aa6c5a1f1d

test/helpers/config_scanning.rb
kaylareopelle: Thanks for pointing this out. I updated the patterns to use backreferences in 1aa3e5709649f1998a9696f306bdcf47f16f9708

tannalynn (approved):
🎉

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -749,6 +749,29 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'If `true`, the agent captures log records emitted by your application.'
         },
+        :'application_logging.forwarding.log_level' => {
```
hannahramadan, line 752: Do we want to add this to newrelic.yml?

hannahramadan (approved):

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -749,6 +749,29 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :description => <<~DESCRIPTION
+            Sets the minimum level a log event must have to be forwarded to New Relic.
+
+            This is based on the integer values of Ruby's `Logger::Severity` constants: https://github.com/ruby/ruby/blob/master/lib/logger/severity.rb
+
+            The intention is to forward logs with the level given to the configuration, as well as any logs with a higher level of severity.
+
+            For example, setting this value to "debug" will forward all log events to New Relic. Setting this value to "error" will only forward log events with the levels "error", "fatal", and "unknown".
+
+            Valid values (ordered lowest to highest):
```
fallwith, line 766: Do we want to enforce this adherence to this allowlist? One possible use case: I typed in "warm" by mistake. Should I be told about the problem?

fallwith (dismissed):

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: There's some enforcement in the [LogEventAggregator#minimum_log_level method](https://github.com/newrelic/newrelic-ruby-agent/pull/2051/files#diff-680b1e6a8547301e66b77f383e273ef141166e5986c9a1180625081dc8e271bcR236), where values outside the valid range will be overridden with `:DEBUG`. It's a little flexible for caps/lowercase values since the configs are coerced to upcased symbols before evaluation. 

I like the idea of being told about the problem and added a log_once message to the method: 3fc3c2a1d7b8230e3b9b1b75448d4967b296a936 

If you think there are additional steps for enforcement that would benefit customers, please let me know.

fallwith (approved):

tannalynn (approved):
👍

Issue #2048, opened 2023-06-05
False warning: Unable to determine source code info for <Controller>


jcrisp:
## Description
In newrelic_agent.log, I am receiving megabytes of false warnings:
`WARN : Unable to determine source code info for 'XYZController', method 'index' - NameError: undefined method 'index' for class '#<Class:XYZController>'`
(XYZ is standing in for the real name of the controller)

The controller doesn't have a method for 'index' but it does have an 'index' view and it is in routes.rb. No method is required on the Controller in this case - the view is rendered automatically by Rails without a method on the Controller.

This pull request highlights the code that is causing the warning:
https://github.com/newrelic/newrelic-ruby-agent/pull/1593

I am using newrelic_rpm (9.2.0).

## Expected Behavior
No logging of this false warning.
Or, a way to turn this warning off in the yaml config file without turning off all warnings.

## Steps to Reproduce
Set up a new rails app with a single controller and view. Add the view to routes.rb so it can be accessed (or make it the website root), but do not add a method for the view to the controller.

## For Maintainers Only or Hero Triaging this bug
*Suggested Priority (P1,P2,P3,P4,P5):*
*Suggested T-Shirt size (S, M, L, XL, Unknown):*

hannahramadan:
Hi @jcrisp - thank you for submitting this issue. This shouldn't be happening and we are currently working on a fix. In the meantime, you can disable these specific warnings by turning off Code Level Metrics. In your `newrelic.yml`, add the following line:

`code_level_metrics.enabled: false`

jcrisp:
Perfect, thank you. Disabling the code_level_metrics has stopped the incorrect log warnings. I'm not using the CodeStream extension - does disabling this affect anything else?

hannahramadan:
@jcrisp Yes! CodeStream is currently the only consumer of Code Level Metrics, so disabling shouldn't affect anything else.

hannahramadan:
Hi @jcrisp. Thanks again for reporting this issue! A fix will be available in our next release [PR#2061](https://github.com/newrelic/newrelic-ruby-agent/pull/2061)

jcrisp:
Thank you!! I appreciate your help and fixing it so quickly :-)
James

tannalynn:
9.3.0, which includes the fix for this, has been released. You should be able to update to this version now and have this be resolved. Please let us know if you run into any other issues in the future!

jcrisp:
Thank you!!
James

PR #2061, opened 2023-06-08
Controller methods not in controller class don't warn

hannahramadan:
Rails controllers can be a special case because by default, controllers in Rails automatically render views with names that correspond to valid routes. This means that a controller method may not have a corresponding method in the controller class.

This PR adds a check to see if we have a valid method that isn't attached to a controller, and if so, can we figure out what controller it belongs to. 

The check makes an assumption of a simple rails app with all controllers living inside app/controllers, so if a customer's app breaks this pattern, the check fails, we don't report on method metrics, and we also don't get false warnings. If we can determine the controller, the method information is inserted at the top of the file.

Community issue #2048

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ Version <dev> of the agent adds the ability to filter logs by level and expands 
     * `transmit_subscription_confirmation.action_cable`
     * `transmit_subscription_rejection.action_cable`
 
+- **Bugfix: Report Code Level Metrics for Rails controller methods**
+
+  Controllers in Rails automatically render views with names that correspond to valid routes. This means that a controller method may not have a corresponding method in the controller class. Code Level Metrics now report on these methods and don't log false warnings. Thanks to [@jcrisp](https://github.com/jcrisp) for reporting this issue. [PR#2061](https://github.com/newrelic/newrelic-ruby-agent/pull/2061)
+
 ## v9.2.2
 
   Version 9.2.2 of the agent fixes a bug with the `Transaction#finished?` method.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## dev
 
-Version <dev> of the agent adds the ability to filter logs by level and expands instrumentation for Action Cable.
+Version <dev> of the agent adds the ability to filter logs by level, expands instrumentation for Action Cable, and provides a bugfix for Code-Level Metrics.
 
 - **Feature: Filter forwarded logs based on level**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,9 +20,9 @@ Version <dev> of the agent adds the ability to filter logs by level, expands ins
     * `transmit_subscription_confirmation.action_cable`
     * `transmit_subscription_rejection.action_cable`
 
-- **Bugfix: Report Code Level Metrics for Rails controller methods**
+- **Bugfix: Report Code-Level Metrics for Rails controller methods**
 
-  Controllers in Rails automatically render views with names that correspond to valid routes. This means that a controller method may not have a corresponding method in the controller class. Code Level Metrics now report on these methods and don't log false warnings. Thanks to [@jcrisp](https://github.com/jcrisp) for reporting this issue. [PR#2061](https://github.com/newrelic/newrelic-ruby-agent/pull/2061)
+  Controllers in Rails automatically render views with names that correspond to valid routes. This means that a controller method may not have a corresponding method in the controller class. Code-Level Metrics now report on these methods and don't log false warnings. Thanks to [@jcrisp](https://github.com/jcrisp) for reporting this issue. [PR#2061](https://github.com/newrelic/newrelic-ruby-agent/pull/2061)
 
 ## v9.2.2
 
--- a/lib/new_relic/agent/method_tracer_helpers.rb
+++ b/lib/new_relic/agent/method_tracer_helpers.rb
@@ -46,13 +46,15 @@ module NewRelic
         cache_key = "#{object.object_id}#{method_name}".freeze
         return @code_information[cache_key] if @code_information.key?(cache_key)
 
-        namespace, location, is_class_method = namespace_and_location(object, method_name.to_sym)
+        info = namespace_and_location(object, method_name.to_sym)
+        return ::NewRelic::EMPTY_HASH if info.empty?
 
+        namespace, location, is_class_method = info
         @code_information[cache_key] = {filepath: location.first,
                                         lineno: location.last,
                                         function: "#{'self.' if is_class_method}#{method_name}",
                                         namespace: namespace}.freeze
-      rescue => e
+      rescue StandardError => e
         ::NewRelic::Agent.logger.warn("Unable to determine source code info for '#{object}', " \
                                         "method '#{method_name}' - #{e.class}: #{e.message}")
         ::NewRelic::Agent.increment_metric(SOURCE_CODE_INFORMATION_FAILURE_METRIC, 1)
@@ -101,6 +103,9 @@ module NewRelic
         klass = object.singleton_class? ? klassify_singleton(object) : object
         name = klass.name || '(Anonymous)'
         is_class_method = false
+
+        return controller_info(klass, name, is_class_method) if controller_without_method?(klass, method_name)
+
         method = if (klass.instance_methods + klass.private_instance_methods).include?(method_name)
           klass.instance_method(method_name)
         else
@@ -108,6 +113,22 @@ module NewRelic
           klass.method(method_name)
         end
         [name, method.source_location, is_class_method]
+      end
+
+      # Rails controllers can be a special case because by default, controllers in Rails
+      # automatically render views with names that correspond to valid routes. This means
+      # that a controller method may not have a corresponding method in the controller class.
+      def controller_without_method?(klass, method_name)
+        defined?(Rails) &&
+          defined?(ApplicationController) &&
+          klass < ApplicationController &&
+          !klass.method_defined?(method_name)
+      end
+
+      def controller_info(klass, name, is_class_method)
+        path = Rails.root.join("app/controllers/#{klass.name.underscore}.rb")
+
+        File.exist?(path) ? [name, [path.to_s, 1], is_class_method] : []
       end
     end
   end
--- a/test/new_relic/agent/method_tracer_helpers_test.rb
+++ b/test/new_relic/agent/method_tracer_helpers_test.rb
@@ -147,4 +147,46 @@ class NewRelic::Agent::MethodTracerHelpersTest < Minitest::Test
       assert_predicate memoized.values.first, :frozen?
     end
   end
+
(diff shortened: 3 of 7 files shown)


lib/new_relic/agent/method_tracer_helpers.rb
```diff
@@ -109,6 +116,22 @@ def namespace_and_location(object, method_name)
...
+      # automatically render views with names that correspond to valid routes. This means
+      # that a controller method may not have a corresponding method in the controller class.
+      def controller_without_method?(klass, method_name)
+        defined?(ApplicationController) && klass < ApplicationController && !klass.method_defined?(method_name)
+      end
+
+      def controller_namespace_and_location(klass, method_name, is_class_method, name)
+        return [] unless defined?(Rails)
+
+        path = Rails.root.join("app/controllers/#{klass.name.underscore}.rb")
```
hannahramadan, line 130: This makes an assumption we're looking at a Rails app without nested controllers (haven't found a way to determine the file location of a class). If this fails, we at least will be muting the false warnings.

test/new_relic/agent/method_tracer_helpers_test.rb
```diff
@@ -13,6 +13,19 @@ def private_method; end
...
+
+class String
+  def underscore
+    'test_controller'
+  end
+
+  def join(*args)
+    'User/fake_app/app/controllers/test_controller.rb'
+  end
+end
```
kaylareopelle, line 27: Cool idea! Are these methods still being used? If so, can you show me how they work?

lib/new_relic/agent/method_tracer_helpers.rb
```diff
@@ -46,13 +46,15 @@ def code_information(object, method_name)
...
+        info = namespace_and_location(object, method_name.to_sym)
+        return ::NewRelic::EMPTY_HASH if info.empty?
 
+        namespace, location, is_class_method = info
         @code_information[cache_key] = {filepath: location.first,
                                         lineno: location.last,
                                         function: "#{'self.' if is_class_method}#{method_name}",
                                         namespace: namespace}.freeze
-      rescue => e
+      rescue StandardError => e
```
fallwith, line 57: The use of `StandardError` here is a best practice and should be the default approach. This leads me to believe that the previously shipped code without it went without for a good reason. Do you happen to know what the various error classes that could be raised are? If they are all of type StandardError, then we are ok.

lib/new_relic/agent/method_tracer_helpers.rb
```diff
@@ -109,6 +116,22 @@ def namespace_and_location(object, method_name)
...
+        defined?(ApplicationController) && klass < ApplicationController && !klass.method_defined?(method_name)
+      end
+
+      def controller_namespace_and_location(klass, method_name, is_class_method, name)
+        return [] unless defined?(Rails)
+
+        path = Rails.root.join("app/controllers/#{klass.name.underscore}.rb")
+        return [] unless File.exist?(path)
+
+        [name, [path.to_s, 1], is_class_method]
```
fallwith, line 133: `path` should be a string already and not need the `#to_s` call

lib/new_relic/agent/method_tracer_helpers.rb
kaylareopelle: Have you tried assuming namespacing leads to nesting? For example, if I saw `Users::LoginController` I would look for the file in `app/controllers/users/login_controller.rb`

lib/new_relic/agent/method_tracer_helpers.rb
fallwith: It looks like out of the 4 method arguments, only `klass` is used. Maybe we should refactor this method to only return either a path or a nil, and then the caller itself can say "if I got a nil, return [] but if I got a path, return a 4 item array using the returned path and the 3 items I already have".

lib/new_relic/agent/method_tracer_helpers.rb
```diff
@@ -109,6 +116,22 @@ def namespace_and_location(object, method_name)
...
+
+      # Rails controllers can be a special case because by default, controllers in Rails
+      # automatically render views with names that correspond to valid routes. This means
+      # that a controller method may not have a corresponding method in the controller class.
+      def controller_without_method?(klass, method_name)
+        defined?(ApplicationController) && klass < ApplicationController && !klass.method_defined?(method_name)
+      end
+
+      def controller_namespace_and_location(klass, method_name, is_class_method, name)
+        return [] unless defined?(Rails)
```
kaylareopelle, line 128: What do you think about moving the `Rails` check into `#controller_without_method?` and removing this line? Do we need both `Rails` and `ApplicationController` to be installed in an application to make this namespace finding valid?

lib/new_relic/agent/method_tracer_helpers.rb
kaylareopelle, line 57: 🎉

lib/new_relic/agent/method_tracer_helpers.rb
fallwith: I recommended extreme caution that an app might use `ApplicationController` directly somehow (which we need to be defined for the `<` inheritance check) without having `Rails` itself defined (which we need to be defined for the `Rails.root` call).

But it living in the checker method is a good idea.

lib/new_relic/agent/method_tracer_helpers.rb
fallwith: Yep, @hannahramadan whipped up a classic admin controller with content in an "admin" subdirectory. For a given `Admin::UsersController` class, she found that calling `.name.underscore` on the class would correctly yield the desired `admin/users_controller` value.

fallwith: Oh I see - @kaylareopelle is responding to @hannahramadan's comment here more than the code. @hannahramadan you got nesting working. The only thing you don't have working from my vantage point is support for controllers living outside of `app/controllers`. I'm not sure what hackery is involved to offer that or how common it is, but I think that's ok for now.

lib/new_relic/agent/method_tracer_helpers.rb
kaylareopelle: That makes a lot of sense! I think it's fine to leave as-is. I mostly wanted to know more about the decision and its tradeoffs.

lib/new_relic/agent/method_tracer_helpers.rb
kaylareopelle: You got it, @fallwith! Thanks for providing some context and correcting my misreading of the initial comment. I also think it's okay to skip automatic detection for controllers that live outside `app/controllers`.

hannahramadan: @fallwith @kaylareopelle thanks both! My comment isn't accurate, thanks for correcting it ◡̈

lib/new_relic/agent/method_tracer_helpers.rb
tannalynn: please correct me if i'm wrong, but I believe that not specifying a class, like we were doing before, defaults to `StandardError` being assumed. I think the reason for the best practice being this way is probably more for style/clarity than a difference in actual behavior, so we should be totally fine with this change

kaylareopelle: You're totally right, Tanna. I can't find the source code in Ruby, but [this article](https://www.honeybadger.io/blog/a-beginner-s-guide-to-exceptions-in-ruby/) reiterates what you're saying under "Rescuing All Errors (The Right Way)".

I thought `rescue => e` was rescuing the `Exception` class.

kaylareopelle: Here we are: https://ruby-doc.org/3.2.2/Exception.html#class-Exception-label-Defaults

tannalynn (commented):
Other than things already being discussed, I think things look good! We should add a changelog entry for this before merging though

lib/new_relic/agent/method_tracer_helpers.rb
hannahramadan: When testing, the result of calling `Rails.root` is a [Pathname](https://ruby-doc.org/stdlib-2.6.3/libdoc/pathname/rdoc/Pathname.html). I'm not sure of all the implications of that, but because in other cases we get a String, I thought it would be safer. Can you see the downsides to it being kept as a Pathname object?

I'll work on that refactor!

fallwith: Ah,ok. Check to see if `#join` returns a string instead of another Pathname object. A string won't need a to_s but another Pathname will.

hannahramadan: Looks like `.join` returns another Pathname

fallwith: Perfect. Sorry for the confusion. The to_s stays.

hannahramadan: @fallwith thoughts on this refactor https://github.com/newrelic/newrelic-ruby-agent/pull/2061/commits/a9201f18defa664cc36cf80d5152b6609f69f490

lib/new_relic/agent/method_tracer_helpers.rb
hannahramadan: Sweet sounds like this change is ok! Cool to learn about both raise and rescue defaults.

fallwith (dismissed):

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ Version <dev> of the agent adds the ability to filter logs by level and expands
     * `transmit_subscription_confirmation.action_cable`
     * `transmit_subscription_rejection.action_cable`
 
+- **Bugfix: Report Code Level Metrics for Rails controller methods**
```
kaylareopelle, line 23:
```suggestion
- **Bugfix: Report Code-Level Metrics for Rails controller methods**
```

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ Version <dev> of the agent adds the ability to filter logs by level and expands
...
+
+  Controllers in Rails automatically render views with names that correspond to valid routes. This means that a controller method may not have a corresponding method in the controller class. Code Level Metrics now report on these methods and don't log false warnings. Thanks to [@jcrisp](https://github.com/jcrisp) for reporting this issue. [PR#2061](https://github.com/newrelic/newrelic-ruby-agent/pull/2061)
```
kaylareopelle, line 25:
```suggestion
  Controllers in Rails automatically render views with names that correspond to valid routes. This means that a controller method may not have a corresponding method in the controller class. Code-Level Metrics now report on these methods and don't log false warnings. Thanks to [@jcrisp](https://github.com/jcrisp) for reporting this issue. [PR#2061](https://github.com/newrelic/newrelic-ruby-agent/pull/2061)
```

CHANGELOG.md
kaylareopelle: @hannahramadan - Could you add this to the overview description at the top of the Changelog?

Issue #1141, opened 2022-05-06
Adding custom attributes to the logs via the agent


rajpawar02:
I am addressing the collection of application logs by New Relic APM agents.

I have referred this [document](https://docs.newrelic.com/docs/logs/logs-context/configure-logs-context-ruby) to configure APM agents on one of the [application](https://onenr.io/0kjnLrrZJwo).

However, logs add five attributes by default, namely entity.guid, entity.name, hostname, trace.id, and span.id.

Because we are running several services/containers for the same application, it is extremely difficult to identify a specific ECS service or container log using the above attributes.

Can we use the Ruby APM [agent](https://docs.newrelic.com/docs/release-notes/agent-release-notes/ruby-release-notes/ruby-agent-860) to add custom attributes like container_id, ecs_service_name, and so on? If we do so, it will be extremely beneficial.
Please let us know what you think about this.
Thanks

kaylareopelle:
Hi @rajpawar02! 👋 

Thanks for submitting this feature request. We're intrigued by your idea to add custom attributes to log events and would like to learn more about how you would like this feature to behave.

I'm hoping you can provide some insight for us on a few questions: 

* How do you expect this feature to behave? 

  - For example, we have the ability to add custom attributes to transaction traces by calling the method [NewRelic::Agent#add_custom_attributes](https://www.rubydoc.info/gems/newrelic_rpm/NewRelic/Agent#add_custom_attributes-instance_method) and to span events using the method [NewRelic::Agent#add_custom_span_attributes](https://www.rubydoc.info/gems/newrelic_rpm/NewRelic/Agent#add_custom_span_attributes-instance_method). Would a method similar to these achieve your goal? 

  - Alternatively, we have the ability to set the app name (entity.name in the default attributes) and hostname by configuration in either the `newrelic.yml` file or by environment variable. Would you prefer to set attributes like container_id and ecs_service_name using this strategy?

* Is there a specific list of attributes you would always like to have added to a log event? 

  - If this is the case, what are your thoughts about having your desired attributes included in log events by default instead of as custom attributes? This feature wouldn't necessarily have to replace custom attributes. Rather, it's helpful for us to know if there's linking metadata customers expect to have on events that we do not currently support

* Are the custom attributes you're hoping to add static values that could be determined at startup, dynamic attributes that should be added at runtime, or both? 

Your responses to these questions would be very helpful! We're investigating this prospect and may have more questions for you as our efforts continue.

[...]

PR #2069, opened 2023-06-13
Create add_custom_log_attributes API

kaylareopelle:
**See #2084 instead** (git mistakes made on this PR)

The `NewRelic::Agent.add_custom_log_attributes` API allows users to add attributes to all log events for a given agent run. It accepts key/value pairs as arguments.

This API abides by log API documentation on [limits](https://docs.newrelic.com/docs/logs/log-api/introduction-log-api/#limits), and [event data](https://docs.newrelic.com/docs/logs/log-api/log-event-data/):
* No more than 255 attributes can be added (limited to 240 for the agent at this time)
* Keys must have less than 64 characters
* Values must have less than 4094 characters
* Key/value pairs with empty content will be dropped
* Values longer than 4094 characters that aren't Strings/Symbols will be dropped

These attributes will be sent to the logs API in the ["common"["attributes"] JSON block](https://docs.newrelic.com/docs/logs/log-api/introduction-log-api/#log-attribute-example).

Relates to: #1141


lib/new_relic/agent.rb
```diff
@@ -649,6 +649,23 @@ def add_custom_span_attributes(params)
       end
     end
 
+    # Add custom attributes to log events for the current agent instance.
+    #
+    # @param [Hash] params    A Hash of attributes to be attached to log events.
+    #                         Keys should be Strings or Symbols, and values
+    #                         may be Strings, Symbols, numeric values or
+    #                         Booleans.
```
kaylareopelle, line 657: Should I add specifics about the attribute limitations here?

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -160,6 +180,64 @@ def enabled?
 
       private
 
+      def add_custom_attribute(key, value)
```
kaylareopelle, line 183: I may make a separate PR when this wraps up to organize all the private methods alphabetically.

lib/new_relic/agent.rb
```diff
@@ -649,6 +649,40 @@ def add_custom_span_attributes(params)
...
+    #                         4095 characters. If the value is a String
+    #                         or a Symbol, values longer than 4094
+    #                         characters will be truncated.
+    #
+    #                         This API can be called multiple times.
+    #                         If the same key is passed more than once,
+    #                         the value associated with the last call
+    #                         will be preserved.
+    #
+    #                         Key/value pairs with empty or nil values
```
kaylareopelle, line 673:
```suggestion
    #                         Key/value pairs with empty or nil contents
```

test/new_relic/agent/log_event_attributes_test.rb
```diff
@@ -0,0 +1,206 @@
...
+
+      refute_includes(common_attributes_from_melt, nil)
+      refute_includes(common_attributes_from_melt, '')
+    end
+
+    def test_log_attr_nil_value_drops_attribute
+      key = 'key'
+      NewRelic::Agent.add_custom_log_attributes(key => nil)
+
+      refute_includes(common_attributes_from_melt, [key])
```
kaylareopelle, line 168:
```suggestion
      refute_includes(common_attributes_from_melt, key)
```

lib/new_relic/agent.rb
```diff
@@ -649,6 +649,43 @@ def add_custom_span_attributes(params)
...
+    #                         or a Symbol, values longer than 4094
+    #                         characters will be truncated. If the value
+    #                         exceeds 4094 characters and is of a
+    #                         different class, the attribute pair will
+    #                         be dropped.
+    #
+    #                         This API can be called multiple times.
+    #                         If the same key is passed more than once,
+    #                         the value associated with the last call
+    #                         will be preserved.
```
fallwith, line 674: These comments seem to suggest that calling the API multiple times will have an additive effect (hash merge). Is there any need to a) reveal what the current set of attributes is and/or b) offer the ability to remove existing attributes?

lib/new_relic/agent.rb
```diff
@@ -649,6 +649,43 @@ def add_custom_span_attributes(params)
...
+    #
+    #                         Attribute pairs with empty or nil contents
+    #                         will be dropped.
+    # @api public
+    def add_custom_log_attributes(params)
+      record_api_supportability_metric(:add_custom_log_attributes)
+
+      if params.is_a?(Hash)
```
fallwith, line 682: I'm not familiar with our API coding best practices, so maybe this is covered somewhere... but as a customer I'm thinking it would be really cool if you returned one thing to me if my call was successful (perhaps return my original input? or perhaps the current full list of attributes?) and another thing if the call failed (perhaps return `nil`).

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -151,6 +161,8 @@ def reset!
           @seen = 0
           @seen_by_severity.clear
         end
+
+        attributes.reset!
```
tannalynn, line 165: When is `reset!` called for this class? Are we sure that we would also want to reset the attributes that were set by the customer for future logs if this happens? 
If this is only a test thing, then it totally makes sense, but if we also do this when we reset the connection with the collector, i'm not sure.

lib/new_relic/agent/log_event_attributes.rb
```diff
@@ -0,0 +1,123 @@
...
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    class LogEventAttributes
+      MAX_ATTRIBUTE_COUNT = 240 # limit is 255, assume we send 15
+      ATTRIBUTE_KEY_CHARACTER_LIMIT = 255
+      ATTRIBUTE_VALUE_CHARACTER_LIMIT = 4094
+
+      attr_reader :custom_attributes, :already_warned_custom_attribute_count_limit
```
hannahramadan, line 12: The naming for `already_warned_custom_attribute_count_limit` kinda leads me to think there's just a warning about reaching a limit but it's actually making a cut.  Could be `custom_attribute_count_limit_reached`? Unless the important takeaway is that we sent a warning off.

lib/new_relic/agent/log_event_attributes.rb
```diff
@@ -0,0 +1,123 @@
...
+        return if already_warned_custom_attribute_count_limit
+
+        attributes.each do |key, value|
+          next if absent?(key) || absent?(value)
+
+          add_custom_attribute(key, value)
+        end
+      end
+
+      def reset!
```
fallwith, line 29: I have an idea to run by you... how about lazily setting the instance variables so that we don't have them taking up memory by default?

Here's what I'm thinking:

1) The `attr_reader` line gets completely removed

2) The `initialize` method gets completely removed

3) The `reset!` method changes to this:

```ruby
def reset!
  %i[custom_attributes already_warned_custom_attribute_count_limit].each do |attr|
    remove_instance_variable(attr) if instance_variable_defined?(attr)
  end
end
```

4) Define a new `custom_attributes` method:

```ruby
def custom_attributes
  @custom_attributes ||= {}
end
```

I'm thinking that those changes would be all that would be required with no other changes needed. The `@already_warned_custom_attribute_count_limit` instance var would be `nil` / undefined, which would evaluate as falsey, which we want by default.

lib/new_relic/agent.rb
```diff
@@ -649,6 +649,43 @@ def add_custom_span_attributes(params)
...
+    #                         be less than 256 characters. Keys longer
+    #                         than 255 characters will be truncated.
+    #
+    #                         Values may be Strings, Symbols, numeric
+    #                         values or Booleans and must be less than
+    #                         4095 characters. If the value is a String
+    #                         or a Symbol, values longer than 4094
+    #                         characters will be truncated. If the value
+    #                         exceeds 4094 characters and is of a
+    #                         different class, the attribute pair will
```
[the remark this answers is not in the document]
kaylareopelle: Yes, that's the intent of this phrase. 

Strings and symbols can be truncated. A value that isn't a string or symbol can't be truncated if it exceeds the limit. In that case, the key/value pair will not be sent. I think the "of a different class" was to provide a heads-up that floats, integers, etc. will just be dropped if they're too large.

Happy to change the phrasing to something that's clearer.

lib/new_relic/agent.rb
kaylareopelle: I'm not sure where I'd reveal the current set of attributes to a user in a meaningful way. Our logs are the only way we communicate with our customers. I could log a message with the current attributes each time a change is made. Would you like me to add that?

To remove existing attributes, a user could submit a `nil` or empty value to an existing key:
`NewRelic::Agent.add_custom_log_attributes(existing: nil)` or `NewRelic::Agent.add_custom_log_attributes(existing: '')`

Would you like me to describe how to remove an attribute with the current functionality? Is that suggestion too janky and it needs an API of its own? 

I'm also a bit confused about the case where you'd want to remove an attribute. You call the API to add the attribute, wouldn't you remove the API call or the key/value from the API call if you didn't want the attribute to be sent anymore?

tannalynn: It seems like removing attributes doesn't seem like something we should put much effort into support. No one is requesting that currently, so if someone does want it they can always submit a feature request for it, but otherwise it seems like unnecessary effort to add that functionality to the existing PR.

lib/new_relic/agent.rb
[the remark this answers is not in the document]
fallwith: I see now. The sentence as-is works great when paired with the sentence immediately preceding it.

Separate from value truncation.,. If the hash key itself is not on the allowlist of class types, does that key/value pair get dropped or perhaps the entire hash get rejected?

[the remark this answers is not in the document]
kaylareopelle: At this time, all keys are coerced into strings before they're evaluated for truncation. This follows the logs API: all keys must be strings. If it can respond to `#to_s`, we'll make it a string for you.

Would you like me to stop coercing before truncation? 

I could see an issue if you had a key that was an integer that should be truncated. It'd be coerced into a string and then truncated with a different value that would be sent to New Relic.

CHANGELOG.md
```diff
@@ -14,6 +14,14 @@ Version <dev> of the agent adds the ability to filter logs by level, expands ins
 
   This setting uses [Ruby's Logger::Severity constants integer values](https://github.com/ruby/ruby/blob/master/lib/logger/severity.rb#L6-L17) to determine precedence.
 
+- **Feature: Custom attributes for logs API**
+
+  You can now add custom attributes to your log events using `NewRelic::Agent.add_custom_log_attributes`.
+
+  For example: `NewRelic::Agent.add_custom_log_attributes(dyno: ENV.fetch('DYNO'), pod_name: ENV.fetch('POD_NAME'))`, will add the attributes `dyno` and `pod_name` to your log events. Attributes passed to this API will be added to all log events.
```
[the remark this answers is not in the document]
kaylareopelle:
```suggestion
  For example: `NewRelic::Agent.add_custom_log_attributes(dyno: ENV['DYNO'], pod_name: ENV['POD_NAME'])`, will add the attributes `dyno` and `pod_name` to your log events. Attributes passed to this API will be added to all log events.
```

The changelog made a whole bunch of other changes in my text editor this time, so I'm going for a GH edit.

lib/new_relic/agent/log_event_aggregator.rb
kaylareopelle: Removed in b83c43b20d62e9d6a0341aa0199c53825c409bfa

lib/new_relic/agent/log_event_attributes.rb
kaylareopelle: Great idea! Thanks for sharing. Incorporated in 3622f6a9b2a2ea8c74b3ec75ffac391fd303587d

lib/new_relic/agent/log_event_attributes.rb
kaylareopelle: Renamed here: 8fe40f1ad13aab82fc488efe0bf0f0bf7629b0e2 
I took out "count" because it didn't seem necessary to me, but it does match the constant name more closely. Let me know if you'd like me to put it back in.

lib/new_relic/agent.rb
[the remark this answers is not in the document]
fallwith: Got it. If length (not type) is the only concern because we're coercing and we're allowing 4095 for `#size`, I think we're ok.

lib/new_relic/agent.rb
kaylareopelle: I agree with Tanna and would also prefer not to add an attribute removal API at this time.

@fallwith, would you like me to change the return value here to be the current attributes?

lib/new_relic/agent.rb
fallwith: > Would you like me to change the return value to match your suggestion?

Given that a) we don't have a standard, b) you've modeled this new API closely off of a fairly similar one, and c) our communication back to the caller is traditionally done via logged messages, no. Thank you for the explanation!

fallwith (dismissed):

kaylareopelle:
If you’re up for it, I’d like to meet synchronously tomorrow to talk this through. I’ve been in a weird place emotionally today and I think my problems have seeped into this PR. Sent from my iPhoneOn Jun 14, 2023, at 18:02, James Bunch ***@***.***> wrote:﻿
@fallwith commented on this pull request.



In lib/new_relic/agent.rb:
> +    #                         exceeds 4094 characters and is of a
+    #                         different class, the attribute pair will
+    #                         be dropped.
+    #
+    #                         This API can be called multiple times.
+    #                         If the same key is passed more than once,
+    #                         the value associated with the last call
+    #                         will be preserved.
+    #
+    #                         Attribute pairs with empty or nil contents
+    #                         will be dropped.
+    # @api public
+    def add_custom_log_attributes(params)
+      record_api_supportability_metric(:add_custom_log_attributes)
+
+      if params.is_a?(Hash)


Would you like me to change the return value to match your suggestion?

Given that a) we don't have a standard, b) you've modeled this new API closely off of a fairly similar one, and c) our communication back to the caller is traditionally done via logged messages, no. Thank you for the explanation!

fallwith:
> If you’re up for it, I’d like to meet synchronously tomorrow to talk this through.

Sure thing.

kaylareopelle:
**See #2084 instead** (git mistakes made on this PR)

PR #2075, opened 2023-06-13
Rake: Generate newrelic.yml

### LINKED ISSUE #1147: Create command to automatically update newrelic.yml
When adding a new configuration option to `default_source.rb`, we currently need to manually update newrelic.yml. It would be beneficial to add a new rake task that can automatically do this for us.

hannahramadan:
A new rake task to update `newrelic.yml` with information from `default_source.rb`. A freshly generated newrelic.yml using this task takes public configs and their description, config option, and default value and populates them into a new file.

resolves #1147


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1592,7 +1591,7 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :public => true,
           :type => String,
           :allowed_from_server => false,
-          :description => 'Specify a custom host name for [display in the New Relic UI](/docs/apm/new-relic-apm/maintenance/add-rename-remove-hosts#display_name).'
+          :description => 'Specify a custom host name for [display in the New Relic UI](/docs/apm/new-relic-apm/maintenance/add-rename-remove-hosts#display_name). Be aware that you cannot rename a hostname, so please rename carefully.'
```
fallwith, line 1594: Maybe there's another way to word the warning here. Seeing "you can't rename" sitting right next to "rename carefully" stands out.

test/new_relic/newrelicyml_test.rb
```diff
@@ -0,0 +1,144 @@
...
+  def test_default_value_use_documentation_default
+    assert_equal 2, NewRelicYML.default_value(:lily, FAKE_CONFIGS[:lily])
+  end
+
+  def test_default_value_string
+    assert_equal '""', NewRelicYML.default_value(:monstera, FAKE_CONFIGS[:monstera])
+  end
+
+  def test_default_value_nil
+    assert_equal 'nil', NewRelicYML.default_value(:begonia, FAKE_CONFIGS[:begonia])
```
fallwith, line 26: Is the 'nil' string expected? I would have thought either `''` or `nil` would be here.

test/new_relic/newrelicyml_test.rb
```diff
@@ -0,0 +1,144 @@
...
+      }
+    }
+
+    config_hash
+  end
+
+  # rubocop:disable Style/RedundantStringEscape
+  def final_string
+    <<~YML
+      \ \ # If true, green with white polka dots.
```
fallwith, line 129: Are the backslashes used to preserve spaces? If so, maybe a different style HEREDOC declaration would work that doesn't trim away the spaces.

lib/tasks/helpers/newrelicyml.rb
```diff
@@ -0,0 +1,142 @@
...
+
+    configs.each do |key, value|
+      yml_string += "#{value[:description]}\n  # #{key}: #{value[:default]}\n\n"
+    end
+
+    yml_string
+  end
+
+  def self.write_file(defaults = DEFAULTS)
+    File.write('newrelic.yml', HEADER + build_string(defaults) + FOOTER)
```
fallwith, line 140: It would be helpful if you could include as a PR comment a diff of how a newly generated .yml file looks compared to what we have now.

lib/new_relic/agent/configuration/default_source.rb
tannalynn: Yes, I also find that confusing. I don't see a note like that on the docs site either, where is that coming from?

lib/tasks/helpers/newrelicyml.rb
```diff
@@ -0,0 +1,142 @@
...
+      app_name: <%= app_name %> (Staging)
+
+    production:
+      <<: *default_settings
+  FOOTER
+
+  def self.get_configs(defaults)
+    final_configs = {}
+
+    defaults.sort.each do |key, value|
```
fallwith, line 72: You could smoosh the need for the new hash object into an `each_with_object` call: `defaults.sort.each_with_object({}) do |(key, value), final_config|`

lib/tasks/helpers/newrelicyml.rb
kaylareopelle: This LOC is the only one that looks like it isn't covered by tests according to Simplecov.

lib/tasks/helpers/newrelicyml.rb
```diff
@@ -0,0 +1,142 @@
...
+      next if CRITICAL.include?(key) || SKIP.include?(key)
+
+      if public_config?(value)
```
fallwith, line 75: It looks like there isn't an `else` scenario. Maybe this should be guarded with a `next unless` similar to the `next if` above it.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1692,14 +1691,14 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
-          :description => 'Sets the maximum number of span events to buffer when streaming to the trace observer.'
+          :description => 'Sets the maximum number of span events to buffer when streaming to the Trace Observer.'
         },
         :'span_events.max_samples_stored' => {
           :default => 2000,
           :public => true,
           :type => Integer,
           :allowed_from_server => true,
-          :description => 'Defines the maximum number of span events reported from a single harvest. Any Integer between `1` and `10000` is valid.'
+          :description => 'Defines the maximum number of span events reported from a single harvest. Any Integer between `1` and `10000` is valid. Increasing this value may impact memory usage'
```
tannalynn, line 1701: I'm not sure we really need to add this here. It hasn't really been an issue people have asking about in the past and this is generally true for a lot of settings in the agent. So calling out this one config in particular for it doesn't make much sense to me.

lib/tasks/helpers/newrelicyml.rb
```diff
@@ -0,0 +1,142 @@
...
+        default = default_value(key, value)
+        final_configs[key] = {description: description, default: default}
+      end
+    end
+
+    final_configs
+  end
+
+  def self.public_config?(value)
+    value[:public] == true
```
fallwith, line 87: It'd be neat if we could have a config object class and ask an instance if it were public with `#public?` but I understand we don't want to alter the existing default source logic.

lib/tasks/helpers/newrelicyml.rb
```diff
@@ -0,0 +1,142 @@
...
+    description.gsub!(/`([^`]+)`/, '\1')
+    # removed hrefs links
+    description.gsub!(/<a href="(.*)">(.*)<\/a>/, '\2')
+
+    description
+  end
+
+  def self.format_description(description)
+    # remove leading and trailing whitespace
+    description = description.strip
```
fallwith, line 107: Is there a `#strip!` to make this easier?

lib/tasks/helpers/newrelicyml.rb
```diff
@@ -0,0 +1,142 @@
...
+
+    description
+  end
+
+  def self.default_value(key, config_hash)
+    if PROCS.include?(key)
+      PROCS[key]
+    else
+      default = config_hash[:documentation_default].nil? ? config_hash[:default] : config_hash[:documentation_default]
+      default = 'nil' if default.nil?
```
fallwith, line 121: Ah here's the `'nil'`. I'd like to know why this is needed over `nil` or `''`.

lib/tasks/helpers/newrelicyml.rb
```diff
@@ -0,0 +1,142 @@
...
+      default = '""' if default == ''
```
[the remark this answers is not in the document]
tannalynn: Yes, `'' == ""`  is true, so this should work for both.

lib/tasks/helpers/newrelicyml.rb
tannalynn: In my rake task PR, I also didn't test the lines that did the actual file writing and reading, instead I just tested that what I was planning to write was correct.

fallwith: For brand new code...

ideally we would either a) have 100% code coverage, or b) use `:nocov:` sparingly to make it explicitly clear that we don't want to bother with covering something. See [skipping code](https://github.com/simplecov-ruby/simplecov#ignoringskipping-code).

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: I added this because at some point, someone found it relevant to add to our current `newrelic.yml`! We don't have to keep this, just as I was going through, this is the kind of thing that would be overwritten from changes we made to newrelic.yml overtime.

test/new_relic/newrelicyml_test.rb
hannahramadan: Yes! Unless I use the string version of `nil` or `""`, a blank shows up in `newrelic.yml` since its literally using nil.

lib/tasks/helpers/newrelicyml.rb
```diff
@@ -0,0 +1,142 @@
...
+
+module NewRelicYML
+  CRITICAL = [:'agent_enabled', :'app_name', :'license_key', :'log_level']
+  DEFAULTS = NewRelic::Agent::Configuration::DEFAULTS
+  # Skip because not configurable via yml
+  SKIP = [:'defer_rails_initialization']
+  # Don't evaluate Procs, instead use set values
+  PROCS = {:'config_path' => 'newrelic.yml',
+           :'process_host.display_name' => 'default hostname',
+           :'transaction_tracer.transaction_threshold' => 1.0}
```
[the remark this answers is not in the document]
hannahramadan: None of these have a documentation default. Our docs website will say `(DYNAMIC`). There must be code at some point that tells the docs website to use DYNAMIC if one can't be found. [Example](https://docs.newrelic.com/docs/apm/agents/ruby-agent/configuration/ruby-agent-configuration/#process-host).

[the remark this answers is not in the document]
hannahramadan: I would be nervous to hard code a documentation default since the values may not actually default to, for example, `default hostname`

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: I've reverted this change. We can address these specifics when we run this task against our current newrelic.yml 
https://github.com/newrelic/newrelic-ruby-agent/pull/2075/commits/5476256550f0b182eb525ff29d93e7bddc0ffcaf

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1264,9 +1263,9 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :type => String,
           :allowed_from_server => false,
           :external => :infinite_tracing,
-          :description => 'Configures the hostname for the trace observer Host. ' \
+          :description => 'Configures the hostname for the Trace Observer Host. ' \
             'When configured, enables tail-based sampling by sending all recorded spans ' \
-            'to a trace observer for further sampling decisions, irrespective of any usual ' \
+            'to a Trace Observer for further sampling decisions, irrespective of any usual ' \
```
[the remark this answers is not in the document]
hannahramadan: I've changed them all to lowercase https://github.com/newrelic/newrelic-ruby-agent/pull/2075/commits/5476256550f0b182eb525ff29d93e7bddc0ffcaf

lib/new_relic/agent/configuration/default_source.rb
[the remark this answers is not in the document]
hannahramadan: I've reverted this change. We can review changes like this once the task is actually ran and a new newrelic.yml is generated https://github.com/newrelic/newrelic-ruby-agent/pull/2075/commits/5476256550f0b182eb525ff29d93e7bddc0ffcaf

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: I've reverted this change. 
https://github.com/newrelic/newrelic-ruby-agent/commit/5476256550f0b182eb525ff29d93e7bddc0ffcaf

lib/tasks/helpers/newrelicyml.rb
hannahramadan: Nice suggestion ◡̈ https://github.com/newrelic/newrelic-ruby-agent/pull/2075/commits/2d0e7236d726b06c6f8187d3aa08d92c083d5aed

lib/tasks/helpers/newrelicyml.rb
hannahramadan: I like - done! https://github.com/newrelic/newrelic-ruby-agent/pull/2075/commits/2d0e7236d726b06c6f8187d3aa08d92c083d5aed

lib/tasks/helpers/newrelicyml.rb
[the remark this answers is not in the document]
kaylareopelle: The code that tells the docs site to use dynamic is in our config generation task [(link)](https://github.com/newrelic/newrelic-ruby-agent/blob/3140248bbba114602e22e46742fe0ad3667005d6/lib/tasks/helpers/format.rb#L61). 

We've overwritten a lot of the procs now with documentation defaults, but I can understand why we've left these. Thanks for sharing your thought process!

lib/tasks/helpers/newrelicyml.rb
hannahramadan: We need the string version of `nil` and the string of double quotes`""` because we are putting these in the yml as strings. If I don't stringify these, the values get evaluated and then we end up with blank values. ie `default: ` vs `default: nil`

lib/tasks/helpers/newrelicyml.rb
hannahramadan: I was getting a frozen string error in my tests when specifically calling `strip!`.

lib/tasks/newrelicyml.rake
```diff
@@ -0,0 +1,13 @@
...
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require_relative './helpers/newrelicyml'
+
+namespace :newrelic do
+  desc 'Update newrelic.yml with latest config options from default_source.rb'
+  task :update_newrelicyml do
+    NewRelicYML.write_file
+    puts 'newrelic.yml updated'
```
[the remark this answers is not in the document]
hannahramadan: Hmm I guess if there was something put into `default_source` that wasn't expected. Are you thinking I should add a `rescue`?

lib/tasks/helpers/newrelicyml.rb
hannahramadan: Are we okay with skipping code coverage on rake tasks? Or perhaps we can pair on it.

lib/tasks/newrelicyml.rake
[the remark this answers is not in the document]
tannalynn: I think we would want to know if it fails though, that way we can address it. If we rescue it, then it would just silently fail, which doesn't seem ideal.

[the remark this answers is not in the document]
hannahramadan: Agree Tanna! Is this fine as is then?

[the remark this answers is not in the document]
tannalynn: I think so. I think unless there is a specific error we would want to have logic around handling, the error failing the task seems good to me.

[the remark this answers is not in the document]
kaylareopelle: Yep, you're both right. I'm being paranoid. This is fine.

lib/tasks/helpers/newrelicyml.rb
kaylareopelle: I think this is okay since everything else is so well-tested. 

I'm also fine with James's recommendation to use `:nocov:` where we're aware test coverage is missing and intentionally didn't add it.

fallwith (approved):

tannalynn (approved):
🎉

test/new_relic/newrelicyml_test.rb
```diff
@@ -0,0 +1,148 @@
...
+    :'pothos' => {
+      :default => '',
+      :public => false,
+      :description => 'Low <InlinePopover type="maintenance" />.'
+    },
+    :'ivy' => {
+      :default => '',
+      :public => true,
+      :deprecated => true,
+      :description => 'This is deprecated'
```
kaylareopelle, line 90: I like that you added this for the deprecated case (also hilarious), but I don't see it used anywhere in the tests. Am I missing something?

hannahramadan: oh oh. I guess because another method uses the entire config hash and runs it through a method that calls the new `deprecated?` method

kaylareopelle: Ah, got it! I see it now. Up to you if you want to add a test to make sure deprecated is ignored. I'll approve as-is.

kaylareopelle (approved):

PR #2083, opened 2023-06-15
Fix for compatibility issue with Async 2.6.2

### LINKED ISSUE #2062: Async 2.6.0 & fiber-annotation
Async 2.6.0 brings in a new dependency with [`fiber-annotation`](https://github.com/ioquatix/fiber-annotation) which I believe is causing some issues with the fiber instrumentation:

```ruby
Exception: ArgumentError
Error: wrong number of arguments (given 1, expected 0)
/usr/local/bundle/gems/fiber-annotation-0.2.0/lib/fiber/annotation.rb:13:in `initialize'
/usr/local/bundle/gems/newrelic_rpm-9.2.2/lib/new_relic/agent/instrumentation/fiber/prepend.rb:14:in `block in initialize'
/usr/local/bundle/gems/newrelic_rpm-9.2.2/lib/new_relic/agent/instrumentation/fiber/instrumentation.rb:11:in `initialize_with_newrelic_tracing'
/usr/local/bundle/gems/newrelic_rpm-9.2.2/lib/new_relic/agent/instrumentation/fiber/prepend.rb:14:in `initialize'
/usr/local/bundle/gems/async-2.6.0/lib/async/task.rb:346:in `new'
/usr/local/bundle/gems/async-2.6.0/lib/async/task.rb:346:in `schedule'
/usr/local/bundle/gems/async-2.6.0/lib/async/task.rb:179:in `run'
/usr/local/bundle/gems/async-2.6.0/lib/async/scheduler.rb:283:in `async'
/usr/local/bundle/gems/async-2.6.0/lib/async/scheduler.rb:247:in `run'
/usr/local/bundle/gems/async-2.6.0/lib/kernel/async.rb:32:in `Async'
```

This is on Ruby 3.2.2 and `newrelic_rpm` 9.2.2.

tannalynn:
The issue was that we were capturing `*args` and passing them on for our Fiber instrumentation. Actually though, you can't actually pass arguments to `Fiber.new` the same way you can as with `Thread.new` (you would accomplish it like `fiber.resume(args)`). This wasn't causing any issues on it's own, however Async released a new version with [fiber-annotation](https://github.com/ioquatix/fiber-annotation), that defines the Fiber initialize method as accepting only **kwargs, and not *args. 

[`lib/fiber/annotation.rb`](https://github.com/ioquatix/fiber-annotation/blob/main/lib/fiber/annotation.rb#L13)
```
def initialize(annotation: nil, **options, &block)
	@annotation = annotation
	super(**options, &block)
end
```

In older Fiber versions, it seems there are no params you can pass to `Fiber.new`, but in newer ruby versions, there are keyword arguments that it accepts. Since we still support older ruby versions, we will define our fiber instrumentation initialize method differently depending on ruby version. This will allow our instrumentation to work on all supported ruby versions, and will not cause errors when using Async 2.6.2. 

closes #2062

- I also made the way we pass args to thread.new more correct with how it should be working
- I also discovered our thread multiverse suite wasn't actually running any tests because if the string we pass to gemfile is empty, we just don't do anything. So i added a comment in the string. 
- I also added tests for checking to make sure args are being correctly passed for both fiber and thread

### DIFF
--- a/lib/new_relic/agent/instrumentation/fiber/chain.rb
+++ b/lib/new_relic/agent/instrumentation/fiber/chain.rb
@@ -10,9 +10,16 @@ module NewRelic::Agent::Instrumentation
 
         alias_method(:initialize_without_new_relic, :initialize)
 
-        def initialize(*args, &block)
-          traced_block = add_thread_tracing(*args, &block)
-          initialize_with_newrelic_tracing { initialize_without_new_relic(*args, &traced_block) }
+        if RUBY_VERSION < '2.7.0'
+          def initialize(*args, &block)
+            traced_block = add_thread_tracing(&block)
+            initialize_with_newrelic_tracing { initialize_without_new_relic(&traced_block) }
+          end
+        else
+          def initialize(**kwargs, &block)
+            traced_block = add_thread_tracing(&block)
+            initialize_with_newrelic_tracing { initialize_without_new_relic(**kwargs, &traced_block) }
+          end
         end
       end
     end
--- a/lib/new_relic/agent/instrumentation/fiber/chain.rb
+++ b/lib/new_relic/agent/instrumentation/fiber/chain.rb
@@ -11,7 +11,7 @@ module NewRelic::Agent::Instrumentation
         alias_method(:initialize_without_new_relic, :initialize)
 
         if RUBY_VERSION < '2.7.0'
-          def initialize(*args, &block)
+          def initialize(*_args, &block)
             traced_block = add_thread_tracing(&block)
             initialize_with_newrelic_tracing { initialize_without_new_relic(&traced_block) }
           end
--- a/test/multiverse/suites/thread/Envfile
+++ b/test/multiverse/suites/thread/Envfile
@@ -5,5 +5,5 @@
 instrumentation_methods :chain, :prepend
 
 gemfile <<~RB
-  # this can't be empty or the tests wont run
+  # this can't be empty or the tests won't run
 RB
--- a/test/multiverse/suites/thread/Envfile
+++ b/test/multiverse/suites/thread/Envfile
@@ -4,4 +4,6 @@
 
 instrumentation_methods :chain, :prepend
 
-gemfile nil
+gemfile <<~RB
+  # this can't be empty or the tests wont run
+RB
--- a/test/multiverse/suites/thread/thread_fiber_instrumentation_test.rb
+++ b/test/multiverse/suites/thread/thread_fiber_instrumentation_test.rb
@@ -73,4 +73,16 @@ class ThreadFiberInstrumentationTest < Minitest::Test
   def test_parents_fiber_thread
     run_nested_parent_test(Fiber, Thread)
   end
+
+  def test_thread_instrumentation_args_preserved
+    Thread.new('arg 1', 2) do |arg1, arg2|
+      assert_equal ['arg 1', 2], [arg1, arg2]
+    end.join
+  end
+
+  def test_fiber_instrumentation_args_preserved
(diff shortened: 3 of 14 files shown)


test/multiverse/suites/thread/thread_fiber_instrumentation_test.rb
```diff
@@ -73,4 +73,16 @@ def test_parents_thread_fiber
...
+    Thread.new('arg 1', 2) do |arg1, arg2|
+      assert_equal ['arg 1', 2], [arg1, arg2]
+    end.join
+  end
+
+  def test_fiber_instrumentation_args_preserved
+    Fiber.new do |arg1, arg2|
+      assert_equal ['arg 1', 2], [arg1, arg2]
+    end.resume('arg 1', 2)
+  end
```
kaylareopelle, line 87: Is the agent running for these test cases?

tannalynn: Yes, the other tests in the files are making sure the agent is recording stuff and nesting spans correctly, so the agent is running and the instrumentation is installed. I didn't add a transaction to these tests since I didn't need to assert anything about a transaction and only cared about making sure the instrumentation was passing the args properly.

lib/new_relic/agent/instrumentation/fiber/chain.rb
```diff
@@ -10,9 +10,16 @@ def self.instrument!
 
         alias_method(:initialize_without_new_relic, :initialize)
 
-        def initialize(*args, &block)
-          traced_block = add_thread_tracing(*args, &block)
-          initialize_with_newrelic_tracing { initialize_without_new_relic(*args, &traced_block) }
+        if RUBY_VERSION < '2.7.0'
+          def initialize(*args, &block)
```
fallwith, line 14: I think `*args` should be `*_args` here to convey that we don't make use of it.

tannalynn: oh right! you mentioned that earlier and I totally blanked on it, my bad. thanks for the reminder.
updated in https://github.com/newrelic/newrelic-ruby-agent/pull/2083/commits/ec8a03b12a23daeebf96a737ecd9cdc71a6fb922

fallwith (approved):

kaylareopelle (approved):
🎉

test/multiverse/suites/thread/Envfile
```diff
@@ -4,4 +4,6 @@
 
 instrumentation_methods :chain, :prepend
 
-gemfile nil
+gemfile <<~RB
+  # this can't be empty or the tests won't run
```
hannahramadan, line 8: Ooo good to know. Were they not running before?

hannahramadan (approved):

test/multiverse/suites/thread/Envfile
tannalynn: apparently not! I know they ran at some point in the past, because they did when i was working on them, but I think i ended up switching to passing nil to gemfile as an afterthought before merging in that PR iirc, because we thought that would work fine, so that would have have broken it.

fallwith (approved):

isaacbowen:
@tannalynn can y'all get this out in a patch release soon? this is preventing us from using Async. :)

tannalynn:
This fix will be included in 9.3.0, which we plan on releasing next week. I'll let you know when it's available once it's been published. Thank you!

tannalynn:
@isaacbowen 9.3.0, which includes this fix, has been released! You should be able to use it now, please let me know if you have any other issues.

PR #2092, opened 2023-06-21
Bugfix for code level metrics on active record classes

### LINKED ISSUE #2064: Custom method tracing for class methods shows error when code_level_metrics are enabled and class overrides inspect method
## Description
Recently when updating `newrelic_rpm` from version 8.8.0 to 9.2.0 in our Rails 7 project, we encountered errors for all of our custom method traces that were applied to class methods. These were set up in an initializer and had been working for all previous versions.

```
WARN : Unable to determine source code info for '#<Class:Foo(id: integer, created_at: datetime, updated_at: datetime, name: string)>', method 'drive' - NameError: wrong constant name Foo(id: integer, created_at: datetime, updated_at: datetime, name: string)

        Object.const_get(klass_name(object))
              ^^^^^^^^^^
```

## Expected Behavior

Class methods should support custom method tracing as per the documentation.

## Troubleshooting results

After some digging, we determined that what was shown in the logs was the output of the `ActiveRecord` override of `inspect`, which was not in the format expected by the `klass_name` method:

https://github.com/newrelic/newrelic-ruby-agent/blob/85fcaa0e8937b566734b2feb52078540c3e3a016/lib/new_relic/agent/method_tracer_helpers.rb#L68-L76

This method is only called when `code_level_metrics.enabled` is `true` and the default value was changed in the 8.10.0 release, introducing this bug.

## Steps to Reproduce

Add custom method tracing to a class method for a class that inherits from ActiveRecord:

```ruby
Foo.class_eval do
  class << self

    add_method_tracer :drive, "Custom/Foo.drive"
  end
end
```

## Additional context

Class methods for classes that don't inherit from `ActiveRecord` did not generate any errors or warnings; disabling `code_level_metrics` eliminated the problem. While this was encountered in a Rails app, we believe it will affect any class that overrides `inspect`.

tannalynn:
The regex the agent was using didn't account for the possibility of active record adding attributes to inspect, which resulted in trying to call `Object.const_get` with a class name that was like `"Foo(id: integer, created_at: datetime, updated_at: datetime, name: string)"`, instead of just `"Foo"`. This would cause an error `NameError: wrong constant name` that prevent code level metrics from gathering the information and would log a warning.
Now, the regex will only grab word characters and :: for the name, preventing the `(....` from being included as the class name. 
A test was also added to confirm this works properly and prevent issues in the future. 

closes #2064

### DIFF
--- a/lib/new_relic/agent/method_tracer_helpers.rb
+++ b/lib/new_relic/agent/method_tracer_helpers.rb
@@ -71,7 +71,7 @@ module NewRelic
       # '#<Class:MyModule::MyClass>', or '#<Class:MyModule::MyClass(id: integer, attrbiute: string)>'
       # Return the 'MyModule::MyClass' part of that string
       def klass_name(object)
-        name = Regexp.last_match(1) if object.to_s =~ /^#<Class:([\w|:{2}]*).*>$/
+        name = Regexp.last_match(1) if object.to_s =~ /^#<Class:([\w:]*).*>$/
         return name if name
 
         raise "Unable to glean a class name from string '#{object}'"
--- a/lib/new_relic/agent/method_tracer_helpers.rb
+++ b/lib/new_relic/agent/method_tracer_helpers.rb
@@ -71,7 +71,7 @@ module NewRelic
       # '#<Class:MyModule::MyClass>', or '#<Class:MyModule::MyClass(id: integer, attribute: string)>'
       # Return the 'MyModule::MyClass' part of that string
       def klass_name(object)
-        name = Regexp.last_match(1) if object.to_s =~ /^#<Class:([\w:]*).*>$/
+        name = Regexp.last_match(1) if object.to_s =~ /^#<Class:([\w:]+).*>$/
         return name if name
 
         raise "Unable to glean a class name from string '#{object}'"
--- a/lib/new_relic/agent/method_tracer_helpers.rb
+++ b/lib/new_relic/agent/method_tracer_helpers.rb
@@ -68,10 +68,10 @@ module NewRelic
       end
 
       # The string representation of a singleton class looks like
-      # '#<Class:MyModule::MyClass>'. Return the 'MyModule::MyClass' part of
-      # that string
+      # '#<Class:MyModule::MyClass>', or '#<Class:MyModule::MyClass(id: integer, attrbiute: string)>'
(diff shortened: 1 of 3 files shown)


lib/new_relic/agent/method_tracer_helpers.rb
```diff
@@ -68,10 +68,10 @@ def clm_enabled?
       end
 
       # The string representation of a singleton class looks like
-      # '#<Class:MyModule::MyClass>'. Return the 'MyModule::MyClass' part of
-      # that string
+      # '#<Class:MyModule::MyClass>', or '#<Class:MyModule::MyClass(id: integer, attrbiute: string)>'
```
fallwith, line 71:
```suggestion
      # '#<Class:MyModule::MyClass>', or '#<Class:MyModule::MyClass(id: integer, attribute: string)>'
```

lib/new_relic/agent/method_tracer_helpers.rb
```diff
@@ -68,10 +68,10 @@ def clm_enabled?
...
+      # Return the 'MyModule::MyClass' part of that string
       def klass_name(object)
-        name = Regexp.last_match(1) if object.to_s =~ /^#<Class:(.*)>$/
+        name = Regexp.last_match(1) if object.to_s =~ /^#<Class:([\w:]*).*>$/
```
fallwith, line 74: I think we should require (`+`) the matched value instead of having it be optional (`*`):

```suggestion
        name = Regexp.last_match(1) if object.to_s =~ /^#<Class:([\w:]+).*>$/
```

(in retrospect, I also think the original pattern should have used `.+` instead of `.*`)

fallwith, line 74: With `[\w:]`, you can end up with colons in the match text. Any need to strip those away?

fallwith (approved):

lib/new_relic/agent/method_tracer_helpers.rb
tannalynn: I added the colon because we needed it in order to handle a class name like `Something::Class`, otherwise it would only match `Something` instead of the full thing, which would not be the correct name to use. So we would want the colons in the name to make sure our `Object.const_get` works correctly.

kaylareopelle (approved):
🎉

hannahramadan (approved):

PR #2093, opened 2023-06-21
CI: remove test:multiverse:self

### LINKED ISSUE #1163: test:multiverse:self is broken
Our multiverse [documentation](https://github.com/newrelic/newrelic-ruby-agent/tree/dev/test/multiverse#testing-multiverse) suggests that one can have the multiverse test itself with: 

```shell
rake test:multiverse:self
```

That task simply ends up calling `ruby test/multiverse/test/multiverse_test.rb`. The `test/multiverse/test/multiverse_test.rb` is `test/unit` based, and `test/unit` is not currently one of the agent's development dependencies.

We should do one of the following:

- port `test/multiverse/test/multiverse_test.rb` to Minitest
- declare `test/unit` as a dev dependency
- remove the tests and all related documentation

fallwith:
Remove the `test:multiverse:self` Rake task and its related test-unit test code and runner script.

This task has been broken for quite some time, owing to the fact that it requires test-unit which is no longer available as part of the default installation for modern Rubies.

We could declare test-unit as a dev dependency to restore the tests, but of the 3 tests 2 of them will fail because of the way we have reworked Multiverse testing with things like SimpleCov and the reporting of the slowest tests. The Rake task's tests expect the Multiverse process itself to exit with a non-zero code in the event of a test failure and it no longer does so by design.

resolves #1163


test/multiverse/README.md
```diff
@@ -159,13 +159,12 @@ The default gemfile line is
 
 ### Test files
 
-All files in a test suite directory that end with .rb will be executed as test
-files.  These should use test unit.
+All files in a test suite directory (`test/multiverse/suites/*`) that end with
+`_test.rb` will be executed as test files. These should use MiniTest.
```
kaylareopelle, line 163: I think lowercase "t" is the current flavor:
```suggestion
`_test.rb` will be executed as test files. These should use Minitest.
```

kaylareopelle (approved):

kaylareopelle (approved):

PR #2096, opened 2023-06-22
CI: permit `serialize!` in Envfile

### LINKED ISSUE #1154: CI: Add before_all_suites/after_all_suites functionality to multiverse
Currently, when running the `active_record_pg` mutliverse suite, in the before_suite.rb, we are dropping and creating the database before each environment run.  This is there to prevent the database from growing too large since the tests don't all clean up after themselves. In the CI this is fine because the CI runs multiverse serialized rather than parallel. However, when running this suite locally this can cause this error to occur when attempting to drop the database while another process is running the suite as well. 
```
PG::ObjectInUse: ERROR:  database "multiverse_activerecord_6_1_6_2_7_4_ruby" is being accessed by other users
DETAIL:  There is 1 other session using the database.
Couldn't drop database 'multiverse_activerecord_6_1_6_2_7_4_ruby'
```
Also, it looks like this error doesn't seem to happen when running on ruby version 2.5, but does happen on 2.7.
It doesn't look like multiverse has the ability to do a before_all_suites/after_all_suites which is where this type of behavior should be instead to avoid parallelism issues.

fallwith:
Satisfy the Multiverse CI testing TODO to permit individual suites to insist on running serialized.

For any suite that needs to be ran serialized (and we only have 1 currently), simply call `serialize!` somewhere within `Envfile`.

Updated the relevant `Envfile` documentation.

resolves #1154

### DIFF
--- a/test/multiverse/README.md
+++ b/test/multiverse/README.md
@@ -133,13 +133,15 @@ contain at least two files.
 The Envfile is a meta gem file.  It allows you to specify one or more gemset
 that the tests in this directory should be run against.  For example:
 
-    gemfile <<~GEMFILE
-      gem "rails", "~>6.1.0"
-    GEMFILE
+```ruby
+gemfile <<~GEMFILE
+  gem "rails", "~>6.1.0"
+GEMFILE
 
-    gemfile <<~GEMFILE
-      gem "rails", "~>6.0.0"
-    GEMFILE
+gemfile <<~GEMFILE
+  gem "rails", "~>6.0.0"
+GEMFILE
+```
 
 This will run these tests against 2 environments, one running rails 6.1, the
 other running rails 6.0.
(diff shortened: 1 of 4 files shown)


test/multiverse/README.md
```diff
@@ -150,12 +152,44 @@ using two environment variables.
...
+end
+```
+
+- Option 2 for larger prep: In the suite directory, create a `before_suite.rb`
+file:
+
+```ruby
+# before_suite.rb
+complex_prep_operation_to_be_ran_once
+```
```
kaylareopelle, line 191: Thanks for documenting this!

hannahramadan (approved):
This is nice! Much cleaner

kaylareopelle (approved):
Provided the tests pass, this looks great!

PR #2118, opened 2023-07-07
Create test for agent_helper in gem files

### LINKED ISSUE #2117: Add test for gemspec files
#2113 revealed our recent changes to the files included in the Ruby agent unintentionally broke the `NewRelic::Agent.require_test_helper` public API. We have some tests that try to protect this from happening, but they didn't catch the problem in this case. 

The lines of code to test are around the `s.files` assignment:
https://github.com/newrelic/newrelic-ruby-agent/blob/9fdb4396600ccdf45df1ae662219593536561c06/newrelic_rpm.gemspec#L42-L46

At a minimum, we need to test whether `test/agent_helper.rb` is included in the build.

kaylareopelle:
This PR adds a test to ensure the `test/agent_helper.rb` file is included in the distributed gem.

Thank you, @ajesler for your creative solution! 🌈  

Resolves #2117

### DIFF
--- a/test/new_relic/gemspec_files_test.rb
+++ b/test/new_relic/gemspec_files_test.rb
@@ -1,0 +1,20 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'minitest/autorun'
+
+class GemspecFilesTest < Minitest::Test
+  def test_the_test_agent_helper_is_shipped_in_the_gem_files
+    skip if defined?(Rails::VERSION)
+
+    agent_helper_file = 'test/agent_helper.rb'
+
+    gem_spec_file_path = File.expand_path('../../../newrelic_rpm.gemspec', __FILE__)
+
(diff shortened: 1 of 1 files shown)


test/new_relic/gemspec_files_test.rb
```diff
@@ -0,0 +1,18 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'minitest/autorun'
```
kaylareopelle, line 5: @ajesler - I swapped out the `test_helper` require statement with `minitest/autorun` because we don't want/need the agent to be loaded for this particular test. 

This is what we do with our other utilities, like our [`CveNotifierTests`](https://github.com/newrelic/newrelic-ruby-agent/blob/2fa9a6385f6ee29cda7873890584b5aba6e585f3/test/new_relic/slack_notifications_tests/cve_notifier_tests.rb)

fallwith (approved):

fallwith (dismissed):

kaylareopelle:
Hmm... local tests and the CI aren't matching. I'm going to come back to this after the release.

PR #2130, opened 2023-07-11
CI: gemspec files test - bypass cache

fallwith:
Fixes to get the .gemspec test to work with our automated CI system:

- effectively fork `Gem::Specification.load` to bypass its internal caching mechanism
  - In so doing, the CI logs now report the actual underlying issue that caused a .gemspec eval to fail
- Assert that the parse succeeded before working with the result
- Update `newrelic_rpm.gemspec` to use an absolute path to the `.build_ignore` file
- `chdir` to the agent root for the evaluation of the .gemspec file

### DIFF
--- a/test/new_relic/gemspec_files_test.rb
+++ b/test/new_relic/gemspec_files_test.rb
@@ -6,13 +6,16 @@ require 'minitest/autorun'
 
 class GemspecFilesTest < Minitest::Test
   def test_the_test_agent_helper_is_shipped_in_the_gem_files
-    skip if defined?(Rails::VERSION) || File.basename(Dir.pwd).match?('rails')
+    skip if defined?(Rails::VERSION)
 
     gem_spec_file_path = File.expand_path('../../../newrelic_rpm.gemspec', __FILE__)
-    gem_spec = eval(Gem.open_file(gem_spec_file_path, 'r:UTF-8:-', &:read))
 
-    assert gem_spec, "Failed to parse '#{gem_spec_file_path}'"
-    assert_equal('newrelic_rpm', gem_spec.name)
-    assert_includes(gem_spec.files, 'test/agent_helper.rb')
+    Dir.chdir(File.dirname(gem_spec_file_path)) do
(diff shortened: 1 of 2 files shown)


fallwith:
resolves #2117

test/new_relic/gemspec_files_test.rb
```diff
@@ -6,15 +6,13 @@
 
 class GemspecFilesTest < Minitest::Test
   def test_the_test_agent_helper_is_shipped_in_the_gem_files
-    skip if defined?(Rails::VERSION)
-
-    agent_helper_file = 'test/agent_helper.rb'
+    skip if defined?(Rails::VERSION) || File.basename(Dir.pwd).match?('rails')
```
kaylareopelle, line 9: @fallwith  - If we skip running this test in the `norails` environment, when does this test run?

fallwith: @kaylareopelle Ugh. Apparently we only run `bundle exec rake test` locally and via the automated CI it's always in `env`. Will address.

kaylareopelle: My understanding is that `norails` is the default environment that gets run when we execute `bundle exec rake test` locally.

kaylareopelle: That might not be the case though, haven't looked in a while.

fallwith: Addressed with [7fcd753](https://github.com/newrelic/newrelic-ruby-agent/pull/2130/commits/7fcd753b89d7f39ca8904a1450c08a3d9a97deeb)

fallwith: I was able to repro locally with `bundle exec rake test:env[norails]` while having it work fine with `bundle exec rake test`.

I believe `bundle exec rake test` will preserve PWD as the agent root, but all of the `env` tests will chdir to `test/environments/<name>`, bundle from there, and then execute from there.

That key difference seemed to be the problem. Now the new test should work in and out of `env`.

We could even allow the test to run on the Rails envs, but there's not much value in having the test run more than once.

kaylareopelle (approved):

PR #2139, opened 2023-07-19
Remove old method

hannahramadan:
`new_relic_trace_controller_action` is defined and added to customer Sinatra applications, but is never called.


tannalynn (approved):
We literally can't find where this method has ever been used. Even when it was first introduced, it didn't do anything and wasn't used. ¯\\_(ツ)_/¯

PR #2140, opened 2023-07-21
Perf test updates

fallwith:
- introduce SImpleCov
- change the default to run the suites inline instead of in a subshell
- for every method in every test suite, use a fixed iteration count instead of the 5 secs time based value


tannalynn (approved):
Thanks for working on this!

Also, I am curious why you chose the specific numbers that you did for the iterations for the different tests?

fallwith:
> Also, I am curious why you chose the specific numbers that you did for the iterations for the different tests?

Most (all but 1?) of the methods defined in the suites were running as many iterations as they could in 5 seconds. I recorded the iteration counts for each and then did some rounding to be able to switch to a fixed iteration count but still try to stay roughly at the same amount of time spent.

tannalynn (approved):

PR #2142, opened 2023-07-21
run perf tests within Rails

fallwith:
Conduct the perf tests via a Rails runner so that the agent will operate as it does when running within a Rails application


tannalynn (approved):
🎉 

Did you notice any difference in the results in running with rails vs without?

fallwith:
> 🎉
> 
> Did you notice any difference in the results in running with rails vs without?

@tannalynn the code coverage increased for some relevant Rails specific content, but it wasn't a huge bump or not. Fortunately there isn't much load introduced by the Rails runner wrapper, so we can see how things go for awhile.

Issue #1774, opened 2023-01-27
Investigate multiverse failures with webrick 1.8.1


tannalynn:
Webrick released new versions on 1/27/22 (1.8.0 and 1.8.1) that break many of our multiverse suites. [Failing CI](https://github.com/newrelic/newrelic-ruby-agent/actions/runs/4023013925)
We need to investigate these failures and figure out what changes in webrick have caused these issues. 
We will pin our webrick versions to 1.7.0 until this is resolved.

PR #2143, opened 2023-07-25
CI: multiverse - unpin webrick

fallwith:
- Update the fake server based classes used for unit tests to run on the latest version of WEBrick (previously we were not compatible with v1.8.0+ and locked to v1.7.0)
- Update the fake server class to respect `ENV['DEBUG']`. Going forward if there are any errors happening that are fake server related, we need only to set the `DEBUG` environment variable to see what is going on.
- Update the fake collector class to stop attempting to call `#rewind` on objects that don't support it (for improved WEBrick compatibility)


fallwith:
manual nightly run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/5663343396

fallwith:
resolves #1774

kaylareopelle (approved):

hannahramadan (approved):

PR #2144, opened 2023-07-25
Add Roda instrumentation

### LINKED ISSUE #1783: Add instrumentation for Roda
[Roda](https://roda.jeremyevans.net) is a web framework that aims to be extremely safe, efficient, and easy. It [compares well ](https://roda.jeremyevans.net/compare_to_sinatra.html)against Sinatra.

We should get after some Roda adoption numbers and evaluate the priority for some instrumentation.

hannahramadan:
Adds instrumentation for [Roda](https://roda.jeremyevans.net/) versions 3.19.0+. Roda will be automatically detected and instrumented.

closes #1783


lib/new_relic/agent/instrumentation/roda/roda_transaction_namer.rb
```diff
@@ -0,0 +1,38 @@
...
+          def initial_transaction_name(request)
+            transaction_name(::NewRelic::Agent::UNKNOWN_METRIC, request)
+          end
+
+          ROOT = '/'.freeze
+
+          def transaction_name(path, request)
+            verb = http_verb(request)
+            path = request.path if request.path
+            name = path.gsub(%r{^[/^\\A]*(.*?)[/\$\?\\z]*$}, '\1') # remove any rouge slashes
```
fallwith, line 21: 1. I think if you're going to the trouble of freezing something as simple as the `ROOT` slash, you should freeze this regex pattern here and have it live outside the method as well to avoid redefining it every time the method gets called.
2. Perhaps consider removing slashes of other colors as well...

lib/new_relic/agent/instrumentation/roda/instrumentation.rb
```diff
@@ -0,0 +1,54 @@
...
+          @_request.params
+        rescue => e
+          NewRelic::Agent.logger.debug('Failed to get params from Rack request.', e)
+          nil
+        end
+      end
+
+      def _roda_handle_main_route_with_tracing(*args)
+        request_params = rack_request_params
+        filtered_params = ::NewRelic::Agent::ParameterFiltering::apply_filters(request.env, request_params || {})
```
fallwith, line 43: Do you plan on modifying the filtered params hash later? If not, you can replace `{}` with `NewRelic::EMPTY_HASH`.

lib/new_relic/agent/instrumentation/roda.rb
```diff
@@ -0,0 +1,40 @@
...
+require_relative 'roda/instrumentation'
+require_relative 'roda/chain'
+require_relative 'roda/prepend'
+
+DependencyDetection.defer do
+  named :roda
+
+  depends_on do
+    defined?(Roda) &&
+      Roda::RodaVersion >= '3.19.0' &&
```
fallwith, line 14: `Roda::RodaVersion` is a simple string, and you won't want to compare 2 strings for their numeric content.

I believe you are attempting to support Roda v3.19.0 and above, but currently this check succeeds for Roda v3.2.0 as well given that `2` >= `1`. Using a `Gem::Version` instance for comparison should yield your desired behavior.

```suggestion
      Gem::Version.new(Roda::RodaVersion) >= '3.19.0' &&
```

hannahramadan: Thank you—we've definitely been over this one! [All fixed](https://github.com/newrelic/newrelic-ruby-agent/pull/2144/files#diff-7f80c12f838c8bc306fce6b66036c7760f13728f57ce6622f1de94a966c40c6aR14).

lib/new_relic/agent/instrumentation/roda/instrumentation.rb
hannahramadan: Nice call. [Fixed](https://github.com/newrelic/newrelic-ruby-agent/pull/2144/files).

lib/new_relic/agent/instrumentation/roda/roda_transaction_namer.rb
hannahramadan: [Froze the regex](https://github.com/newrelic/newrelic-ruby-agent/pull/2144/files#diff-b4632c237fe6e6069aa8e3bf333b378bb34b6c86f63df1ad4c1b24d198fd7887R17). Ah shoot, I didn't mean to leave all the other colors out—how embarrassing. Now we get to capture all rogue slashes, not just rouge.

lib/new_relic/agent/instrumentation/roda/chain.rb
```diff
@@ -0,0 +1,47 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+# rubocop:disable Lint/DuplicateMethods
```
hannahramadan, line 5: Rubocop was upset about defining duplicate methods. I'm not sure why we are getting yelled at this time but seems to pass in other chain instrumentations.

lib/new_relic/agent/instrumentation/roda/chain.rb
```diff
@@ -0,0 +1,47 @@
...
+module NewRelic::Agent::Instrumentation
+  module Roda
+    module Chain
+      def self.instrument!
+        ::Roda.class_eval do
+          include ::NewRelic::Agent::Instrumentation::Roda::Tracer
+
+          alias_method(:_roda_handle_main_route_without_tracing, :_roda_handle_main_route)
+          alias_method(:_roda_handle_main_route, :_roda_handle_main_route_with_tracing)
```
tannalynn, line 14: iirc, I believe this line isn't necessary. 
So basically the line above it makes `:_roda_handle_main_route_without_tracing` = `:_roda_handle_main_route`, and then below this you're redefining the method with the original name, which overwrites what this does.
`_roda_handle_main_route_with_tracing` is a different method that we are calling inside of the definition of `_roda_handle_main_route` below, it wouldn't actually be a replacement for this method since it doesn't actually call the original code itself, which is why we use it below and pass it a block to call the original roda code. So this is pointing to that code, but when you define the `_roda_handle_main_route` method below this, you're actually changing where the method is pointing to the new code you wrote on line 16, so it ends up just acting like this never happened anways since it gets overwritten. So this doesn't harm anything being here or anything, it's just overwritten right away, so it's not needed.

I'm guessing it's probably like this here because it's like this in the instrumentation template files. It's probably like that because some of our instrumentation DOES have it like this, but those situations would have defined things in a different order (we can talk about how it works more later if you want! just lmk). But in this case with the order things are defined in, just removing line 14 would be good to do. 
Also, this would be the same thing for the other section below this in the file as well.

lib/new_relic/agent/instrumentation/roda/chain.rb
```diff
@@ -0,0 +1,47 @@
...
+                end
+              end
+            end
+          end
+        end
+      end
+    end
+  end
+end
+# rubocop:enable Lint/DuplicateMethods
```
tannalynn, line 47: Hmm, i wonder why this was necessary, I don't think we usually have to do that. I wonder if it's because of the extra alias_methods i mentioned above? I'd be interested in looking at why this is a thing with you.

lib/new_relic/agent/instrumentation/roda/roda_transaction_namer.rb
```diff
@@ -0,0 +1,39 @@
...
+            transaction_name(::NewRelic::Agent::UNKNOWN_METRIC, request)
+          end
+
+          ROOT = '/'.freeze
+          REGEX_MUTIPLE_SLASHES = %r{^[/^\\A]*(.*?)[/\$\?\\z]*$}.freeze
+
+          def transaction_name(path, request)
+            verb = http_verb(request)
+            path = request.path if request.path
+            name = path.gsub(REGEX_MUTIPLE_SLASHES, '\1') # remove any rogue slashes
```
tannalynn, line 22: what does the `'\1'` mean? Does it just replace things with literally a `\1` in the string? I'm just curious about what this is for since it's not ringing any bells for me.

lib/new_relic/agent/instrumentation/roda/chain.rb
hannahramadan: Thanks for this explanation! I'm seeing how this works better now and I've removed it. This is exactly what the Rubocop Duplicate Method cop was complaining about and removing the extra lines has resolved that warning.

lib/new_relic/agent/instrumentation/roda/roda_transaction_namer.rb
hannahramadan: The `\1` is referencing the first capture group or the stuff inside the first set of parentheses. If we had two capture groups, then we'd get to that second capture group with `\2`

So we're saying to look for any amount of beginning and ending `\`, capture only the text between, and replace the whole thing with the first capture group.

newrelic.yml
```diff
@@ -372,6 +372,10 @@ common: &default_settings
   # prepend, chain, disabled.
   # instrumentation.bunny: auto
 
+  # Controls auto-instrumentation of roda at start up.
+  # May be one of [auto|prepend|chain|disabled]
+  # instrumentation.roda: auto
```
kaylareopelle, line 377: Should we still manually add things to this file? I would have expected the change to happen via our CI.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1227,6 +1228,13 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'If `true`, disables [Sidekiq instrumentation](/docs/agents/ruby-agent/background-jobs/sidekiq-instrumentation).'
         },
+        :disable_roda_auto_middleware => {
+          :default => false,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => 'If `true`, disables agent middleware for Roda. This middleware is responsible for advanced feature support such as [page load timing](/docs/browser/new-relic-browser/getting-started/new-relic-browser) and [error collection](/docs/apm/applications-menu/events/view-apm-error-analytics).'
+        },
```
kaylareopelle, line 1237: After discussing `disable_middleware_tracing` this morning, I'm curious if Roda instrumentation will work if `disable_middleware_tracing` is set to `true`, but `disable_roda_auto_middleware` is set to `false`. Has this come up in your research so far?

newrelic.yml
hannahramadan: This was auto-added from the thor generator. I've deleted it and will let the CI handle it at a later date ◡̈

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: All seems to be working fine if `disable_middleware_tracing` is true, but `disable_roda_auto_middleware` is set to false. [This test](https://github.com/newrelic/newrelic-ruby-agent/pull/2144/files#diff-aeeae0fa20827c88507de124359d0418691e7f4bf94b2ef6ae335f3ff95d3889R114-R121) demonstrates. I also spun up a test app to be certain—happy to demo live ◡̈

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## dev
 
-Version <dev> of the agent introduces improved error tracking functionality by associating a transaction id with each error, and uses more reliable network timeout logic.
+Version <dev> of the agent introduces improved error tracking functionality by associating a transaction id with each error, uses more reliable network timeout logic, and adds Roda instrumentation.
```
fallwith, line 5:
```suggestion
Version <dev> of the agent introduces improved error tracking functionality by associating a transaction id with each error, uses more reliable network timeout logic, and adds [Roda](https://roda.jeremyevans.net/) instrumentation.
```

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ Version <dev> of the agent introduces improved error tracking functionality by a
 
   In line with current Ruby best practices, make use of Net::HTTP's own timeout logic and avoid the use of `Timeout.timeout()` when possible. The agent's data transmissions and cloud provider detection routines have been updated accordingly. [PR#2147](https://github.com/newrelic/newrelic-ruby-agent/pull/2147)
 
+- **Feature: Add Roda instrumentation**
+
+  Roda is a now an instrumented framework. The agent currently supports Roda versions 3.19.0+. [PR#2144](https://github.com/newrelic/newrelic-ruby-agent/pull/2144)
```
fallwith, line 17:
```suggestion
  [Roda](https://roda.jeremyevans.net/) is a now an instrumented framework. The agent currently supports Roda versions 3.19.0+. [PR#2144](https://github.com/newrelic/newrelic-ruby-agent/pull/2144)
```

lib/new_relic/agent/instrumentation/roda/instrumentation.rb
```diff
@@ -0,0 +1,55 @@
...
+        yield
+      end
+
+      # Roda makes use of Rack, so we can get params from the request object
+      def rack_request_params
+        begin
+          @_request.params
+        rescue => e
+          NewRelic::Agent.logger.debug('Failed to get params from Rack request.', e)
+          nil
```
fallwith, line 37:
```suggestion
          NewRelic::EmptyHash
```

from what I can tell, the result here winds up getting passed in as the `params` (second) argument to `NewRelic::Agent::ParameterFiltering.filter_rack_file_data`, which in turns invokes `#inject` on it. `#inject` won't work on `nil`...

fallwith (changes requested):

test/multiverse/suites/roda/roda_instrumentation_test.rb
```diff
@@ -0,0 +1,156 @@
...
+
+  def test_transaction_name_error
+    NewRelic::Agent.stub(:logger, NewRelic::Agent::MemoryLogger.new) do
+      begin
+        NewRelic::Agent::Instrumentation::Roda::TransactionNamer.transaction_name({})
+      rescue
+        # NOOP - Allow error to be raised
+      ensure
+        assert_logged(/Error encountered trying to identify Roda transaction name/)
+      end
```
fallwith, line 131: You can:

- Do away with the `rescue`, as the instrumentation swallows the underlying exception and doesn't re-raise it
- Enhance your existing logger output check to confirm that the expected exception was the one that was logged
- Add an additional assertion to confirm that the return value is what's expected

We can also update `test_rack_request_params_error` in a similar fashion, but I'm waiting for the suggested change to the code being tested to be accepted/rejected first.

```suggestion
      # pass in {} to produce an error, because {} doesn't support #path and
      # confirm that the desired error handling took place
      result = NewRelic::Agent::Instrumentation::Roda::TransactionNamer.transaction_name({})

      assert_equal NewRelic::Agent::UNKNOWN_METRIC, result
      assert_logged(/NoMethodError.*Error encountered trying to identify Roda transaction name/)
```

fallwith: Oh and finally:

- Add a comment to let future reviewers know that the use of `{}` as input was intended to raise a specific type of error.

lib/new_relic/agent/instrumentation/roda/instrumentation.rb
fallwith: Ah, I missed the `||` you refactored out with [160a64b](https://github.com/newrelic/newrelic-ruby-agent/pull/2144/commits/160a64b6fbbc3f6dfb13af9c2770c2f76287a29b). That explains it. Sorry about that. I think I remember seeing it in pairing but forgot when it came to the PR. I think the refactor looks better, so we should keep it if we can get the tests in line.

lib/new_relic/agent/instrumentation/roda/instrumentation.rb
```diff
@@ -0,0 +1,54 @@
...
+          NewRelic::EMPTY_HASH
+        end
+      end
+
+      def _roda_handle_main_route_with_tracing(*args)
+        request_params = rack_request_params
+        filtered_params = ::NewRelic::Agent::ParameterFiltering::apply_filters(request.env, request_params)
+        name = TransactionNamer.transaction_name(request)
+
+        perform_action_with_newrelic_trace(:category => :roda,
```
kaylareopelle, line 46: nit: When breaking up the args for a method across multiple lines, I prefer to put the first arg on the line below the method definition.

lib/new_relic/agent/instrumentation/roda/roda_transaction_namer.rb
```diff
@@ -0,0 +1,35 @@
...
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    module Instrumentation
+      module Roda
+        module TransactionNamer
+          extend self
+
+          ROOT = '/'.freeze
```
kaylareopelle, line 12: We have similar constants elsewhere in the codebase set to `'/'`. We could take on a refactoring project to consolidate and move them into the New Relic constants file.

lib/new_relic/agent/instrumentation/roda/instrumentation.rb
fallwith: I'd write it like this (align second argument with first), for a third option. I recognize all 3 as being idiomatic enough to represent "how a Rubyist would write it", so all are fine as long as our RuboCop settings permit them.

```ruby
perform_action_with_newrelic_trace(category: :roda,
                                   name: name,
                                   params: filtered_params) do
```

test/multiverse/suites/roda/roda_instrumentation_test.rb
```diff
@@ -0,0 +1,155 @@
...
+      begin
+        RodaTestApp::RodaRequest.any_instance
+          .stubs(:params).raises(StandardError.new)
+        get('/home?')
+      rescue
+        # NOOP - Allow error to be raised
+      ensure
+        assert_logged(/Failed to get params from Rack request./)
+      end
+    end
```
fallwith, line 145: Same deal here on not needing to rescue or noop.

```suggestion
    NewRelic::Agent.stub(:logger, NewRelic::Agent::MemoryLogger.new) do
      # Have the #params call made on the request raise an exception to test
      # the error handling
      RodaTestApp::RodaRequest.any_instance.stubs(:params).raises(StandardError.new)
      get('/home?')

      assert_logged(/Failed to get params from Rack request./)
    end
```

hannahramadan: THank you—great suggestion and makes sense. Committed!

lib/new_relic/agent/instrumentation/roda/roda_transaction_namer.rb
hannahramadan: I love that idea. I'll make a ticket for it!

lib/new_relic/agent/instrumentation/roda/instrumentation.rb
hannahramadan: Nice suggestions! I've updated this to the Rubocop preference, which is each arg on its own line + the closing paren on its own line

test/multiverse/suites/roda/roda_instrumentation_test.rb
```diff
@@ -0,0 +1,156 @@
...
+
+  setup_and_teardown_agent
+
+  def app
+    RodaTestApp
+  end
+
+  def app2
+    RodaNoMiddleware
+  end
```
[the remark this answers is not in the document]
hannahramadan: You're right! This was totally leftover from testing. Its been removed.

CHANGELOG.md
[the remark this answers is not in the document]
hannahramadan: Good idea - moved it to the beginning ◡̈

lib/new_relic/agent/instrumentation/roda/instrumentation.rb
```diff
@@ -0,0 +1,56 @@
...
+      def _roda_handle_main_route_with_tracing(*args)
+        request_params = rack_request_params
+        filtered_params = ::NewRelic::Agent::ParameterFiltering::apply_filters(request.env, request_params)
+        name = TransactionNamer.transaction_name(request)
+
+        perform_action_with_newrelic_trace(
+          :category => :roda,
+          :name => name,
+          :params => filtered_params
+        ) do
```
kaylareopelle: A slightly different suggestion, building off @fallwith's recommendation:

```suggestion
        perform_action_with_newrelic_trace(
          category: :roda,
          name: TransactionNamer.transaction_name(request),
          params: ::NewRelic::Agent::ParameterFiltering::apply_filters(request.env, rack_request_params)
        ) do
```
This eliminates all unnecessary object allocations, though readability may suffer.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1564,6 +1572,14 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of resque at start up. May be one of: `auto`, `prepend`, `chain`, `disabled`.'
         },
+        :'instrumentation.roda' => {
+          :default => 'auto',
+          :public => true,
+          :type => String,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => 'Controls auto-instrumentation of Roda at start up. May be one of: `auto`, `prepend`, `chain`, `disabled`.'
```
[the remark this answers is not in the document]
hannahramadan: lol ```No one, however, gives start up as acceptable.```
I'll make a ticket for this. A good first issue.

lib/new_relic/agent/instrumentation/roda/instrumentation.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def _roda_handle_main_route_with_tracing(*args)
+        filtered_params = ::NewRelic::Agent::ParameterFiltering::apply_filters(request.env, rack_request_params)
+        name = TransactionNamer.transaction_name(request)
+
+        perform_action_with_newrelic_trace(
+          category: :roda,
+          name: name,
+          params: filtered_params
+        ) do
```
kaylareopelle, line 49: Another suggestion here, building off @fallwith's recommendation, would eliminate all unnecessary variable assignments, though readability may suffer:
```suggestion
        perform_action_with_newrelic_trace(
          category: :roda,
          name: TransactionNamer.transaction_name(request),
          params: ::NewRelic::Agent::ParameterFiltering::apply_filters(request.env, rack_request_params)
        ) do
```

test/multiverse/suites/roda/roda_instrumentation_test.rb
```diff
@@ -0,0 +1,146 @@
...
+      assert_equal NewRelic::Agent::UNKNOWN_METRIC, result
+      assert_logged(/NoMethodError.*Error encountered trying to identify Roda transaction name/)
+    end
+  end
+
+  def test_rack_request_params_error
+    NewRelic::Agent.stub(:logger, NewRelic::Agent::MemoryLogger.new) do
+      # Have the #params call made on the request raise an exception to test
+      # the error handling
+      RodaTestApp::RodaRequest.any_instance.stubs(:params).raises(StandardError.new)
```
fallwith, line 134: Following up on offline discussions on this one...

- Minitest doesn't support (and one might even infer it discourages) the stubbing of any instance
- The `any_instance` here is provided by the Mocha testing gem
- Personally I think introducing Mocha to a Minitest based project detracts from the Minitest experience as much as it adds to it. Perhaps it's like adding a sidecar to a motorcycle. It's clear that the sidecar offers a distinct advantage over a motorcycle without a sidecar. But there are also fundamental changes to the motorcycling experience that could be considered as drawbacks.
- Testing any instance is well known anti-pattern given the risk of having stubbing behavior bleed over to other instances in play in a testing context where the stubbing behavior would be problematic.
    - The risk is mitigated here in two ways:
        - We are assured of only having one Rack request in play during the test.
        - We do not (currently) make use of parallelized, concurrent tests.
- Typically the best practice is to have a handle on the specific instance under test. The need to reach for the stubbing of any instance could itself indicate complexity of code. This complexity might come from not following TDD or could indicate a focus / prioritization on performance over testability. For whatever reason, tests that require stubbing any instance might indicate difficulties in maintainability.
- We use this `any_instance` method quite a lot in this project. It is completely acceptable to use it again here from a PR approval standpoint. The maintainers may at one point decide to dissuade future usage or refactor out existing usage of `any_instance`, but for now there's a green light.
    - We bring in a RuboCop gem specific to Minitest, but unfortunately given that `any_instance` comes from Mocha, its usage is not flagged by any cops.
- Looking at the context for this exact line of code, the benefit of using `any_instance` is pretty clear. We invoke the `Rack::Test` `get` method, which via `Forwardable` goes through the `Rack::Test` flow of calling the user defined `app` method to get an app class. An instance of that class is instantiated and the request params object is prepped. For us to be able to get at the exact single instance of the class (as opposed to any instance) that will field the web request, we would need to change the overall flow of this test and have it be different from all the others. Perhaps we could stub `new` on the Rack app and use that stubbing for a foothold while retaining the Rack test integration style of using `get`. Alternatively, we could have this test not leverage `Rack::Test` and we could instead directly invoke the `rack_request_params` method being tested. (We'd use a unit test instead of an integration test). The benefits of having this test look and act like the others is a pretty clear advantage of `any_instance`.

test/multiverse/suites/roda/Envfile
```diff
@@ -0,0 +1,20 @@
...
+
+RODA_VERSIONS = [
+  [nil, 2.4],
+  ['3.19.0', 2.4]
+]
+
+def gem_list(roda_version = nil)
+  <<~RB
+    gem 'roda'#{roda_version}
+    gem 'rack', '~> 2.2'
```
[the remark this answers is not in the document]
hannahramadan: It was unnecessary to pin Rack. Good call out ◡̈

lib/new_relic/agent/instrumentation/roda/instrumentation.rb
hannahramadan: Accepted suggestion. Thank you!

test/multiverse/suites/roda/roda_instrumentation_test.rb
hannahramadan: Thanks for the detailed thoughts—loving the motorcycle analogy. I agree it's weird to mix Mocha and Minitest. I've updated this and to go with the unit-style test— just calling `rack_request_params` directly, which is going to fail since no Rack request will be in existence. 

Updating our codebase to get rid of `ay_instance` would be a fun project. I think it would help with learning.

lib/new_relic/agent/instrumentation/roda.rb
```diff
@@ -0,0 +1,40 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require_relative 'roda/instrumentation'
+require_relative 'roda/chain'
```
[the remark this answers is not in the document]
hannahramadan: I like this call out. Updated

lib/new_relic/agent/instrumentation/roda.rb
```diff
@@ -0,0 +1,40 @@
...
+  depends_on do
+    defined?(Roda) &&
+      Gem::Version.new(Roda::RodaVersion) >= Gem::Version.new('3.19.0') &&
+      Roda::RodaPlugins::Base::ClassMethods.private_method_defined?(:build_rack_app) &&
+      Roda::RodaPlugins::Base::InstanceMethods.method_defined?(:_roda_handle_main_route)
+  end
+
+  executes do
+    # These requires are inside an executes block because they require rack, and
+    # we can't be sure that rack is available when this file is first required.
```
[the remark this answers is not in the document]
hannahramadan: Hmm I see what you're saying. [Moved these ](https://github.com/newrelic/newrelic-ruby-agent/pull/2144/files#diff-8067a27ede8ceb13b66d3b4c1389e9cb9994ec127bc1820ce451975d6c156aecR5-R6)to where roda instrumentation logic lives.

lib/new_relic/agent/instrumentation/roda.rb
```diff
@@ -0,0 +1,38 @@
...
+    NewRelic::Agent.logger.info('Installing Roda instrumentation')
+
+    if use_prepend?
+      require_relative 'roda/prepend'
+      prepend_instrument Roda, NewRelic::Agent::Instrumentation::Roda::Prepend
+    else
+      require_relative 'roda/chain'
+      chain_instrument NewRelic::Agent::Instrumentation::Roda::Chain
+    end
+  end
```
kaylareopelle, line 37: I understand that we often have multiple executes blocks, but now that we're calling the `require_relative` in both blocks, condensing the work into one seems more efficient. Are you open to combining them?

hannahramadan: Oh heck yes good point!!

lib/new_relic/agent/instrumentation/roda.rb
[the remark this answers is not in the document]
hannahramadan: Backup - our tests didn't like them existing in roda instrumentation logic. I've [moved](https://github.com/newrelic/newrelic-ruby-agent/pull/2144/files#diff-7f80c12f838c8bc306fce6b66036c7760f13728f57ce6622f1de94a966c40c6aR18-R19) them back so that I can just call these requires once vs putting them in the chain and prepend files.

lib/new_relic/agent/instrumentation/roda/roda_transaction_namer.rb
```diff
@@ -0,0 +1,35 @@
...
+          REGEX_MUTIPLE_SLASHES = %r{^[/^\\A]*(.*?)[/\$\?\\z]*$}.freeze
+
+          def transaction_name(request)
+            verb = http_verb(request)
+            path = request.path || ::NewRelic::Agent::UNKNOWN_METRIC
+            name = path.gsub(REGEX_MUTIPLE_SLASHES, '\1') # remove any rogue slashes
+            name = ROOT if name.empty?
+            name = "#{verb} #{name}" unless verb.nil?
```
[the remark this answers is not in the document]
hannahramadan: Updated this method to do this logic inside of the method.

lib/new_relic/agent/instrumentation/roda/roda_transaction_namer.rb
```diff
@@ -0,0 +1,31 @@
...
+
+          ROOT = '/'.freeze
+          REGEX_MULTIPLE_SLASHES = %r{^[/^\A]*(.*?)[/$?\z]*$}.freeze
+
+          def transaction_name(request)
+            verb = request.request_method if request.respond_to?(:request_method)
+            path = request.path || ::NewRelic::Agent::UNKNOWN_METRIC
+            name = path.gsub(REGEX_MULTIPLE_SLASHES, '\1') # remove any rogue slashes
+            name = ROOT if name.empty?
+            name = "#{verb} #{name}" unless verb.nil?
```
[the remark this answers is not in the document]
hannahramadan: Good call I like this. Accepted

lib/new_relic/control/frameworks/roda.rb
```diff
@@ -0,0 +1,20 @@
...
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'new_relic/control/frameworks/ruby'
+module NewRelic
+  class Control
+    module Frameworks
+      # Contains basic control logic for Roda
+      class Roda < NewRelic::Control::Frameworks::Ruby
+        protected
```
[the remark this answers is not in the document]
fallwith: Hmm... yeah on the "something we don't want people to use" tip, I'm thinking it should be `private` instead.

I don't see this `NewRelic::Control::Frameworks::Roda` class getting subclassed anywhere, and I don't see anything else that would make the `protected` bit make sense. But it does indeed keep the Roda instrumentation inline with the syntax used for every other framework. So let's keep it and address the issue of "using `protected` for frameworks" as its own separate thing if we think it necessary in future.

fallwith (approved):

PR #2147, opened 2023-08-01
don't use Timeout.timeout()

### LINKED ISSUE #975: Replace calls to Timeout.timeout()
`Timeout.timeout()` is no longer a recommended practice by the Ruby community as it is prone to unexpected behavior. 

* https://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/
* https://jvns.ca/blog/2015/11/27/why-rubys-timeout-is-dangerous-and-thread-dot-raise-is-terrifying/
* https://adamhooper.medium.com/in-ruby-dont-use-timeout-77d9d4e5a001

### Feature Description
Replace `Timeout.timeout()` calls with more specific timeouts, as outlined in [Mike Perham's article](https://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/)

### Describe Alternatives
There is more than one way to replace `Timeout.timeout`, and we're open to additional solutions.

### Priority
[Really Want]

###T-Shirt Size

fallwith:
(outside of `test/`) don't use `Timeout.timeout()` when the underlying library has timeout functionality itself.

resolves #975

### DIFF
--- a/lib/new_relic/agent/new_relic_service.rb
+++ b/lib/new_relic/agent/new_relic_service.rb
@@ -19,7 +19,10 @@ module NewRelic
       # These include Errno connection errors, and all indicate that the
       # underlying TCP connection may be in a bad state.
       CONNECTION_ERRORS = [Net::OpenTimeout, Net::ReadTimeout, EOFError, SystemCallError, SocketError]
-      # Net::WriteTimeout is Ruby 2.6+
+      # TODO: MAJOR VERSION - Net::WriteTimeout wasn't defined until Ruby 2.6.
+      #       Once support for Ruby 2.5 is dropped, we should simply include
+      #       Net::WriteTimeout in the connection errors array directly instead
+      #       of with a conditional
       CONNECTION_ERRORS << Net::WriteTimeout if defined?(Net::WriteTimeout)
       CONNECTION_ERRORS.freeze
 
@@ -327,8 +330,8 @@ module NewRelic
       def setup_connection_timeouts(conn)
         conn.open_timeout = @request_timeout
         conn.read_timeout = @request_timeout
-        conn.ssl_timeout = @request_timeout
-        # #write_timeout= requires Ruby 2.6+
+        # TODO: MAJOR VERSION - #write_timeout= requires Ruby 2.6+, so remove
+        #       the conditional check once support for Ruby 2.5 is dropped
         conn.write_timeout = @request_timeout if conn.respond_to?(:write_timeout=)
 
         if conn.respond_to?(:keep_alive_timeout) && NewRelic::Agent.config[:aggressive_keepalive]
@@ -473,7 +476,13 @@ module NewRelic
         when Net::HTTPGone
           handle_gone_response(response, endpoint)
         else
-          # Net::WriteTimeout requires Ruby 2.6+
+          # TODO: MAJOR VERSION - Net::WriteTimeout wasn't defined until
+          #       Ruby 2.6, so it can't be included in the case statement
+          #       as a constant and instead needs to be found here. Once
+          #       support for Ruby 2.5 is dropped, we should have
+          #       Net::WriteTimeout sit in the 'when' clause above alongside
+          #       Net::OpenTimeout and Net::ReadTimeout and this entire if/else
+          #       conditional can be removed.
           if response.respond_to?(:name) && response.name == 'Net::WriteTimeout'
             handle_server_connection_exception(response, endpoint)
           else


lib/new_relic/agent/new_relic_service.rb
```diff
@@ -19,7 +18,10 @@ class NewRelicService
 
       # These include Errno connection errors, and all indicate that the
       # underlying TCP connection may be in a bad state.
-      CONNECTION_ERRORS = [Timeout::Error, EOFError, SystemCallError, SocketError].freeze
+      CONNECTION_ERRORS = [Net::OpenTimeout, Net::ReadTimeout, EOFError, SystemCallError, SocketError]
+      # Net::WriteTimeout is Ruby 2.6+
```
tannalynn, line 22: Should we add a todo here to keep a note to update this when 2.5 support is dropped someday? I'm thinking that when we search for things related to 2.5 we can remove we potentially won't notice this one, but if it's a todo we'd probably find it. Same with the other one for WriteTimeout below?

lib/new_relic/agent/new_relic_service.rb
```diff
@@ -471,9 +473,14 @@ def handle_error_response(response, endpoint)
         when Net::HTTPGone
           handle_gone_response(response, endpoint)
         else
-          record_endpoint_attempts_supportability_metrics(endpoint)
-          record_error_response_supportability_metrics(response.code)
-          raise UnrecoverableServerException, "#{response.code}: #{response.message}"
+          # Net::WriteTimeout requires Ruby 2.6+
```
tannalynn, line 476: I understand why it's like this with the if/else inside of the case else, but maybe adding the todo I mentioned above would be good to do here as well to make sure it's not missed when this can be updated when 2.5 is dropped (someday)

test/multiverse/suites/agent_only/utilization_data_collection_test.rb
```diff
@@ -96,12 +96,12 @@ def redirect_link_local_address(port)
       @dummy_port = p
 
       class << self
-        def start_with_patch(address, port = nil, p_addr = nil, p_port = nil, p_user = nil, p_pass = nil, &block)
+        def start_with_patch(address, port, *_args, &block)
```
tannalynn, line 99: were these arguments never used by whatever is calling this method?

lib/new_relic/agent/new_relic_service.rb
```diff
@@ -362,8 +366,8 @@ def create_and_start_http_connection
         conn = create_http_connection
         start_connection(conn)
         conn
-      rescue Timeout::Error
-        ::NewRelic::Agent.logger.info('Timeout while attempting to connect. You may need to install system-level CA Certificates, as the ruby agent no longer includes these.')
+      rescue Net::OpenTimeout
+        ::NewRelic::Agent.logger.info('Timed out while attempting to connect. For SSL issues, you may need to install system-level CA Certificates to be used by Net::HTTP.')
```
tannalynn, line 370: I like the clarification about ssl, thanks for updating that log message

lib/new_relic/agent/new_relic_service.rb
```diff
@@ -362,8 +366,8 @@ def create_and_start_http_connection
         conn = create_http_connection
         start_connection(conn)
         conn
-      rescue Timeout::Error
-        ::NewRelic::Agent.logger.info('Timeout while attempting to connect. You may need to install system-level CA Certificates, as the ruby agent no longer includes these.')
+      rescue Net::OpenTimeout
```
tannalynn, line 369: Is this the only type of timeout that needs to be rescued here? I'm curious about the other types of timeouts

lib/new_relic/agent/new_relic_service.rb
```diff
@@ -436,21 +440,19 @@ def relay_request(request, opts)
...
 
       def handle_error_response(response, endpoint)
         case response
         when Net::HTTPRequestTimeOut,
              Net::HTTPTooManyRequests,
              Net::HTTPInternalServerError,
-             Net::HTTPServiceUnavailable
+             Net::HTTPServiceUnavailable,
+             Net::OpenTimeout,
+             Net::ReadTimeout
```
tannalynn, line 455: In the method `setup_connection_timeouts` that you updated, I saw that one of the new things there was an ssl_timeout being set as well. Is that something we would need to handle here as well?

lib/new_relic/agent/new_relic_service.rb
fallwith: Yeah, good call here. Will do.

hannahramadan (approved):
This makes sense to me!

lib/new_relic/agent/new_relic_service.rb
fallwith: Yep, but great question here. Net::HTTP will only throw 1 of 3 exceptions: an open timeout, a read timeout, or a write timeout (and that last one wasn't added until Ruby 2.6). The part that confused me was the ability to specify an "ssl timeout" without having a corresponding ssl timeout error to worry about.

We rescue all 3 error types elsewhere, but in this exact context we are only opening the connection and not reading or writing across it, so just the open timeout is used.

lib/new_relic/agent/new_relic_service.rb
fallwith: Great question!!! Net::HTTP itself will only raise 1 of 3 error types for open, read, and write. But when setting the timeout values, it supports an "ssl" type, so shouldn't there be a corresponding error for that ssl timeout being hit? I thought the same thing and forgot to proactively bring it up here in the PR.

My current understanding: Net::HTTP simply passes the ssl timeout to the underlying OpenSSL::SSL::SSLContext class. If that class has a timeout error, then something ssl specific has gone wrong and I'm currently thinking we want to let the `else` flow of the `case` statement handle it. Let me know if you have other thoughts. One idea I had was to get out of the business of setting the ssl timeout anywhere in the first place and just using OpenSSL's presumably reasonable defaults.

test/multiverse/suites/agent_only/utilization_data_collection_test.rb
fallwith: That's my understanding, yes. This method is defined in this test class and only used by the 2 test cases within - neither of which ever cared about the extra arguments. Without this change, the new Net::HTTP options hash was getting incorrectly mapped to the `p_addr` argument.

lib/new_relic/agent/new_relic_service.rb
fallwith: TODOs added with [607f6d5](https://github.com/newrelic/newrelic-ruby-agent/pull/2147/commits/607f6d5c9466bffb725d9151205fd411637f9ad9)

lib/new_relic/agent/new_relic_service.rb
fallwith: I've decided it's better to leave SSL timeouts up to OpenSSL (which we engage separately elsewhere in this class with code untouched by the PR) and I removed the Net::HTTP specific `ssl_timeout` value.

lib/new_relic/agent/new_relic_service.rb
fallwith: Removed the Net::HTTP specific SSL timeout with [607f6d5](https://github.com/newrelic/newrelic-ruby-agent/pull/2147/commits/607f6d5c9466bffb725d9151205fd411637f9ad9)

lib/new_relic/agent/new_relic_service.rb
fallwith: TODOs added with [607f6d5](https://github.com/newrelic/newrelic-ruby-agent/pull/2147/commits/607f6d5c9466bffb725d9151205fd411637f9ad9)

test/multiverse/suites/agent_only/utilization_data_collection_test.rb
tannalynn: that makes sense, thank you for explaining!

lib/new_relic/agent/new_relic_service.rb
tannalynn: That makes sense, thanks for explaining why this one was just the open timeout!

tannalynn (approved):
Awesome, thanks for explaining everything and adding the todos!

PR #2152, opened 2023-08-08
NewRelic::Control - drop competing `camelize`

### LINKED ISSUE #2151: Reconcile the 2 competing "camelize" methods
Both `NewRelic::Control` and `NewRelic::LanguageSupport` define their own `camelize` methods.

Decide on one to keep and get rid of the other.

fallwith:
The LanguageSupport's `camelize` method appears to be faster, so let's just standardize on using that whenever we need to convert from snake to camel.

```shell
newrelic-ruby-agent 17:03:40 $ ruby test.rb
Rehearsal ---------------------------------------------------
LanguageSupport  20.492525   0.126348  20.618873 ( 20.645389)
Control          29.277619   0.177723  29.455342 ( 29.524243)
----------------------------------------- total: 50.074215sec

                      user     system      total        real
LanguageSupport  21.267743   0.145529  21.413272 ( 21.493573)
Control          30.793050   0.221045  31.014095 ( 31.085878)
```

resolves #2151

### DIFF
--- a/lib/new_relic/control/class_methods.rb
+++ b/lib/new_relic/control/class_methods.rb
@@ -49,18 +49,12 @@ module NewRelic
           # maybe it is already loaded by some external system
           # i.e. rpm_contrib or user extensions?
         end
-        NewRelic::Control::Frameworks.const_get(camelize(framework.to_s))
+        NewRelic::Control::Frameworks.const_get(NewRelic::LanguageSupport.camelize(framework.to_s))
       end
 
       # The root directory for the plugin or gem
       def newrelic_root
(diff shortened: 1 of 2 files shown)


tannalynn (approved):
Unless I'm mistaken, it does seem like the only situation this is used for is loading our own classes for frameworks, which is a pretty limited group of inputs. Because of that, the differences we've noticed in the behavior between this `camelize` method vs `LanguageSupport.camelize` won't cause any actual differences in behavior for this situation.
Thanks for noticing this and getting it updated!

kaylareopelle (approved):

hannahramadan (approved):
Nice!

PR #2158, opened 2023-08-11
Roda instrumentation

hannahramadan:
Fix CI failure with Ruby 3.1.4

### DIFF
--- a/test/multiverse/suites/roda/roda_instrumentation_test.rb
+++ b/test/multiverse/suites/roda/roda_instrumentation_test.rb
@@ -111,7 +111,7 @@ class RodaInstrumentationTest < Minitest::Test
       # pass in {} to produce an error, because {} doesn't support #path and
       # confirm that the desired error handling took place
       result = NewRelic::Agent::Instrumentation::Roda::TransactionNamer.transaction_name({})
-      # binding.irb
+
       assert_equal NewRelic::Agent::UNKNOWN_METRIC, result
       assert_logged(/NoMethodError.*Error encountered trying to identify Roda transaction name/m)
     end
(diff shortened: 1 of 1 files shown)


test/multiverse/suites/roda/roda_instrumentation_test.rb
```diff
@@ -113,7 +113,7 @@ def test_transaction_name_error
       result = NewRelic::Agent::Instrumentation::Roda::TransactionNamer.transaction_name({})
 
       assert_equal NewRelic::Agent::UNKNOWN_METRIC, result
-      assert_logged(/NoMethodError.*Error encountered trying to identify Roda transaction name/)
+      assert_logged(/NoMethodError.*Error encountered trying to identify Roda transaction name/m)
```
fallwith, line 116: Bizarrely, we end up with a newline in the middle of the log message only with Ruby 3.1.4 (not with Ruby 2.7, 3.0, 3.2, etc.), so let's have `.*` match the newline.

test/multiverse/suites/roda/roda_instrumentation_test.rb
```diff
@@ -128,7 +128,10 @@ def test_rack_request_params_error
   end
 
   def assert_logged(expected)
-    found = NewRelic::Agent.logger.messages.flatten.any? { |m| m.match?(expected) }
+    # Example logger array:
+    # [[:debug, ["NoMethodError : undefined method `path' for \
+    # {}:Hash - Error encountered trying to identify Roda transaction name"], nil]]
```
fallwith, line 133: Long term we could possibly benefit from some logger helpers in the `test/` dir, but for now it really helps to have the example of the array so it is clear what `[1][0]` does.

fallwith (approved):

PR #2160, opened 2023-08-11
LogEventAttributes: instance var check

fallwith:
Don't attempt to read `@custom_attribute_limit_reached` until it has been defined

Addresses the following Ruby warning:

```shell
/Users/jbond007/git/public/newrelic-ruby-agent/lib/new_relic/agent/log_event_attributes.rb:17: warning: instance variable @custom_attribute_limit_reached not initialized
```

### DIFF
--- a/lib/new_relic/agent/log_event_attributes.rb
+++ b/lib/new_relic/agent/log_event_attributes.rb
@@ -10,7 +10,7 @@ module NewRelic
       ATTRIBUTE_VALUE_CHARACTER_LIMIT = 4094
 
       def add_custom_attributes(attributes)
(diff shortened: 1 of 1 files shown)


kaylareopelle (approved):

kaylareopelle:
Should we update the tests? If so, I don't think that needs to be taken care of as part of this PR.

fallwith:
> Should we update the tests? If so, I don't think that needs to be taken care of as part of this PR.

Conceptually I like that idea given that I've introduced a new branch with the new conditional. Code coverage wise, I'm currently thinking that we're exercising both branches of the conditional well even though we don't have actual assertions on both sides.

PR #2172, opened 2023-08-18
Fix docker compose when running it on m1 mac

tannalynn:
the mysql container won't run on the new m1 macs, and fails with the error
```
no matching manifest for linux/arm64/v8 in the manifest list entries
```

I found this fix for it, and now it's running on my new laptop.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,6 +22,7 @@ services:
     mem_limit: 1g
   mysql:
     image: mysql:5.7
+    platform: linux/x86_64
     restart: always
     environment:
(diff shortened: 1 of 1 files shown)


fallwith (approved):

kaylareopelle (approved):
Nice find! Thanks for blazing this trail!

kaylareopelle:
This fix works on a non-M1 laptop (Intel) as well. Tested by running `docker-compose up` in a clean environment.

tannalynn:
awesome, thank you for checking!

PR #2179, opened 2023-08-29
Describe Rails subscriber implementation

hannahramadan:
Adds a code comment for future readers. The agent uses a non-traditional way to subscribe to Rails events. Instead of passing a block to `ActiveSupport::Notifications.subscribe`, the agent defines a `#start` and `#finish` method, which Rails responds to. This is due to a threading issue discovered on initial instrumentation.
- [Original Issue](https://github.com/rails/rails/issues/12069 )
- [Rails Code](https://github.com/rails/rails/blob/ed5af004598fa7645fec2210453cca00f4b59168/activesupport/lib/active_support/notifications/fanout.rb#L320)


kaylareopelle (dismissed):
I think this addition is great as-is. The only change I'd recommend is to add a link to the Rails issue in the comment. It could look like "See: https://github.com/rails/rails/issues/12069"

kaylareopelle (approved):

PR #2180, opened 2023-08-30
Stripe instrumentation

### LINKED ISSUE #2057: Ruby - Add Instrumentation Support for Stripe
### Description
Add new instrumentation support for Stripe - https://rubygems.org/gems/stripe

hannahramadan:
Adds instrumentation for Stripe 5.38.0+. Makes use of Stripe's pub/sub instrumentation pattern.

Stripe users can attach custom attributes to Stripe events via the `user_data` hash created on a Stripe request event initialization. No attributes are collected by default. Two new config options allow customers to fine-tune reported attributes: `stripe.user_data.include` and `stripe.user_data.exclude`

closes #2057


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1595,6 +1595,15 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of Sinatra at start up. May be one of: `auto`, `prepend`, `chain`, `disabled`.'
         },
+        :'instrumentation.stripe' => {
+          :default => 'enabled',
+          :documentation_default => 'enabled',
+          :public => true,
+          :type => String,
+          :dynamic_name => true,
```
fallwith, line 1603: Your new `lib/new_relic/agent/instrumentation/stripe.rb` file uses the `'instrumentation.stripe'` string statically, so there's no need for the `dynamic_name` key/value pair here. Rather than switch the value to `false`, you can just remove both the key and value.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1595,6 +1595,15 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :allowed_from_server => false,
+          :description => 'Controls auto-instrumentation of Stripe at start up. May be one of: `enabled`, `disabled`.'
```
fallwith, line 1605: The lack of chain/prepend options that signals that the library is observability friendly and not in need of monkeypatching is great to see!

lib/new_relic/agent/configuration/default_source.rb
fallwith: Alternatively, you might be able to keep `dynamic_name: true` here, and remove the potentially redundant `depends_on` check in the new instrumentation class.

hannahramadan: I see how there was no need for the dynamic naming, so just removed it ◡̈ https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/135282a44fd26491287f8389645522497145eb77

lib/new_relic/agent/instrumentation/stripe_subscriber.rb
```diff
@@ -0,0 +1,46 @@
...
+          NewRelic::Agent::Tracer.state
+        end
+
+        def start_segment(event)
+          return unless state.is_execution_traced?
+
+          segment = Tracer.start_segment(name: metric_name(event))
+          event.user_data[:newrelic_segment] = segment
+        rescue => e
+          NewRelic::Agent.logger.debug("Error starting New Relic Stripe segment: #{e}")
```
fallwith, line 22: I think this event warrants something more severe than `debug`.

hannahramadan: I guess if this fails it means something is pretty wrong here. Agree, I've made this an `error` level. https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/9c8998032d130be04257693e0cf0a33fce2acd7d

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1595,6 +1595,14 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of Sinatra at start up. May be one of: `auto`, `prepend`, `chain`, `disabled`.'
         },
+        :'instrumentation.stripe' => {
+          :default => 'enabled',
+          :documentation_default => 'enabled',
```
kaylareopelle, line 1600: I don't think you have to add `documentation_default` when the `default` value is not dynamic (ex. uses a method or a proc to resolve the value)

lib/new_relic/agent/instrumentation/stripe.rb
```diff
@@ -0,0 +1,27 @@
...
+      Gem::Version.new(Stripe::VERSION) >= Gem::Version.new('5.15.0') # Stripe subscribers added 5.15.0
+  end
+
+  executes do
+    NewRelic::Agent.logger.info('Installing Stripe instrumentation')
+  end
+
+  executes do
+    Stripe::Instrumentation.subscribe(:request_begin) { |event| NewRelic::Agent::Instrumentation::StripeSubscriber.new.start_segment(event) }
+    Stripe::Instrumentation.subscribe(:request_end) { |event| NewRelic::Agent::Instrumentation::StripeSubscriber.new.finish_segment(event) }
```
kaylareopelle, line 25: Could you create one `NewRelic::Agent::Instrumentation::StripeSubscriber` instance and use it for both subscriptions?

lib/new_relic/agent/instrumentation/stripe_subscriber.rb
```diff
@@ -0,0 +1,46 @@
...
+module NewRelic
+  module Agent
+    module Instrumentation
+      class StripeSubscriber
+        DEFAULT_DESTINATIONS = AttributeFilter::DST_SPAN_EVENTS
+        EVENT_ATTRIBUTES = %i[http_status method num_retries path request_id].freeze
+
+        def state
+          NewRelic::Agent::Tracer.state
+        end
```
kaylareopelle, line 14: Two things:
1. I see you use `Tracer` directly on line 19. Can you do that here too? 
2. It looks like whenever this method is invoked, `is_execution_traced?` is called on it. Do you see a use for `state` outside this context? What do you think about changing this method to call `is_execution_traced?` and renaming it?

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: Removed! https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/d577fddc85e8d4f31c64ec12e128fa6e4ab2f5f9

lib/new_relic/agent/instrumentation/stripe.rb
hannahramadan: Oh yes love this idea... done and tested. https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/d577fddc85e8d4f31c64ec12e128fa6e4ab2f5f9

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1602,6 +1602,39 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of Sinatra at start up. May be one of: `auto`, `prepend`, `chain`, `disabled`.'
         },
+        :'instrumentation.stripe' => {
+          :default => 'enabled',
+          :public => true,
+          :type => String,
+          :allowed_from_server => false,
+          :description => 'Controls auto-instrumentation of Stripe at start up. May be one of: `enabled`, `disabled`.'
```
fallwith, line 1610: This isn't new for Stripe, but I think it ought to be `startup`.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1602,6 +1602,39 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          default: NewRelic::EMPTY_ARRAY,
+          public: true,
+          type: Array,
+          dynamic_name: true,
+          allowed_from_server: false,
+          :transform => DefaultSource.method(:convert_to_list),
+          :description => <<~DESCRIPTION
+            An array of strings to specify which keys inside a Stripe event's `user_data` hash should be reported
+            to New Relic. Each string in this array will be turned into a regular expression via `Regexp.new` to
+            permit advanced matching. Setting the value to `"."` will report all `user_data`.
```
fallwith, line 1622: > Setting the value to `"."` will report all `user_data`.

Technically - given that the value is an array of strings and not a single string - it would be `["."]`. I think you should either make that change to the backticked code and leave the rest of the sentence unchanged or rewrite it to say something like "Add `"."` to the array to match all hash keys".

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1602,6 +1602,39 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          default: NewRelic::EMPTY_ARRAY,
+          public: true,
+          type: Array,
+          dynamic_name: true,
+          allowed_from_server: false,
+          :transform => DefaultSource.method(:convert_to_list),
+          :description => <<~DESCRIPTION
+            An array of strings to specify which keys inside a Stripe event's `user_data` hash should not be reported
+            to New Relic. Each string in this array will be turned into a regular expression via `Regexp.new` to
+            permit advanced matching. By default, no `user_data` is reported.
```
fallwith, line 1635: Maybe

```diff
- By default, no `user_data` is reported.
+ By default, no `user_data` is reported, so this option should only be used if the `stripe.user_data.include` option is being used.
```

lib/new_relic/agent/instrumentation/stripe.rb
```diff
@@ -0,0 +1,28 @@
...
+  depends_on do
+    defined?(Stripe) &&
+      Gem::Version.new(Stripe::VERSION) >= Gem::Version.new('5.15.0') # Stripe subscribers added 5.15.0
+  end
+
+  executes do
+    NewRelic::Agent.logger.info('Installing Stripe instrumentation')
+  end
+
+  executes do
```
fallwith, line 23: I think this block of code needs a `rescue` with an error level log message, so as to not crash the agent if either `subscribe` call fails. Optionally, a new method can be defined in `stripe_subscriber.rb` that handles the rescuing and that method could be called with a one-liner here.

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: I think there is a ticket out there to change all these... hoping someone picks it up for hacktoberfest! I'll make this change for Stripe though.

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: Good point! Updated to the hash option https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/8cd8f196a25cb78da3da8b2130fb11b3d9202cb2

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: Nice suggestion! Accepted https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/8cd8f196a25cb78da3da8b2130fb11b3d9202cb2

lib/new_relic/agent/instrumentation/stripe.rb
hannahramadan: Yup yup yup this is important--good catch https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/8cd8f196a25cb78da3da8b2130fb11b3d9202cb2

hannahramadan: Ooo on second look, the `executes` method already has a rescue. So I think we are covered here. Reverted. https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/89b9db0edf085ab245b05b21e78b9610639f7e99

lib/new_relic/agent/instrumentation/stripe_subscriber.rb
```diff
@@ -0,0 +1,71 @@
...
+
+        def finish_segment(event)
+          begin
+            return unless is_execution_traced?
+
+            segment = event.user_data[:newrelic_segment]
+            add_stripe_attributes(segment, event)
+            add_custom_attributes(segment, event)
+          ensure
+            segment.finish
```
fallwith, line 63: It is possible to reach the `ensure` block with a `nil` segment, and calling `#finish` on `nil` will raise another error which won't be handled. We can use the safe navigation operator to proactively anticipate and defend against this possibility.

```suggestion
            segment&.finish
```

test/multiverse/suites/stripe/config/newrelic.yml
```diff
@@ -0,0 +1,22 @@
...
+development:
+  error_collector:
+    enabled: true
+  apdex_t: 0.5
+  agent_enabled: true
+  monitor_mode: true
+  license_key: bootstrap_newrelic_admin_license_key_000
+  ca_bundle_path: ../../../config/test.cert.crt
+  instrumentation:
+    sinatra: <%= $instrumentation_method %>
```
fallwith, line 11:
```suggestion
```

You don't need these.

test/multiverse/suites/stripe/stripe_instrumentation_test.rb
```diff
@@ -0,0 +1,160 @@
...
+    # later, helps us get around needing to provide a valid API key
+    @connection = Stripe::ConnectionManager.new
+    @response = Net::HTTPResponse.new('1.1', '200', 'OK')
+    # Bypass #stream_check ("attempt to read body out of block")
+    @response.instance_variable_set(:@read, true)
+    @response.body = {
+      object: 'list',
+      data: [{'id': '12134'}],
+      has_more: false,
+      url: '/v1/charges'
```
fallwith, line 24: It doesn't _really_ matter given that the tests don't operate on the url contained in the response, but from a "the specs should tell the story" perspective, perhaps you should store `'v1/customers'` as a constant and use it both here and in the `#detect` calls that the tests perform.

lib/new_relic/agent/instrumentation/stripe_subscriber.rb
```diff
@@ -0,0 +1,71 @@
...
+          return unless is_execution_traced?
+
+          segment = NewRelic::Agent::Tracer.start_segment(name: metric_name(event))
+          event.user_data[:newrelic_segment] = segment
+        rescue => e
+          NewRelic::Agent.logger.error("Error starting New Relic Stripe segment: #{e}")
+        end
+
+        def metric_name(event)
+          "Stripe#{event.path} #{event.method}"
```
kaylareopelle, line 28: Can metric names have spaces in them? (I have no idea, but it struck me as odd)

test/multiverse/suites/stripe/stripe_instrumentation_test.rb
```diff
@@ -0,0 +1,160 @@
...
+
+  def test_subscribed_request_end
+    subcribers = Stripe::Instrumentation.send(:subscribers)
+    newrelic_begin_subscriber = subcribers[:request_end].detect { |_k, v| v.to_s.include?('instrumentation/stripe') }
+
+    assert(newrelic_begin_subscriber)
+  end
+
+  def test_newrelic_segment
+    Stripe::StripeClient.stub(:default_connection_manager, @connection) do
```
fallwith, line 47: Seeing as you use this wrapper multiple times, it'd be a good candidate for a test helper.

```ruby
def with_stubbed_connection_manager(&block)
  Stripe::StripeClient.stub(:default_connection_manager, @connection) do
    @connection.stub(:execute_request, @response) do
      block.call
    end
  end
end
```

then for each test you could use it:

```ruby
with_stubbed_connection_manager do
  in_transaction do |txn|
    Stripe::Customer.list({limit: 3})
    stripe_segment = txn.segments.detect { |s| s.name == 'Stripe/v1/customers get' }

    assert(stripe_segment)
  end
end
```

lib/new_relic/agent/instrumentation/stripe_subscriber.rb
```diff
@@ -0,0 +1,71 @@
...
+
+module NewRelic
+  module Agent
+    module Instrumentation
+      class StripeSubscriber
+        DEFAULT_DESTINATIONS = AttributeFilter::DST_SPAN_EVENTS
+        EVENT_ATTRIBUTES = %i[http_status method num_retries path request_id].freeze
+        ATTRIBUTE_NAMESPACE = 'stripe.user_data'
+        ATTRIBUTE_FILTER_TYPES = %i[include exclude].freeze
+
```
kaylareopelle, line 13: I think some of the methods in this class could be made private. `start_segment` and `finish_segment` should probably stay public, but I think the others could be made private.

test/multiverse/suites/stripe/stripe_instrumentation_test.rb
```diff
@@ -0,0 +1,160 @@
...
+      @connection.stub(:execute_request, @response) do
+        in_transaction do |txn|
+          Stripe::Customer.list({limit: 3})
+          stripe_segment = txn.segments.detect { |s| s.name == 'Stripe/v1/customers get' }
```
fallwith, line 51: I could see you doing something like:

```ruby
stripe_segment = stripe_segment_from_txn(txn)
```

for this repeated pattern.

test/multiverse/suites/stripe/stripe_instrumentation_test.rb
```diff
@@ -0,0 +1,160 @@
...
+
+  def test_start_when_not_traced
+    Stripe::StripeClient.stub(:default_connection_manager, @connection) do
+      @connection.stub(:execute_request, @response) do
+        NewRelic::Agent::Tracer.state.stub(:is_execution_traced?, false) do
+          in_transaction do |txn|
+            Stripe::Customer.list({limit: 3})
+            stripe_segment = txn.segments.detect { |s| s.name == 'Stripe/v1/customers get' }
+
+            assert_empty txn.segments
```
fallwith, line 139: There's always a single "dummy" segment when you use `in_transaction`. You should be able to do:

```ruby
refute stripe_segment
```

but you don't want the transaction's segment array itself to be empty, as it'll always have at least 1 entry.

fallwith (changes requested):

test/multiverse/suites/stripe/config/newrelic.yml
kaylareopelle, line 11: I think you want this to be stripe?

test/multiverse/suites/stripe/stripe_instrumentation_test.rb
```diff
@@ -0,0 +1,160 @@
...
+    @response.body = {
+      object: 'list',
+      data: [{'id': '12134'}],
+      has_more: false,
+      url: '/v1/charges'
+    }.to_json
+  end
+
+  def test_version_supported
+    assert(Stripe::VERSION >= '5.38.0')
```
kaylareopelle, line 29: Does this need `Gem::Version.new` to compare the values?

test/multiverse/suites/stripe/stripe_instrumentation_test.rb
```diff
@@ -0,0 +1,160 @@
...
+    newrelic_begin_subscriber = subcribers[:request_end].detect { |_k, v| v.to_s.include?('instrumentation/stripe') }
+
+    assert(newrelic_begin_subscriber)
+  end
+
+  def test_newrelic_segment
+    Stripe::StripeClient.stub(:default_connection_manager, @connection) do
+      @connection.stub(:execute_request, @response) do
+        in_transaction do |txn|
+          Stripe::Customer.list({limit: 3})
```
kaylareopelle, line 50: Could you add the `start a Stripe event` comment here too? I loved seeing it on the later tests.

kaylareopelle: Alternatively, you could name a method `start_stripe_event` and have it call `Stripe::Customer.list({limit: 3})`, if the event is started the same way in all the tests.

kaylareopelle (changes requested):
This is so cool, Hannah! Awesome work. I found a few small things for you to consider before approval.

test/multiverse/suites/stripe/config/newrelic.yml
kaylareopelle: Totally, I forgot we're pub/sub here!

lib/new_relic/agent/instrumentation/stripe_subscriber.rb
hannahramadan: Nice suggestion.. done! https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/c59f04c4a06e2c6e404ec8f9daf89be8c6dbbfa7

lib/new_relic/agent/instrumentation/stripe_subscriber.rb
```diff
@@ -0,0 +1,71 @@
...
+        def add_stripe_attributes(segment, event)
+          EVENT_ATTRIBUTES.each do |attribute|
+            segment.add_agent_attribute("stripe_#{attribute}", event.send(attribute), DEFAULT_DESTINATIONS)
+          end
+        end
+
+        def add_custom_attributes(segment, event)
+          return if NewRelic::Agent.config[:'stripe.user_data.include'].empty?
+
+          event.user_data.delete(:newrelic_segment)
```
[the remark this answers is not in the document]
hannahramadan: Made some changes. Let me know if this works https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/c59f04c4a06e2c6e404ec8f9daf89be8c6dbbfa7

lib/new_relic/agent/instrumentation/stripe_subscriber.rb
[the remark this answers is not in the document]
hannahramadan: I've updated this. https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/c59f04c4a06e2c6e404ec8f9daf89be8c6dbbfa7

lib/new_relic/agent/instrumentation/stripe_subscriber.rb
hannahramadan: Great call https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/c59f04c4a06e2c6e404ec8f9daf89be8c6dbbfa7

test/multiverse/suites/stripe/stripe_instrumentation_test.rb
hannahramadan: We were running fine without it, but I've definitely run into issues with this in the past. To be safe and match instrumentation, I've updated https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/c997c511306d9a07fb8f68c839f653cdc3dd3cc3

lib/new_relic/agent/instrumentation/stripe_subscriber.rb
hannahramadan: Both you and James called this out. I've removed the space https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/c59f04c4a06e2c6e404ec8f9daf89be8c6dbbfa7

test/multiverse/suites/stripe/stripe_instrumentation_test.rb
hannahramadan: That makes sense. I've updated what we're looking for https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/63014dd5d7236c3fb38390840d7cfc0ff0ed70a4

test/multiverse/suites/stripe/stripe_instrumentation_test.rb
hannahramadan: Love this. Kayla had a similar suggestion to put repeated patterns into methods. I've updated https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/63014dd5d7236c3fb38390840d7cfc0ff0ed70a4

test/multiverse/suites/stripe/stripe_instrumentation_test.rb
hannahramadan: Pattern into a method.. love and done! https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/63014dd5d7236c3fb38390840d7cfc0ff0ed70a4

test/multiverse/suites/stripe/stripe_instrumentation_test.rb
hannahramadan: updated https://github.com/newrelic/newrelic-ruby-agent/pull/2180/commits/63014dd5d7236c3fb38390840d7cfc0ff0ed70a4

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1602,6 +1602,40 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          type: Array,
+          dynamic_name: true,
+          allowed_from_server: false,
+          :transform => DefaultSource.method(:convert_to_list),
+          :description => <<~DESCRIPTION
+            An array of strings to specify which keys inside a Stripe event's `user_data` hash should be reported
+            to New Relic. Each string in this array will be turned into a regular expression via `Regexp.new` to
+            permit advanced matching. Setting the value to `["."]` will report all `user_data`.
+          DESCRIPTION
+        },
```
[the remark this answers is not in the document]
fallwith: Discussed offline, addressed with https://github.com/newrelic/newrelic-ruby-agent/pull/2192

fallwith (dismissed):

kaylareopelle (approved):

kaylareopelle (approved):

PR #2181, opened 2023-08-31
Update Elasticsearch port_path_or_id value

kaylareopelle:
Previously, the `port_path_or_id` value for Elasticsearch segments was set to the path given to the `request_with_tracing` method. This caused a metrics grouping issue for a customer because a new instance metric (ex. "Datastore/instance/Elasticsearch/<host>/<port>") was created for every Elasticsearch document ID.

The source of the host value is a hash that also contains a port key. This updates the value of `port_path_or_id` to use the port from the `nr_hosts` hash.


kaylareopelle:
Closing in favor of #2182

PR #2183, opened 2023-09-01
CI: Sidekiq multiverse - remove old dependencies

fallwith:
Our CI system has recently errored out when running the Sidekiq multiverse suite for 2 reasons:

- `json` gem dependency issues: The `json` version specified by `Gemfile.n` conflicts with the one already activated. Given the recent reworking of the Sidekiq suite, we should no longer need anything but Sidekiq and the agent.
- CLI instance issues. For some reason, we occasionally end up with a nil `config` object for `Sidekiq::CLI.instance`. Set a config if this happens. Also fix memoization in the test helper.

### DIFF
--- a/test/multiverse/suites/sidekiq/Envfile
+++ b/test/multiverse/suites/sidekiq/Envfile
@@ -14,9 +14,6 @@ SIDEKIQ_VERSIONS = [
 
 def gem_list(sidekiq_version = nil)
   <<-RB
-    gem 'rack', "~> 2.2.4"
-    gem 'json'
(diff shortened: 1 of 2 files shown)


kaylareopelle (approved):

fallwith:
scheduled run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/6054242769

kaylareopelle (dismissed):
🕵️

hannahramadan (approved):

kaylareopelle (approved):
🤞

Issue #2055, opened 2023-06-08
Ruby - Support New Synthetics Header


kford-newrelic:
### Description
Support Q3 Major Launch for Automated Testing.

Ruby team will deliver the CDD, Agent Spec update, and Ruby agent implementation in Q2 to pave the way for agent teams to deliver in Q3 to support the launch.

kford-newrelic:
Additional tech convos/details : https://issues.newrelic.com/browse/NR-121367

PR #2203, opened 2023-09-11
Add synthetics info header

tannalynn:
This adds the logic for handling the new synthetics header, `X-NewRelic-Synthetics-Info`, adding the attributes to the transaction as intrinsic attributes, and adds the header to outgoing external request headers. These attributes are included in transaction events and error events. They will also automatically be included in transaction/error traces, as those include all intrinsic attributes already.

[spec](https://source.datanerd.us/agents/agent-specs/pull/632)
closes https://github.com/newrelic/newrelic-ruby-agent/issues/2055


lib/new_relic/language_support.rb
```diff
@@ -83,6 +83,10 @@ def camelize_with_first_letter_downcased(string)
       camelized[0].downcase.concat(camelized[1..-1])
     end
 
+    def snakeize(string)
+      string.gsub(/(.)([A-Z])/, '\1_\2').downcase
+    end
```
kaylareopelle, line 88: Nice! We have a language support test file. Could you add tests for this method to it?

tannalynn: added a test for this here https://github.com/newrelic/newrelic-ruby-agent/pull/2203/commits/436f5d26cabaa70775d009a80d8be99d85ad16ea

lib/new_relic/agent/monitors/synthetics_monitor.rb
```diff
@@ -6,6 +6,7 @@ module NewRelic
   module Agent
     class SyntheticsMonitor < InboundRequestMonitor
       SYNTHETICS_HEADER_KEY = 'HTTP_X_NEWRELIC_SYNTHETICS'.freeze
+      SYNTHETICS_INFO_HEADER_KEY = 'HTTP_X_NEWRELIC_SYNTHETICS_INFO'
```
fallwith, line 9: Just for uniformity, I think it'd be good to remove the `#freeze` call on the other constant's value.

lib/new_relic/agent/monitors/synthetics_monitor.rb
```diff
@@ -16,6 +17,7 @@ def on_finished_configuring(events)
 
       def on_before_call(request) # THREAD_LOCAL_ACCESS
         encoded_header = request[SYNTHETICS_HEADER_KEY]
+        info_header = request[SYNTHETICS_INFO_HEADER_KEY]
         return unless encoded_header
```
fallwith, line 21: 2 things here...

- Naming-wise, now looking at "encoded_header" and "info_header", I'm tempted to consider "info_header" as non-encoded. Given that the new header is an "info" type header, is there anything that be used to help with naming of the original header? Perhaps "base_header" or something?
- The early `return` will continue to happen unless the original header is present. Is it possible to have the new header present without the original header?

lib/new_relic/agent/monitors/synthetics_monitor.rb
```diff
@@ -27,7 +29,16 @@ def on_before_call(request) # THREAD_LOCAL_ACCESS
...
         txn.raw_synthetics_header = encoded_header
+        txn.raw_synthetics_info_header = info_header
         txn.synthetics_payload = incoming_payload
+        txn.synthetics_info_payload = load_json(info_header, SYNTHETICS_INFO_HEADER_KEY)
+      end
+
+      def load_json(header, key)
+        ::JSON.load(header)
+      rescue => err
+        NewRelic::Agent.logger.debug("Failure loading json header '#{key}' in #{self.class}, #{err.class}, #{err.message}")
```
fallwith, line 40: How likely is it to have invalid JSON reach this rescue? Would a `warn` or `error` level log statement be better?

lib/new_relic/agent/monitors/synthetics_monitor.rb
tannalynn: > Naming-wise, now looking at "encoded_header" and "info_header", I'm tempted to consider "info_header" as non-encoded. Given that the new header is an "info" type header, is there anything that be used to help with naming of the original header? Perhaps "base_header" or something?
You're correct that info_header is not encoded, it is just straight json. I didn't originally change the name because of that, but if you think base header is better I can totally do that.


> The early return will continue to happen unless the original header is present. Is it possible to have the new header present without the original header?
No, this isn't possible, the new header is just additional information and can't be used on its own.

lib/new_relic/agent/monitors/synthetics_monitor.rb
tannalynn: Probably not super likely, since we'll be taking it right out of the header. Also, I don't think this merits a warn or error level log though. Warn and error level both would be logged by default, unlike debug. And I don't think this information is something that could be helpful the the default logs since this only would be relevant to synthetics events, not the agent functioning in general.

lib/new_relic/agent/monitors/synthetics_monitor.rb
fallwith: Excellent, thank you. No actions needed here then.

lib/new_relic/agent/monitors/synthetics_monitor.rb
fallwith: Gotcha, sounds good.

lib/new_relic/agent/monitors/synthetics_monitor.rb
tannalynn: updated in https://github.com/newrelic/newrelic-ruby-agent/pull/2203/commits/705bcd9fe0c9a991c3bb3638491020a704d2b3ae

kaylareopelle (approved):

test/new_relic/agent/monitors/synthetics_monitor_test.rb
```diff
@@ -78,6 +78,72 @@ def test_records_synthetics_state_from_header
...
+    end
+
+    def test_load_json
+      info_payload = <<~PAYLOAD
+        {
+          "version": "1",
+          "type": "automatedTest",
+          "initiator": "cli",
+          "attributes": {
+            "attribute1": "on0e"
```
kaylareopelle, line 121: Maybe a typo?

```suggestion
            "attribute1": "one"
```

tannalynn: fixed this in https://github.com/newrelic/newrelic-ruby-agent/pull/2203/commits/99d2734be84dbdb47b73d626980f4831432b48e3

kaylareopelle (approved):

CHANGELOG.md
```diff
@@ -2,12 +2,17 @@
...
 ## dev
 
-Version <dev> brings support for gleaning a Docker container id from cgroups v2 based containers.
+Version <dev> brings support for gleaning a Docker container id from cgroups v2 based containers and records additional synthetics attributes.
 
 - **Feature: Enhance Docker container id reporting**
 
   Previously, the agent was only capable of determining a host Docker container's ID if the container was based on cgroups v1. Now, containers based on cgroups v2 will also have their container IDs reported to New Relic. [PR#2229](https://github.com/newrelic/newrelic-ruby-agent/issues/2229).
 
+- **Feature: Updates events with additional synthetics attributes when available**
```
kaylareopelle, line 11:
```suggestion
- **Feature: Update events with additional synthetics attributes when available**
```

fallwith (approved):

kaylareopelle (approved):
🎉

PR #2229, opened 2023-09-21
Support cgroups v1 and v2 for Docker containers

### LINKED ISSUE #2026: Docker: ensure the container id can be gleaned
With [chatter](https://stackoverflow.com/questions/68816329/how-to-get-docker-container-id-from-within-the-container-with-cgroup-v2) picking up about alternatives to using cgroups to obtain a Docker container's id, we should track with internal and external best practices to make sure we're able to glean a container's id in as many contexts as possible.

fallwith:
Previously the agent was only capable of gleaning a Docker container id by leveraging a cgroups v1 approach. Now, it will first attempt to leverage a cgroups v2 approach and fall back to the v1 approach if that doesn't work. All other Docker related behavior such as that seen with containers that don't have cgroups functionality enabled is not impacted.

resolves #2026

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+Version <dev> brings support for gleaning a Docker container id from cgroups v2 based containers.
+
+- **Feature: Enhance Docker container id reporting**
+
+  Previously the agent was only capable of determining a host Docker container's id if the container was based on cgroups v1. Now containers based on cgroups v2 will also have their container ids reported to New Relic. [PR#2026](https://github.com/newrelic/newrelic-ruby-agent/issues/2026).
+
 ## v9.5.0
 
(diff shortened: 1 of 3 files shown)


kaylareopelle (dismissed):

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+Version <dev> brings support for gleaning a Docker container id from cgroups v2 based containers.
+
+- **Feature: Enhance Docker container id reporting**
+
+  Previously the agent was only capable of determining a host Docker container's id if the container was based on cgroups v1. Now containers based on cgroups v2 will also have their container ids reported to New Relic. [PR#2229](https://github.com/newrelic/newrelic-ruby-agent/issues/2229).
```
```suggestion
  Previously, the agent was only capable of determining a host Docker container's ID if the container was based on cgroups v1. Now, containers based on cgroups v2 will also have their container IDs reported to New Relic. [PR#2229](https://github.com/newrelic/newrelic-ruby-agent/issues/2229).
```

kaylareopelle (approved):

PR #2232, opened 2023-09-21
CI: Ruby v3.3.0+ requires Puma v6.4.0+

fallwith:
Now that Puma v6.4.0 is out, we can resume testing with "nil" (latest) on Ruby v3.3.0, but we need to prevent Ruby v3.3.0 and future rubies from ever testing Pumas older than v6.4.0.

### DIFF
--- a/test/multiverse/suites/rack/Envfile
+++ b/test/multiverse/suites/rack/Envfile
@@ -6,19 +6,14 @@ instrumentation_methods :chain, :prepend
 
 # The Rack suite also tests Puma::Rack::Builder
 # Which is why we also control Puma tested versions here
-PUMA_VERSIONS = [
-  'nil',
-  '5.6.4',
-  '4.3.12',
-  '3.12.6'
-]
-
 # Ruby v3.3.0+ requires Puma v6.4.0+
 # https://github.com/puma/puma/commit/188f5da1920ff99a8689b3e9b46f2f26b7c62d66
-if RUBY_VERSION >= Gem::Version.new('3.3.0')
-  puma_min = Gem::Version.new('6.4.0')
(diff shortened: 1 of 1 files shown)


tannalynn (approved):

hannahramadan (approved):

kaylareopelle (commented):
@fallwith  - There's an option to handle minimum Ruby versions in an Envfile's versions array. You can make a given version row a nested array to include minimum and maximum Ruby versions for testing.  

For example: https://github.com/newrelic/newrelic-ruby-agent/blob/62ee0257b81f229dc2c03c7c1bf564edec6b8ee8/test/multiverse/suites/active_record_pg/Envfile#L14-L22 

Can you try refactoring to remove the `'nil'` and replace with `nil` (no string), and add the nested array for that line? 

```diff 
- 'nil',
+ [nil, 3.3],
```

kaylareopelle (changes requested):
Sorry, meant to hit "request changes" rather than "comment" earlier.

fallwith:
> @fallwith - There's an option to handle minimum Ruby versions in an Envfile's versions array. You can make a given version row a nested array to include minimum and maximum Ruby versions for testing.
>
> Can you try refactoring to remove the `'nil'` and replace with `nil` (no string), and add the nested array for that line?
> 
> ```diff
> - 'nil',
> + [nil, 3.3],
> ```

Addressed with [bbf283d](https://github.com/newrelic/newrelic-ruby-agent/pull/2232/commits/bbf283d93b664a34cee7887b8e51bc0f71c4e237)

kaylareopelle (approved):
Thanks for your updates!

PR #2236, opened 2023-09-25
CI: fix flapping Sidekiq test

### LINKED ISSUE #2227: CI: fix flapping Sidekiq test
We have a unit test to confirm that we properly record Sidekiq "internal" errors (errors that take place within Sidekiq's own handling of things and not job specific). That unit test performs the following:

```ruby
Sidekiq::CLI.instance.handle_exception(exception)
```
Occasionally - and apparently only via our GitHub Actions runners and not locally - the instance yielded is `nil`, which doesn't define `#handle_exception`.

We should see if there's a way to guard against this error with a retry or whatnot.

```shell
  1) Error:
SidekiqInstrumentationTest#test_captures_sidekiq_internal_errors:
NoMethodError: undefined method `handle_exception' for nil:NilClass

      config.handle_exception(ex, ctx)
            ^^^^^^^^^^^^^^^^^
    /the/ruby/path/Ruby/3.1.4/x64/lib/ruby/gems/3.1.0/gems/sidekiq-7.1.4/lib/sidekiq/component.rb:48:in `handle_exception'
    /the/runner/path/newrelic-ruby-agent/newrelic-ruby-agent/test/multiverse/suites/sidekiq/sidekiq_instrumentation_test.rb:50:in `block in test_captures_sidekiq_internal_errors'
```

fallwith:
The singleton instance of `Sidekiq::CLI` needs to have a Sidekiq configuration set on it for certain operations to work.

Previously we had been lucking out by having the `cli` test helper method be called prior to the `test_captures_sidekiq_internal_errors` test (which does not make use of the helper) being ran.

To prevent flapping, have the test make use of the `cli` helper (which preps a CLI instance with a config) to make it always pass regardless of the run order.

resolves #2227

### DIFF
--- a/test/multiverse/suites/sidekiq/sidekiq_instrumentation_test.rb
+++ b/test/multiverse/suites/sidekiq/sidekiq_instrumentation_test.rb
@@ -47,7 +47,7 @@ class SidekiqInstrumentationTest < Minitest::Test
     exception = StandardError.new('bonk')
     noticed = []
     NewRelic::Agent.stub :notice_error, proc { |e| noticed.push(e) } do
-      Sidekiq::CLI.instance.handle_exception(exception)
+      cli.handle_exception(exception)
     end
 
     assert_equal 1, noticed.size


kaylareopelle (approved):
Nice find! 🤞

PR #2239, opened 2023-09-29
Disable autostart for non 'server' Rails commands

### LINKED ISSUE #2086: Update agent defaults to ignore rails 7 commands
Rails 7 uses Thor for the rails commands instead of rake. Currently we're ignoring these using our config `autostart.denylisted_rake_tasks`, but this doesn't work for the Thor commands. One option we could explore is adding the classes for the commands we want to ignore to our default for `autostart.denylisted_constants`.

Comment surfacing this issue: https://github.com/newrelic/newrelic-ruby-agent/issues/1313#issuecomment-1594303083

fallwith:
Previously booting a Rails app via launching the Rails console or running a Rake task would prevent the agent from autostarting, but other Rails commands such as `rails routes` would still see the agent autostart.

Now the `:'autostart.denylisted_constants'` config option's default list of constants has been updated to include all of the `Rails::Command` based command classes that either boot a Rails app or otherwise don't hide via `hide_command!`.

This existing denylisted constants based approach was chosen given what little we have to work with when taking alternative approaches. When a Rails user runs `bin/rails routes` in their app directory, the `railties/lib/rails/commands.rb` file ends up `shift`-ing the `'routes'` value out of `ARGV` and `ARGV` is empty (`[]`) when it reaches the agent. With `ARGV` not providing us anything to hook into, we have to either rely on walking the callstack to determine a caller or checking for constants to be defined. Given the precedent for checking for constants to ignore Rails console contexts, that approach was built upon to ignore the other Rails CLI/TUI contexts.

resolves #2086


fallwith:
NOTE that these additional commands were not skipped over previously due to an oversight or anything. The agent simply predates their existence. The additional commands either did not exist at all previously or were handled differently as Rake tasks, and the agent already ignores Rake tasks by default.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1014,7 +1014,22 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+            Rails::Command::DestroyCommand
+            Rails::Command::DevCommand
+            Rails::Command::EncryptedCommand
+            Rails::Command::GenerateCommand
+            Rails::Command::InitializersCommand
+            Rails::Command::NotesCommand
+            Rails::Command::RakeCommand
+            Rails::Command::RoutesCommand
+            Rails::Command::SecretsCommand
+            Rails::Command::TestCommand
```
kaylareopelle, line 1030: Some customers choose to run New Relic in their test environments by manually setting `monitor_mode: true`. If they use `rails test` to run their tests, would their experience change?

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 Version <dev> brings support for gleaning a Docker container id from cgroups v2 based containers and records additional synthetics attributes.
 
+- **Feature: Prevent the agent from starting in additional Rails CLI/TUI contexts**
```
tannalynn, line 7: maybe we should clarify that this fixes an issue on rails 7? since older rails versions already have this working the the rake task config?

lib/new_relic/agent/configuration/default_source.rb
fallwith: Thank you for zooming me out and reminding me that we already have a convention for handling the "test" environment. We have a convention for handling Rake tasks as well. Removed the `TestCommand` and `RakeCommand` command classes from the denylist. We don't want to redundantly disable functionality or as you said cause issues for existing users who rely on using the agent in a test or Rake task context. Addressed with [c6dae08](https://github.com/newrelic/newrelic-ruby-agent/pull/2239/commits/c6dae081ace9af1295445138e2aeb9073568bb02)

CHANGELOG.md
fallwith: I thought about that. Unfortunately I learned that `bin/rails routes` isn't particularly new and has been around since at least Rails 5.0. So while it's true that the agent and its autostart denylist logic predates Rails 5.0 and that's why we never denylisted the `bin/rails` subcommands previously, it's likely that this change applies to the majority of current Rails shops.

fallwith: (From an offline conversation) I was incorrect in my understanding of the problem. Even though Rails 5 knew what to do with `rails routes`, the `routes` bit was actually being handed off to Rake.

- Beginning of Rails - had very limited rails binary functionality, use rake for everything except starting the server
- Version ? - the rails binary is now a one stop shop. Run rails foo and Rails will either know about foo and do something, or work as a pass-through to hand foo off to Rake
- Version 7 - certain subcommands such as routes are now known to Rails and handled by Rails commands instead of being handed off to Rake

So it's Rails 7 users who will benefit from this PR, with 5 and 6 users already having autostart disabled via Rake related denylist functionality.

Thanks for clearing things up for me, @tannalynn!

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
...
+
+  The agent already knew how to not start up when recognizing that `rails console` or a Rake task was running, but it did not recognize other Rails CLI/TUI contexts resulting from commands such as `rails routes`. Now the agent will additionally recognize other Rails commands such as `rails routes`, `rails dbconsole`, `rails test`, etc. and not start up in those contexts. Note that the agent will continue to start up when the `rails server` and `rails runner` commands and invoked. [PR#2239](https://github.com/newrelic/newrelic-ruby-agent/pull/2239)
```
```suggestion
  The agent already knew how to not start up when recognizing that `rails console` or a Rake task was running, but it did not recognize other Rails CLI/TUI contexts resulting from commands such as `rails routes`. Now the agent will additionally recognize other Rails commands such as `rails routes`, `rails dbconsole`, etc. and not start up in those contexts. Note that the agent will continue to start up when the `rails server` and `rails runner` commands and invoked. [PR#2239](https://github.com/newrelic/newrelic-ruby-agent/pull/2239)
```

```suggestion
- **Feature: Prevent the agent from starting in rails commands in Rails 7**

  Previously, the agent ignored many rails commands by default, such as `rails routes`, using rake specific logic`. This was accomplished by setting these commands as default values for the config option `autostart.denylisted_rake_tasks`.However, rails 7 no longer uses rake for these commands, causing the agent to start running and attempting to record data when running these commands. The commands have now been added to the default value for the config option `autostart.denylisted_constants`, which will allow the agent to recognize these commands correctly in rails 7 and prevent the agent from starting during ignored tasks. Note that the agent will continue to start up when the `rails server` and `rails runner` commands and invoked. [PR#2239](https://github.com/newrelic/newrelic-ruby-agent/pull/2239)
```

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 Version <dev> brings support for gleaning a Docker container id from cgroups v2 based containers and records additional synthetics attributes.
 
+- **Feature: Prevent the agent from starting in additional Rails CLI/TUI contexts**
+
+  The agent already knew how to not start up when recognizing that `rails console` or a Rake task was running, but it did not recognize other Rails CLI/TUI contexts resulting from commands such as `rails routes`. Now the agent will additionally recognize other Rails commands such as `rails routes`, `rails dbconsole`, etc. and not start up in those contexts. Note that the agent will continue to start up when the `rails server` and `rails runner` commands and invoked. [PR#2239](https://github.com/newrelic/newrelic-ruby-agent/pull/2239)
```
tannalynn, line 9:
```suggestion
- **Feature: Prevent the agent from starting in rails commands in Rails 7**

  Previously, the agent ignored many rails commands by default, such as `rails routes`, using rake specific logic`. This was accomplished by setting these commands as default values for the config option `autostart.denylisted_rake_tasks`. However, rails 7 no longer uses rake for these commands, causing the agent to start running and attempting to record data when running these commands. The commands have now been added to the default value for the config option `autostart.denylisted_constants`, which will allow the agent to recognize these commands correctly in rails 7 and prevent the agent from starting during ignored tasks. Note that the agent will continue to start up when the `rails server` and `rails runner` commands and invoked. [PR#2239](https://github.com/newrelic/newrelic-ruby-agent/pull/2239)
```

CHANGELOG.md
fallwith: committed the [d66a3a9](https://github.com/newrelic/newrelic-ruby-agent/pull/2239/commits/d66a3a9a20873cee668403c1a1f9a155e7f31e9a) version

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 Version <dev> brings support for gleaning a Docker container id from cgroups v2 based containers and records additional synthetics attributes.
 
+- **Feature: Prevent the agent from starting in rails commands in Rails 7**
+
+  Previously, the agent ignored many rails commands by default, such as `rails routes`, using rake specific logic`. This was accomplished by setting these commands as default values for the config option `autostart.denylisted_rake_tasks`. However, rails 7 no longer uses rake for these commands, causing the agent to start running and attempting to record data when running these commands. The commands have now been added to the default value for the config option `autostart.denylisted_constants`, which will allow the agent to recognize these commands correctly in rails 7 and prevent the agent from starting during ignored tasks. Note that the agent will continue to start up when the `rails server` and `rails runner` commands and invoked. [PR#2239](https://github.com/newrelic/newrelic-ruby-agent/pull/2239)
```
```suggestion
  Previously, the agent ignored many Rails commands by default, such as `rails routes`, using Rake-specific logic. This was accomplished by setting these commands as default values for the config option `autostart.denylisted_rake_tasks`. However, Rails 7 no longer uses Rake for these commands, causing the agent to start running and attempting to record data when running these commands. The commands have now been added to the default value for the config option `autostart.denylisted_constants`, which will allow the agent to recognize these commands correctly in Rails 7 and prevent the agent from starting during ignored tasks. Note that the agent will continue to start up when the `rails server` and `rails runner` commands are invoked. [PR#2239](https://github.com/newrelic/newrelic-ruby-agent/pull/2239)
```

kaylareopelle (approved):

PR #2242, opened 2023-10-05
CI: permit tests to be specified by line number

### LINKED ISSUE #2241: CI: Support the referencing of one or more individual tests with colons and line numbers
While reviewing a [recent write-up about new Ruby test framework TLDR](https://blog.testdouble.com/posts/2023-10-03-introducing-the-tldr-ruby-test-framework/), @kaylareopelle and I decided that we're not ready to use that framework with this project yet but that we could benefit from the ability to specify line numbers for where individual tests are defined near.

For example. Given this file:

```ruby
 1. # test/my_test.rb
 2.
 3. class MyTest < Minitest::Test
 4.   def test_something_specific
 5.     # do something
 6.     # verify something
 7.   end
 8.
 9.   def test_something_else
10.     # do something
11.     # verify something
12.   end
13. end
``` 

we'd like to be able to do this:

```shell
TEST=test/my_test.rb:5 bundle exec rake test
```

with the `5` referencing line 5. Any of the lines 4-7 should behave the same if given as the value. Upon receiving this input, our Rake task should run only the `test_something_specific` test and nothing else.

fallwith:
Permit one or more individual tests to be referenced by line number instead of by name, permitting one to do the following:

```
TEST=test/my_test.rb:5 bundle exec rake test

bert test/my_test.rb:5
```

resolves #2241

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1265,6 +1265,7 @@ Style/MethodCallWithArgsParentheses:
     - stub_const
     - throw
     - use
+    - warn
   AllowedPatterns: [^assert, ^refute]
 
 Style/MethodCallWithoutArgsParentheses:
--- a/lib/tasks/tests.rake
+++ b/lib/tasks/tests.rake
@@ -11,6 +11,75 @@ rescue LoadError
 end
 
 if defined? Rake::TestTask
+  def name_for_number(content, number)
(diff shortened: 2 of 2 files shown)


kaylareopelle (approved):
Works great, thank you!

Issue #2245, opened 2023-10-06
Rails 7.1: Duplicate log events created per broadcasted logger


kaylareopelle:
When using Rails 7.1 with the Ruby agent, ActiveSupport::Logger instrumentation no longer blocks broadcasted loggers from sending duplicate events. 

This issue seeks to resolve the duplication.

See: 
* https://github.com/rails/rails/issues/49494 
* https://github.com/rails/rails/pull/49518

PR #2252, opened 2023-10-10
Rails 7.1: Record only one copy of a broadcasted log message

kaylareopelle:
In Rails, you have one central logger, `Rails.logger`, but often want those messages to be output to multiple destinations. In a development environment, this includes STDOUT and `development.log` by default. This works by having one central logger that creates the logs, and "broadcasts" those logs to multiple destinations.

This posed a problem when we added APM automatic log forwarding, because New Relic emitted an identical log event for each broadcast, often causing duplicate events to appear in the UI.

In earlier versions of Rails, we leverage [`ActiveSupport::Logger.broadcast`](https://github.com/rails/rails/blob/7-0-stable/activesupport/lib/active_support/logger.rb#L23) to set the `@skip_instrumenting` instance variable to `true` to avoid recording duplicated logs. This was a private method, prone to change at any time.

Rails 7.1 replaced the `ActiveSupport::Logger.broadcast` method with a new public class, [`ActiveSupport::BroadcastLogger`](https://github.com/rails/rails/blob/7-1-stable/activesupport/lib/active_support/broadcast_logger.rb). This class operates in largely the same way as the old method, but since it is a class and no longer a method, we cannot monkey-patch it in the same way. However, since it is public, we can have greater faith in the stability of our instrumentation.

This POC provides one option to prevent duplicate log records from being sent to New Relic when using `ActiveSupport::BroadcastLogger`. Logs emitted by `debug`, `info`, or `add` methods set `@skip_instrumenting` to `true` for all but the first logger in the broadcasts array. At the end of the method, all loggers in the broadcasts array have `@skip_instrumenting` set to `false`.

Some things to consider:
* The loggers included in `ActiveSupport::BroadcastLogger#broadcasts` can change after startup. New loggers may be added or removed from the array.
* The `add` method in `ActiveSupport::BroadcastLogger` calls the `add` method on each individual logger. This means the BroadcastLogger version of the method [doesn't call the severity method (ex. `debug`) like it does in the Ruby logger](https://github.com/ruby/logger/blob/master/lib/logger.rb#L691-L693).
* If we decide this POC is on the right track, we'll need to monkey-patch all the severity methods to get our desired behavior (`error`, `warn`, etc.)

Closes: #2245


lib/new_relic/agent/instrumentation/active_support_broadcast_logger/instrumentation.rb
```diff
@@ -0,0 +1,13 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic::Agent::Instrumentation
+  module ActiveSupportBroadcastLogger
+    def add_with_new_relic(*args)
```
kaylareopelle, line 7: This method name should probably change to something like "filter broadcasts", but since this is a POC, I'm leaving those details until I get approval on this idea.

lib/new_relic/agent/instrumentation/active_support_logger.rb
```diff
@@ -10,13 +10,7 @@
...
 
   depends_on do
-    defined?(ActiveSupport::Logger) &&
-      # TODO: Rails 7.1 - ActiveSupport::Logger#broadcast method removed
-      # APM logs-in-context automatic forwarding still works, but sends
-      # log events for each broadcasted logger, often causing duplicates
-      # Issue #2245
-      defined?(Rails::VERSION::STRING) &&
-      Gem::Version.new(Rails::VERSION::STRING) < Gem::Version.new('7.1.0')
+    defined?(ActiveSupport::Logger) && defined?(ActiveSupport::Logger.broadcast)
```
kaylareopelle, line 13: `ActiveSupport::Logger` is still defined in Rails 7.1, but the method we instrument, `ActiveSupport::Logger.broadcast` is not. This seemed more precise than using a Rails version.

test/multiverse/suites/active_support_logger/Envfile
```diff
@@ -0,0 +1,23 @@
+# This file is distributed under New Relic's license terms.
```
kaylareopelle, line 1: I realized our `chain` instrumentation for ActiveSupport::Logger was never being tested, so I created a multiverse suite to make sure it gets checked.

test/multiverse/suites/rails/rails_logger_test.rb
```diff
@@ -0,0 +1,43 @@
+# This file is distributed under New Relic's license terms.
```
kaylareopelle, line 1: Hopefully, this test will give us a heads-up if something changes with the Rails logger that causes the duplication issue to resurface.

fallwith (dismissed):

fallwith (dismissed):

lib/new_relic/agent/instrumentation/active_support_broadcast_logger/chain.rb
```diff
@@ -0,0 +1,69 @@
...
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic::Agent::Instrumentation
+  module ActiveSupportBroadcastLogger::Chain
+    def self.instrument!
+      ::ActiveSupportBroadcastLogger.class_eval do
+        include NewRelic::Agent::Instrumentation::ActiveSupportBroadcastLogger
+
+        alias_method(:add_without_new_relic, :add)
```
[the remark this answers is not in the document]
fallwith: For others who might come across this with curiosity... `define_method` works fine with `super` if the arguments are fixed. But the splatted `*args` needed here won't work:

```
implicit argument passing of super from method defined by
define_method() is not supported. Specify all arguments
explicitly. (RuntimeError)
```

test/multiverse/suites/rails/rails_logger_test.rb
```diff
@@ -0,0 +1,43 @@
...
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require './app'
```
fallwith, line 5: Equivalent to `require_relative 'app'` with fewer keystrokes?

CHANGELOG.md
```diff
@@ -20,6 +20,12 @@ Version <dev> brings support for gleaning a Docker container id from cgroups v2
...
 
+- **Fix: Stop sending duplicate log events for Rails 7.1 users**
+
+  Rails 7.1 introduced the public API [`ActiveSupport::BroadcastLogger`](https://api.rubyonrails.org/classes/ActiveSupport/BroadcastLogger.html). This logger replaces a private API, `ActiveSupport::Logger.broadcast`. In Rails versions below 7.1, the agent uses the `broadcast` method to stop duplicate logs from being recoded by broadcasted loggers. Now, we've updated the code to provide a similar duplication fix with the new `ActiveSupport::BroadcastLogger` class.
+
+  Previously, the agent prevented broadcasted loggers from recording duplicate log events by skipping instrumentation for broadcasted loggers, identified by calling `ActiveSupport::Logger.broadcast`. In Rails 7.1, this method has been refactored into a class. The agent now records log events for only the first logger found in an `ActiveSupport::BroadcastLogger.broadcasts` array.
```
kaylareopelle, line 27: I'm only planning to keep one of these two paragraphs, @fallwith. Do you have a preference?

test/multiverse/suites/rails/rails_logger_test.rb
kaylareopelle: Great catch! I prefer require_relative. This was a copy/paste carryover.

kaylareopelle: Updated in: 0f9ebe3eb1545bc8d5279f193e608134ed4e5478

lib/new_relic/agent/instrumentation/active_support_broadcast_logger/chain.rb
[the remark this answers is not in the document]
kaylareopelle: Thank you for clarifying, @fallwith!

CHANGELOG.md
fallwith: If I can only have 1, the first one seems best. I link the link to BroadcastLogger documentation.

fallwith (approved):

nathan-appere:
This causes an exception for us on:
```
NoMethodError
undefined method `each' for nil:NilClass (NoMethodError)

      broadcasts[1..-1].each { |broadcasted_logger| broadcasted_logger.instance_variable_set(:@skip_instrumenting, true) }
```

With:
- `newrelic_rpm (9.7.1)`
- `rails (7.1.3.2)`

I have no idea how `broadcasts` can be nil (should be an Array looking at the implementation of `ActiveSupportBroadcastLogger`)

Only happens in production with low volume, so tricky to identify the cause (the volume makes me think it could be thread related?)

PR #2256, opened 2023-10-12
Add and Replace SLASH and ROOT constants

### LINKED ISSUE #2153: Add `'/'.freeze` to constants.rb
We have a few instances of `'/'.freeze` throughout the agent. We should add this to `constants.rb` and replace calls with this newly created constant.

chahmedejaz:
# Overview

- This PR adds the SLASH and ROOT constants and replaces them throughout the agent.
- I created the ROOT (`/`) constant because I felt that we are using it in the path's context in many places as well
- I thought it would be handy to separate it out into a new constant.

Submitter Checklist:

Resolves #2153

### DIFF
--- a/lib/new_relic/agent/rules_engine.rb
+++ b/lib/new_relic/agent/rules_engine.rb
@@ -9,7 +9,7 @@ require 'new_relic/language_support'
 module NewRelic
   module Agent
     class RulesEngine
-      SEGMENT_SEPARATOR = '/'.freeze
+      SEGMENT_SEPARATOR = NewRelic::SLASH
       LEADING_SLASH_REGEX = %r{^/}.freeze
 
       include Enumerable
--- a/lib/new_relic/agent/instrumentation/active_record_helper.rb
+++ b/lib/new_relic/agent/instrumentation/active_record_helper.rb
@@ -232,7 +232,6 @@ module NewRelic
 
           DEFAULT = 'default'.freeze
           UNKNOWN = 'unknown'.freeze
-          SLASH = '/'.freeze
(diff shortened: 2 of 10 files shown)


lib/new_relic/agent/rules_engine.rb
```diff
@@ -9,7 +9,7 @@
 module NewRelic
   module Agent
     class RulesEngine
-      SEGMENT_SEPARATOR = '/'.freeze
+      SEGMENT_SEPARATOR = NewRelic::SLASH
```
chahmedejaz, line 12: I assigned the `SLASH` to `SEGMENT_SEPARATOR` because I felt, the `SEGMENT_SEPARATOR` is the best name for the context for `SLASH` here. Please let me know if you think otherwise.

fallwith: Having one constant reference another gives us the benefit of using a more context-specific name while still minimizing resources and maintaining two copies of the same frozen string. I really like your approach here, @chahmedejaz.

fallwith (approved):
Thank you very much for your contribution, @chahmedejaz! I'll let the automated unit tests run and as long as they pass with the changes, we'll get this one merged. These changes will provide improved performance for customers and are appreciated!

chahmedejaz:
> Thank you very much for your contribution, @chahmedejaz! I'll let the automated unit tests run and as long as they pass with the changes, we'll get this one merged. These changes will provide improved performance for customers and are appreciated!

It was a pleasure working on this. Thanks for giving me the opportunity! :smile:

PR #2269, opened 2023-10-17
send_stream subscriber

### LINKED ISSUE #1588: Add instrumentation for ActionController::Live#send_stream
### Description

* #send_stream is a new feature in Rails 7.0 Action Controller
* This isn't listed on the guide for Rails::Notifications (yet)
* We don't have any instrumentation for ActionController::Live yet. There may be some extra steps that need to be taken to instrument this class
* Instead of monkey-patching the method New Relic-style, we could contribute a notification event PR for this method
* Consider applying what we'd like to see Rails do via a monkey-patch ourselves and rip it back out again once the PR gets merged 
* The agent reacts to the presence of ActionController::Live by not installing the browser agent

Resources
------------
* https://api.rubyonrails.org/classes/ActionController/Live.html#method-i-send_stream 
* https://blog.appsignal.com/2021/12/15/whats-new-in-rails7.html (search for send_stream)

hannahramadan:
A [PR to add ](https://github.com/rails/rails/pull/49297)`send_stream` notifications is sitting with the Rails team and has been labeled to go out in Rails 7.2. Rather than hold this PR, let's just skip the send_stream subscriber test until Rails 7.2. is released. 

Closes #1588

### DIFF
--- a/lib/new_relic/agent/instrumentation/rails_notifications/action_controller.rb
+++ b/lib/new_relic/agent/instrumentation/rails_notifications/action_controller.rb
@@ -34,6 +34,7 @@ DependencyDetection.defer do
 
     subs = %w[send_file
       send_data
+      send_stream
       redirect_to
       halted_callback
       unpermitted_parameters]
--- a/test/multiverse/suites/rails/action_controller_other_test.rb
+++ b/test/multiverse/suites/rails/action_controller_other_test.rb
@@ -17,6 +17,11 @@ if defined?(ActionController::Live)
       send_data('wow its a adata')
     end
 
+    # send_stream
+    def send_test_stream
+      send_stream(filename: 'dinosaurs.html')
+    end
+
     # halted_callback
     before_action :do_a_redirect, only: :halt_my_callback
     def halt_my_callback; end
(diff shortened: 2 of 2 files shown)


lib/new_relic/agent/instrumentation/rails_notifications/action_controller.rb
```diff
@@ -38,6 +38,9 @@
       halted_callback
       unpermitted_parameters]
 
+    # TODO: Add send_steam to subs list when Rails 7.2 becomes avaliable
+    subs << 'send_stream' if Rails::VERSION::MAJOR.to_i >= 7 && Rails::VERSION::MINOR.to_i >= 2
```
fallwith, line 42: I suggest going ahead and satisfying the TODO right now, and only having the conditional check live in the tests where you have it.

The ever-so-slightly-more complex regex pattern that results from "or"-ing (`|`) in the new `'send_stream'` string will make for negligible performance impact on Rails < 7.2 users while they wait for 7.2 to be released.

And then when 7.2 drops, they can automatically benefit from the new subscription type without needing to upgrade their APM agent.

hannahramadan: Thank you! I like this suggestion and have updated ◡̈

fallwith (approved):

kaylareopelle (approved):

Issue #2263, opened 2023-10-16
CI: Improve logged error visibility


fallwith:
When the unit tests run, errors and warnings may appear in either the New Relic agent log and/or the Rails log when Rails is in scope.

Sometimes these errors are expected, because behavior that would produce a warning or an error was artificially induced under test.

Sometimes these errors are not expected, and unless a human reviews the log(s), the errors and warnings will go unnoticed.

For improved developer visibility of unexpected errors and for increased scrutiny performed by test automation, it would be ideal if errors and warnings that are found in the logs will be surfaced to the human viewer and cause a test failure when under automation.

Here's a proposal for a linear sequence of passes to take:

1. After a test completes, quote any WARN and ERROR lines from newrelic_agent.log to STDOUT. These should be hidden unless `VERBOSE_TEST_OUTPUT` is enabled for now.
2. After a test completes, also quote any WARN and ERROR lines from the relevant Rails log to STDOUT. These should be hidden unless `VERBOSE_TEST_OUTPUT` is enabled for now.
3. Create an "expect this action to log with this severity" type helper that will monitor our own internal logger class's behavior AND also mark any generated log file entries to be ignored and not output to STDOUT.
4. Once 1-3 are satisfied, consider the presence of a WARN or ERROR message to be cause for a test failure. Fail the test and output the log message(s). Do this regardless of `VERBOSE_TEST_OUTPUT`.

chynesNR:
Not needed.

PR #2271, opened 2023-10-18
Rails fixes

### LINKED ISSUE #2253: Confirm good compatibility with and value add for ActionController#proc_render
With the release of Rails v7.1, we saw the multiverse `rails` test suite start failing due to changes related to `ActionController#proc_render`.

Our test class `test/multiverse/suites/rails/view_instrumentation_test.rb` actually defines a customer `proc_render` method on our controller class for backwards compatibility with Rails 3 whose `ActionController` did not offer it.

With https://github.com/newrelic/newrelic-ruby-agent/pull/2248 we stopped testing `proc_render` because of Rails v7.1 compatibility issues.

We should do the following:

- Determine if we can get rid of our custom `proc_render` method in the test class and simply exclude `proc_render` testing from Rails versions that are too hold to support it
- Determine if there is anything the agent is not compatible with that would cause errors or warnings or other undesirable behavior if a customer is to leverage `proc_render` in their Rails app
- Determine if there is anything the agent could do additionally to provide better instrumentation around `proc_render` that it isn't currently offering.

### LINKED ISSUE #2254: Rails v7.1+ support: get to the bottom of ActionController::Streaming::Body#to_ary issues in CI
With https://github.com/newrelic/newrelic-ruby-agent/pull/2248, we introduced a hack to improve Rails v7.1 compatibility with our Rails multiverse testing by adding a `test/multiverse/suites/rails/before_suite.rb` file.

This hack was introduced with the following assumptions having been made:

- The hack is only necessary for our multiverse testing, and not for customers using the agent with production Rails v7.1 code bases
- The hack is only necessary because of some of the wacky things we do to test Rails. See https://github.com/newrelic/newrelic-ruby-agent/issues/2251 for helping us be less wacky.

We should confirm these assumptions and address the `TODO` comment in the `before_suite.rb` file - which would mean that we could remove the entire `before_suite.rb` file itself from the project.

fallwith:
These Rails fixes were driven by extensive testing of the newly released Rails version 7.1, but some may be relevant to older Rails versions as well.

*lib/*
- When adding new segment attributes, anticipate and ignore numeric parameter keys
- For `Transaction#finish`, don't attempt to invoke methods on a non-existent initial segment

*test/*
- Inline controller method rendering with a proc was a thing in early Rails versions that was no longer supported in Rails 3. We were defining our own `proc_render` controller method to simulate the old style behavior and it finally caused issues for us in Rails 7. Given that we don't even support Rails versions older than 4 now, let's just stop testing the Rails 2 and below proc functionality.
- `ActiveJobTest`: a single test needed an additional `expect`
- Remove the temporary `before_suite.rb` based hacks for the Rails multiverse suite.
- Modify the `ActionController::Live` RUM test to set a `last-modified` header on the streaming response. This prevents Rack v3 from attempting to call `to_ary` on the streaming response object returned. In the future we could author a better streaming test that properly chunks data and possibly even uses a layout. But for now, this header fix gives us Rack 3 compatibility with the old test.

resolves #2253
resolves #2254


tannalynn (approved):

lib/new_relic/agent/transaction.rb
```diff
@@ -520,7 +520,7 @@ def needs_middleware_summary_metrics?(name)
       end
 
       def finish
-        return unless state.is_execution_traced?
+        return unless state.is_execution_traced? && initial_segment
```
hannahramadan, line 523: Nice change... so this is making sure at least one segment exists. Just curious was this an issue that came up in Rails testing or an enhancement you found while in code?

hannahramadan (approved):
Your PR summary was very helpful ✨

lib/new_relic/agent/transaction.rb
fallwith: This fix was driven by the `rails` multiverse suite. Related to https://github.com/newrelic/newrelic-ruby-agent/issues/2263, this change was needed to eliminate an error that showed up only in the logs but did not cause test assertions to fail.

kaylareopelle (approved):

Issue #1176, opened 2022-05-19
Instrumenting Async::HTTP


trevorturk:
### Is your feature request related to a problem? Please describe. 

I migrated a Rails application from [Puma](https://github.com/puma/puma) and [Typhoeus](https://github.com/typhoeus/typhoeus) to [Falcon](https://github.com/socketry/falcon) and [Async::HTTP](https://github.com/socketry/async-http). 

New Relic seems to be mostly working well after the switch, but I miss the "Web External" details that I had before on the New Relic dashboard. This would show me the average response time for the external API endpoints I was hitting. I found it useful for setting timeout levels per external API.

(For a little more context, in my app, I'm hitting weather data APIs, and some are slower than others. If I can see the average response time depending on the API hostname, it helps me set appropriate timeouts per data source, to limit my app's autoscaling.)

Only time will tell, but I expect more people to consider using Async::HTTP in the future (I've heard about it in multiple talks at RailsConf 2022 so far) so I would love to see New Relic add support here. 

### Feature Description

"Web External" instrumentation for the Async::HTTP gem, so I can see average response times etc per external API. 

### Additional context

Perhaps related to #850, /cc @kaylareopelle who reminded me of this issue during the RailsConf 2022 talk from New Relic. 

I'd be happy to help with this feature in any way!

### Priority

Really Want. (I've been living without it for a while, but it was useful information that I miss.)

Thank you!

kaylareopelle:
Hi, @trevorturk! 👋  

Thanks for submitting this issue! It was great chatting with you at RailsConf. I could see a spike on this issue fitting in well with #850! I don't believe our team will be able to start work on that until Q3 (approx. October 2022). If you're interested in contributing code to instrument this gem in the meantime, we'd love to assist you! 

The first step would be to identify the methods that should be instrumented (like the process in the part of the talk about Ruby's Logger class). What methods are you calling when you use Async::HTTP? Do those methods call something else more central that all external requests pass through? 

When you have some ideas, please bring your research to this issue and our team can take a look to verify whether they seem like the right ones.

Thanks again for your submission!

trevorturk:
Thanks for the reply! I don't think I have the time and/or ability to help before Q3 honestly, and I understand if it's not a top priority in any case. I think you'll have more `async` users over time, but I think most people still reach for `typhoeus` at this time, so it's not urgent.  

[...]

PR #2272, opened 2023-10-18
Add async http instrumentation

tannalynn:
Adds instrumentation for [async-http](https://github.com/socketry/async-http)

resolves https://github.com/newrelic/newrelic-ruby-agent/issues/1176


test/multiverse/suites/async_http/async_http_instrumentation_test.rb
```diff
@@ -0,0 +1,131 @@
...
+  end
+
+  def request_and_wait(method, url, headers = nil, body = nil)
+    resp = nil
+    Async do
+      internet = Async::HTTP::Internet.new
+      resp = internet.send(method, url, headers)
+      @read_resp = resp&.read
+    rescue => e
+      puts "**************************ERROR: #{e}"
```
fallwith, line 34: Could this `puts` statement be commented out - at least conditionally?

```shell
# Running:

.................**************************ERROR: Could not read line!
...............................**************************ERROR: read timeout reached
............

Fabulous run in 14.474338s, 4.1453 runs/s, 8.5669 assertions/s.
```

lib/new_relic/agent/http_clients/async_http_wrappers.rb
```diff
@@ -0,0 +1,83 @@
...
+module NewRelic
+  module Agent
+    module HTTPClients
+      class AsyncHTTPResponse < AbstractResponse
+        def get_status_code
+          get_status_code_using(:status)
+        end
+
+        def [](key)
+          @wrapped_response.headers.to_h[key.downcase]&.first
```
fallwith, line 17:
```suggestion
          to_hash[key.downcase]&.first
```

lib/new_relic/agent/instrumentation/async_http.rb
```diff
@@ -0,0 +1,31 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require_relative 'async_http/instrumentation'
+require_relative 'async_http/chain'
+require_relative 'async_http/prepend'
+
+DependencyDetection.defer do
+  named :'async_http'
```
fallwith, line 10:
```suggestion
  named :async_http
```

lib/new_relic/agent/instrumentation/async_http.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  depends_on do
+    # The class that needs to be defined to prepend/chain onto. This can be used
+    # to determine whether the library is installed.
+    defined?(Async::HTTP)
+    # Add any additional requirements to verify whether this instrumentation
+    # should be installed
```
fallwith, line 17:
```suggestion
```

lib/new_relic/agent/instrumentation/async_http.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  executes do
+    NewRelic::Agent.logger.info('Installing async_http instrumentation')
+
+    require 'async/http/internet'
+    require 'new_relic/agent/http_clients/async_http_wrappers'
```
fallwith, line 24: The first `require` is needed here for the code block that follows it. As for the wrappers file, I recommend moving that one to `async_http/instrumentation.rb` to make it more clear that it's that code that needs it.

lib/new_relic/agent/instrumentation/async_http.rb
```diff
@@ -0,0 +1,31 @@
...
+    if use_prepend?
+      prepend_instrument Async::HTTP::Internet, NewRelic::Agent::Instrumentation::AsyncHttp::Prepend
```
fallwith, line 26: Do we anticipate instrumenting any additional `Async::HTTP` classes in future? If not, this is great as-is. If so, maybe our own class name should include `Internet` in its name.

lib/new_relic/agent/instrumentation/async_http/instrumentation.rb
```diff
@@ -0,0 +1,35 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic::Agent::Instrumentation
+  module AsyncHttp
+    def call_with_new_relic(method, url, headers = nil, body = nil)
+      headers ||= {} # if it is nil, we need to make it a hash so we can insert headers
```
fallwith, line 8: Will there be an actual `nil` value passed as the third argument, or just the absence of a third argument? If it's the latter, you could do this:

```suggestion
    def call_with_new_relic(method, url, headers = {}, body = nil)
```

newrelic.yml
```diff
@@ -382,7 +382,11 @@ common: &default_settings
   # prepend, chain, disabled.
   # instrumentation.bunny: auto
 
-  # Controls auto-instrumentation of the concurrent-ruby library at start-up. May be
+# Controls auto-instrumentation of Async::HTTP at start up.
+# May be one of [auto|prepend|chain|disabled]
+# instrumentation.async_http: auto
```
fallwith, line 387:
```suggestion
  # Controls auto-instrumentation of Async::HTTP at start up.
  # May be one of [auto|prepend|chain|disabled]
  # instrumentation.async_http: auto
```

test/multiverse/suites/async_http/Envfile
```diff
@@ -0,0 +1,10 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+instrumentation_methods :chain, :prepend
+
+gemfile <<~RB
+  gem 'async-http'
+  gem 'rack'
+RB
```
fallwith, line 10: I think we should test 3 years back and latest stable...

```suggestion
ASYNC_HTTP_VERSIONS = [
  nil,
  '0.53.1'
]

def gem_list(async_http_version = nil)
  <<~GEM_LIST
    gem 'async-http'#{async_http_version}
    gem 'rack'
  GEM_LIST
end

create_gemfiles(ASYNC_HTTP_VERSIONS)
```

test/multiverse/suites/async_http/async_http_instrumentation_test.rb
tannalynn: Thanks, I had this originally for debugging purposes and just forgot to remove it. 
JSYK though, the async gem will catch the error now and output it as well, but that doesn't cause any actual issues with the tests

lib/new_relic/agent/instrumentation/async_http/instrumentation.rb
tannalynn: It will be passed an actual nil since we always call this method like `call_with_new_relic(method, url, headers, body)`, the headers = nil is just because I copied the original call method args exactly, but since it goes through another method to get here, all of those should always be passed in, even if they are nil.

lib/new_relic/agent/instrumentation/async_http.rb
tannalynn: I don't think so, I don't see any reason we'd need to intrument anything else since all the http requests will go through this instrumentation

lib/new_relic/agent/instrumentation/async_http.rb
tannalynn: Moved in https://github.com/newrelic/newrelic-ruby-agent/pull/2272/commits/fef6ba793936f2d306009c643c5ebdc316704704

newrelic.yml
```diff
@@ -382,7 +382,11 @@ common: &default_settings
   # prepend, chain, disabled.
   # instrumentation.bunny: auto
 
-  # Controls auto-instrumentation of the concurrent-ruby library at start-up. May be
+  # Controls auto-instrumentation of Async::HTTP at start up.
+  # May be one of [auto|prepend|chain|disabled]
+  # instrumentation.async_http: auto
+
+  # Controls auto-instrumentation of the concurrent-ruby library at start up. May be
```
kaylareopelle, line 389: @tannalynn - Out of curiosity, did you add this manually or use the script? (If you used the script, I'm a bit concerned about why the first version was outdented one level)

lib/new_relic/agent/http_clients/async_http_wrappers.rb
```diff
@@ -0,0 +1,83 @@
...
+          @connection = connection
+          @method = method
+          @url = ::NewRelic::Agent::HTTPClients::URIUtil.parse_and_normalize_url(url)
+          @headers = headers
+        end
+
+        ASYNC_HTTP = 'Async::HTTP'
+        LHOST = 'host'
+        UHOST = 'Host'
+        COLON = ':'
```
kaylareopelle, line 36: Not for this PR, but maybe it's time to move these constants to a shared location.

kaylareopelle (approved):
GitHub is showing me "4" comments, so I think some of what shows up will be moot based on more recent changes!

test/multiverse/suites/async_http/Envfile
```diff
@@ -0,0 +1,19 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+instrumentation_methods :chain, :prepend
+
+ASYNC_HTTP_VERSIONS = [
+  nil,
+  '0.59.0'
+]
```
kaylareopelle, line 10: Looks like the RubyGems site is misleading about the lowest supported version. Could you add constraints to test only Ruby 2.5 and above? 

```
Bundler could not find compatible versions for gem "ruby":
  In Gemfile.1:
    ruby

    async-http (~> 0.59.0) was resolved to 0.59.5, which depends on
      async-io (>= 1.28) was resolved to 1.30.0, which depends on
        ruby (~> 2.5)

    newrelic_rpm was resolved to 9.5.0, which depends on
      ruby (>= 2.4.0)
```

newrelic.yml
tannalynn: I used the generator originally, but there was a merge conflict so i had to copy it and put it back, it's very possible that i messed up the indentation when i resolved the merge conflict.

test/multiverse/suites/async_http/Envfile
tannalynn: added 2.5 as the min ruby version https://github.com/newrelic/newrelic-ruby-agent/pull/2272/commits/e911c78feb1ffc87524bc380577b133ab37820d7

kaylareopelle (approved):

lib/new_relic/agent/instrumentation/async_http.rb
fallwith: This one is marked as resolved without a change. Did you decide on keeping the quotes?

tannalynn: oh weird, no thats my bad, sorry about that! I meant to hit commit suggestion but i accidentally hit resolve. I did that a couple times this morning when i was sleepy and thought i had fixed them all but must have missed this one.

fallwith (approved):

trevorturk:
Exciting stuff, thank you!

/cc @ioquatix in case he's interested in having a look 👀

ioquatix:
Thanks for working on this.

I don't want to come across as negative as I think it's awesome someone is getting value from this, but:

1. There is a chance instrumentation that's injected like this will break due to internal changes or interface changes.
2. In order to deal with this, I created a gem, [`traces`](https://github.com/socketry/traces).
3. That gem provides a generic interface for tracing traces and spans, and provides implementation specific backends e.g. [traces-backend-open_telemetry](https://github.com/socketry/traces-backend-open_telemetry), 
[traces-backend-datadog](https://github.com/socketry/traces-backend-datadog).
4. This ensures any tracing (including across service boundaries) is handled correctly and is compatible, as the tracing is provided by the library itself.

I'm cautious about monkey patching approaches as unless tested in non-production (non-instrumented) environments, can suddenly break on deploying to production. Ideally, we can implement `traces-backend-newrelic`?

tannalynn:
@ioquatix 
Thanks for bringing this to our attention! I didn't realize that existed, I must have missed it when I looked originally, sorry about that! Before I started, I looked at the readme for async-http to see if there was anything built in for observability (like subscriptions or something like this) but didn't see anything mentioned so I ended up just approaching it the way we instrument most gems (with monkey patching and prepend options). 

An initial question I have after a quick look at it: Is there a way to insert into the headers? I see there is a w3c trace context method built in, but the new relic ruby agent has other types of headers that we will need to insert into outgoing request headers in certain situations. I'm not sure if that would be possible though. 

I'll have to spend a little time setting up my test app to use traces and see how the things looks when creating a span with the new relic ruby agent with the traces interface. Once I've had a chance to do that I'll have a better idea of if we'll be able to move our instrumentation there. 

Thank you for pointing this out and letting us know about traces!

ioquatix:
If you use `traces`, you will get instrumentation across all the socketry gems where it's implemented.

tannalynn:
Creating a new relic implementation of traces seems like it would be a great thing for us to do to get more visibility, so I've created an issue https://github.com/newrelic/newrelic-ruby-agent/issues/2283 for us to work on it and tentatively plan to work on it next quarter.


As for being able to insert headers, the agent does that on the HTTP client side in order to then parse it out on the HTTP server side (presuming both client and server are instrumented by New Relic). This is something we need to do to support other new relic products (for example: synthetics monitoring https://newrelic.com/platform/synthetics). If that's not something we can do with traces, we'll have to carry on monkey patching for async http still.

trevorturk:
Hello, I know it sounds like even more good stuff is coming later with #2283 but I wanted to say thank you for working on this feature. I just rolled out a NR gem update in an app where I'm using Async::HTTP heavily and things are working great. It's lovely having this extra visibility now -- I really missed that when migrating from Typhoeus. Thank you!

PR #2282, opened 2023-10-23
CI: Rework JRuby testing

fallwith:
- Upgrade to JRuby v9.4.5.0
- Remove all "services" from the JRuby workflow (databases, memory stores, etc.)
- Stop using JRuby `--debug`
- Rename `unit_tests` to `jruby_unit_tests` to match the `jruby_multiverse` name
- Replace calls to our `scripts/retry_command` with the use of `nick-fields/retry`
- Define 2 JRuby specific `add-opens` Java opts and specify `Xmx` and `Xms` values
- Stop testing Rails v6.0 (but continue testing v6.1)
- Switch away from running groups of multiverse suites to running individual suites
- Stop testing all suites known to require a service or be problematic with JRuby
- Introduce a `dup` for config hash resetting, as JRuby correctly points to it as having a potential race condition without the `dup`
- Define a `first_call_for` test helper that permits JRuby to have more than 1 call recorded while still examining the first call
- Skip JRuby for the `async-http` suite as it requires native C extensions

JRuby CI run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/6964342780


.github/workflows/ci_jruby.yml
```diff
@@ -36,85 +28,19 @@ jobs:
         with:
           timeout_minutes: 30
           max_attempts: 2
-          command:  bundle exec rake test:env[norails,rails61,rails60]
+          command:  bundle exec rake test:env[norails,rails61]
```
kaylareopelle, line 31: From your commits, it looks like you tried to use Rails 7.0 here. Why did you decide to go with Rails 6.1 instead?

Theoretically, [JRuby supports 7.0 as of 9.4.4.0](https://www.reddit.com/r/ruby/comments/z2z7qd/jruby_9400_released_now_supporting_ruby_31_and/), and [version 9.4.5.0 is fixing bugs related to Rails 7.1](https://github.com/jruby/jruby/issues/7982), so support might exist for that version too.

.github/workflows/ci_jruby.yml
```diff
@@ -123,22 +49,24 @@ jobs:
...
-        run: ./.github/workflows/scripts/retry_command 
+        uses: nick-fields/retry@943e742917ac94714d2f408a0e8320f2d1fcafcd # tag v2.8.3
+        with:
+          timeout_minutes: 20
+          max_attempts: 3
+          command: "bundle exec rake test:multiverse[${{ matrix.suite }}]"
         env: 
-          TEST_CMD: "bundle exec rake test:multiverse[group=${{ matrix.multiverse }}]"
           VERBOSE_TEST_OUTPUT: true
-          RETRY_ATTEMPTS: 5
```
kaylareopelle, line 138: What's your thinking behind removing the retry attempts? I apologize if this has already been discussed!

test/multiverse/suites/async_http/Envfile
```diff
@@ -16,4 +16,4 @@ def gem_list(async_http_version = nil)
   GEM_LIST
 end
 
-create_gemfiles(ASYNC_HTTP_VERSIONS)
+create_gemfiles(ASYNC_HTTP_VERSIONS) unless defined?(JRUBY_VERSION) # async-http needs native extensions
```
kaylareopelle, line 19: This works as-is, but Envfiles generally skip JRuby using the `Multiverse::Envfile#suite_condition` method. Here's how it's used in Sidekiq:
https://github.com/newrelic/newrelic-ruby-agent/blob/c959f40959bffcfc8aca5478926f89ebedbe3308/test/multiverse/suites/sidekiq/Envfile#L5-L7

kaylareopelle (dismissed):

kaylareopelle (approved):

PR #2303, opened 2023-11-01
Add tests for activerecord async queries

### LINKED ISSUE #2220: Rails v7.1: Author tests to confirm async AR queries are instrumented well
Shopify's https://github.com/rails/rails/pull/44446 delivers new Active Record async support for aggregate and `find_by_sql` queries. While the instrumentation is handled by Rails notifications, we should ensure that the data resulting from these async queries is presented to our users in a coherent and consistent way.

Use rails 7.1.0.beta1 or EDGE to test with.

hannahramadan:
Add tests for activerecord async queries.

Closes #2220


test/multiverse/suites/active_record_pg/active_record_test.rb
```diff
@@ -414,6 +414,24 @@ def test_metrics_for_direct_sql_select
     assert_generic_rollup_metrics('select')
   end
 
+  def test_metrics_for_find_by_sql
+    skip unless active_record_major_version >= 7
```
fallwith, line 418: If there's time / interest for a side project, we could open a PR against `rubocop-minitest` to treat `skip` the same way that `rubocop` treats `raise` (and require a blank line after it appears). No need to make that change here before that PR though.

hannahramadan: Went ahead and added the blank line. I love the contribution idea too!

tannalynn (approved):

fallwith (approved):

PR #2306, opened 2023-11-02
stop testing 8T with Ruby < 2.7

fallwith:
Infinite Tracing depends on the `grpc` gem, which requires Ruby 2.5+.

Frequently, the gRPC Ruby gem maintainers inadvertently break compatibility with Ruby 2.5 and 2.6 by publishing a new version of `google-protobuf` (which `grpc` depends on) that itself requires Ruby 2.7+.

We do not wish to pin our `grpc` dependency to an exact version that is known to be compatible with Ruby 2.5 and Ruby 2.6 so that we leave users free to use whatever 1.x version of `grpc` that they wish, especially if they are themselves using `grpc` in their application.

Whether the gRPC maintainers' frequent breaking of Ruby 2.5 and 2.6 compatibility is accidental or deliberate, supporting these Rubies with our CI system has simply caused too many issues.

We are not actively introducing any new Ruby code to Infinite Tracing that could break Ruby 2.5 and 2.6 compatibility, so we will switch to manual testing of these Rubies as appropriate moving forward.

### DIFF
--- a/infinite_tracing/newrelic-infinite_tracing.gemspec
+++ b/infinite_tracing/newrelic-infinite_tracing.gemspec
@@ -72,7 +72,7 @@ Gem::Specification.new do |s|
   s.summary = 'New Relic Infinite Tracing for the Ruby agent'
 
   s.add_dependency 'newrelic_rpm', NewRelic::VERSION::STRING
-  s.add_dependency 'grpc', '~> 1.59'
+  s.add_dependency 'grpc', '~> 1.34'
 
   s.add_development_dependency 'rake', '12.3.3'
   s.add_development_dependency 'rb-inotify'
--- a/infinite_tracing/newrelic-infinite_tracing.gemspec
+++ b/infinite_tracing/newrelic-infinite_tracing.gemspec
@@ -72,7 +72,7 @@ Gem::Specification.new do |s|
   s.summary = 'New Relic Infinite Tracing for the Ruby agent'
 
   s.add_dependency 'newrelic_rpm', NewRelic::VERSION::STRING
-  s.add_dependency 'grpc', '~> 1.34'
+  s.add_dependency 'grpc', '~> 1.59'
 
   s.add_development_dependency 'rake', '12.3.3'
   s.add_development_dependency 'rb-inotify'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -286,7 +286,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby-version: [2.5.9, 3.2.2]
+        ruby-version: [2.7.8, 3.2.2]
     steps:
       - name: Configure git
         run: 'git config --global init.defaultBranch main'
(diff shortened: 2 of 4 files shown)


infinite_tracing/newrelic-infinite_tracing.gemspec
```diff
@@ -72,7 +72,7 @@ Gem::Specification.new do |s|
   s.summary = 'New Relic Infinite Tracing for the Ruby agent'
 
   s.add_dependency 'newrelic_rpm', NewRelic::VERSION::STRING
-  s.add_dependency 'grpc', '~> 1.34'
+  s.add_dependency 'grpc', '~> 1.59'
```
kaylareopelle, line 75: Would this effectively make new versions of the 8T gem require Ruby 2.5 and above, since the required `grpc` version is one that doesn't work with Ruby 2.5 in our testing?

kaylareopelle (commented):
I have a suggestion for another approach to solve this problem, maybe you've already considered it.

I think the reason why 2.6 failed and 2.5 did not is because of this condition in the infinite tracing Envfile:
https://github.com/newrelic/newrelic-ruby-agent/blob/8c412243f34f008d3fecc06882ca0004de7ca4b2/test/multiverse/suites/infinite_tracing/Envfile#L5-L7

We could change this condition to limit the version for Ruby 2.6 as well. When I wrote that code, I wasn't aware of any issues with 2.6 and the current grpc gem version.

However, any choice we make here would only impact [a very small number of accounts](https://onenr.io/0Zw0NbblOjv), so this is only a non-blocking suggestion.

infinite_tracing/newrelic-infinite_tracing.gemspec
fallwith: This change does not alter anything functionally from a Bundler / Ruby perspective. It still tells Bundler to grab any 1.x version, just the same as before. The only real impact this has is that the `~> 1.59` string will appear in the gemspec file and out on the RubyGems.org listing and convey to a human reader that we have tested with 1.59 and signed off on it.

fallwith: Scratch that, @kaylareopelle. I tested it and you are right. The new constraint is effectively saying `>= 1.59 and < 2.0`, which is imposing a newer version requirement. We don't want that. I will revert this.

fallwith: reversed with [077f66a](https://github.com/newrelic/newrelic-ruby-agent/pull/2306/commits/077f66ae2d2162de621a30dce90928e95fc86b72)

fallwith:
> I have a suggestion for another approach to solve this problem, maybe you've already considered it.

@kaylareopelle I really like this suggestion, but my preference would be to make that change AND the .yml changes. Applied your suggestion with [8ea394e](https://github.com/newrelic/newrelic-ruby-agent/pull/2306/commits/8ea394e9da221cab142c3d6565daa86e852dae36).

If you want the unchanging code tested continually with the unchanging `grpc` gem version, let me know. Perhaps if so we could treat it like JRuby and only conditionally fire the CI on changes to specific content.

kaylareopelle:
> I really like this suggestion, but my preference would be to make that change AND the .yml changes. 
@fallwith - great point! It doesn't have to be an either/or situation. Thank you!

kaylareopelle (approved):

PR #2326, opened 2023-11-18
avoid compiling gRPC with Ruby 3.3

fallwith:
for now, don't test gRPC with Ruby 3.3 to prevent CI issues

CI cron run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/6911039829


kaylareopelle (approved):
Thanks for fixing this.

PR #2367, opened 2023-12-14
ViewComponent instrumentation

### LINKED ISSUE #2205: Add support for getting render timing information with view_component
### Feature Description
When viewing a response time chart in new relic you can see how much time the view takes to render, since view_components are is a very popular library used with rails, it would be beneficial to view_component users to see how much time is spent rendering the component just like they can do with their views

### Describe Alternatives
none that I know of(maybe something like server timing headers in the chrome devtools in dev mode)
### Additional context
none

### Priority
Really Want

hannahramadan:
Add New Relic instrumentation for [ViewComponent](https://viewcomponent.org/). 

Closes #2205


.rubocop_todo.yml
```diff
@@ -24,6 +24,11 @@ Metrics/AbcSize:
     - lib/new_relic/cli/commands/deployments.rb
     - test/**/*
 
+# Offense count: 1
+Metrics/CollectionLiteralLength:
+  Exclude:
+    - 'lib/new_relic/agent/configuration/default_source.rb'
```
hannahramadan, line 30: [Reminder](https://github.com/newrelic/newrelic-ruby-agent/pull/2369#discussion_r1428446469) to inline disable this cop before finalizing.

hannahramadan: Done.

lib/new_relic/agent/instrumentation/view_component/instrumentation.rb
```diff
@@ -0,0 +1,40 @@
...
+    def render_in_with_tracing(*args)
+      NewRelic::Agent.record_instrumentation_invocation(INSTRUMENTATION_NAME)
+
+      component = self.class.name
+      identifier = self.class.identifier
+
+      begin
+        segment = NewRelic::Agent::Tracer.start_segment(
+          name: metric_name(identifier, component)
+        )
```
kaylareopelle, line 18: To reduce object allocations at the cost of making things less readable, you could remove this variable assignment and access the component and identifier names in the method arguments: 
```suggestion
      begin
        segment = NewRelic::Agent::Tracer.start_segment(
          name: metric_name(self.class.name, self.class.identifier)
        )
```

lib/new_relic/agent/instrumentation/view_component/instrumentation.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    def metric_path(identifier)
+      if identifier.nil?
+        'component'
+      elsif identifier && (parts = identifier.split('/')).size > 1
+        parts[-2..-1].join('/') # Get filepath by assuming the Rails' structure: app/components/home/example_component.rb
+      else
+        NewRelic::Agent::UNKNOWN_METRIC
+      end
```
kaylareopelle, line 37: Another way to think about this:
```suggestion
      return 'component' if identifier.nil?

      if (parts = identifier.split('/')).size > 1
        parts[-2..-1].join('/') # Get filepath by assuming the Rails' structure: app/components/home/example_component.rb
      else
        NewRelic::Agent::UNKNOWN_METRIC
      end
```

It makes the conditions a bit simpler by ensuring identifier is not nil in the guard clause. You could also go a step further and use `return 'component' unless identifier` to just check for identifier's truthiness.

test/multiverse/suites/rails/rails3_app/my_app.rb
```diff
@@ -94,6 +94,8 @@ class MyApp < Rails::Application
 
     post '/parameter_capture', :to => 'parameter_capture#create'
 
+    get '/view_components', :to => 'view_component#index'
```
kaylareopelle, line 97: Maybe add a comment here about what this route is doing? I could see someone thinking it wasn't used and accidentally deleting it.

test/multiverse/suites/rails/view_instrumentation_test.rb
```diff
@@ -1,227 +0,0 @@
-# This file is distributed under New Relic's license terms.
```
kaylareopelle, line 1: Did you mean to delete this file?

test/multiverse/suites/view_component/Envfile
```diff
@@ -0,0 +1,20 @@
...
+
+instrumentation_methods :chain, :prepend
+
+VIEW_COMPONENT_VERSIONS = [
+  [nil, 2.7],
+  ['2.0.0', 2.4]
+]
+
+def gem_list(view_component_version = nil)
+  <<-RB
```
kaylareopelle, line 13: Why did you choose to use `<<-` over `<<~`? Do you want the leading whitespace?

test/multiverse/suites/view_component/instrumentation_test.rb
```diff
@@ -0,0 +1,40 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require_relative '../rails/app'
+
+class ExampleComponent < ViewComponent::Base
+  <<-ERB
```
kaylareopelle, line 8: Same question as in the Envfile, why use `<<-` instead of `<<~`?

kaylareopelle (commented):
This looks great overall. 🖼️ 🎉  

I know you're still working on things, but don't forget to add a CHANGELOG entry!  

I have a few other comments below 🙂

test/multiverse/suites/rails/view_instrumentation_test.rb
hannahramadan: Yes! This is when I was attempting to hookup this Rails app, before deciding that ViewComponent should be its own suite.

test/multiverse/suites/rails/rails3_app/my_app.rb
hannahramadan: Added! https://github.com/newrelic/newrelic-ruby-agent/pull/2367/commits/9fa2fb0d30b2ed490f571515083ba0da09a0cbe3

test/multiverse/suites/view_component/Envfile
hannahramadan: No good reason lol I think it might have been a copy & paste. I've updated https://github.com/newrelic/newrelic-ruby-agent/pull/2367/commits/9fa2fb0d30b2ed490f571515083ba0da09a0cbe3

test/multiverse/suites/rails/view_instrumentation_test.rb
kaylareopelle: Hmm... I think that might've been another file that copied this one? Since it's showing as deleted here, that means it predates this PR. I'm seeing some commits over 10 years old when I look at this file on `dev`.

hannahramadan: Oh no!! You're right that is a terrible mistake. I'll get that back in.

test/multiverse/suites/view_component/view_component_instrumentation_test.rb
```diff
@@ -0,0 +1,40 @@
...
+end
+
+class ViewComponentInstrumentationTest < ActionDispatch::IntegrationTest
+  include MultiverseHelpers
+  setup_and_teardown_agent
+
+  def test_metric_recorded
+    get('/view_components')
+
+    assert_metrics_recorded('View/view_component/view_component_instrumentation_test.rb/ExampleComponent')
```
fallwith, line 30: The metric name looked odd to me with the .rb file path, but it does match the convention we've long had in place for `View/` based metrics:

```shell
'View/model/index.html.erb/Rendering'
'View/model/_list.html.erb/Partial'
```

fallwith (dismissed):

lib/new_relic/agent/instrumentation/view_component/instrumentation.rb
```diff
@@ -0,0 +1,37 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic::Agent::Instrumentation
+  module ViewComponent
+    INSTRUMENTATION_NAME = NewRelic::Agent.base_name(name)
+
+    def render_in_with_tracing(*args)
```
tannalynn, line 9: I'm a bit confused, i'm probably missing something obvious (sorry!).
But this method doesn't seem to take a block or yield or anything, so i'm curious what the segment thats being created is measuring?

lib/new_relic/agent/instrumentation/view_component/instrumentation.rb
```diff
@@ -0,0 +1,37 @@
...
+      NewRelic::Agent.record_instrumentation_invocation(INSTRUMENTATION_NAME)
+
+      begin
+        segment = NewRelic::Agent::Tracer.start_segment(
+          name: metric_name(self.class.identifier, self.class.name)
+        )
```
tannalynn, line 15: Normally I would expect a yield after this line, which would call `super/render_in_without_tracing` from the do blocks.

hannahramadan: Great catch @tannalynn. This wasn't yielding and preventing views from showing up 😱 Added the yield, tested, and we get both the NR metric and app view now.

test/multiverse/suites/view_component/view_component_instrumentation_test.rb
hannahramadan: Agree! I don't love the look and thought it was a bit weird, but matches 🤷‍♀️ It is also how ViewComponent itself has chosen to do [their instrumentation](https://github.com/ViewComponent/view_component/blob/c8462c2113ca1771b681fc89eb025d8e8b685727/lib/view_component/instrumentation.rb#L16) (`self.class.identifier` is the full path including `.rb`)

tannalynn (approved):

tannalynn (approved):

lib/new_relic/agent/instrumentation/view_component/instrumentation.rb
```diff
@@ -0,0 +1,38 @@
...
+        yield
+      rescue => e
+        ::NewRelic::Agent.logger.debug('Error capturing ViewComponent segment', e)
```
mjacobus, line 18: Late to the party, but... is suppressing errors here the intended behavior? 

https://github.com/ViewComponent/view_component/issues/1981

Issue #2360, opened 2023-12-12
Elasticsearch: Instrumentation adds potentially dangerous `GET _cluster/health` request to every initial request


erikkessler1:
## Description

We recently updated a Rails service to a version of this agent that includes the Elasticsearch instrumentation added in #1525. We have a large Elasticsearch cluster backing the service, and when the service with the updated agent version was deployed, our Elasticsearch master nodes began to experience extreme CPU and network usage which caused the cluster to become unstable. 

When sampling the [hot threads][1] on the cluster we found the following trace sample indicating the node was spending a lot of time generating a `ClusterHealthResponse`:

> app//org.elasticsearch.action.admin.cluster.health.ClusterHealthResponse.<init>(ClusterHealthResponse.java:188)

After investigation, we found that this agent injects a [`GET _cluster/health`][2] request before the first request from the application in order to resolve the `cluster_name`:

https://github.com/newrelic/newrelic-ruby-agent/blob/713200b286d35fda33a64fc4a5b6d77ea52cd4ac/lib/new_relic/agent/instrumentation/elasticsearch/instrumentation.rb#L59

In a high throughput environment with many clients, these `_cluster/health` checks can overwhelm the single master that has to handle the requests and make requests to other nodes in the cluster to gather the status.

[1]: https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-hot-threads.html
[2]: https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html

## Expected Behavior

Adding instrumentation to our service should not cause the Elasticsearch cluster to become unstable. Given that `_cluster/health` _can_ be so expensive and disruptive in certain cases, it is dangerous to have the instrumentation silently making the request by default.

## Steps to Reproduce

The following script gives a demonstration of what is happening:

```ruby
c = Elasticsearch::Client.new(url: "...", log: true)
c.cluster.stats # or any other action
```

```
GET https://***/_cluster/health [status:200, request:0.033s, query:n/a] <- from the NewRelic agent
GET https://***/ [status:200, request:0.010s, query:n/a] <- from `Elasticsearch::Client#verify_elasticsearch`
GET https://***/_cluster/stats [status:200, request:0.028s, query:n/a] <- the real request
```

## Your Environment

* Ruby Version: 3.2.2
* Agent Version: 9.6.0
* Elasticsearch Version: 7.17.5

kaylareopelle:
[...]

PR #2369, opened 2023-12-15
Create :'elasticsearch.perform_health_check' config

kaylareopelle:
To access the cluster name to assign to the `database_name` attribute on the Elasticsearch datastore segment, we currently make a request to the `'_cluster/health'` endpoint. This can cause CPU and network spikes when large clusters start. 

To avoid this performance issue, we're introducing the `:'elasticsearch.perform_health_check'` config value. This value can be set to false to stop the agent from making requests to requests to `'_cluster/health'`. 

When this config is false, the `database_name` attribute on all Elasticsearch datastore segments will be 'Unknown'.

Relates to: #2360

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -313,6 +313,7 @@ module NewRelic
         'webpacker:compile'
       ].join(',').freeze
 
+      # rubocop:disable Metrics/CollectionLiteralLength
       DEFAULTS = {
         # Critical
         :agent_enabled => {
@@ -1324,6 +1325,13 @@ module NewRelic
           :type => Boolean,
           :allowed_from_server => true,
           :description => 'If `true`, the agent obfuscates Elasticsearch queries in transaction traces.'
+        },
+        :'elasticsearch.perform_health_check' => {
+          :default => true,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => "If true, the agent will query the cluster's `_cluster/health` endpoint to determine information about the cluster, including its name. When false, the database_name attribute for Elasticsearch datastore segements will be nil."
         },
         # Heroku
         :'heroku.use_dyno_names' => {
@@ -2405,6 +2413,7 @@ module NewRelic
           :description => 'This value represents the total amount of memory available to the host (not the process), in mebibytes (1024 squared or 1,048,576 bytes).'
         }
       }.freeze
+      # rubocop:enable Metrics/CollectionLiteralLength
     end
   end
 end
--- a/lib/new_relic/agent/instrumentation/elasticsearch/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/elasticsearch/instrumentation.rb
@@ -52,11 +52,13 @@ module NewRelic::Agent::Instrumentation
     end
 
     def nr_cluster_name
-      return @nr_cluster_name if @nr_cluster_name
+      return @nr_cluster_name if defined?(@nr_cluster_name)
       return if nr_hosts.empty?
 
       NewRelic::Agent.disable_all_tracing do
-        @nr_cluster_name ||= perform_request('GET', '_cluster/health').body['cluster_name']
+        @nr_cluster_name ||= if NewRelic::Agent.config[:'elasticsearch.perform_health_check']
+          perform_request('GET', '_cluster/health').body['cluster_name']
+        end
       end
     rescue StandardError => e
       NewRelic::Agent.logger.error('Failed to get cluster name for elasticsearch', e)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # New Relic Ruby Agent Release Notes
 
+## v9.7.0
+
+Version 9.7.0 adds a configuration value that toggles whether the agent will perform health check requests on Elasticsearch clusters.
+
+- **Feature: Add elasticsearch.perform_health_check configuration**
+
+  In [#2360](https://github.com/newrelic/newrelic-ruby-agent/pull/2360), it was brought to our attention that large clusters starting the agent can be overwhelmed by requests to `_cluster/health` performed by the agent to acquire the cluster name. The cluster name, once acquired, is assigned to the `database_name` attribute on Elasticsearch datastore segments. Now, users can disable these requests using a new configuration value. When set to `false`, the `database_name` will be `nil`. Thank you [@erikkessler1](https://github.com/erikkessler1) for bringing this to our attention.
+
+  | Configuration name          | Default | Behavior                                               |
+  | --------------------------- | ------- | ------------------------------------------------------ |
+  | `elasticsearch.perform_health_check` | `true` | If true, the agent will query the cluster's `_cluster/health` endpoint to determine information about the cluster, including its name. When false, the database_name attribute will be `nil``. |
+
 ## v9.6.0
 
 Version 9.6.0 adds instrumentation for Async::HTTP, Ethon, and HTTPX, adds the ability to ignore specific routes with Roda, gleans Docker container IDs from cgroups v2-based containers, records additional synthetics attributes, fixes an issue with Rails 7.1 that could cause duplicate log records to be sent to New Relic, fixes a deprecation warning for the Sidekiq error handler, adds additional attributes for OpenTelemetry compatibility, and resolves some technical debt, thanks to the community.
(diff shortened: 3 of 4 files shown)


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -313,6 +313,7 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
         'webpacker:compile'
       ].join(',').freeze
 
+      # rubocop:disable Metrics/CollectionLiteralLength
```
kaylareopelle, line 316: The `NewRelic::Agent::Configuration::DEFAULTS` hash now exceeds 250 lines, which violates the `Metrics/CollectionLiteralLength` rule. Moving `DEFAULTS` into a separate data storage file, like YAML or JSON would impact many other parts of our system, such as the generation of our documentation using HEREDOCs.

kaylareopelle: ^-- This is a summary of a discussion with @fallwith and @hannahramadan about temporarily disabling this rule. Should it be included in a source code comment?

lib/new_relic/agent/instrumentation/elasticsearch/instrumentation.rb
```diff
@@ -56,7 +56,11 @@ def nr_cluster_name
       return if nr_hosts.empty?
 
       NewRelic::Agent.disable_all_tracing do
-        @nr_cluster_name ||= perform_request('GET', '_cluster/health').body['cluster_name']
+        @nr_cluster_name ||= if NewRelic::Agent.config[:'elasticsearch.perform_health_check']
+          perform_request('GET', '_cluster/health').body['cluster_name']
+        else
+          NewRelic::UNKNOWN
```
kaylareopelle, line 62: Technically the `database_name` attribute on a datastore segment can be nil. Would that make more sense here than `Unknown`? I think it's closer to the API's expectations.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1325,6 +1326,13 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => true,
           :description => 'If `true`, the agent obfuscates Elasticsearch queries in transaction traces.'
         },
+        :'elasticsearch.perform_health_check' => {
+          :default => true,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => true,
```
tannalynn, line 1333: There is nothing on the server side that would update this config, so this should probably be false.

lib/new_relic/agent/instrumentation/elasticsearch/instrumentation.rb
tannalynn: Hmm I think one of the benefits to not using nil is that we won't have to check the config every single time. If we did set it to nil, then line 55 wouldn't return early anymore, so we'd have to do everything else every time and itd take a bit longer. Not sure if it would be a concerning increase though, or not.

lib/new_relic/agent/configuration/default_source.rb
tannalynn: It seems fine to me, but i'm good with it either way!

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle, line 1333:
```suggestion
          :allowed_from_server => false,
```

lib/new_relic/agent/instrumentation/elasticsearch/instrumentation.rb
kaylareopelle: That's right! Thanks for setting me straight :) 

We could add another guard clause between lines 54 and 55 to check for the config and return early if it's false, but that might still be less performant than just setting "Unknown".

fallwith: I think `'Unknown'` vs `nil` is worth discussing for what gets sent up in the db span. Originally I felt that `'Unknown'` would convey a failure to obtain the name or a deliberate short-circuit preventing the discovery, whereas `nil` would convey that the name is unset (given that Rails apps don't require the name in `database.yml`).

If `nil` is thought to be more appropriate, we can work around the performance implications. Perhaps `defined?` is faster than a hash lookup.

```diff
- @nr_cluster_name ||= if NewRelic::Agent.config[:'elasticsearch.perform_health_check']
+ return @nr_cluster_name if defined?(@nr_cluster_name)
+
+ @nr_cluster_name = if NewRelic::Agent.config[:'elasticsearch.perform_health_check']
```

lib/new_relic/agent/configuration/default_source.rb
fallwith: Reminder for @hannahramadan to https://github.com/newrelic/newrelic-ruby-agent/pull/2367/commits/6181fe1a91f742080b8513e21632d20456ebaf09 to match

fallwith (approved):

lib/new_relic/agent/instrumentation/elasticsearch/instrumentation.rb
kaylareopelle: `start_datastore_segment`'s default arguments set `database_name` to `nil`, which leads me to believe that's a more standard value than `Unknown`. 

The code change I alluded to slightly differs from what you've suggested above.
Using the method as it exists on `dev`, the change would be:
```diff
    def nr_cluster_name
+    return unless NewRelic::Agent.config[:'elasticsearch.perform_health_check']
      return @nr_cluster_name if @nr_cluster_name
      return if nr_hosts.empty?

      NewRelic::Agent.disable_all_tracing do
        @nr_cluster_name ||= perform_request('GET', '_cluster/health').body['cluster_name']
      end
    rescue StandardError => e
      NewRelic::Agent.logger.error('Failed to get cluster name for elasticsearch', e)
      nil
    end
```

So the `NewRelic::Agent.config` check would occur every time `#nr_cluster_name` is called, instead of when the `@nr_cluster_name` variable is defined/assigned. I don't know if there's a way to make that config check more performant.

fallwith: Ah, I was definitely thinking from the perspective of a change to config option only taking effect on a complete restart of the agent, so we'd only check the config hash one time and then either cache the string cluster name value or `nil`. And then use `defined?` for memoization help with `nil`. If you actually want to check the config freshly each time, then a hash lookup in a relatively small hash with symbols as keys shouldn't be too bad and the code as-is works great.

kaylareopelle: Ohh, I understand now. Thanks for clarifying! Your latest comment made me realize we don't need to check the config that often. It's reasonable to expect the config won't change during Elasticsearch client initialization. I was getting overly fixated on the test scenario where I needed to create a new client. Thank you!

gremerritt:
@kaylareopelle Out of curiosity, why are we using the `_cluster/health` endpoint to pull the cluster name? Am I reading it correctly that it's _only_ for pulling the cluster name, and does not actually check the health (ie. it is not using `wait_for_status`)? Would it be possible to use a safer endpoint, like [`_cluster/stats`](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-stats.html) (or [`/_info/_all`](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-info.html) on 8)?

kaylareopelle:
Hi @gremerritt, those are great suggestions! 

We considered trying out `_cluster/stats`, but hadn't noticed the `master:false` option. That seems like it'd be a great fit for the scenario in #2360. I see that you're from the same company as the reporter of #2360. 

If we set that up in a branch, would you be willing to test it in a staging environment to see how that endpoint impacts your cluster?

We'd probably prefer to use `_cluster/stats` over `_info/_all` to have fewer conditions related to versions of Elasticsearch code, but if `_info/_all` is more performant, we'd probably default to that for users of version 8.

gremerritt:
@kaylareopelle thanks for the update. Yes @erikkessler1 have been working together on this issue. Thanks for posting that branch, I think we'll have some bandwidth to give it a test.

From some more discussion on our team, it sounds like the "canonical" way to fetch the cluster name is by hitting the base `/` endpoint. For example, from the [ruby client](https://github.com/elastic/elasticsearch-ruby) this is just `client.info`. So I'm quite sure that you could use the following for an even simpler approach:
```ruby
@nr_cluster_name ||= perform_request('GET', '/').body['cluster_name']
```

We're happy to test out either. Thanks for digging into this!

kaylareopelle:
Hi @gremerritt! Thanks for your response! I appreciate the work you and your team are doing to help us find the right endpoint! 

Out of curiosity, what helped your team determine the base endpoint, `/`, was the canonical way to fetch the cluster name? 

I think it's a great next test! I've created #2377 on branch `test_root_for_cluster_name` using the snippet you suggested to access the cluster name. Let me know how the performance testing goes! I'll post a similar update on #2360.

kaylareopelle:
Closing in favor of #2377

PR #2374, opened 2023-12-19
Use _cluster/stats to access cluster name

kaylareopelle:
Unlike `_cluster/health`, this request does not call `wait_for_status`. With the addition of `master:false`, stats will not be generated for the master node.

See: https://github.com/newrelic/newrelic-ruby-agent/pull/2369#issuecomment-1860909721 

Relates to #2360


kaylareopelle:
Closing in favor of #2377

Issue #2235, opened 2023-09-24
Update stale public docs


kford-newrelic:
### Description
Timeboxed team activity to update possibly stale public documentation.

### Acceptance Criteria
Select public docs are reviewed as a team and updated if appropriate.

### Design Consideration/Limitations
N/A

### Dependencies
N/A

### Additional context
[Tracking spreadsheet](https://docs.google.com/spreadsheets/d/1waNtEyvfsYrX74WQaMK7Inry5weUzNkWvzKnwHLLWE0/edit?usp=sharing) for the docs to be reviewed. Update as appropriate when doc updates are made.

kaylareopelle:
In addition to the docs in the [newrelic/docs-website](https://github.com/newrelic/docs-website/) repo, we should also review the documentation in our repo, such as our README, CONTRIBUTING.md, etc.

kaylareopelle:
PR https://github.com/newrelic/docs-website/pull/15599 fixes many of the outdated documents in the tracking spreadsheet. 

Some notable remaining documents are:
* Ruby agent requirements and supported frameworks
* Record deployments (see #2373)
* Ruby custom instrumentation
* Install the Ruby agent, specifically enhanced instructions on installing with Docker

We also still need to analyze documents in the "Instrumented Gems" and "Troubleshooting" groups to identify outdated content. 

Documentation in our repo has also not been audited/updated.

kaylareopelle:
Other related PRs: 
* https://github.com/newrelic/newrelic-ruby-agent/pull/2375 
* https://github.com/newrelic/newrelic-ruby-agent/pull/2376

PR #2376, opened 2023-12-20
Update config doc opening

kaylareopelle:
This updates outdated content in the ERB template that generates the config documentation.

Relates to #2235 

These changes won't take place on [the docs website](https://docs.newrelic.com/docs/apm/agents/ruby-agent/configuration/ruby-agent-configuration/) until we release and the config_docs workflow runs.


lib/tasks/helpers/config.html.erb
```diff
@@ -42,9 +42,9 @@ In other words, environment variables override all other configuration settings
 
 ## View and edit config file options [#Edit]
 
-The Ruby agent's `newrelic.yml` is a standard YAML configuration file. It typically includes a `Defaults` section at the top, plus sections below for each application environment; for example, `Development`, `Testing`, and `Production`.
+The Ruby agent's `newrelic.yml` is a standard YAML configuration file. It typically includes a `Defaults` section at the top, plus sections below for each application environment (for example, `Development`, `Staging`, and `Production`).
```
hannahramadan, line 45: An improvement as-is! Since we're listing 3 of the 4 envs, thoughts on removing the 'for example' and calling them all out? 

(`Development`, `Test`, `Staging`, and `Production`)

hannahramadan (approved):

lib/tasks/helpers/config.html.erb
kaylareopelle: Great catch! Thank you! 
```suggestion
The Ruby agent's `newrelic.yml` is a standard YAML configuration file. It typically includes a `Defaults` section at the top, plus sections below for each application environment (`Development`, `Test`, `Staging`, and `Production`).
```

tannalynn (approved):

PR #2380, opened 2024-01-02
Rails edge can only run on ruby 3.1 and above

tannalynn:
Our nightly Ci is failing on 2.7 and 3.0 because of a bundler error with the rails edge env tests. 

```
Because every version of rails depends on Ruby >= 3.1.0
  and Gemfile depends on rails >= 0,
  Ruby >= 3.1.0 is required.
So, because current Ruby version is = 3.0.6,
  version solving has failed.
Could not find compatible versions
```

Since rails edge needs 3.1+, I've removed the rails edge tests from 2.7 and 3.0 in our ci_cron


kaylareopelle (approved):
Thanks for fixing this!

kaylareopelle (approved):
🎉

hannahramadan (approved):

Issue #2188, opened 2023-09-06
Add Support for Falcon


elucus:
Add support for the [Falcon](https://github.com/socketry/falcon) webserver

kford-newrelic:
Only adding a dispatcher initially. May want to investigate a deeper instrumentation later.

PR #2383, opened 2024-01-03
Falcon instrumentation

hannahramadan:
The agent will now report Falcon as a recognized webserver.

Closes [#2188](https://github.com/newrelic/newrelic-ruby-agent/issues/2188)

Note: This PR defers adding tests to a separate [ticket](https://github.com/newrelic/newrelic-ruby-agent/issues/2384).


lib/new_relic/local_environment.rb
```diff
@@ -138,6 +139,13 @@ def check_for_puma
       end
     end
 
+    def check_for_falcon
+      if defined?(::Falcon::Server) && NewRelic::LanguageSupport.object_space_usable?
+        v = find_class_in_object_space(::Falcon::Server)
```
hannahramadan, line 144: A similar object space check that we do for Unicorn and Thin.

tannalynn (commented):
Thanks for working on this! Could you add a changelog entry for this?

lib/new_relic/local_environment.rb
```diff
@@ -138,6 +139,13 @@ def check_for_puma
...
+        @discovered_dispatcher = :falcon if v
```
fallwith, line 145:
```suggestion
        @discovered_dispatcher = :falcon if find_class_in_object_space(::Falcon::Server)
```

OR

```ruby
        return unless find_class_in_object_space(::Falcon::Server)
        
        @discovered_dispatcher = :falcon
```

hannahramadan: Nice edit! I took your direct comitt suggestion https://github.com/newrelic/newrelic-ruby-agent/pull/2383/commits/5e2ff03db7c78f245013caff7bdcbea1f783a60d

lib/new_relic/agent/agent_helpers/special_startup.rb
```diff
@@ -10,7 +10,7 @@ module SpecialStartup
         # requests, we need to wait until the children are forked
         # before connecting, otherwise the parent process sends useless data
         def using_forking_dispatcher?
-          if [:puma, :passenger, :unicorn].include?(Agent.config[:dispatcher])
+          if [:puma, :passenger, :unicorn, :falcon].include?(Agent.config[:dispatcher])
```
fallwith, line 13: Falcon was added to this list because of its ability to fork. Falcon can also run in a threaded mode instead of a forking one, but given that:

> There is no guarantee that your process would be forked or threaded.

https://github.com/socketry/falcon/issues/53#issuecomment-461979683

It makes sense to always defer when Falcon is in play.

lib/new_relic/local_environment.rb
```diff
@@ -138,6 +139,12 @@ def check_for_puma
       end
     end
 
+    def check_for_falcon
+      if defined?(::Falcon::Server) && NewRelic::LanguageSupport.object_space_usable?
+        @discovered_dispatcher = :falcon if find_class_in_object_space(::Falcon::Server)
+      end
```
fallwith, line 145: An optional way to avoid having the entire method be a single `if` statement.

```suggestion
      return unless defined?(::Falcon::Server) && NewRelic::LanguageSupport.object_space_usable?
      
      @discovered_dispatcher = :falcon if find_class_in_object_space(::Falcon::Server)
```

fallwith (approved):

lib/new_relic/local_environment.rb
hannahramadan: I like this cleanup! All the different webserver patterns follow this check. I'll make a cleanup ticket and label 'good first issue' to remove these giant `if` blocks in a consistent way.

hannahramadan: New ticket! https://github.com/newrelic/newrelic-ruby-agent/issues/2388

Issue #2371, opened 2023-12-18
Remove the Ruby/Thread + other spans


tannalynn:
While our thread instrumentation has allowed customers to have visibility into the work happening inside of the threads, the actual spans created by the thread instrumentation seem to cause confusion frequently. 
We should update the agent to no longer create the spans for 'Ruby/Thread', 'Ruby/Fiber', and perhaps for the concurrent-ruby instrumentation as well. This change will NOT impact the agents ability to record instrumented code that is run inside of these threads, this will only prevent the specified spans from being created.

PR #2389, opened 2024-01-05
Remove Ruby/Thread and Ruby/Fiber spans

tannalynn:
To reduce confusing information being reported, the agent no longer creates the new spans for `Ruby/Thread` and `Ruby/Fiber`. The agent will still record anything happening inside of these threads/fibers correctly. The parents will still also be created correctly. We did have to make some adjustment to segment parent logic to accommodate the removal of these spans while maintaining consistently correct parent nesting. 

I decided to leave the concurrent ruby spans as is, since those are more likely to be useful information for customers compared to the ruby/thread spans.

resolves https://github.com/newrelic/newrelic-ruby-agent/issues/2371

### DIFF
--- a/lib/new_relic/agent/tracer.rb
+++ b/lib/new_relic/agent/tracer.rb
@@ -418,16 +418,17 @@ module NewRelic
           NewRelic::Agent.config[:'instrumentation.thread.tracing']
         end
 
-        def thread_block_with_current_transaction(segment_name:, parent: nil, &block)
+        def thread_block_with_current_transaction(segment_name: nil, parent: nil, &block)
           parent ||= current_segment
           current_txn = ::Thread.current[:newrelic_tracer_state]&.current_transaction if ::Thread.current[:newrelic_tracer_state]&.is_execution_traced?
           proc do |*args|
             begin
               if current_txn && !current_txn.finished?
                 NewRelic::Agent::Tracer.state.current_transaction = current_txn
+                ::Thread.current[:newrelic_thread_span_parent] = parent
                 current_txn.async = true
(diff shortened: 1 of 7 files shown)


lib/new_relic/agent/tracer.rb
```diff
@@ -418,16 +418,17 @@ def thread_tracing_enabled?
...
           current_txn = ::Thread.current[:newrelic_tracer_state]&.current_transaction if ::Thread.current[:newrelic_tracer_state]&.is_execution_traced?
           proc do |*args|
             begin
               if current_txn && !current_txn.finished?
                 NewRelic::Agent::Tracer.state.current_transaction = current_txn
+                ::Thread.current[:newrelic_thread_span_parent] = parent
                 current_txn.async = true
-                segment_name += "/Thread#{::Thread.current.object_id}/Fiber#{::Fiber.current.object_id}" if NewRelic::Agent.config[:'thread_ids_enabled']
-                segment = NewRelic::Agent::Tracer.start_segment(name: segment_name, parent: parent)
+                segment_name = segment_name.to_s + "/Thread#{::Thread.current.object_id}/Fiber#{::Fiber.current.object_id}" if NewRelic::Agent.config[:'thread_ids_enabled']
```
fallwith, line 430: We can save on spawning and concatenating 2 strings:

```suggestion
                segment_name = "#{segment_name}/Thread#{::Thread.current.object_id}/Fiber#{::Fiber.current.object_id}" if NewRelic::Agent.config[:'thread_ids_enabled']
```

tannalynn: oh yeah for sure, thank you. i didnt even think of that because it had started with the += lol

PR #2391, opened 2024-01-08
Skip Async-http instrumentation installation when Traces::Backend::NewRelic is defined

tannalynn:
If the traces-backend-newrelic instrumentation is being used, then skip async-http instrumentation installation since this gem provides instrumentation. 
https://github.com/newrelic/traces-backend-newrelic

### DIFF
--- a/lib/new_relic/agent/instrumentation/async_http.rb
+++ b/lib/new_relic/agent/instrumentation/async_http.rb
@@ -10,7 +10,9 @@ DependencyDetection.defer do
   named :async_http
 
   depends_on do
-    defined?(Async::HTTP) && Gem::Version.new(Async::HTTP::VERSION) >= Gem::Version.new('0.59.0')
+    defined?(Async::HTTP) &&
+      Gem::Version.new(Async::HTTP::VERSION) >= Gem::Version.new('0.59.0') &&
+      !defined?(Traces::Backend::NewRelic)
   end
 
   executes do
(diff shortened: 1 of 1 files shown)


fallwith (dismissed):

kaylareopelle (approved):
Nice! I have a few questions, and the answers can all be no 🙂 

* Should we add a changelog entry for this? It might be as close as we get to mentioning the new gem in the main agent's release notes.
* Would it be helpful to have a comment around line 15 describing where `Traces::Backend::NewRelic` is defined? I'm worried someone might have trouble finding it since it's not part of the async-http library.
* Do we want to add a test for this? We don't usually test depends_on blocks, so maybe the answer is no, I'm thinking about accidental regression.

tannalynn:
> Nice! I have a few questions, and the answers can all be no 🙂
> 
> * Should we add a changelog entry for this? It might be as close as we get to mentioning the new gem in the main agent's release notes.

I was thinking it would be better to wait until we've released a version of the traces-backend-newrelic before adding something for this to the changelog for whatever the next version of the agent is. 

> * Would it be helpful to have a comment around line 15 describing where `Traces::Backend::NewRelic` is defined? I'm worried someone might have trouble finding it since it's not part of the async-http library.

sure, added one

> * Do we want to add a test for this? We don't usually test depends_on blocks, so maybe the answer is no, I'm thinking about accidental regression.

I was leaning towards not, since we dont normally and it'd add the time of running a whole new multiverse suite. Plus the consequence for it failing is pretty minimal. Since it just won't create the second spans inside of the traces-backend-newrelic gem, but the ones the agent creates would still be present, there isn't a concern about double reporting the external requests.

kaylareopelle (approved):

PR #2398, opened 2024-01-18
OpenAI: instrument embeddings and chat completions

### LINKED ISSUE #2403: OpenAI: Instrument embeddings
Gem method: OpenAI::Client#embeddings
[OpenAI API Documentation: Embeddings](https://platform.openai.com/docs/guides/embeddings/embeddings
)
[OpenAI API Reference: Embeddings](https://platform.openai.com/docs/api-reference/embeddings)

- [ ] Create span when method is called if there’s a current transaction
- [ ] Span name: `Llm/embedding/OpenAI/create`
- [ ] Create custom event named `LlmEmbedding` [(example attributes)](https://newrelic.atlassian.net/wiki/spaces/INST/pages/3029139689/Node.js+OpenAI+Instrumentation+Change+Design+Document#LLM-Events)(internal)
- [ ] Generate MLOps metric when method is called: `Ruby/ML/OpenAI/<version>`

### LINKED ISSUE #2404: OpenAI: Instrument chat completions
Gem method: OpenAI::Client#chat
[OpenAI Chat Completions Documentation](https://platform.openai.com/docs/guides/text-generation/chat-completions-api) 
[OpenAI Chat Completions API Reference](https://platform.openai.com/docs/api-reference/chat)

- [x] Create span when method is called if there’s a current transaction; Span name: `Llm/completion/OpenAI/Create`
- [x] Generate metric when method is called: Metric name: `Ruby/ML/OpenAI/<version>`
- [x] Create custom event `LlmChatCompletionSummary` [(example attributes)](https://newrelic.atlassian.net/wiki/spaces/INST/pages/3029139689/Node.js+OpenAI+Instrumentation+Change+Design+Document#LLM-Events)(Internal)
- [x] Create custom event `LlmChatCompletionMessage` [(example attributes)](https://newrelic.atlassian.net/wiki/spaces/INST/pages/3029139689/Node.js+OpenAI+Instrumentation+Change+Design+Document#LLM-Events)(Internal)

### Notes:
- There’s a deprecated API called completions. That’s not what we’re talking about. We’re talking about “chat completions”, which is accessed in the Ruby gem through OpenAI::Client#chat
- Using the chat completions example linked above, the output should be one summary event and five message events. Four of the message events come from the request and one of the message events comes from the response.
- Summary events and messages events should be created as part of the same unit of work because they are both derived from the OpenAI::Client#chat method.

kaylareopelle:
Initial PR for OpenAI instrumentation. We're merging this into the `openai_instrumentation` branch to preserve one historical PR for our initial instrumentation rollout to hopefully make it easier for others to reference in the future.

Co-authored by @hannahramadan 
Closes #2403
Closes #2404 

This PR includes:
* Basic instrumentation setup for the [`ruby-openai` gem](https://rubygems.org/gems/ruby-openai)
  * Support for versions 3.4.0+, based on versions currently in use by our customers [(dashboard)(internal)](https://onenr.io/0gR7KyqgWRo)
  * Create of an `ai` multiverse runner group
  * Add `ai`/`ruby_openai` testing to the appropriate CI workflows
* Instrumentation for `OpenAI::Client#embeddings`:
  * Record a segment on every method invocation
  * Record a metric on every method invocation 
  * Create LlmEmbedding events
* Instrumentation for `OpenAI::Client#chat`, known in the spec as chat completions:
  * Record a segment on every method invocation
  * Record a metric on every method invocation
  * Create `LlmChatCompletionSummary` events
  * Create `LlmChatCompletionMessage` events
* Add attributes to `LlmEmbedding` and `LlmChatCompletionSummary` events from the `Net::HTTP` response headers  
* Set `llm: true` as an attribute on all transaction events with LLM-related segments 
* Test using mocked responses from real OpenAI requests

It also makes the following updates to the `Llm`-namespaced classes/modules:
* Update the `LlmEvent` and other `Llm` classes to use the attribute key strings expected by the UI
* Add an `llm_event` attribute to `AbstractSegment`
* Organize attribute names to reflect the order of the spec

This PR does not contain the following GA-required tasks:
* Changelog entry
* Tests for the `Net::HTTP` response header assignment to the LLM events
* Tests that validate the OpenAI instrumentation correctly assigns LLM event attribute values
* Error tracing attributes
* Feedback API
* The overall `ai_monitoring.enabled` configuration option

These items will be opened in separate PRs that we'll merge into the `openai_instrumentation` feature branch before GA.


lib/new_relic/agent/instrumentation/net_http/instrumentation.rb
```diff
@@ -32,12 +32,27 @@ def request_with_tracing(request)
             end
 
             wrapped_response = NewRelic::Agent::HTTPClients::NetHTTPResponse.new(response)
```
kaylareopelle, line 34: This is the POC for adding headers from the response; it's broken without the LLM events PR atm.

lib/new_relic/agent/instrumentation/net_http/instrumentation.rb
```diff
@@ -32,12 +32,28 @@ def request_with_tracing(request)
...
+          result
+        end
+
+        def add_llm_response_headers(response, parent)
+          return unless parent.instance_variable_defined?(:@llm_summary) # and maybe log a warning??
+
+          event = parent.instance_variable_get(:@llm_summary)
+          event.instance_variable_set(:@request_id, response['x-request-id'])
+          event.instance_variable_get(:@reponse_headers).populate_response_headers(response.to_hash)
+        end
```
kaylareopelle, line 56: This stuff all relies on the LLM events PR

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -0,0 +1,51 @@
...
+      elsif path == CHAT_COMPLETIONS_PATH
+        # chat_completions_instrumentation(parameters, headers)
+        segment = NewRelic::Agent::Tracer.start_segment(name: 'Llm/completion/OpenAI/create')
+        event = NewRelic::Agent::LlmEvent::ChatCompletion::Summary.new
+        segment.instance_variable_set(:@llm_summary, event)
+        begin
+          NewRelic::Agent::Tracer.capture_segment_error(segment) { yield }
+        ensure
+          segment&.finish
+          event.record
```
kaylareopelle, line 26: This stuff needs the LLM Events PR

lib/new_relic/agent/instrumentation/net_http/instrumentation.rb
```diff
@@ -32,12 +32,28 @@ def request_with_tracing(request)
...
             segment.process_response_headers(wrapped_response)
+
             response
           ensure
             segment&.finish
           end
         end
+
+        def openai_parent?(segment)
+          segment&.parent&.name&.match?(/Llm\/.*\/OpenAI\/.*/)
```
fallwith, line 49: We should freeze the pattern and keep reusing the same frozen object. We can also use `%r{}` to prevent having to escape the forward slashes.

```diff
- segment&.parent&.name&.match?(/Llm\/.*\/OpenAI\/.*/)

+ OPENAI_SEGMENT_PATTERN = %r{Llm/.*/OpenAI/.*}.freeze
+
+ segment&.parent&.name&.match?(OPENAI_SEGMENT_PATTERN)
```

lib/new_relic/agent/instrumentation/ruby_openai.rb
```diff
@@ -0,0 +1,37 @@
...
+  depends_on do
+    # add a config check for ai_monitoring.enabled
+    # maybe add DT check here eventually?
+    defined?(OpenAI) && defined?(OpenAI::Client) &&
+      OPENAI_VERSION >= Gem::Version.new('3.4.0')
+  end
+
+  executes do
+    if use_prepend?
+      if OPENAI_VERSION >= Gem::Version.new('5.0.0')
```
fallwith, line 23: Cool to see the support for older versions.

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -0,0 +1,175 @@
...
+    end
+
+    def add_embeddings_response_params(response, event)
+      event.response_model = response['model']
+      event.response_usage_total_tokens = response['usage']['total_tokens']
+      event.response_usage_prompt_tokens = response['usage']['prompt_tokens']
+    end
+
+    def parse_api_key
+      'sk-' + headers['Authorization'][-4..-1]
```
fallwith, line 106: Hmm... it looks like the ability to use `[-4..]` instead of `[-4..-1]` wasn't introduced until Ruby v2.6.

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -0,0 +1,175 @@
...
+
+    def nr_supportability_metric
+      @nr_supportability_metric ||= "Supportability/Ruby/ML/OpenAI/#{::OpenAI::VERSION}"
+    end
+
+    def finish(segment, event)
+      segment&.finish
+      event&.error = true if segment_noticed_error?(segment)
+      event&.duration = segment&.duration
+      event&.record
```
fallwith, line 172: Currently this code always calls `segment_noticed_error?`, even when there isn't an event (the `if` is calculated first, then the `&.`). I'm also wondering if `#record` should ever be called if neither `error=` nor `duration=` were called. Here's an attempt to address both issues:

```suggestion
      return unless segment
      
      segment.finish
      return unless event
      
      event.error = true if segment_noticed_error?(segment)
      event.duration = segment.duration
      event.record
```

lib/new_relic/agent/llm/chat_completion_summary.rb
```diff
@@ -24,6 +32,13 @@ def attributes
           LlmEvent::ATTRIBUTES + ChatCompletion::ATTRIBUTES + ResponseHeaders::ATTRIBUTES + ATTRIBUTES
         end
 
+        def attribute_name_exceptions
+          # TODO: OLD RUBIES < 2.6
+          # Hash#merge accepts multiple arguments in 2.6, so we can reduce this
+          # to a single Hash#merge call with two arguments at that point
```
fallwith, line 38: This is where better granular perf tests would help. If there's a significant performance gain to be had, it'd be worth writing the code both ways:

```ruby
if RUBY_VERSION >= 2.6
  new_way
else
  old way
end
```

lib/new_relic/agent/instrumentation/net_http/instrumentation.rb
hannahramadan: Learning moment! What is the benefit of freezing the pattern/creating a new object here if we're only doing this type of check once?

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -0,0 +1,175 @@
...
+      end
+    end
+
+    def create_chat_completion_summary(parameters)
+      NewRelic::Agent::Llm::ChatCompletionSummary.new(
+        # TODO: POST-GA: Add metadata from add_custom_attributes if prefixed with 'llm.', except conversation_id
+        vendor: VENDOR,
+        conversation_id: conversation_id,
+        api_key_last_four_digits: parse_api_key,
+        request_max_tokens: parameters[:max_tokens] || parameters['max_tokens'],
```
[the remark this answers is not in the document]
hannahramadan: Both Strings and Symbols are accepted as keys, so it's all up to the user on which they use. And yes—both types can be mix and matched in the same request. 

Side note: We did performance testing of `||` vs `kind_of?` vs `is_a?` and found `||` to be the most performant by about 1.2x, which is why we went with an "or" check vs creating a helper method that would decide which to use.

lib/new_relic/agent/llm/chat_completion_summary.rb
hannahramadan: Great call out! Ran a perf test a few times and `item.merge(item1).merge(item2)` is consistently ~1.25x slower. We will break this out into a conditional.

hannahramadan: Done! I wanted to use a ternary operator but it got super long https://github.com/newrelic/newrelic-ruby-agent/pull/2398/commits/36d6cab921169da77ee3bccd2e83fd0da3ea728c

lib/new_relic/agent/instrumentation/net_http/instrumentation.rb
fallwith: @hannahramadan From what I can tell, the `openai_parent?` method gets invoked once for every single wrapped `Net::HTTP` call made to check to see if OpenAI was involved or not. A Ruby server running in a loop could potentially perform thousands of `Net::HTTP` calls - especially for non AI use cases - and every call would cause Ruby to spawn a fresh object to hold the same regex pattern over and over again.

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
kaylareopelle: @fallwith, thank you for these suggestions! I think the intended behavior is a little different than your description. 

We still want to record an event if a segment is not present. The `error` and `duration` attributes are optional, and a valid event can be created without them. 

I'm not sure how likely it is to have an event without a segment, but I'd like to protect for that case regardless since they are instances of different event types that can be sent independently of each other.

Given this additional context, what are your thoughts about the following refactor?
```suggestion
      segment&.finish

      return unless event

      if segment
        event.error = true if segment_noticed_error?(segment)
        event.duration = segment.duration
      end

      event.record
```

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -0,0 +1,175 @@
...
+
+    def create_chat_completion_messages(parameters, summary_id)
+      (parameters[:messages] || parameters['messages']).map.with_index do |message, index|
+        NewRelic::Agent::Llm::ChatCompletionMessage.new(
+          content: message[:content] || message['content'],
+          role: message[:role] || message['role'],
+          sequence: index,
+          completion_id: summary_id,
+          vendor: VENDOR,
+          is_response: false
```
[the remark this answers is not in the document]
kaylareopelle: Great callout! Thank you!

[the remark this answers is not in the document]
hannahramadan: This has been updated! https://github.com/newrelic/newrelic-ruby-agent/pull/2398/commits/cdd42f68793a70501bd4b740caffd2f141793220

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
fallwith: Given your explanation, this code looks good, yep!

lib/new_relic/agent/instrumentation/net_http/instrumentation.rb
hannahramadan: Thank you @fallwith! That makes a lot of sense. I've updated this: https://github.com/newrelic/newrelic-ruby-agent/pull/2398/commits/d43b1a6bb130f181b3b7a05c41d04b3cb219afb3

fallwith (approved):

tannalynn (approved):

PR #2410, opened 2024-01-23
Don't suppress view component rendering errors

mjacobus:

# Overview

Fixes a bug introduced in https://github.com/newrelic/newrelic-ruby-agent/pull/2367, where the render_in behavior was changed, suppressing errors.

We should be able to se error pages, specially in when Rails.env is development. The piece of code that was removed here prevented from happening. Instead, it would have the `render_in` method return `nil` when an error happened, resulting in either a blank page, or an incomplete page.

I believe we also want to raise errors in production. We should not make a decision here, whether the error should be suppressed. Otherwise we are making the assumption "incomplete pages are better than error pages", which may be the correct assumption under some circumstances, but not under other circumstances.

Fixes https://github.com/ViewComponent/view_component/issues/1981

Submitter Checklist:

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label

### DIFF
--- a/lib/new_relic/agent/instrumentation/view_component/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/view_component/instrumentation.rb
@@ -15,7 +15,8 @@ module NewRelic::Agent::Instrumentation
         )
         yield
       rescue => e
-        ::NewRelic::Agent.logger.debug('Error capturing ViewComponent segment', e)
+        NewRelic::Agent.notice_error(e)
+        raise
       ensure
         segment&.finish
       end


lib/new_relic/agent/instrumentation/view_component/instrumentation.rb
```diff
@@ -14,8 +14,6 @@ def render_in_with_tracing(*args)
           name: metric_name(self.class.identifier, self.class.name)
         )
         yield
-      rescue => e
-        ::NewRelic::Agent.logger.debug('Error capturing ViewComponent segment', e)
```
hannahramadan, line 18: Hi @mjacobus! Thank you for catching this—we shoudn't be swallowing errors! But we do want to capture and report them to New Relic. Could you update your code to the following?

```ruby
rescue => e
  NewRelic::Agent.notice_error(e)
  raise
```

Then we can get this merged ◡̈

hannahramadan (changes requested):

lib/new_relic/agent/instrumentation/view_component/instrumentation.rb
mjacobus: Sure thing! There you go. Thank you @hannahramadan

hannahramadan (approved):

hannahramadan:
Thank you @mjacobus! We will followup once these changes are released.

hannahramadan:
Hi @mjacobus—Ruby agent 9.7.1 has been released and includes this fix. Thanks again for your help on this!

PR #2412, opened 2024-01-23
Update thor scaffold command

tannalynn:
Fixes the insert point for the default_source.rb, since it was no longer working after updates we made to default_source.rm
Also, I fixed the extra line that we've had in the chain instrumentation. I always remove it, and I know we've discussed how that line is redundant in other PRs that used the scaffold, so I updated it. (The method gets redefined immediately after that line.)


kaylareopelle (approved):
Thank you!

fallwith (approved):

PR #2414, opened 2024-01-23
[WIP] Create LLM Events

### LINKED ISSUE #2408: OpenAI: Create LLM Events classes
Create new classes to hold the LLM events that will be generated for all LLM instrumentation (OpenAI and Bedrock)

One approach would be to create one base class and subclasses for each specific LLM event type. These events would get generated in the instrumentation instead of calling record_custom_event directly. If we need to move away from recording custom events in the future, this will make it easier 

Llm events:
- LlmEmbedding
- LlmChatCompletionSummary
- LlmChatCompletionMessage
- LlmChatCompletionFeedback

kaylareopelle:
We need to create events with specific attributes regardless of what library is instrumented.

These classes are one strategy to solve this problem. There might be a better way to do this.

To reduce duplication, some attributes are passed down from parents/grandparents to child classes, ex. LlmEvents::ChatCompletion::Message has some of its attributes defined in LlmEvents, ChatCompletion and Messages. They all get merged in an attributes method in the child class, which in this example would be messages. 

The child classes have their own record methods. These methods are intended to create the custom events. The record method is defined in parent/grandparent classes, but it is empty. Following patterns elsewhere in the agent, we don't raise an error if a child class hasn't defined the record method.

One way to use these classes: 
```
event = NewRelic::Agent::LlmEvents::ChatCompletion::Summary.new(...pass attributes)
event.instance_variable_set(:@attribute, value_not_available_at_creation)
event.record
```

Closes #2408 

co-authored-by: @hannahramadan

### DIFF
--- a/lib/new_relic/agent/llm_event.rb
+++ b/lib/new_relic/agent/llm_event.rb
@@ -29,7 +29,7 @@ module NewRelic
 
       def llm_event_attributes
         {id: @id, app_name: @app_name, request_id: @request_id, span_id: @span_id, transaction_id: @transaction_id,
-        trace_id: @trace_id, response_model: @response_model, vendor: @vendor }
+         trace_id: @trace_id, response_model: @response_model, vendor: @vendor}
       end
 
       # Method for subclasses to override
--- a/lib/new_relic/agent/llm_event.rb
+++ b/lib/new_relic/agent/llm_event.rb
@@ -3,18 +3,21 @@
 # frozen_string_literal: true
 
 require_relative 'llm_event/chat_completion'
+# require_relative 'llm_event/embedding'
+# require_relative 'llm_event/feedback'
+require_relative 'llm_event/chat_completion/message'
+# require_relative 'llm_event/chat_completion/summary'
 
 module NewRelic
   module Agent
     class LlmEvent
-
       # response_model looks like repsonse.model
       attr_accessor :id, :app_name, :request_id, :span_id, :transaction_id, :trace_id, :response_model, :vendor
       INGEST_SOURCE = 'Ruby'
 
-      def initialize(id:, app_name:, request_id:, span_id:, transaction_id:, trace_id:, response_model:, vendor:, **args)
+      def initialize(id: nil, request_id: nil, span_id: nil, transaction_id: nil, trace_id: nil, response_model: nil, vendor: nil, **args)
         @id = id
-        @app_name = app_name
+        @app_name = NewRelic::Agent.config[:app_name]
         @request_id = request_id
         @span_id = span_id
         @transaction_id = transaction_id
@@ -23,8 +26,13 @@ module NewRelic
         @vendor = vendor
       end
 
+      def llm_event_attributes
+        {id: @id, app_name: @app_name}
+      end
+
       # Method for subclasses to override
-      def record; end
+      def record
+      end
     end
   end
 end
--- a/lib/new_relic/agent/llm_event.rb
+++ b/lib/new_relic/agent/llm_event.rb
@@ -14,7 +14,7 @@ module NewRelic
       # response_model looks like repsonse.model
       attr_accessor :id, :app_name, :request_id, :span_id, :transaction_id, :trace_id, :response_model, :vendor
       INGEST_SOURCE = 'Ruby'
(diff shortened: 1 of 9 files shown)


lib/new_relic/agent/llm_event.rb
```diff
@@ -0,0 +1,39 @@
...
+require_relative 'llm_event/feedback'
+require_relative 'llm_event/response_headers'
+
+module NewRelic
+  module Agent
+    class LlmEvent
+      # response_model looks like repsonse.model
+      INGEST_SOURCE = 'Ruby'
+
+      def initialize(id: nil, request_id: nil, span_id: nil, transaction_id: nil, trace_id: nil, response_model: nil, vendor: nil, **args)
```
fallwith, line 18: Is the double splat intended to catch something in the future? If not, I'm thinking it can be removed or perhaps the method input can be changed to simply be a hash instead of named arguments.

lib/new_relic/agent/llm_event/chat_completion.rb
```diff
@@ -0,0 +1,28 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    class LlmEvent
+      class ChatCompletion < LlmEvent
+        # TODO: should any of the attrs be required on initialization?
```
tannalynn, line 9: since there are so many attributes, i'm down for not having any be required immediately, and just letting things be settable, so we could create the object first and then add attributes to it as we go? Maybe adding a []= method or something would be nice so we could just be like `event[:conversation_id] = 2`, or something like that? I think something like that would make the creation of all the attributes a little nicer to do than just having to provide everything in the initializer.
idk what does everyone think about something like that?

lib/new_relic/agent/llm_event/chat_completion.rb
```diff
@@ -0,0 +1,28 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    class LlmEvent
+      class ChatCompletion < LlmEvent
```
fallwith, line 8: Is `ChatCompletion` meant to be nested inside the `LlmEvent` class AND inherit from it as well?

lib/new_relic/agent/llm_event/chat_completion/message.rb
```diff
@@ -0,0 +1,32 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    class LlmEvent
+      class ChatCompletion
+        class Message < ChatCompletion
```
fallwith, line 9: I think I prefer or perhaps am simply accustomed to module namespacing instead of class namespacing. What does the `class class class` give us here?

lib/new_relic/agent/llm_event.rb
```diff
@@ -0,0 +1,39 @@
...
+        @id = id
+        @request_id = request_id
+        @span_id = span_id
+        @transaction_id = transaction_id
```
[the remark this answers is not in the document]
kaylareopelle: 💯 

thanks for catching this. meant to pull that out of the assignment and forgot! trace_id can probably be accessed that way too.

lib/new_relic/agent/llm_event.rb
```diff
@@ -0,0 +1,39 @@
...
+require_relative 'llm_event/embedding'
+require_relative 'llm_event/feedback'
+require_relative 'llm_event/response_headers'
+
+module NewRelic
+  module Agent
+    class LlmEvent
+      # response_model looks like repsonse.model
+      INGEST_SOURCE = 'Ruby'
+
```
kaylareopelle, line 17: does **args work below 2.7

kaylareopelle:
Closing in favor of #2415

PR #2415, opened 2024-01-25
Create LLM event structure (team refactor)

### LINKED ISSUE #2408: OpenAI: Create LLM Events classes

kaylareopelle:
All the LLM event classes are nested within the `NewRelic::Agent::Llm` module. See the notes in `NewRelic::Agent::Llm::LlmEvent` for use.

Closes #2408 
Supersedes #2414

### DIFF
--- a/lib/new_relic/agent/llm/chat_completion.rb
+++ b/lib/new_relic/agent/llm/chat_completion.rb
@@ -1,0 +1,15 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    module Llm
+      module ChatCompletion
+        ATTRIBUTES = %i[conversation_id]
+
+        attr_accessor(*ATTRIBUTES)
+      end
+    end
+  end
+end
--- a/lib/new_relic/agent/llm/chat_completion_message.rb
+++ b/lib/new_relic/agent/llm/chat_completion_message.rb
@@ -17,8 +17,8 @@ module NewRelic
           LlmEvent::ATTRIBUTES + ChatCompletion::ATTRIBUTES + ATTRIBUTES
         end
 
-        def record
-          NewRelic::Agent.record_custom_event(EVENT_NAME, event_attributes)
+        def event_name
+          EVENT_NAME
         end
       end
     end
--- a/lib/new_relic/agent/llm/chat_completion_message.rb
+++ b/lib/new_relic/agent/llm/chat_completion_message.rb
@@ -1,0 +1,26 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    module Llm
+      class ChatCompletionMessage < LlmEvent
+        include ChatCompletion
+
+        ATTRIBUTES = %i[content role sequence completion_id is_response]
+        EVENT_NAME = 'LlmChatCompletionMessage'
+
+        attr_accessor(*ATTRIBUTES)
+
+        def attributes
+          LlmEvent::ATTRIBUTES + ChatCompletion::ATTRIBUTES + ATTRIBUTES
+        end
+
+        def record
+          NewRelic::Agent.record_custom_event(EVENT_NAME, event_attributes)
+        end
+      end
+    end
+  end
+end
--- a/lib/new_relic/agent/llm/chat_completion_summary.rb
+++ b/lib/new_relic/agent/llm/chat_completion_summary.rb
@@ -1,0 +1,33 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require_relative 'response_headers'
+
+module NewRelic
+  module Agent
+    module Llm
+      class ChatCompletionSummary < LlmEvent
+        include ChatCompletion
+        include ResponseHeaders
+
+        ATTRIBUTES = %i[api_key_last_four_digits response_number_of_messages
+          response_usage_completion_tokens response_choices_finish_reason
+          request_temperature request_max_tokens request_model
+          response_organization response_usage_total_tokens
+          response_usage_prompt_tokens duration error]
+        EVENT_NAME = 'LlmChatCompletionSummary'
+
+        attr_accessor(*ATTRIBUTES)
+
+        def attributes
+          LlmEvent::ATTRIBUTES + ChatCompletion::ATTRIBUTES + ResponseHeaders::ATTRIBUTES + ATTRIBUTES
+        end
+
+        def record
+          NewRelic::Agent.record_custom_event(EVENT_NAME, event_attributes)
+        end
+      end
+    end
+  end
+end
--- a/lib/new_relic/agent/llm/chat_completion_summary.rb
+++ b/lib/new_relic/agent/llm/chat_completion_summary.rb
@@ -11,11 +11,11 @@ module NewRelic
         include ChatCompletion
         include ResponseHeaders
 
-        ATTRIBUTES = %i[api_key_last_four_digits response_number_of_messages
+        ATTRIBUTES = %i[api_key_last_four_digits request_max_tokens
+          response_number_of_messages request_model response_organization
+          response_usage_total_tokens response_usage_prompt_tokens
           response_usage_completion_tokens response_choices_finish_reason
-          request_temperature request_max_tokens request_model
-          response_organization response_usage_total_tokens
-          response_usage_prompt_tokens duration error]
+          request_temperature duration error]
         EVENT_NAME = 'LlmChatCompletionSummary'
 
         attr_accessor(*ATTRIBUTES)
@@ -24,8 +24,8 @@ module NewRelic
           LlmEvent::ATTRIBUTES + ChatCompletion::ATTRIBUTES + ResponseHeaders::ATTRIBUTES + ATTRIBUTES
         end
 
-        def record
-          NewRelic::Agent.record_custom_event(EVENT_NAME, event_attributes)
+        def event_name
+          EVENT_NAME
         end
       end
     end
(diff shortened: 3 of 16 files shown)


lib/new_relic/agent/llm/response_headers.rb
```diff
@@ -0,0 +1,28 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    module Llm
+      module ResponseHeaders
```
kaylareopelle, line 8: Since we don't necessarily need to generate a ResponseHeaders event, I thought making this a module would be a good way to keep the unique attributes while also allowing unique methods to be defined to set values.

lib/new_relic/agent/llm/llm_event.rb
```diff
@@ -0,0 +1,41 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    module Llm
+      class LlmEvent
```
kaylareopelle, line 8: I don't know if this is the best name for this class. `Default` or `Event` also come to mind. However, I'm not opposed to keeping `LlmEvent` as-is.

lib/new_relic/agent/llm/response_headers.rb
```diff
@@ -0,0 +1,28 @@
...
+
+        # Headers is a hash of Net::HTTP response headers
+        def populate_openai_response_headers(headers)
+          llm_version = headers['openai-version'][0]
+          rate_limit_requests = headers['x-ratelimit-limit-requests'][0]
+          rate_limit_tokens = headers['x-ratelimit-limit-tokens'][0]
+          rate_limit_reset_requests = headers['x-ratelimit-reset-requests'][0]
+          rate_limit_reset_tokens = headers['x-ratelimit-reset-tokens'][0]
+          rate_limit_remaining_requests = headers['x-ratelimit-remaining-requests'][0]
+          rate_limit_remaining_tokens = headers['x-ratelimit-remaining-tokens'][0]
```
kaylareopelle, line 23: This throws assigned, but unused variable warnings. I'll take a look tomorrow afternoon. We might need to go back to using `@llm_version`.

test/new_relic/agent/llm/chat_completion_message_test.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    def test_args_passed_to_init_not_set_as_instance_vars_when_not_in_attributes_constant
+      event = NewRelic::Agent::Llm::ChatCompletionMessage.new(fake: 'fake')
+
+      refute_includes event.attributes, :fake
+      refute event.instance_variable_defined?(:@fake)
+    end
+
+    def test_record_creates_an_event
```
kaylareopelle, line 53: Tomorrow I'll update the `test_record_creates_an_event` tests include values for every attributes on the event to make sure they're defined

lib/new_relic/agent/llm/chat_completion.rb
```diff
@@ -0,0 +1,15 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    module Llm
+      module ChatCompletion
```
kaylareopelle, line 8: After taking another look at the attributes list, it seems that fewer unique attributes are shared among ChatCompletionSummary and ChatCompletionMessage than we originally thought. Since there's only one, I thought this would be better suited as a module, and then we can simplify the inheritance for the classes that actually create events.

lib/new_relic/agent/llm/chat_completion_summary.rb
```diff
@@ -0,0 +1,33 @@
...
+    module Llm
+      class ChatCompletionSummary < LlmEvent
+        include ChatCompletion
+        include ResponseHeaders
+
+        ATTRIBUTES = %i[api_key_last_four_digits response_number_of_messages
+          response_usage_completion_tokens response_choices_finish_reason
+          request_temperature request_max_tokens request_model
+          response_organization response_usage_total_tokens
+          response_usage_prompt_tokens duration error]
```
kaylareopelle, line 18: A lot of these attributes are also shared with Embedding. I think we should break them out into a shared module, but I'm not sure about the name of that module or whether it should be combined with the ResponseHeaders module.

tannalynn (approved):

fallwith (approved):

lib/new_relic/agent/llm/chat_completion.rb
hannahramadan: Ah! I didn't realize so few were shared. I like this move.

lib/new_relic/agent/llm/llm_event.rb
```diff
@@ -0,0 +1,55 @@
...
+          end
+        end
+
+        # Subclasses define an attributes method to concatenate attributes
+        # defined across their ancestors and other modules
+        def attributes
+          ATTRIBUTES
+        end
+
+        # Subclasses that record events will override this method
```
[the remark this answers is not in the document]
kaylareopelle: Great point! I'll update

lib/new_relic/agent/llm/llm_event.rb
hannahramadan: I don't mind `LlmEvent`, but I don't know if thats just because I've seen it a lot. I like that in a way, the longer name seems more 'grand' and as the parent of all the other events I notice it more.

lib/new_relic/agent/llm/chat_completion_message.rb
```diff
@@ -0,0 +1,26 @@
...
+module NewRelic
+  module Agent
+    module Llm
+      class ChatCompletionMessage < LlmEvent
+        include ChatCompletion
+
+        ATTRIBUTES = %i[content role sequence completion_id is_response]
+        EVENT_NAME = 'LlmChatCompletionMessage'
+
+        attr_accessor(*ATTRIBUTES)
```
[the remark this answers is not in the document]
fallwith: Ah ok. I myself always put the space after the asterisk for multiplication (ex: `'hey' * 3`) and don't encounter the ambiguity, but I respect the intent of the cop. Let's keep the parens.

lib/new_relic/agent/llm/response_headers.rb
```diff
@@ -0,0 +1,28 @@
...
+
+        # Headers is a hash of Net::HTTP response headers
+        def populate_openai_response_headers(headers)
+          self.llm_version = headers['openai-version']&.first
+          self.rate_limit_requests = headers['x-ratelimit-limit-requests']&.first
+          self.rate_limit_tokens = headers['x-ratelimit-limit-tokens']&.first
+          self.rate_limit_remaining_requests = headers['x-ratelimit-remaining-requests']&.first
+          self.rate_limit_remaining_tokens = headers['x-ratelimit-remaining-tokens']&.first
+          self.rate_limit_reset_requests = headers['x-ratelimit-reset-requests']&.first
+          self.rate_limit_reset_tokens = headers['x-ratelimit-reset-tokens']&.first
```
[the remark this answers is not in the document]
kaylareopelle: 1. Without the `self.` prefix, a warning is printed to the console that the variable `llm_version` is defined but never used. This fails one of our unit tests. Alternatively, I could replace `self` with `@`.

2. Good point, I'll adjust

hannahramadan (dismissed):

lib/new_relic/agent/llm/response_headers.rb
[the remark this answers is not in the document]
fallwith: For #1, sorry about that! Of course, without the `self` or `@` you'd be assigning a value to a local variable instead of lighting up the setter. Got it.

lib/new_relic/agent/llm/llm_event.rb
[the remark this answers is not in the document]
kaylareopelle: Updated in: 2f40a7bf89cb36da1bbae664151960ab818f048d

lib/new_relic/agent/llm/response_headers.rb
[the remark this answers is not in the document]
kaylareopelle: Updated in: 2f40a7bf89cb36da1bbae664151960ab818f048d

lib/new_relic/agent/llm/chat_completion_message.rb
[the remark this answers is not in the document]
kaylareopelle: Copy that! Thanks!

lib/new_relic/agent/llm/chat_completion_message.rb
```diff
@@ -0,0 +1,26 @@
...
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    module Llm
+      class ChatCompletionMessage < LlmEvent
+        include ChatCompletion
+
+        ATTRIBUTES = %i[content role sequence completion_id is_response]
```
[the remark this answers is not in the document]
kaylareopelle: They were close to matching the docs before, but now they match them specifically: 2f40a7bf89cb36da1bbae664151960ab818f048d 

At least during this initial development phase, I'd like to keep them matching the documentation order to make things easier to check off lists. I'd be down to move to alphabetical later on.

lib/new_relic/agent/llm/llm_event.rb
kaylareopelle: Thanks! Let's leave it for now!

tannalynn (approved):

fallwith (approved):

PR #2420, opened 2024-01-30
Update SQL obfuscation tests to include cases for Active Record Query Logs

kaylareopelle:
# Overview
SQL Obfuscation work in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/529 revealed we may have a gap in testing for queries prepended by comments in the style of [Marginalia](https://github.com/basecamp/marginalia) (pre-Rails 7) and [Active Record Query Logs](https://api.rubyonrails.org/classes/ActiveRecord/QueryLogs.html) (Rails 7+). 

This PR adds tests for those cases. If we choose to merge this PR, we'll need to follow-up with an update to the cross-agent tests, since these fixtures are synced with that repository. If we don't want to go that route, I'll move these tests so they're separate from the cross-agent tests.

This type of test may have cross-agent value because Google's [sqlcommenter](https://google.github.io/sqlcommenter/) tool supports many different languages and databases and uses a similar syntax.

### DIFF
--- a/test/fixtures/cross_agent_tests/sql_obfuscation/sql_obfuscation.json
+++ b/test/fixtures/cross_agent_tests/sql_obfuscation/sql_obfuscation.json
@@ -674,5 +674,35 @@
       "cassandra",
       "mssql"
     ]
+  },
+  {
+    "name": "prepended_comments_with_quotes.postgres",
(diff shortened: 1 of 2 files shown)


fallwith (approved):

tannalynn (approved):

kaylareopelle:
I'm going to go ahead and merge this. [PR to add this to our cross agent tests (internal)](https://source.datanerd.us/agents/cross_agent_tests/pull/155) to avoid these new cases from being accidentally overwritten.

PR #2424, opened 2024-02-02
Update attribute names to period-delimited strings

kaylareopelle:
Some attribute names for LLM events are expected to include periods. We can't use names with periods in setter/getter method names. To resolve this, a new `ATTRIBUTE_NAME_EXCEPTIONS` constant, which is accompanied by an `attribute_name_exceptions` method exists to connect the string value of the attribute names expected upstream with our Ruby-metaprogramming-friendly name.

Then, `replace_attr_with_string` returns the correct string name when we're creating the final event_attributes hash.

### DIFF
--- a/lib/new_relic/agent/llm/llm_event.rb
+++ b/lib/new_relic/agent/llm/llm_event.rb
@@ -12,6 +12,8 @@ module NewRelic
           trace_id response_model vendor ingest_source]
         # These attributes should not be passed as arguments to initialize and will be set by the agent
         AGENT_DEFINED_ATTRIBUTES = %i[span_id transaction_id trace_id ingest_source]
+        ATTRIBUTE_NAME_EXCEPTIONS = {response_model: 'response.model'}
+
         INGEST_SOURCE = 'Ruby'
 
         attr_accessor(*ATTRIBUTES)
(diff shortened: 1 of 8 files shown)


hannahramadan (approved):

lib/new_relic/agent/llm/llm_event.rb
```diff
@@ -36,7 +38,7 @@ def initialize(opts = {})
         # attributes and their values
         def event_attributes
           attributes.each_with_object({}) do |attr, hash|
-            hash[attr] = instance_variable_get(:"@#{attr}")
+            hash[replace_attr_with_string(attr)] = instance_variable_get(:"@#{attr}")
```
hannahramadan, line 41: Nifty way to do this!

test/new_relic/agent/llm/chat_completion_summary_test.rb
```diff
@@ -64,7 +64,6 @@ def test_record_creates_an_event
         summary.response_number_of_messages = 5
         summary.request_model = 'gpt-4-turbo-preview'
         summary.response_model = 'gpt-4'
-        summary.response_organization = '98338'
```
hannahramadan, line 67: No longer needed attribute?

kaylareopelle: It wasn't when I first wrote this code, but now it's back in 🙃 
The attribute is present in our `openai` branch, so if you're comfortable leaving this as-is for now, I'd prefer to go that route.

kaylareopelle:
Closing, as this is included in #2398

PR #2428, opened 2024-02-07
Skip flaky scoped metric reporting test

kaylareopelle:
This test has failed on Ruby versions ranging from 2.5 to 3.2 (perhaps more, but I didn't look too deeply into this). Skip the test until we can figure out why it's broken.

#2429 is an issue to fix this test when we can get around to it.


hannahramadan (approved):
thank you!!!

tannalynn (approved):

PR #2442, opened 2024-02-14
AIM Instrumentation - OpenAI

kaylareopelle:
Introduces OpenAI instrumentation for the [ruby-openai](https://github.com/alexrudall/ruby-openai) gem versions 3.4.0 and above. 

OpenAI instrumentation:

**Provides metrics on chat summary/message events and embedding events**

**Introduces two APIs**
- `NewRelic::Agent.record_llm_feedback_event` - records user feedback events for LLM applications
- `NewRelic::Agent.set_llm_token_count_callback` - Sets a callback proc for calculating `token_count` attributes for LlmEmbedding and LlmChatCompletionMessage events

**Configuration options**
- `ai_monitoring.enabled` - disabled by default, controls LLM instrumentation. If disabled, no LLM metrics, events, or spans will be sent.
- `ai_monitoring.record_content.enabled` - enabled by default, allows the capture input and output content on specific LLM events.
- `instrumentation.ruby_openai` - enabled by default, controls the instrumentation of the ruby-openai gem

**Notes**
- In this PR, AI Monitoring is disabled if High-Security Mode is enabled
- LLM content-related events are exempt from 4095 byte attribute limit
- To collect request headers, AI monitoring requires doing a check in Net HTTP instrumentation to see if the parent segment belongs to an AI request
- Custom attributes can be added to ai events, but must be prefixed with `llm.`
- The supportability metric is: `Supportability/Ruby/ML/OpenAI/#{::OpenAI::VERSION}`


CHANGELOG.md
```diff
@@ -4,6 +4,18 @@
 
 Version <dev> introduces the option to store tracer state on the thread-level and hardens the browser agent insertion logic to better proactively anticipate errors.
 
+- **Feature: Add instrumentation for ruby-openai**
+
+  Instrumentation has been added for the [ruby-openai](https://github.com/alexrudall/ruby-openai) gem, supporting versions 3.4.0 and higher (PR#2442)[https://github.com/newrelic/newrelic-ruby-agent/pull/2442]. While ruby-openai instrumentation is enabled by default, the configuration option `ai_monitoring.enabled` is disabled by default and controls all AI monitoring. `ai_monitoring.enabled` must be set to `true` in order to receive ruby-openai instrumentation.
+
+  Calls to embedding and chat completion endpoints are automatically traced. Two new APIs allow users to record additional information on events:
```
tannalynn, line 11: Maybe this part should start a new section, like `Add LLM API methods` or something? since it'll be able to be used for bedrock too once we release the bedrock stuff

lib/new_relic/agent.rb
```diff
@@ -389,6 +392,92 @@ def record_custom_event(event_type, event_attrs)
...
+    #
+    # @api public
+    #
+    def record_llm_feedback_event(trace_id:,
+      rating:,
+      category: nil,
+      message: nil,
+      metadata: NewRelic::EMPTY_HASH)
+
+      record_api_supportability_metric(:record_llm_feedback_event)
```
fallwith, line 423: Perhaps we should hold off on sending the supportability metric until we've made it past the "is DT on?" check that is performed. Otherwise we'll see a metric suggesting use that isn't actually taking place.

lib/new_relic/agent.rb
```diff
@@ -389,6 +392,92 @@ def record_custom_event(event_type, event_attrs)
...
+
+      feedback_message_event = {
+        'trace_id': trace_id,
+        'rating': rating,
+        'category': category,
+        'message': message,
+        'id': NewRelic::Agent::GuidGenerator.generate_guid,
+        'ingest_source': NewRelic::Agent::Llm::LlmEvent::INGEST_SOURCE
+      }
+      feedback_message_event.merge!(metadata) if !metadata.empty?
```
fallwith, line 436:
```suggestion
      feedback_message_event.merge!(metadata) unless metadata.empty?
```

fallwith: I'm curious of the performance differences between `hash = {}; %w{one two three}.merge!(hash)` and `hash = {}; %w{one two three}.merge unless hash.empty?`. Is `merge!` efficient enough to outpace the penalty incurred by performing the `#empty?` check?

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -0,0 +1,192 @@
...
+
+      NewRelic::Agent.record_instrumentation_invocation(INSTRUMENTATION_NAME)
+      NewRelic::Agent::Llm::LlmEvent.set_llm_agent_attribute_on_transaction
+      record_openai_metric
+
+      if path == EMBEDDINGS_PATH
+        embeddings_instrumentation(parameters) { yield }
+      elsif path == CHAT_COMPLETIONS_PATH
+        chat_completions_instrumentation(parameters) { yield }
+      end
```
fallwith, line 25: Is it possible to have a `path` argument fail both `==` checks?

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -0,0 +1,192 @@
...
+
+    private
+
+    def embeddings_instrumentation(parameters)
+      segment = NewRelic::Agent::Tracer.start_segment(name: EMBEDDINGS_SEGMENT_NAME)
+      event = create_embeddings_event(parameters)
+      segment.llm_event = event
+      begin
+        response = NewRelic::Agent::Tracer.capture_segment_error(segment) { yield }
+        # TODO: Remove !response.include?('error') when we drop support for versions below 4.0.0
```
fallwith, line 36: This TODO seems helpful. I didn't see a TODO for similar version related logic elsewhere; particularly in `lib/new_relic/agent/instrumentation/ruby_openai.rb` which checks for  v5+. I think we should have TODOs for all versioned checks being performed.

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -0,0 +1,192 @@
...
+        model = message.is_response ? message.response_model : summary.request_model
+
+        message.token_count = calculate_token_count(model, message.content)
+      end
+    end
+
+    def calculate_token_count(model, content)
+      return unless NewRelic::Agent.llm_token_count_callback
+
+      count = NewRelic::Agent.llm_token_count_callback.call({model: model, content: content})
```
fallwith, line 147: This current approach puts the onus of all error handling on the customer, which they might not be inclined to do when authoring a simple proc. I think we ought to wrap the `#call` call with a `rescue`.

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -0,0 +1,192 @@
...
+    def llm_custom_attributes
+      NewRelic::Agent::Tracer.current_transaction&.attributes&.custom_attributes&.select { |k| k.to_s.match(/llm.*/) }
+    end
+
+    def record_openai_metric
+      NewRelic::Agent.record_metric(nr_supportability_metric, 0.0)
+    end
+
+    def segment_noticed_error?(segment)
+      segment&.instance_variable_get(:@noticed_error)
```
fallwith, line 172: If we need to get at the noticed error belonging to a segment, I think we should update the segment class to better expose the noticed error.

lib/new_relic/agent/llm.rb
```diff
@@ -3,9 +3,47 @@
...
+
+      # LLM content-related attributes are exempt from the 4095 byte limit
+      def self.exempt_event_attribute?(type, key)
+        return false unless instrumentation_enabled?
+
+        (type == NewRelic::Agent::Llm::Embedding::EVENT_NAME && key == INPUT) ||
+          (type == NewRelic::Agent::Llm::ChatCompletionMessage::EVENT_NAME && key == CONTENT)
+      end
+
+      def self.openai
```
fallwith, line 30: I still think the method should have a question mark:

```suggestion
      def self.openai?
```

(note that the memoization variable does not get the question mark)

lib/new_relic/agent/llm.rb
```diff
@@ -3,9 +3,47 @@
...
+        @openai ||= %i[prepend chain].include?(NewRelic::Agent.config[:'instrumentation.ruby_openai']) &&
+          NewRelic::Agent.config[:'ai_monitoring.enabled']
+      end
+
+      # Used in NetHTTP instrumentation
+      def self.openai_parent?(segment)
+        return false unless openai
```
fallwith, line 37:
```suggestion
        return false unless openai?
```

lib/new_relic/agent/llm.rb
```diff
@@ -3,9 +3,47 @@
...
 require_relative 'llm/embedding'
-require_relative 'llm/feedback'
 require_relative 'llm/response_headers'
+
+module NewRelic
+  module Agent
+    class LLM
+      INPUT = 'input'
+      CONTENT = 'content'
+      SEGMENT_PATTERN = %r{Llm/.*/OpenAI/.*}.freeze
```
fallwith, line 16: I think we can be stricter with the pattern, especially doing so results in a performance boost. One idea is to not use `*`:

```suggestion
      SEGMENT_PATTERN = %r{Llm/.+/OpenAI/.+}.freeze
```

CHANGELOG.md
hannahramadan: I like that @tannalynn! Its been updated

lib/new_relic/agent/llm.rb
```diff
@@ -3,9 +3,47 @@
...
+      def self.instrumentation_enabled?
+        NewRelic::Agent.config[:'ai_monitoring.enabled']
+      end
+
+      # LLM content-related attributes are exempt from the 4095 byte limit
+      def self.exempt_event_attribute?(type, key)
+        return false unless instrumentation_enabled?
+
+        (type == NewRelic::Agent::Llm::Embedding::EVENT_NAME && key == INPUT) ||
+          (type == NewRelic::Agent::Llm::ChatCompletionMessage::EVENT_NAME && key == CONTENT)
```
fallwith, line 27: I was originally worried about `type` being a string. One concern is that a symbol might appear which would fail the `==` check, and another concern is that for performance the use of a symbol might actually be preferred.

From what I can tell, `CustomEventAggregator#record` takes care of forcing `type` to become a string by routing it through a `@type_strings` lookup. We can evaluate a switch to symbols for performance post GA.

lib/new_relic/agent/llm/chat_completion_summary.rb
```diff
@@ -8,25 +8,59 @@ module NewRelic
...
+          end
         end
 
         def event_name
           EVENT_NAME
         end
+
+        def error_attributes(exception)
+          attrs = {}
+          attrs[ERROR_COMPLETION_ID] = id
```
fallwith, line 48:
```suggestion
          attrs = {ERROR_COMPLETION_ID => id}
```

lib/new_relic/agent.rb
tannalynn: I think the intention of these api supportability metric is to indicate that the user is calling them, not necessarily that they are successful. There are other api methods where we record the metric before we check whether we actually want to do something or not (`record_custom_event` for example, records the api metric even if custom events are disabled and nothing will happen). So because of that, i think leaving it as is will keep things in line with how we're treating these metrics for other api methods.

fallwith: Excellent. Alignment with existing agent behavior is certainly the goal here with this PR. I'll address any lingering supportability concerns I have with that approach in a different context. Resolved.

lib/new_relic/agent/llm/llm_event.rb
```diff
@@ -50,9 +67,30 @@ def attributes
...
+        def error_attributes(exception)
+          NewRelic::EMPTY_HASH
+        end
+
+        private
+
+        def replace_attr_with_string(attr)
+          return attribute_name_exceptions[attr] if attribute_name_exceptions.key?(attr)
+
+          attr
```
fallwith, line 92:
```suggestion
          attribute_name_exceptions.fetch(attr, attr)
```

lib/new_relic/agent/llm/response_headers.rb
```diff
@@ -6,29 +6,73 @@ module NewRelic
...
       module ResponseHeaders
-        ATTRIBUTES = %i[llm_version rate_limit_requests rate_limit_tokens
-          rate_limit_remaining_requests rate_limit_remaining_tokens
-          rate_limit_reset_requests rate_limit_reset_tokens]
+        ATTRIBUTES = %i[response_organization llm_version ratelimit_limit_requests
+          ratelimit_limit_tokens ratelimit_remaining_requests
+          ratelimit_remaining_tokens ratelimit_reset_requests
+          ratelimit_reset_tokens ratelimit_limit_tokens_usage_based
+          ratelimit_reset_tokens_usage_based
+          ratelimit_remaining_tokens_usage_based]
```
fallwith, line 14:
```suggestion
          ratelimit_remaining_tokens_usage_based].freeze
```

lib/new_relic/agent/llm/response_headers.rb
```diff
@@ -6,29 +6,73 @@ module NewRelic
...
+          ratelimit_limit_requests: 'response.headers.ratelimitLimitRequests',
+          ratelimit_limit_tokens: 'response.headers.ratelimitLimitTokens',
+          ratelimit_remaining_requests: 'response.headers.ratelimitRemainingRequests',
+          ratelimit_remaining_tokens: 'response.headers.ratelimitRemainingTokens',
+          ratelimit_reset_requests: 'response.headers.ratelimitResetRequests',
+          ratelimit_reset_tokens: 'response.headers.ratelimitResetTokens',
+          ratelimit_limit_tokens_usage_based: 'response.headers.ratelimitLimitTokensUsageBased',
+          ratelimit_reset_tokens_usage_based: 'response.headers.ratelimitResetTokensUsageBased',
+          ratelimit_remaining_tokens_usage_based: 'response.headers.ratelimitRemainingTokensUsageBased'
+        }
```
fallwith, line 28:
```suggestion
        }.freeze
```

lib/new_relic/agent.rb
hannahramadan: I like the `unless` change. 

The performance difference here is a great call out. This surprised me, but it's consistently 1.12x - 1.25x slower to merge an empty hash vs checking if it is empty first.

lib/new_relic/agent/llm.rb
hannahramadan: So the `*` is saying anything cane come before/after `/OpenAI/`, but what is the `+` saying?

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
hannahramadan: Is possible! We're doing an early `return yield `on line 15 if the path doesn't match the embeddings or chat completions path

lib/new_relic/agent/llm.rb
fallwith: > So the * is saying anything

`*` = 0 or more
`+` = 1 or more

The existing pattern matches `'Llm//OpenAI/'` with nothing between the `Llm` and `OpenAI` bits and nothing following the 'OpenAI' bit. Switching to `.+` to ensure something is present in both places could make it a little better. I'd need to have a better understanding of the type of data that is expected to be present but not match in order to suggest a better pattern.

hannahramadan: There are only two segment names as of right now that we are expecting: `Llm/embedding/OpenAI/embeddings` and `Llm/completion/OpenAI/chat`, so your suggested pattern works ◡̈

lib/new_relic/agent/llm.rb
hannahramadan: Added a new story for this: https://github.com/newrelic/newrelic-ruby-agent/issues/2520

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
hannahramadan: @fallwith I updated with a rescue, can you take a look?: https://github.com/newrelic/newrelic-ruby-agent/pull/2442/commits/e09a9672bc1068a2147f6e06060f6f4c50a4072b#diff-ddc2a8a97b6d6ef3a85b32d4ae4f738d2caa310cc5f652734ce904096fadbeaeR147-R152

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
hannahramadan: Added a TODO in the mentioned file: https://github.com/newrelic/newrelic-ruby-agent/pull/2442/commits/e09a9672bc1068a2147f6e06060f6f4c50a4072b#diff-8aced1668f9411100e08ab91d237e03ba38650fb2af79721c13744384f10bea1R20

Took a look and no other version conditions stood out as needing a TODO

lib/new_relic/agent/llm.rb
hannahramadan: This is helpful. Should have a `?` as a true/false method and helps distinguish from the variable. Nice nice

lib/new_relic/agent/llm/chat_completion_summary.rb
hannahramadan: Nice condensing!

lib/new_relic/agent/llm/llm_event.rb
hannahramadan: This required some lookup but very fun! So with `fetch`, we look into the `attribute_name_exceptions` hash for the `attr` key and return that if avaliable. If unavaliable, we just return the original `attr`.

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
fallwith: Looks good. I think this logic was modeled off of `lib/new_relic/agent/error_collector.rb`'s `update_error_group_name` method, and I checked to confirm it has a `rescue` too. Good to go.

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
hannahramadan: Thinking something like a method that just returns any noticed errors?

fallwith: We shouldn't need to make use of `Segment#instance_variable_get` when we control `Segment`. I'm thinking you could simply add a new `attr_reader` line to `Segment` and then change this line to read `segment&.noticed_error`.

fallwith (dismissed):

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
hannahramadan: Oh funny the segement class already has an `attr_reader` for `noticed_error`. I can make that line change

fallwith (approved):

fallwith (approved):

PR #2443, opened 2024-02-15
Add `ai_monitoring.enabled` configuration

### LINKED ISSUE #2439: Add new config `ai_monitoring.enabled`
Add a new config option that indicates whether LLM (OpenAI, Bedrock, LangChain) instrumentation will be registered. The new config should be `ai_monitoring.enabled` and the default is `true`. 

Ensure this check is completed in OpenAI instrumentation.

Also, `ai_monitoring.enabled` must be set to `false` if high security mode is enabled.

hannahramadan:
Add new configuration `ai_monitoring.enabled`. This configuration will automatically be set to `false` if high-security mode is on and has been added to the Security Policy Source, which allows customers to customize high-security settings. 

closes #2439

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -365,7 +365,7 @@ module NewRelic
           :public => true,
           :type => Boolean,
           :allowed_from_server => false,
-          :description => 'If `false`, all LLM (OpenAI, Bedrock) instrumentation will be disabled and no metrics, events, or spans will be sent. AI Monitoring is automatically disabled if `high_security` mode is enabled.'
+          :description => 'If `false`, all LLM (OpenAI) instrumentation will be disabled and no metrics, events, or spans will be sent. AI Monitoring is automatically disabled if `high_security` mode is enabled.'
         },
         # this is only set via server side config
         :apdex_t => {
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -364,7 +364,7 @@ module NewRelic
           :default => true,
           :public => true,
           :type => Boolean,
-          :allowed_from_server => true,
+          :allowed_from_server => false,
           :description => 'If `false`, all LLM (OpenAI, Bedrock) instrumentation will be disabled and no metrics, events, or spans will be sent. If `high_security` mode is enabled, no LLM events will be sent.'
         },
         # this is only set via server side config
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -360,6 +360,13 @@ module NewRelic
               - a.third.event
           DESCRIPTION
         },
+        :'ai_monitoring.enabled' => {
+          :default => true,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => true,
+          :description => 'If `false`, all LLM (OpenAI, Bedrock) instrumentation will be disabled and no metrics, events, or spans will be sent. If `high_security` mode is enabled, no LLM events will be sent.'
+        },
         # this is only set via server side config
         :apdex_t => {
           :default => 0.5,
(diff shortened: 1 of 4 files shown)


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -360,6 +360,13 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
               - a.third.event
           DESCRIPTION
         },
+        :'ai_monitoring.enabled' => {
+          :default => true,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => 'If `false`, all LLM (OpenAI, Bedrock) instrumentation will be disabled and no metrics, events, or spans will be sent. AI Monitoring is automatically disabled if `high_security` mode is enabled.'
```
kaylareopelle, line 368: What do you think about waiting to include `Bedrock` in this description once we have instrumentation for it? 

Maybe we add a TODO or something near the config to help us remember...

lib/new_relic/agent/instrumentation/ruby_openai.rb
```diff
@@ -12,9 +12,10 @@
   OPENAI_VERSION = Gem::Version.new(OpenAI::VERSION) if defined?(OpenAI)
 
   depends_on do
-    # add a config check for ai_monitoring.enabled
-    # maybe add DT check here eventually?
-    defined?(OpenAI) && defined?(OpenAI::Client) &&
+    NewRelic::Agent.config[:'ai_monitoring.enabled'] != false &&
+      NewRelic::Agent.config[:'instrumentation.ruby_openai'] != 'disabled' &&
```
kaylareopelle, line 16:
```suggestion
```

I think `instrumentaiton.ruby_openai` is checked as part of the `DependencyDetection` class. Did you find something to the contrary in your testing?

lib/new_relic/agent/instrumentation/ruby_openai.rb
```diff
@@ -12,9 +12,10 @@
...
+      # maybe add DT check here eventually?
```
kaylareopelle, line 17: I think we can safely remove this comment based on today's [DT update](https://newrelic.slack.com/archives/C05DWELBP0F/p1708098719787269) (internal)

kaylareopelle (changes requested):

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: That's a good call out! I think I'll make a note for tanna to add it here

lib/new_relic/agent/instrumentation/ruby_openai.rb
hannahramadan: Removed! https://github.com/newrelic/newrelic-ruby-agent/pull/2443/commits/95e5d803ee09c0e69fb2db7f4fc673a0df46a691

lib/new_relic/agent/instrumentation/ruby_openai.rb
hannahramadan: Removed ◡̈ https://github.com/newrelic/newrelic-ruby-agent/pull/2443/commits/95e5d803ee09c0e69fb2db7f4fc673a0df46a691

kaylareopelle (approved):
🦕

lib/new_relic/agent/configuration/security_policy_source.rb
```diff
@@ -35,6 +35,8 @@ def change_setting(policies, option, new_value)
           end
         end
 
+        # The Language Security Policy Source gives customers the ability to
+        # configure high security mode settings.
```
kaylareopelle, line 39: Two small, totally optional, suggestions:
1. If you move this comment between lines 9 and 10, just before the class declaration, it'll get added to RubyDoc's YARD documentation for the class. That's where I expect to see comments like this.

2. One of the things I found most helpful about your description of this class was its dubious context. I recommend adding something like: 
"This source never made it to general availability (no public docs) and work on it has been abandoned. We're uncertain if any of our customers use this."
This may help future engineers more quickly know how to prioritize work in this class.

Issue #2436, opened 2024-02-13
Refactor: Embedding and ChatCompletionSummary shared attributes


hannahramadan:
Embedding and ChatCompletionSummary share many common attributes, which we add to each event separately. Consider refactoring these attributes into their own module to be shared.

tannalynn:
Closing since this is AI instrumentation related

PR #2457, opened 2024-02-22
OpenAI: Error tracing

### LINKED ISSUE #2405: OpenAI: Trace errors
Record errors during chat and embedding creation with the following attributes:

- [ ] http.statusCode
- [ ] error.message 
- [ ] error.param 
- [ ] completion_id
- [ ] embedding_id

kaylareopelle:
When an error is raised by a segment that has an associated LLM event, the agent will add the following to the NoticedError's `attributes_from_notice_error` instance variable:
* `http.statusCode` => status code of the response with the error
* `error.code` => Value of `code` attribute returned from exception body
* `error.param` => Value of `param` attribute returned from exception body
* `completion_id` => `id`  of the related Llm::ChatCompletionSummary if the LlmEvent was of that type
* `embedding_id` => `id` of the related Llm::Embedding if the LlmEvent was of that type

This currently is formatted specifically for OpenAI responses, and may need to change when Bedrock is supported.

Closes #2405

### DIFF
--- a/lib/new_relic/agent/error_collector.rb
+++ b/lib/new_relic/agent/error_collector.rb
@@ -216,6 +216,10 @@ module NewRelic
       def notice_segment_error(segment, exception, options = {})
         return if skip_notice_error?(exception)
 
+        if segment.llm_event
+          options.merge!(segment.llm_event.error_attributes(exception))
+        end
+
         segment.set_noticed_error(create_noticed_error(exception, options))
         exception
       rescue => e
--- a/lib/new_relic/agent/error_collector.rb
+++ b/lib/new_relic/agent/error_collector.rb
@@ -216,9 +216,7 @@ module NewRelic
       def notice_segment_error(segment, exception, options = {})
         return if skip_notice_error?(exception)
 
-        if segment.llm_event
-          options.merge!(segment.llm_event.error_attributes(exception))
-        end
+        options.merge!(segment.llm_event.error_attributes(exception)) if segment.llm_event
 
         segment.set_noticed_error(create_noticed_error(exception, options))
         exception
--- a/lib/new_relic/agent/llm/chat_completion_summary.rb
+++ b/lib/new_relic/agent/llm/chat_completion_summary.rb
@@ -23,7 +23,7 @@ module NewRelic
           response_choices_finish_reason: 'response.choices.finish_reason',
           temperature: 'request.temperature'
         }
-
+        ERROR_COMPLETION_ID = 'completion_id'
         EVENT_NAME = 'LlmChatCompletionSummary'
 
         attr_accessor(*ATTRIBUTES)
@@ -45,6 +45,29 @@ module NewRelic
 
         def event_name
           EVENT_NAME
+        end
+
+        def error_attributes(exception)
+          attrs = {}
+
+          if exception.respond_to?(:response)
+            attrs[ERROR_ATTRIBUTE_STATUS_CODE] = exception.response.dig(:status)
+
+            attrs[ERROR_ATTRIBUTE_CODE] = exception.response.dig(
+              :body,
+              LlmEvent::ERROR_STRING,
+              LlmEvent::CODE_STRING
+            )
+            attrs[ERROR_ATTRIBUTE_PARAM] = exception.response.dig(
+              :body,
+              LlmEvent::ERROR_STRING,
+              LlmEvent::PARAM_STRING
+            )
+          end
+
+          attrs[ERROR_COMPLETION_ID] = id
+
+          attrs
         end
       end
     end
--- a/lib/new_relic/agent/llm/chat_completion_summary.rb
+++ b/lib/new_relic/agent/llm/chat_completion_summary.rb
@@ -49,23 +49,19 @@ module NewRelic
 
         def error_attributes(exception)
           attrs = {}
-
-          if exception.respond_to?(:response)
-            attrs[ERROR_ATTRIBUTE_STATUS_CODE] = exception.response.dig(:status)
-
-            attrs[ERROR_ATTRIBUTE_CODE] = exception.response.dig(
-              :body,
-              LlmEvent::ERROR_STRING,
-              LlmEvent::CODE_STRING
-            )
-            attrs[ERROR_ATTRIBUTE_PARAM] = exception.response.dig(
-              :body,
-              LlmEvent::ERROR_STRING,
-              LlmEvent::PARAM_STRING
-            )
-          end
-
           attrs[ERROR_COMPLETION_ID] = id
+
+          error_attributes_from_response(exception, attrs)
+        end
+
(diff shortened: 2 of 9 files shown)


lib/new_relic/agent/error_collector.rb
```diff
@@ -216,6 +216,10 @@ def extract_stack_trace(exception)
       def notice_segment_error(segment, exception, options = {})
         return if skip_notice_error?(exception)
 
+        if segment.llm_event
+          options.merge!(segment.llm_event.error_attributes(exception))
+        end
```
kaylareopelle, line 221: The noticed error gets recorded separately from the OpenAI segment. When we call `NewRelic::Agent::Tracer.capture_segment_error` in the instrumentation, the exception will be captured and the error will be recorded before we get back to the OpenAI instrumentation. Errors have their own harvest cycle separate from segments.

So far, this was the best place I found to add this data. The NoticedError we want to attach the attributes to has yet to be recorded. This code is only called by private APIs. However, it is called every time an segment error is recorded.

lib/new_relic/agent/llm/llm_event.rb
```diff
@@ -19,6 +19,12 @@ class LlmEvent
         INGEST_SOURCE = 'Ruby'
         LLM = :llm
         CUSTOM_ATTRIBUTE_CONVERSATION_ID = 'llm.conversation_id'
+        ERROR_ATTRIBUTE_STATUS_CODE = 'http.statusCode'
+        ERROR_ATTRIBUTE_CODE = 'error.code'
+        ERROR_ATTRIBUTE_PARAM = 'error.param'
+        ERROR_STRING = 'error'
+        CODE_STRING = 'code'
+        PARAM_STRING = 'param'
```
kaylareopelle, line 27: Since these constants are used by both Embedding and ChatCompletionSummary, I thought I'd put them here. In the future, if we work on https://github.com/newrelic/newrelic-ruby-agent/issues/2436, we might want to move these constants into that shared folder.

lib/new_relic/agent/llm/llm_event.rb
```diff
@@ -73,6 +79,11 @@ def record
           NewRelic::Agent.record_custom_event(event_name, event_attributes)
         end
 
+        # Subclasses that add attributes to noticed errors will override this method
+        def error_attributes(exception)
+          {}
+        end
```
kaylareopelle, line 85: If the `llm_event` attached to a segment is not an `Embedding` nor `ChatCompletionSummary`, this method should be hit, which should harmlessly return an empty hash for merging.

lib/new_relic/agent/error_collector.rb
fallwith, line 221: Perhaps when calling this on every single error it'll seems less invasive when confined to a single line. 😄 

```suggestion
        options.merge!(segment.llm_event.error_attributes(exception)) if segment.llm_event
```

lib/new_relic/agent/llm/chat_completion_summary.rb
```diff
@@ -46,6 +46,29 @@ def attribute_name_exceptions
         def event_name
           EVENT_NAME
         end
+
+        def error_attributes(exception)
+          attrs = {}
+
+          if exception.respond_to?(:response)
```
fallwith, line 53: This is a pretty big `if` block. Perhaps a new `error_attributes_from_response` method could be created and the entire if statement could be replaced with:

```ruby
error_attributes_from_response(exception, attrs)
```

The method itself would then look like:

```ruby
def error_attributes_from_response(exception, attrs)
  return unless exception.respond_to?(:response)

  attrs[ERROR_ATTRIBUTE_STATUS_CODE] = exception.response.dig(:status)
  attrs[ERROR_ATTRIBUTE_CODE] = exception.response.dig(
    :body,
    LlmEvent::ERROR_STRING,
    LlmEvent::CODE_STRING
  )
  attrs[ERROR_ATTRIBUTE_PARAM] = exception.response.dig(
    :body,
    LlmEvent::ERROR_STRING,
    LlmEvent::PARAM_STRING
  )
end
```

Disregard if it impacts any deadlines of course. On a related side note, it'd be cool if we could somehow hold brand new Ruby code to an agreed upon ideal RuboCop config while permitting the old code to use the handicap of having the RuboCop todo file.

lib/new_relic/agent/llm/embedding.rb
```diff
@@ -38,6 +39,29 @@ def attribute_name_exceptions
         def event_name
           EVENT_NAME
         end
+
+        def error_attributes(exception)
+          attrs = {}
+
+          if exception.respond_to?(:response)
```
fallwith, line 46: Similar "that's a big `if` block" comment as the previous one.

lib/new_relic/agent/llm/llm_event.rb
```diff
@@ -73,6 +79,11 @@ def record
           NewRelic::Agent.record_custom_event(event_name, event_attributes)
         end
 
+        # Subclasses that add attributes to noticed errors will override this method
+        def error_attributes(exception)
+          {}
```
fallwith, line 84: Presuming the overriding method will define a structure that can be modified, we should be safe reusing the frozen hash here I think.

```suggestion
          NewRelic::EMPTY_HASH
```

fallwith (approved):

test/new_relic/agent/llm/chat_completion_summary_test.rb
```diff
@@ -119,5 +119,68 @@ def test_record_creates_an_event
...
+    def test_error_attributes_for_irregular_exception
+      event = NewRelic::Agent::Llm::ChatCompletionSummary.new
+      expected = {'completion_id' => event.id}
+      exception = StandardError.new
+
+      assert_equal expected, event.error_attributes(exception)
+    end
+
+    class MockFaradayBadResponseError < StandardError
+      # Return value from an OpenAI chat completions request without a model parameter
```
hannahramadan, line 146: Helpful comment!

lib/new_relic/agent/llm/chat_completion_summary.rb
kaylareopelle: Great suggestion! Resolved in: d91ad510ea952c64c5ff50e3f2470e433245ff9e

lib/new_relic/agent/llm/embedding.rb
kaylareopelle: Great suggestion! Resolved in: d91ad510ea952c64c5ff50e3f2470e433245ff9e

lib/new_relic/agent/llm/chat_completion_summary.rb
kaylareopelle: Also, I totally agree about the Rubocop side note. Maybe we can contribute something like that during SAFT :)

PR #2468, opened 2024-02-27
Add Feedback API

### LINKED ISSUE #2406: OpenAI: Instrument feedback
Implement the record feedback API to allow customers to record feedback events.

Spec is still in progress. The strategy we'll use to link feedback events to the appropriate spans/traces/transactions is still under development.

Update 2/22: customers should be able to grab thread_id and pass that into new API along with Feedback. Documentation should be added along with examples of how this works.

hannahramadan:
Adds support for new a new API `record_llm_feedback_event`. This API requires two args: `trace_id` and `rating`. It accepts additional args `category` and `message`, and allows customers to pass in their own `metadata` hash. 

This new API uses `record_custom_event` to send a Feedback event with the title `LlmFeedbackMessage`.

[Spec](https://source.datanerd.us/agents/agent-specs/pull/661/files).

Closes #2406


lib/new_relic/agent.rb
```diff
@@ -388,6 +389,51 @@ def record_custom_event(event_type, event_attrs)
...
+    # Records user feedback events for LLM applications. This API must pass
+    # the current trace id as a parameter, which can be obtained using:
+    #
+    #   NewRelic::Agent::Tracer.current_trace_id
+    #
+    # @param [String] ID of the trace where the chat completion(s) related
+    #   to the feedback occurred.
+    #
+    # @param [String or Integer] Rating provided by an end user
+    #   (ex: “Good/Bad”, “1-10”).
```
fallwith, line 401: I'm not familiar with the format you guys are adhering to with these API comments, so feel free to disregard but I'm wondering if it might be better to give examples of individual values rather than sets or ranges, and to clearly indicate the optional use of an integer:

```suggestion
    #   (ex: “Good", Bad”, 1, 2, 5, 8, 10).
```

test/new_relic/agent/llm/feedback_test.rb
```diff
@@ -6,5 +6,60 @@
...
+    end
+
+    def test_record_llm_feedback_event_records_required_attributes
+      in_transaction do |txn|
+        NewRelic::Agent.record_llm_feedback_event(trace_id: '01234567890', rating: 5)
+        _, events = NewRelic::Agent.agent.custom_event_aggregator.harvest!
+        type, attributes = events[0]
+
+        assert_equal 'LlmFeedbackMessage', type['type']
+        assert_equal '01234567890', attributes['trace_id']
```
fallwith, line 20: I recommend assigning the `'01234567890'` string to a variable and then having both the test code and the assertion reference that variable instead of repeating the string.

In this particular case with `'01234567890'`, you're actually using that same value all over the place across several tests. So perhaps you could do something like:

```ruby
def setup
  @trace_id = '01234567890'
end
```

or

```ruby
private

def trace_id
  @trace_id ||= '01234567890'
end
```

test/new_relic/agent/llm/feedback_test.rb
```diff
@@ -6,5 +6,60 @@
...
+        assert_equal 'strudel', attributes['toaster']
+      end
+    end
+
+    def test_record_llm_feedback_event_raises_missing_required_parameter
+      assert_raises(ArgumentError) { NewRelic::Agent.record_llm_feedback_event(trace_id: '01234567890') }
+    end
+
+    def test_record_llm_feedback_event_invalid_param
+      assert_raises(ArgumentError) { NewRelic::Agent.record_llm_feedback_event(trace_id: '01234567890', rating: 5, food: 'blueberry') }
```
fallwith, line 46: This line is too long, but the linter stage of the automated CI process doesn't ding you on it because we've relaxed so many rules to work with the 15 year old code base.

@kaylareopelle and I were chatting about this recently. We'd like to come up with a solution for holding new code accountable to our desired RuboCop configuration while allowing the old code to be excepted. Perhaps a starting point would be to have new code forbidden from using the `.rubocop_todo.yml` file, though long term there are a whole heck of lot of exceptions expressed in `.rubocop.yml` too.

But for now for this PR, would you please get these lines at or under 120 columns in width? Thank you.

lib/new_relic/agent.rb
hannahramadan: I grabbed this from the spec, but this is so much more clear. Thank you!

test/new_relic/agent/llm/feedback_test.rb
hannahramadan: Thank you for this callout! I never re-added a vertical line helper to my editor. Thats been done, and this code cleaned up. https://github.com/newrelic/newrelic-ruby-agent/pull/2468/commits/ce0ee2d8818c149a649fda49aebc8cc60615f3c9#diff-bc386c8b22f64519a262c1bde1fb6b92218a86c109d9da8df409a91a0bb5be08R47-R52

test/new_relic/agent/llm/feedback_test.rb
hannahramadan: Thank you! Took your first suggestion https://github.com/newrelic/newrelic-ruby-agent/pull/2468/commits/ce0ee2d8818c149a649fda49aebc8cc60615f3c9

fallwith (approved):

kaylareopelle (approved):
Great job with documentation parallelization!

PR #2470, opened 2024-02-28
Add support for `llm.` custom attributes

### LINKED ISSUE #2437: OpenAI: Attach custom attributes to llm_events when prefixed with `llm.`
Customers should be able to add custom attributes to a transaction via `add_custom_attribute` API. If a custom attribute is prefixed with `llm.`, we should automatically attach it to the llm event, ie `llm.<user_defined_metadata>` 

The conversation_id is now being handled like any other llm.* custom event and should be added to all events when available. Update the references to `conversation_id` to use this new `llm.` style.

Let's make sure the attributes are applied to all custom events. Not just chat completion summaries.

We should make sure this is captured in documentation and may need to additionally describe this functionality in the [reserved keyword](https://docs.newrelic.com/docs/data-apis/custom-data/custom-events/data-requirements-limits-custom-event-data/#reserved-words) docs.

hannahramadan:
Adds custom attributes to LLM events when prefixed with `llm.`. 

Closes #2437

### DIFF
--- a/lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
@@ -151,7 +151,7 @@ module NewRelic::Agent::Instrumentation
     end
 
     def llm_custom_attributes
-      NewRelic::Agent::Tracer.current_transaction.attributes.custom_attributes.select { |k, v| k.to_s.match(/llm.*/) }
+      NewRelic::Agent::Tracer.current_transaction.attributes.custom_attributes.select { |k| k.to_s.match(/llm.*/) }
     end
 
     def record_openai_metric
--- a/lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
@@ -30,7 +30,6 @@ module NewRelic::Agent::Instrumentation
       segment = NewRelic::Agent::Tracer.start_segment(name: EMBEDDINGS_SEGMENT_NAME)
       record_openai_metric
       event = create_embeddings_event(parameters)
-      event.metadata = llm_custom_attributes
       segment.llm_event = event
       begin
         response = NewRelic::Agent::Tracer.capture_segment_error(segment) { yield }
@@ -47,7 +46,6 @@ module NewRelic::Agent::Instrumentation
       segment = NewRelic::Agent::Tracer.start_segment(name: CHAT_COMPLETIONS_SEGMENT_NAME)
       record_openai_metric
       event = create_chat_completion_summary(parameters)
-      event.metadata = llm_custom_attributes
       segment.llm_event = event
       messages = create_chat_completion_messages(parameters, event.id)
 
@@ -73,7 +71,8 @@ module NewRelic::Agent::Instrumentation
         conversation_id: conversation_id,
         request_max_tokens: parameters[:max_tokens] || parameters['max_tokens'],
         request_model: parameters[:model] || parameters['model'],
-        temperature: parameters[:temperature] || parameters['temperature']
+        temperature: parameters[:temperature] || parameters['temperature'],
+        metadata: llm_custom_attributes
       )
     end
 
@@ -82,7 +81,8 @@ module NewRelic::Agent::Instrumentation
         # TODO: POST-GA: Add metadata from add_custom_attributes if prefixed with 'llm.', except conversation_id
         vendor: VENDOR,
         input: parameters[:input] || parameters['input'],
-        request_model: parameters[:model] || parameters['model']
+        request_model: parameters[:model] || parameters['model'],
+        metadata: llm_custom_attributes
       )
     end
 
@@ -151,7 +151,9 @@ module NewRelic::Agent::Instrumentation
     end
 
     def llm_custom_attributes
-      NewRelic::Agent::Tracer.current_transaction.attributes.custom_attributes.select { |k| k.to_s.match(/llm.*/) }
+      attributes = NewRelic::Agent::Tracer.current_transaction&.attributes&.custom_attributes&.select { |k| k.to_s.match(/llm.*/) }
+
+      attributes.transform_keys! { |key| key[4..-1] } if !attributes.nil?
     end
 
     def record_openai_metric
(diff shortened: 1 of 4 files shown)


lib/new_relic/agent/llm/llm_event.rb
```diff
@@ -27,6 +27,7 @@ class LlmEvent
         PARAM_STRING = 'param'
 
         attr_accessor(*ATTRIBUTES)
+        attr_accessor :metadata
```
fallwith, line 30: Would you please explain how the metadata attribute is designed to work? I'm specifically curious about why it is on its own here instead of being included in the `ATTRIBUTES` array.

Currently all other attributes can be sent in at initialize time.

Setting aside metadata, your new test code that looks like this:

```ruby
event = NewRelic::Agent::Llm::LlmEvent.new(id: 123)
event.vendor = 'OpenAI'
event.response_model = 'gpt-4'
```

could instead be written as this:

```ruby
event = NewRelic::Agent::Llm::LlmEvent.new(id: 123, vendor: 'OpenAI', response_model: 'gpt-4')
```

The model class is flexible enough to allow you to specify a vendor and response model later, but if you know them at initialize time, it's easier to just pass them to `new()`.

But now with the new `:metadata` attribute, we're breaking from that pattern and requiring that the attribute writer be used post initialize time. So we'd have to write this:

```ruby
event = NewRelic::Agent::Llm::LlmEvent.new(id: 123, vendor: 'OpenAI', response_model: 'gpt-4')
event.metadata = {'Marathon' => '26.2', 'Ultra Marathon' => 'Ouch'}
```

instead of this:

```ruby
event = NewRelic::Agent::Llm::LlmEvent.new(
  id: 123,
  vendor: 'OpenAI',
  response_model: 'gpt-4',
  metadata: {'Marathon' => '26.2', 'Ultra Marathon' => 'Ouch'})
```

If there's a good reason for having `:metadata` work differently, such as it not being known at initialize time, then we can add a code comment to explain that. Otherwise, we can just add `:metadata` to the existing `ATTRIBUTES` array to match existing attribute behavior. You shouldn't even need to refactor the new unit tests if you do.

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -147,6 +149,10 @@ def update_chat_completion_messages(messages, response, summary)
       end
     end
 
+    def llm_custom_attributes
+      NewRelic::Agent::Tracer.current_transaction.attributes.custom_attributes.select { |k, v| k.to_s.match(/llm.*/) }
```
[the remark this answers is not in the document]
hannahramadan: Great suggestion! Commiting this.

lib/new_relic/agent/llm/llm_event.rb
kaylareopelle: I believe the `metadata` attribute needs to be assigned in a way that’s unique from all the other attributes.

Currently, the other attributes are assigned to the final event using metaprogramming within the `NewRelic::Agent::Llm::LlmEvent#event_attributes` method:  https://github.com/newrelic/newrelic-ruby-agent/blob/ad0dfa4011c1589139902927fd85ed349d1445f8/lib/new_relic/agent/llm/llm_event.rb#L37-L41 

This means, the attributes that get sent to the custom event API are pretty much a direct copy of the attributes we iterate over. The `trace_id` instance var becomes `trace_id: value`.

We can’t assign metadata this way because the custom event API flattens nested hashes and separates the nesting with a period when it assigns attribute names.  

If we assign metadata like the other attributes, a structure like: 
`{ metadata: { child: ‘value’ }`

would have the final custom event attribute: 
`metadata.child: value`

The spec requires the customer-provided key is the key, without any reference to `metadata`, `llm`, or other prefixes. We need a way to assign the attribute as `child: value`.

By separating `metadata` from the rest of the attributes, we can avoid this assignment issue.   However, this isn’t the only way we could achieve this. We could also try to flatten the metadata hash before recording the event attributes. We could update the inititalize method to address the concerns about not being able to assign the value on initialization. 

I acknolwedge that this is a bit weird, so happy to work on something else that seems like a better design.

kaylareopelle: Also, with regards to the test code, it's currently written to test assignment in both ways so that we can make sure passing args to the initalizer and assigning using the attr_writer are both functional. I recognize this isn't well documented and is cluttering the reason behind the test, so we can rewrite those to be more specific.

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -144,9 +146,16 @@ def update_chat_completion_messages(messages, response, summary)
         message.conversation_id = conversation_id
         message.request_id = summary.request_id
         message.response_model = response['model']
+        message.metadata = llm_custom_attributes
       end
     end
 
+    def llm_custom_attributes
+      attributes = NewRelic::Agent::Tracer.current_transaction&.attributes&.custom_attributes&.select { |k| k.to_s.match(/llm.*/) }
```
fallwith, line 154: This one looks like a fun one to hit 100% branch coverage on.

fallwith (approved):

test/new_relic/agent/llm/llm_event_test.rb
```diff
@@ -46,6 +46,17 @@ def test_event_attributes_returns_a_hash_of_assigned_attributes_and_values
...
 
+    def test_event_attributes_adds_custom_attributes
+      event = NewRelic::Agent::Llm::LlmEvent.new(id: 123)
+      event.vendor = 'OpenAI'
+      event.response_model = 'gpt-4'
+      event.metadata = {'Marathon' => '26.2', 'Ultra Marathon' => 'Ouch'}
+      result = event.event_attributes
+
+      assert_equal('26.2', result['Marathon'])
+      assert_equal('Ouch', result['Ultra Marathon'])
```
kaylareopelle, line 57: I'm somewhat interested in adding an assertion to this test that makes sure the other attributes are also correctly assigned. Maybe something like:
```suggestion
      assert_equal('Ouch', result['Ultra Marathon'])
      assert_equal('OpenAI', result[:vendor])
```

I'm not sure if other tests would fail if we accidentally refactored the code in a way that overrode the attributes hash with just the metadata hash.

kaylareopelle (approved):
Great updates, Hannah! One small assertion suggestion, but take it or leave it. 🌻

test/new_relic/agent/llm/llm_event_test.rb
hannahramadan: I like it! Taking this commit suggestion

PR #2473, opened 2024-02-28
Remove references to `conversation_id`

### LINKED ISSUE #2471: Remove plumbing for `conversation_id`
The `conversation_id` is now being handled like any other `llm.*` custom event and should be added to all events when available. 

Update the references to `conversation_id` to use this new llm. style by removing the plumbing for it, giving customers control of when this attribute is added.

hannahramadan:
Removes plumbing and references to `conversation_id`, which should now be handled like any other custom attributes and added to all events when passed to `add_custom_attribute` by a used. 

Because the sole purpose of the module ChatCompletion was to add `conversation_id` to messages, this module has been removed along with it's inclusion in other modules.

Closes #2471


kaylareopelle (approved):
Thank you!

PR #2474, opened 2024-03-01
Introduce support for serverless operations (AWS Lambda for now)

fallwith:
These changes bring the Ruby agent in line with New Relic's cross-APM-agent specification for operating within a serverless  context.

The new content is intended only for use by the function wrapper provided by the New Relic AWS Lambda Ruby layer. The wrapper invokes `NewRelic::Agent::ServerlessHandler#lambda_handler` as the contact point to handing off control to the agent. The agent then invokes the wrapped customer function and writes metric data to a named pipe. The layer delivers a Lambda extension that runs alongside the customer's function, reads from the named pipe, and relays data to New Relic.

Because serverless mode does not involve any network connectivity being performed by the agent itself, `NewRelicService` has been updated to be able to function without an active connection.

Screenshot 2024-02-29 at 9 54 22 PM


kaylareopelle (commented):
I'm still working my way through this, but I thought I'd pass along the comments I have so far.

lib/new_relic/agent/configuration/yaml_source.rb
```diff
@@ -53,6 +53,8 @@ def failed?
         protected
 
         def validate_config_file_path(path)
+          return if NewRelic::Agent.config[:'serverless_mode.enabled']
```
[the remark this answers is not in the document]
fallwith: That's correct, yes. Serverless customers will be sourcing the agent via a Lambda layer that we publish. That layer is built from a Bundler driven build process that grabs the agent from RubyGems.org. The layer won't include a config file and the customer is not expected to hack the existing layer or add another layer to deliver one. According to the spec on this one, all non-default configuration is to be based on environment variables (in addition to not having a config file, there's no server-side config in play either).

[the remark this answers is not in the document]
fallwith: Technically the agent operates just fine without a config file without this change, but without the change the agent does emit warning messages to the logs. Given that the absence of the config file is expected for serverless, I didn't like seeing the warning.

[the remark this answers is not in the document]
kaylareopelle: Sounds good, thank you!

lib/new_relic/agent/connect/request_builder.rb
```diff
@@ -24,7 +24,7 @@ def connect_payload
             :host => local_host,
             :display_host => Agent.config[:'process_host.display_name'],
             :app_name => Agent.config[:app_name],
-            :language => 'ruby',
+            :language => LANGUAGE,
```
kaylareopelle, line 27: If the capitalization isn't important, in the future, this could be combined with the `Ruby` constant we're adding for AIM's ingest source.

kaylareopelle (approved):
This looks good to me, James. I feel like there's more I would need to learn about the spec and observe the code running in a lambda context to be more precise in a review. Maybe we can have a deeper demo sometime next week.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1808,6 +1808,18 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+        :'serverless_mode.enabled' => {
+          :default => false,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :transform => proc { |bool|
+            ENV.key?(NewRelic::Agent::ServerlessHandler::LAMBDA_ENVIRONMENT_VARIABLE) || bool
+          },
+          :description => 'If `true`, the agent will operate in a streamlined mode suitable for use with short-lived ' \
+                          'serverless functions.'
```
fallwith, line 1821:
```suggestion
          :description => 'If `true`, the agent will operate in a streamlined mode suitable for use with short-lived ' \
                          'serverless functions. NOTE: Only AWS Lambda functions are supported currently and this ' \
                          "option is not intended for use without New Relic's Ruby Lambda layer offering."
```

kaylareopelle (dismissed):

lib/new_relic/agent/connect/request_builder.rb
fallwith: My current hypothesis is that the string value is currently only expected to be read by humans so having a capital letter ought to be ok, but I'd like to be consistent with the example provided by the agent spec and the existing agents that are going with all lowercase.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1815,6 +1815,17 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+        # Serverless
+        :'serverless_mode.enabled' => {
+          :default => false,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :transform => proc { |bool| NewRelic::Agent::ServerlessHandler.env_var_set? || bool },
+          :description => 'If `true`, the agent will operate in a streamlined mode suitable for use with short-lived ' \
+                          'serverless functions. NOTE: Only AWS Lambda functions are supported currently and this ' \
+                          "option is not intended for use without New Relic's Ruby Lambda layer offering."
```
hannahramadan, line 1827: Sometimes I see links included when there is mention of a New Relic product/service. Would [this page](https://docs.newrelic.com/docs/serverless-function-monitoring/aws-lambda-monitoring/get-started/monitoring-aws-lambda-serverless-monitoring/) be appropriate to link to a word in "New Relic's Lambda layer offering"?

lib/new_relic/agent/distributed_tracing/distributed_trace_payload.rb
```diff
@@ -101,10 +101,6 @@ def current_segment_id(transaction)
             transaction.current_segment.guid
           end
         end
-
-        def connected?
```
hannahramadan, line 105: Was removing this method and instead calling `Agent.instance.connected?` a cleanup, or related to lambda code?

hannahramadan (approved):

lib/new_relic/agent/agent_helpers/connect.rb
```diff
@@ -62,10 +66,8 @@ def log_error(error)
         # no longer try to connect to the server, saving the
         # application and the server load
         def handle_license_error(error)
-          ::NewRelic::Agent.logger.error( \
-            error.message, \
-            'Visit NewRelic.com to obtain a valid license key, or to upgrade your account.'
-          )
+          ::NewRelic::Agent.logger.error(error.message,
+            'Visit NewRelic.com to obtain a valid license key, or to upgrade your account.')
```
kaylareopelle, line 70: I think our company has changed its preferred capitalization. Not a big deal, but 🤷 
```suggestion
            'Visit newrelic.com to obtain a valid license key, or to upgrade your account.')
```

kaylareopelle (approved):
🍵

lib/new_relic/agent/configuration/default_source.rb
fallwith: This is a good idea. The Ruby agent specific documentation doesn't exist yet to make use of but that high level one will likely eventually offer a click-through path to the Ruby goods.

Please submit an inline suggestion in the format you're thinking of and I'll commit it. Alternatively, feel free to directly commit to the feature branch.

lib/new_relic/agent/distributed_tracing/distributed_trace_payload.rb
fallwith: A little bit of both. For serverless/lambda's sake I originally introduced a new `ready?` method that read better than `connected?` in a serverless context. Later I took the agent's serverless mode even more minimalistic and stopped using the worker thread entirely, so there was no need to perform a `ready?` check anymore. When reverting out the `ready?` work, I decide to leave these refactors because I love being able to search for and find only a single `def connected?` line instead of the number we had before. So it started out lambda specific and now it's just a cleanup.

hannahramadan: Thank you for the explanation! ◡̈

Issue #2452, opened 2024-02-17
AIM/OpenAI: Create config to remove content from LLM events


kaylareopelle:
* when enabled, the agent removes content before sending LLM events, specifically:
  * `LlmChatCompletionMessage` event `content` attribute
  * `LlmFeedbackMessage` event `messages` attribute (TBD may have moved to another event)
  * `LlmEmbedding` event, `input` attribute

* config defaults to `true`
* name `ai_monitoring.record_content.enabled`

Verify these requirements with the spec: https://source.datanerd.us/agents/agent-specs/pull/663/ (Internal)

PR #2482, opened 2024-03-05
Configuration option: `ai_monitoring.record_content.enabled`

hannahramadan:
Introduces a new configuration option, `ai_monitoring.record_content.enabled`, that when disabled, the agent removes certain content before sending LLM events, specifically:
- LlmChatCompletionMessage event content attribute
- LlmEmbedding event, input attribute

name: `ai_monitoring.record_content.enabled`
default: `true`

[Spec](https://source.datanerd.us/agents/agent-specs/pull/663/) (Internal)

Note per the spec, the following should also be removed (we aren't recording these as of now—noting here for completness):
- LlmTool event input attribute
- LlmTool event output attribute
- LlmVectorSearch request.query input attribute
- LlmVectorSearchResult page_content input attribute

closes [#2452](https://github.com/newrelic/newrelic-ruby-agent/issues/2452)


lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -30,6 +30,7 @@ def embeddings_instrumentation(parameters)
       segment = NewRelic::Agent::Tracer.start_segment(name: EMBEDDINGS_SEGMENT_NAME)
       record_openai_metric
       event = create_embeddings_event(parameters)
+      event.remove_instance_variable(:@input) if !record_content_enabled?
```
hannahramadan, line 33: I'm torn between making this a method similar to `def remove_content(messages)` for Message events. The code is small so I left it—but open to creating a method for cleaner consistency.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -367,6 +367,23 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => <<~DESCRIPTION
+            If `false`, LLM (OpenAI) instrumentation will not capture input and output content on specific LLM events.
+
+            The excluded attributes include:
+              * `content` from LlmChatCompletionMessage events
+              * `input` from LlmEmbedding events
+              * `input` from LlmTool events
```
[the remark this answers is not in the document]
kaylareopelle: Yes, let's only reference events that we have at this time. We can update it if we end up supporting other events. 
LlmTool could be in our future since ruby-openai can access that part of the API.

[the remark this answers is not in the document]
hannahramadan: Thank you—Ill get these removed!

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -367,6 +367,19 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'If `false`, all LLM (OpenAI) instrumentation will be disabled and no metrics, events, or spans will be sent. AI Monitoring is automatically disabled if `high_security` mode is enabled.'
         },
+        :'ai_monitoring.record_content.enabled' => {
+          :default => true,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => <<~DESCRIPTION
+            If `false`, LLM (OpenAI) instrumentation will not capture input and output content on specific LLM events.
```
fallwith, line 376: The description states that this option is specific to LLM and even more specific to OpenAI at that. My first instinct would be to express that specificity in the config parameter naming. Is there a strategy behind not having "LLM" nor "OpenAI" in the name, such as keeping it generic for future-proofing?

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -30,6 +30,7 @@ def embeddings_instrumentation(parameters)
       segment = NewRelic::Agent::Tracer.start_segment(name: EMBEDDINGS_SEGMENT_NAME)
       record_openai_metric
       event = create_embeddings_event(parameters)
+      event = remove_input(event) if !record_content_enabled?
```
fallwith, line 33: Does either the spec or the code encourage this approach of conditionally removing content? If not, I wonder if conditionally adding that content beforehand would be better.

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -147,6 +148,22 @@ def update_chat_completion_messages(messages, response, summary)
       end
     end
 
+    def record_content_enabled?
+      NewRelic::Agent.config[:'ai_monitoring.record_content.enabled']
+    end
+
+    def remove_content(messages)
```
fallwith, line 155: I think these would be better implemented as methods that destructively operate on their input and don't return anything.

test/multiverse/suites/ruby_openai/ruby_openai_instrumentation_test.rb
```diff
@@ -233,4 +233,33 @@ def test_set_llm_agent_attribute_on_embedding_transaction
 
     assert_truthy harvest_transaction_events![1][0][2][:llm]
   end
+
+  def test_embeddings_drop_input_when_record_content_disabled
+    with_config(:'ai_monitoring.record_content.enabled' => false) do
```
fallwith, line 238: It would be good to have the 2 `enabled => true` inverse tests here as well. That way we don't rely on the defaults and sometimes even catch things that an all-refute approach might miss.

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: Yes for future-proofing! This will include Bedrock at some point

test/multiverse/suites/ruby_openai/ruby_openai_instrumentation_test.rb
hannahramadan: I like this idea! Ill add the inverse tests

lib/new_relic/agent/configuration/default_source.rb
fallwith: In that case I'm torn between changing `"(OpenAI)"` to `"(OpenAI only for now)"` or simply referring to LLM instrumentation generically now in a future proof way that doesn't mention a vendor at all:

```suggestion
            If `false`, LLM instrumentation will not capture input and output content on specific LLM events.
```

hannahramadan: I liked having the OpenAI reference since I think this was the only place to specifically call out which instrumentations are impacted by this config change. We're okay with the "(OpenAI only for now)" route?

kaylareopelle: I'm fine with "OpenAI only for now". I prefer that over removing "OpenAI" entirely.

fallwith: Definitely go with whatever you both like and makes the most sense to you (given that you're much closer to its usage than I am). My thing was a) LLM is much bigger than OpenAI, and b) AI is much bigger than LLM, and c) this option is described as referring to OpenAI LLM without mentioning either of those words in the parameter name. If you're telling me that the parameter name itself will not need to change in future, I'm fine with the description saying whatever you like for now and getting updated in future.

The reason I advocate for "LLM (OpenAI only for now)" vs "LLM (OpenAI)" is that it wasn't immediately clear to me what the bit in parens was clarifying. Perhaps "(OpenAI) LLM" would have made more sense in my brain.

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
hannahramadan: Methods are now just destroying, no return

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: Changed to "OpenAI only for now" https://github.com/newrelic/newrelic-ruby-agent/pull/2482/files#diff-d116d3c8c7bebe204a18ee3983ac59c5d15bd70da7123f2a2c4a12f144bca181R368

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
hannahramadan: Thanks for your help @fallwith! We are now adding content vs removing.

kaylareopelle (approved):

Issue #2451, opened 2024-02-17
AIM/OpenAI: Create callback to record client-side token count


kaylareopelle:
Implement a callback API for customer to implement a token count client-side if the agent doesn't already have the token count. 

* name: set_llm_token_callback
* return value: integer 
* behavior: 
  * additional calls to the method overwrite the proc
  * accepts arguments for model and content

Spec link: https://source.datanerd.us/agents/agent-specs/pull/662 (internal)

The work to assign the proc to the correct attributes will be handled in #2450

kaylareopelle:
Related to #2450 and #2449

kaylareopelle:
Closed by #2485

PR #2485, opened 2024-03-05
Create `set_llm_token_count_callback` API

kaylareopelle:
The API accepts one argument, a proc, that should contain the callback the agent will invoke if it cannot glean the token count from instrumentation.

A supportability metric is recorded when the API is called.

Closes: #2451

### DIFF
--- a/lib/new_relic/agent.rb
+++ b/lib/new_relic/agent.rb
@@ -109,9 +109,11 @@ module NewRelic
     LLM_FEEDBACK_MESSAGE = 'LlmFeedbackMessage'
 
     attr_reader :error_group_callback
+    attr_reader :llm_token_count_callback
 
     @agent = nil
     @error_group_callback = nil
+    @llm_token_count_callback = nil
     @logger = nil
     @tracer_lock = Mutex.new
     @tracer_queue = []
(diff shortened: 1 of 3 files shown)


fallwith (approved):
This one gave me a bit of proc déjà vu.

kaylareopelle:
> This one gave me a bit of proc déjà vu.

Your work on the error callback made this easy!

lib/new_relic/agent.rb
```diff
@@ -436,6 +438,42 @@ def record_llm_feedback_event(trace_id:,
...
     # @!endgroup
 
+    # @!group LLM callbacks
+
+    # Set a callback proc for calculating `token_count` attributes for
+    # LlmEmbedding and LlmChatCompletionMessage events
+    #
+    # @param callback_proc [Proc] the callback proc
+    #
+    # Typically this method should be called only once to set a callback for
```
hannahramadan, line 448: Do we need the word 'typically'?

hannahramadan (approved):

lib/new_relic/agent.rb
kaylareopelle: I don't think so. This documentation was drawn from another one of our APIs. I'll remove!

kaylareopelle, line 448:
```suggestion
    # This method should be called only once to set a callback for
```

PR #2489, opened 2024-03-06
Create `token_count` attribute

### LINKED ISSUE #2450: OpenAI: Attach `token_count` to LLM events
Related to #2451 and #2449 

Create a new attribute, `token_count` on Llm::Embedding and Llm::ChatCompletionMessage events.

The value should be an integer representing the tokens used on the request, or the result of an llm token callback function if the LLM response doesn't include token counts. 

Spec: https://source.datanerd.us/agents/agent-specs/pull/662 (internal)

kaylareopelle:
* Adds `token_count` to `ChatCompletionMessage` and `Embedding`
* Calculates `token_count` using the customer-provided callback set by `set_llm_token_count_callback` and assigns it in OpenAI instrumentation
* If the callback is not available, or the result is not an Integer greater than zero or nil, don't send the `token_count` attribute.

Closes #2450


lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -101,8 +102,7 @@ def create_chat_completion_messages(parameters, summary_id)
           role: message[:role] || message['role'],
           sequence: index,
           completion_id: summary_id,
-          vendor: VENDOR,
-          is_response: true
```
kaylareopelle, line 105: This is an unrelated fix. This first group of messages are not from a response. They're from a request. The `is_response` should only be attached to the event if it has a truthy value.

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
```diff
@@ -129,9 +129,21 @@ def update_chat_completion_messages(messages, response, summary)
...
         message.metadata = llm_custom_attributes
+
+        model = message.is_response ? message.response_model : summary.request_model
+
+        message.token_count = calculate_token_count(model, message.content)
       end
     end
 
+    def calculate_token_count(model, content)
+      # return unless NewRelic::Agent.config['ai_monitoring.record_content.enabled']
```
kaylareopelle, line 140: This'll get uncommented once the config PR is merged

lib/new_relic/agent/instrumentation/ruby_openai/instrumentation.rb
hannahramadan: nice

lib/new_relic/agent/llm/chat_completion_message.rb
```diff
@@ -6,7 +6,7 @@ module NewRelic
   module Agent
     module Llm
       class ChatCompletionMessage < LlmEvent
-        ATTRIBUTES = %i[content role sequence completion_id is_response]
```
hannahramadan, line 9: I'm sure this is said somewhere, but why are removing completion_id? I'm not remembering/ seeing the connection

kaylareopelle: We aren't 🤦 lol. So glad you asked this. I'm not sure what I was thinking. I'll revert.

hannahramadan (approved):
Looks good after the completion_id stuff is added back ◡̈

PR #2490, opened 2024-03-07
Synchronize config hash

hannahramadan:
The use of JRuby revealed a synchronization issue of threads accessing/editing configuration. Adding a 'synchronize do' block to mediate this.


fallwith (approved):

fallwith:
NOTE: https://github.com/newrelic/newrelic-ruby-agent/pull/2282 replaced the lock and went with a `#dup` instead. We currently feel that we still need that `#dup` for a separate issue that can be reproduced with our JRuby CI. This approach of layering on both the lock and the `#dup` is intentional and approved.

tannalynn (approved):

PR #2507, opened 2024-03-13
OpenAI: Update response headers

### LINKED ISSUE #2461: OpenAI: Verify response headers
The response headers associated with LLM events have changed slightly since we first worked on them. We also now have data type requirements.

Review the LLMs spec and update the code accordingly.

Since some of these have "token" in their name, let's hold off working on it until we get a spec update related to the work documented in #2449, #2450, and #2451

kaylareopelle:
LlmEmbedding and LlmChatCompletionSummary events generated by OpenAI have attributes pulled from response headers.

The following response headers should be captured if they're available:
* response.headers.ratelimitLimitTokensUsageBased
* response.headers.ratelimitResetTokensUsageBased
* response.headers.ratelimitRemainingTokensUsageBased

We haven't been able to craft a request that returns these headers.

In addition, this PR updates the data types of the response headers to match [the spec (internal)](https://source.datanerd.us/agents/agent-specs/blob/main/LLMs.md#response-headers).

Closes #2461

### DIFF
--- a/lib/new_relic/agent/llm/response_headers.rb
+++ b/lib/new_relic/agent/llm/response_headers.rb
@@ -50,8 +50,8 @@ module NewRelic
           self.request_id = headers[X_REQUEST_ID]&.first
           self.response_organization = headers[OPENAI_ORGANIZATION]&.first
           self.llm_version = headers[OPENAI_VERSION]&.first
-          self.rate_limit_requests = headers[X_RATELIMIT_LIMIT_REQUESTS]&.first
-          self.rate_limit_tokens = headers[X_RATELIMIT_LIMIT_TOKENS]&.first
+          self.rate_limit_requests = headers[X_RATELIMIT_LIMIT_REQUESTS]&.first&.to_i
+          self.rate_limit_tokens = headers[X_RATELIMIT_LIMIT_TOKENS]&.first&.to_i
           remaining_headers(headers)
           reset_headers(headers)
           tokens_usage_based_headers(headers)
@@ -60,8 +60,8 @@ module NewRelic
         private
 
         def remaining_headers(headers)
-          self.rate_limit_remaining_requests = headers[X_RATELIMIT_REMAINING_REQUESTS]&.first
-          self.rate_limit_remaining_tokens = headers[X_RATELIMIT_REMAINING_TOKENS]&.first
+          self.rate_limit_remaining_requests = headers[X_RATELIMIT_REMAINING_REQUESTS]&.first&.to_i
+          self.rate_limit_remaining_tokens = headers[X_RATELIMIT_REMAINING_TOKENS]&.first&.to_i
         end
 
         def reset_headers(headers)
@@ -71,8 +71,8 @@ module NewRelic
 
         def tokens_usage_based_headers(headers)
           self.rate_limit_limit_tokens_usage_based = headers[X_RATELIMIT_LIMIT_TOKENS_USAGE_BASED]&.first
-          self.rate_limit_reset_tokens_usage_based = headers[X_RATELIMIT_RESET_TOKENS_USAGE_BASED]&.first
-          self.rate_limit_remaining_tokens_usage_based = headers[X_RATELIMIT_REMAINING_TOKENS_USAGE_BASED]&.first
+          self.rate_limit_reset_tokens_usage_based = headers[X_RATELIMIT_RESET_TOKENS_USAGE_BASED]&.first&.to_i
+          self.rate_limit_remaining_tokens_usage_based = headers[X_RATELIMIT_REMAINING_TOKENS_USAGE_BASED]&.first&.to_i
         end
       end
     end
--- a/lib/new_relic/agent/llm/response_headers.rb
+++ b/lib/new_relic/agent/llm/response_headers.rb
@@ -70,8 +70,8 @@ module NewRelic
         end
 
         def tokens_usage_based_headers(headers)
-          self.ratelimit_limit_tokens_usage_based = headers[X_RATELIMIT_LIMIT_TOKENS_USAGE_BASED]&.first
-          self.ratelimit_reset_tokens_usage_based = headers[X_RATELIMIT_RESET_TOKENS_USAGE_BASED]&.first&.to_i
+          self.ratelimit_limit_tokens_usage_based = headers[X_RATELIMIT_LIMIT_TOKENS_USAGE_BASED]&.first&.to_i
+          self.ratelimit_reset_tokens_usage_based = headers[X_RATELIMIT_RESET_TOKENS_USAGE_BASED]&.first
           self.ratelimit_remaining_tokens_usage_based = headers[X_RATELIMIT_REMAINING_TOKENS_USAGE_BASED]&.first&.to_i
         end
       end
--- a/lib/new_relic/agent/llm/response_headers.rb
+++ b/lib/new_relic/agent/llm/response_headers.rb
@@ -50,8 +50,8 @@ module NewRelic
           self.request_id = headers[X_REQUEST_ID]&.first
           self.response_organization = headers[OPENAI_ORGANIZATION]&.first
           self.llm_version = headers[OPENAI_VERSION]&.first
-          self.ratelimit_limit_requests = headers[X_RATELIMIT_LIMIT_REQUESTS]&.first&.to_i
-          self.ratelimit_limit_tokens = headers[X_RATELIMIT_LIMIT_TOKENS]&.first&.to_i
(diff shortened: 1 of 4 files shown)


lib/new_relic/agent/llm/response_headers.rb
```diff
@@ -43,11 +52,28 @@ def populate_openai_response_headers(headers)
           self.llm_version = headers[OPENAI_VERSION]&.first
           self.rate_limit_requests = headers[X_RATELIMIT_LIMIT_REQUESTS]&.first
           self.rate_limit_tokens = headers[X_RATELIMIT_LIMIT_TOKENS]&.first
+          remaining_headers(headers)
+          reset_headers(headers)
+          tokens_usage_based_headers(headers)
```
kaylareopelle, line 57: The branch condition for the method was getting to be too high, so I grouped headers with similar names into their own methods

test/new_relic/agent/llm/chat_completion_summary_test.rb
```diff
@@ -62,12 +62,15 @@ def test_record_creates_an_event
...
-        summary.rate_limit_tokens = '101'
+        summary.rate_limit_requests = 100
+        summary.rate_limit_tokens = 101
         summary.rate_limit_reset_tokens = '102'
         summary.rate_limit_reset_requests = '103'
-        summary.rate_limit_remaining_tokens = '104'
-        summary.rate_limit_remaining_requests = '105'
+        summary.rate_limit_remaining_tokens = 104
+        summary.rate_limit_remaining_requests = 105
+        summary.rate_limit_limit_tokens_usage_based = '106'
```
hannahramadan, line 71: Why are some strings and some ints

kaylareopelle: It's just to follow what's in the spec. I think some of the headers, like 
ratelimitResetTokens, have values that are a mix of letters and numbers. I was hoping NRDB could do the transformation for us, but unfortunately that's not how it works.

kaylareopelle: Would you like me to update the test data to be more representative of what the actual response headers would contain? the examples in the `response_headers_test` are more precise.

lib/new_relic/agent/llm/response_headers.rb
```diff
@@ -41,13 +50,30 @@ def populate_openai_response_headers(headers)
           self.request_id = headers[X_REQUEST_ID]&.first
           self.response_organization = headers[OPENAI_ORGANIZATION]&.first
           self.llm_version = headers[OPENAI_VERSION]&.first
-          self.rate_limit_requests = headers[X_RATELIMIT_LIMIT_REQUESTS]&.first
-          self.rate_limit_tokens = headers[X_RATELIMIT_LIMIT_TOKENS]&.first
-          self.rate_limit_remaining_requests = headers[X_RATELIMIT_REMAINING_REQUESTS]&.first
-          self.rate_limit_remaining_tokens = headers[X_RATELIMIT_REMAINING_TOKENS]&.first
+          self.rate_limit_requests = headers[X_RATELIMIT_LIMIT_REQUESTS]&.first&.to_i
```
fallwith, line 53: I'm tempted to simply `headers[X_RATELIMIT_LIMIT_REQUESTS]&.first&.to_i` to `headers[X_RATELIMIT_LIMIT_REQUESTS]&.first.to_i`, which will help on branch coverage testing. Can the header really be set to an array with nils inside?

lib/new_relic/agent/llm/response_headers.rb
```diff
@@ -41,13 +50,30 @@ def populate_openai_response_headers(headers)
...
+          self.rate_limit_remaining_tokens = headers[X_RATELIMIT_REMAINING_TOKENS]&.first&.to_i
+        end
+
+        def reset_headers(headers)
           self.rate_limit_reset_requests = headers[X_RATELIMIT_RESET_REQUESTS]&.first
           self.rate_limit_reset_tokens = headers[X_RATELIMIT_RESET_TOKENS]&.first
         end
+
+        def tokens_usage_based_headers(headers)
+          self.rate_limit_limit_tokens_usage_based = headers[X_RATELIMIT_LIMIT_TOKENS_USAGE_BASED]&.first
```
fallwith, line 73: The value stays a string on this one?

lib/new_relic/agent/llm/response_headers.rb
```diff
@@ -18,7 +21,10 @@ module ResponseHeaders
           rate_limit_remaining_requests: 'response.headers.ratelimitRemainingRequests',
           rate_limit_remaining_tokens: 'response.headers.ratelimitRemainingTokens',
           rate_limit_reset_requests: 'response.headers.ratelimitResetRequests',
-          rate_limit_reset_tokens: 'response.headers.ratelimitResetTokens'
+          rate_limit_reset_tokens: 'response.headers.ratelimitResetTokens',
```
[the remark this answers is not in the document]
kaylareopelle: I'm happy to change it. these have always felt weird to me in the naming department.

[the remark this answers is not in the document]
kaylareopelle: I like `ratelimit` here instead. I'll update the names.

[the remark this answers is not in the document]
fallwith: Yeah, after I scrolled down to the one-word `RATELIMIT` in the header strings, I think we should change `rate_limit`  to `ratelimit` just for consistency with the header strings, regardless of how the corresponding camel case values look.

lib/new_relic/agent/llm/response_headers.rb
fallwith: Ah, I see you both chatting about strings vs. ints elsewhere. It's odd to see given the variable name similarities, but that's how it has to be if the spec calls for it that way.

lib/new_relic/agent/llm/response_headers.rb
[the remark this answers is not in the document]
kaylareopelle: Nice. Thank you. Fixed in 091771a3b

lib/new_relic/agent/llm/response_headers.rb
kaylareopelle: I wanted to do this too, but held off because I'm concerned it interferes with our reporting accuracy. 

`nil.to_i` returns zero.

If a custom event attribute has a value of `nil`, it isn't reported.
If it has a value of zero, it is reported.

These attributes could theoretically be equal to zero too. 

I think preserving the nil seems important in this case. However, I'm open to other ideas and might be missing something here.

lib/new_relic/agent/llm/response_headers.rb
kaylareopelle: Yes, it's part of the spec 🤷

I'm bummed that I haven't been able to create a request that returns this header so that I can find out what it actually looks like. I should probably check out what Python and Node are doing in their tests to update the example values. 

For some of the response headers, like `ratelimit_reset_requests`, the response is a time value, so it has a mix of numbers and letters: `4ms`. That might be true for this one as well.

kaylareopelle: Whoops, missed your follow-up when I responded! Glad we're on the same page!

lib/new_relic/agent/llm/response_headers.rb
fallwith: I was thinking you'd only encounter one of these 3 scenarios...

- header is missing
- header is present, with a nil value
- header is present with a value that responds to `#first`

And then drilling down into that 3rd scenario, I was wondering if it is possible for there to be a non-nil value that successfully responds to `#first`, but then gives back a `nil` when `#first` is called. If so, we have a fourth scenario that the second safe-navigation operator is needed to handle. If not, we only need the first operator.

It's definitely totally fine to keep both regardless, I was just thinking of cutting down on one branch for testing's sake if we could.

test/new_relic/agent/llm/chat_completion_summary_test.rb
kaylareopelle: I went ahead and did this anyway. I think it's better for documentation's sake to have something genuine. I looked at Python's tests to get the values added in e5450ba59

fallwith (approved):

lib/new_relic/agent/llm/response_headers.rb
kaylareopelle: Those are great points. I forgot to consider the likelihood that a `nil` value would reach the `#first` call. I think your suggestion is the way to go. I'll update the code.

kaylareopelle: Updated in 9e0224bfa

tannalynn (approved):

PR #2511, opened 2024-03-14
Send full llm content

### LINKED ISSUE #2448: AIM: Send the full LLM content
Currently, the LLM content attributes follow the truncation policy upheld by our other attributes and will be cut at 4096 bytes. The `content` attribute on `LlmChatCompletionMessage` and the `input` attribute on `LlmEmbedding` events should not be subject to this limit. The entire value must be sent to the collector.

It's fine to truncate all the other attributes.

Spec: https://source.datanerd.us/agents/agent-specs/pull/664 (internal)

hannahramadan:
The `content` attribute on `LlmChatCompletionMessage` and the `input` attribute on `LlmEmbedding` events are exempt from truncation rules. They can be over 4096 bytes and entire the value must be sent to the collector.

Spec: https://source.datanerd.us/agents/agent-specs/pull/664 (internal)

closes #2448

### DIFF
--- a/lib/new_relic/agent/custom_event_aggregator.rb
+++ b/lib/new_relic/agent/custom_event_aggregator.rb
@@ -17,8 +17,6 @@ module NewRelic
       MAX_ATTRIBUTE_COUNT = 64
       MAX_ATTRIBUTE_SIZE = 4095
       MAX_NAME_SIZE = 255
-      INPUT = 'input'
-      CONTENT = 'content'
 
       named :CustomEventAggregator
       capacity_key :'custom_insights_events.max_samples_stored'
@@ -72,18 +70,13 @@ module NewRelic
             key = key[0, MAX_NAME_SIZE]
           end
 
-          # value is limited to 4095 expect for LLM content-related events
+          # value is limited to 4095 except for LLM content-related events
           if val.is_a?(String) && val.length > MAX_ATTRIBUTE_SIZE
-            val = val[0, MAX_ATTRIBUTE_SIZE] unless llm_exempt_event_attribute?(type, key)
+            val = val[0, MAX_ATTRIBUTE_SIZE] unless NewRelic::Agent::LLM.llm_exempt_event_attribute?(type, key)
           end
 
           new_result[key] = val
         end
-      end
-
-      def llm_exempt_event_attribute?(type, key)
-        (type == NewRelic::Agent::Llm::Embedding::EVENT_NAME && key == INPUT) ||
-          (type == NewRelic::Agent::Llm::ChatCompletionMessage::EVENT_NAME && key == CONTENT)
       end
 
       def after_initialize
--- a/lib/new_relic/agent/custom_event_aggregator.rb
+++ b/lib/new_relic/agent/custom_event_aggregator.rb
@@ -17,6 +17,10 @@ module NewRelic
       MAX_ATTRIBUTE_COUNT = 64
       MAX_ATTRIBUTE_SIZE = 4095
       MAX_NAME_SIZE = 255
+      EXEMPT_EMBEDDING = 'LlmEmbedding'
+      INPUT = 'input'
+      EXEMPT_CHAT_MESSAGE = 'LlmChatCompletionMessage'
+      CONTENT = 'content'
 
       named :CustomEventAggregator
       capacity_key :'custom_insights_events.max_samples_stored'
@@ -52,11 +56,11 @@ module NewRelic
           {TYPE => type,
            TIMESTAMP => Process.clock_gettime(Process::CLOCK_REALTIME).to_i,
            PRIORITY => priority},
-          create_custom_event_attributes(attributes)
+          create_custom_event_attributes(type, attributes)
         ]
       end
 
-      def create_custom_event_attributes(attributes)
+      def create_custom_event_attributes(type, attributes)
         result = AttributeProcessing.flatten_and_coerce(attributes)
 
         if result.size > MAX_ATTRIBUTE_COUNT
@@ -70,13 +74,17 @@ module NewRelic
             key = key[0, MAX_NAME_SIZE]
           end
 
-          # value is limited to 4095
+          # value is limited to 4095 expect for LLM content-related events
           if val.is_a?(String) && val.length > MAX_ATTRIBUTE_SIZE
-            val = val[0, MAX_ATTRIBUTE_SIZE]
+            val = val[0, MAX_ATTRIBUTE_SIZE] unless llm_exempt_event_attribute?(type, key)
           end
 
           new_result[key] = val
         end
+      end
+
+      def llm_exempt_event_attribute? (type, key)
+        (type == EXEMPT_EMBEDDING && key == INPUT) || (type == EXEMPT_CHAT_MESSAGE && key == CONTENT)
       end
 
       def after_initialize
--- a/lib/new_relic/agent/custom_event_aggregator.rb
+++ b/lib/new_relic/agent/custom_event_aggregator.rb
@@ -17,9 +17,7 @@ module NewRelic
       MAX_ATTRIBUTE_COUNT = 64
       MAX_ATTRIBUTE_SIZE = 4095
       MAX_NAME_SIZE = 255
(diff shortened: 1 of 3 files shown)


lib/new_relic/agent/custom_event_aggregator.rb
```diff
@@ -70,15 +72,20 @@ def create_custom_event_attributes(attributes)
             key = key[0, MAX_NAME_SIZE]
           end
 
-          # value is limited to 4095
+          # value is limited to 4095 expect for LLM content-related events
```
fallwith, line 75:
```suggestion
          # value is limited to 4095 except for LLM content-related events
```

test/new_relic/agent/custom_event_aggregator_test.rb
```diff
@@ -97,6 +97,26 @@ def test_truncates_attributes
       assert_equal(expected, actual)
     end
 
+    def test_does_not_truncates_llm_embedding_input_attribute
+      params = {'input' => 'a' * 5000}
+      expected = {'input' => 'a' * 5000}
```
fallwith, line 102: If you ever want to test going over a maximum threshold, you could always stub the real constant with a lower value. That way you can just test the logic without needing to actually have strings that are crazy long to test with.

In this case...

```ruby
require 'minitest/stub_const'

def test_does_not_truncates_llm_embedding_input_attribute
  NewRelic::Agent::CustomEventAggregator.stub_const(MAX_ATTRIBUTE_SIZE, 5) do
    # ...
  end
end
```

test/new_relic/agent/custom_event_aggregator_test.rb
```diff
@@ -97,6 +97,26 @@ def test_truncates_attributes
       assert_equal(expected, actual)
     end
 
+    def test_does_not_truncates_llm_embedding_input_attribute
```
fallwith, line 100: For the two new tests, I think `truncate` should be singular. I think "does not truncate" reads better than "does not truncates".

lib/new_relic/agent/custom_event_aggregator.rb
```diff
@@ -70,15 +72,20 @@ def create_custom_event_attributes(attributes)
...
           if val.is_a?(String) && val.length > MAX_ATTRIBUTE_SIZE
-            val = val[0, MAX_ATTRIBUTE_SIZE]
+            val = val[0, MAX_ATTRIBUTE_SIZE] unless llm_exempt_event_attribute?(type, key)
           end
 
           new_result[key] = val
         end
       end
 
+      def llm_exempt_event_attribute?(type, key)
```
fallwith, line 84: Just to confirm... are we certain that `type` and `key` will always both be strings 100% of the time?

lib/new_relic/agent/custom_event_aggregator.rb
```diff
@@ -70,15 +72,20 @@ def create_custom_event_attributes(attributes)
             key = key[0, MAX_NAME_SIZE]
           end
 
-          # value is limited to 4095
+          # value is limited to 4095 expect for LLM content-related events
           if val.is_a?(String) && val.length > MAX_ATTRIBUTE_SIZE
-            val = val[0, MAX_ATTRIBUTE_SIZE]
+            val = val[0, MAX_ATTRIBUTE_SIZE] unless llm_exempt_event_attribute?(type, key)
```
fallwith, line 77: I think we should defer LLM specific logic to an LLM class. Then for performance reasons, I think we should also short-circuit the logic when LLM instrumentation isn't even active.

So this line becomes something like:

```ruby
val = val[0, MAX_ATTRIBUTE_SIZE] unless BaseLLMClass.full_length_attribute?(type, key)
```

And then over in your LLM base class if where you'd have what you are currently calling `llm_exempt_event_attribute?`:

```ruby
module NewRelic
  module Agent
    class LLMBaseClass
      def full_length_attribute?(type, key)
        return unless llm_instrumentation_enabled?

        # ...
      end
    end
  end
end

File: lib/new_relic/agent/custom_event_aggregator.rb [hunk 6]
```diff
@@ -70,15 +72,20 @@ def create_custom_event_attributes(attributes)
...
           end
 
           new_result[key] = val
         end
       end
 
+      def llm_exempt_event_attribute?(type, key)
+        (type == NewRelic::Agent::Llm::Embedding::EVENT_NAME && key == INPUT) ||
+          (type == NewRelic::Agent::Llm::ChatCompletionMessage::EVENT_NAME && key == CONTENT)
```
tannalynn, line 86:
```suggestion
        type == NewRelic::Agent::Llm::Embedding::EVENT_NAME && (key == INPUT || key == CONTENT)
```
this will slightly reduce the number of checks we need to make

lib/new_relic/agent/custom_event_aggregator.rb
tannalynn: For type, we do some logic about type earlier in the process that would have already required it to be a string. 
For key, since its a hash key it should only really be string or symbols anyways. plus it wouldn't check the key until after it matches type, which means we're the ones who have defined key, which they are all strings.

test/new_relic/agent/custom_event_aggregator_test.rb
tannalynn: For me, I like to avoid stubbing for things that aren't too difficult to test directly like this. Since making a long string is pretty simple with ruby, I'd prefer testing the actual constant just in case it had been changed for some reason, because then we'd see the failure.

But when things get more complicated, like creating 10,000 spans or something, then I think this is a great way to go.

tannalynn (approved):

lib/new_relic/agent/custom_event_aggregator.rb
hannahramadan: Thanks Tanna!

test/new_relic/agent/custom_event_aggregator_test.rb
hannahramadan: 100% agree. I'll update this

test/new_relic/agent/custom_event_aggregator_test.rb
hannahramadan: OOooo I never thought about the possibility of stubbing very large values. I also kinda like testing directly here, but will keep in mind that practice for the future.

test/new_relic/agent/custom_event_aggregator_test.rb
```diff
@@ -97,6 +97,26 @@ def test_truncates_attributes
...
+    end
+
+    def test_does_not_truncates_llm_chat_message_content_attribute
+      params = {'content' => 'a' * 5000}
+      expected = {'content' => 'a' * 5000}
+
+      @aggregator.record(:LlmChatCompletionMessage, params)
+      actual = @aggregator.harvest![1].first[1]
+
+      assert_equal(expected, actual)
```
fallwith, line 117: Is the inverse (attributes not on the "do not truncate" list) being tested elsewhere? We want to make sure that the new logic only causes certain attributes to have their values truncated, but that all other values still get truncated.

test/new_relic/agent/custom_event_aggregator_test.rb
fallwith: Sounds good. I definitely support @tannalynn in her cautioning against the hiding of bugs that can result from a reliance on stubbing. In cases like this one, the constant predates the PR and the PR doesn't deliver anything that impacts how the constant is being evaluated. I was relying on there already being great, non-stubbed test coverage in place. With everything that is being tested in this PR being downstream of the constant having been checked, I think stubbing is ok. Not stubbing is ok too of course.

test/new_relic/agent/custom_event_aggregator_test.rb
hannahramadan: Yes ◡̈ Tanna added this test earlier when she added the truncation. Just a little [above](https://github.com/newrelic/newrelic-ruby-agent/pull/2511/files#diff-521cca09234a65ab1c95a0146b5022f8b7e6188da82503d130a0d7151b6660ebR84-R96) these tests.

fallwith (approved):

PR #2521, opened 2024-03-22
Errors Inbox transaction tracing updates

fallwith:
- The transaction's guid should be present for both the ErrorTrace and TransactionError types
- Be sure (tests added, no lib code altered) that distributed tracing isn't required for the transaction guid to be populated

### DIFF
--- a/lib/new_relic/agent/transaction_error_primitive.rb
+++ b/lib/new_relic/agent/transaction_error_primitive.rb
@@ -16,26 +16,27 @@ module NewRelic
     module TransactionErrorPrimitive
       extend self
 
-      SAMPLE_TYPE = 'TransactionError'.freeze
-      TYPE_KEY = 'type'.freeze
-      ERROR_CLASS_KEY = 'error.class'.freeze
-      ERROR_MESSAGE_KEY = 'error.message'.freeze
-      ERROR_EXPECTED_KEY = 'error.expected'.freeze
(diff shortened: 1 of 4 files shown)


lib/new_relic/agent/transaction_error_primitive.rb
```diff
@@ -16,26 +16,27 @@ module Agent
...
+      SYNTHETICS_JOB_ID_KEY = 'nr.syntheticsJobId'
+      SYNTHETICS_MONITOR_ID_KEY = 'nr.syntheticsMonitorId'
       SYNTHETICS_TYPE_KEY = 'nr.syntheticsType'
       SYNTHETICS_INITIATOR_KEY = 'nr.syntheticsInitiator'
       SYNTHETICS_KEY_PREFIX = 'nr.synthetics'
-      PRIORITY_KEY = 'priority'.freeze
-      SPAN_ID_KEY = 'spanId'.freeze
+      PRIORITY_KEY = 'priority'
+      SPAN_ID_KEY = 'spanId'
+      GUID_KEY = 'guid'
```
fallwith, line 39: The redundant `String#freeze` calls were all removed given that we're using the special `frozen_string_literal` comment.

The existing CAT specific values had their constants renamed with a `CAT_` prefix.

A new `GUID_KEY` constant was added to hold the transaction id when one is known.

kaylareopelle (approved):

PR #2534, opened 2024-03-28
Net::HTTP OpenAI tests

### LINKED ISSUE #2438: OpenAI: Write Net::HTTP tests for LLM attributes assigned from response headers
The work in #2398 does not yet have tests for the changes to the `Net::HTTP` instrumentation to add attributes from response headers to LLM-events. This ticket exists to add those tests.

kaylareopelle:
* Move the LLM class tests into their own file since they do not rely on the Net::HTTP instrumentation file anymore 
* Add tests for the OpenAI-related behavior in the Net::HTTP instrumentation

Resolves #2438


test/new_relic/net_http_test_cases.rb
```diff
@@ -137,4 +137,39 @@ def test_ipv6_host_get_request_records_metric
...
+      # raise if populate_openai_response_headers is called
+      NewRelic::Agent::LLM.stub(:populate_openai_response_headers, -> { raise OpenAITestError.new }) do
+        in_transaction do
+          get_response
+        end
+      end
+    end
+  end
+
+  def test_attempts_to_populate_response_headers_with_openai
```
fallwith, line 156: I've used the `raise` pattern from `test_does_not_attempt_to_populate_response_headers_without_openai` above to essentially pull off an `assert_does_not_raise` assertion with Minitest. It gives value in that sense, though unfortunately leaves us without any actual real assertions to increase the overall tally with.

But for `test_attempts_to_populate_response_headers_with_openai` where we we're testing the success of something instead of the absence of a raise, I don't think we need to leverage a raise in a proc to get there.

Perhaps something like this would work:

```ruby
expectation = 'cool, you called me'
NewRelic::Agent::LLM.stub(:populate_openai_response_headers, expectation) do
  in_transaction do
    assert_equal expectation, get_response, "Expected LLM to receive `.populate_openai_response_headers`!"
  end
end
```

NOTE: 
- if we perform the assertion inside the `stub` block, then there's no need for a `result` variable at all - nevermind one set to `nil` initially.
- I don't see `wrapped_response` used

test/new_relic/agent/llm_test.rb
```diff
@@ -0,0 +1,132 @@
...
+      end
+
+      def test_populate_openai_response_headers_with_llm_event_calls_llm_method
+        NewRelic::Agent.stub(:config, {:'instrumentation.ruby_openai' => :auto, :'ai_monitoring.enabled' => true}) do
+          response = {}
+          parent = NewRelic::Agent::Transaction::AbstractSegment.new('Llm/embedding/OpenAI/embeddings')
+          mock_llm_event = NewRelic::Agent::Llm::Embedding.new
+          mock_llm_event.expects(:populate_openai_response_headers).with(response)
+          parent.llm_event = mock_llm_event
+          NewRelic::Agent::LLM.populate_openai_response_headers(response, parent)
```
fallwith, line 127: I remember pairing on this one. It's not immediately clear to someone not readily familiar with Mocha, but the `#expects` call is essentially performing a verification on the mock. So that's ultimately where the value is coming from here. Is there really nothing else we can hang an assert on though? Does `response` get modified or anything else take place that we could confirm?

fallwith (approved):

test/new_relic/net_http_test_cases.rb
kaylareopelle: I love how simple this suggestion is. I feel like the current test may be unnecessarily complex. 

The only issue I'm seeing is the return value for `populate_openai_response_headers` is not the return value for `get_response`. 

I tried to get around the return value issue by raising an error within `populate_openai_response_headers` to cut into the middle. We definitely don't need `wrapped_response` so I've removed it. 

I don't have other ideas for adjustments at the moment, but I'm happy to brainstorm with you to make the test simpler.

test/new_relic/agent/llm_test.rb
kaylareopelle: We might be able to refactor this to check to see if headers get assigned to the llm event. 

How does this look? 

```ruby
      def test_populate_openai_response_headers_with_llm_event_adds_headers_to_llm_event
        NewRelic::Agent.stub(:config, {:'instrumentation.ruby_openai' => :auto, :'ai_monitoring.enabled' => true}) do
          request_id = 'PIR8SH1P'
          response = {'x-request-id' => [request_id]}
          parent = NewRelic::Agent::Transaction::AbstractSegment.new('Llm/embedding/OpenAI/embeddings')
          mock_llm_event = NewRelic::Agent::Llm::Embedding.new
          parent.llm_event = mock_llm_event
          NewRelic::Agent::LLM.populate_openai_response_headers(response, parent)

          assert_equal request_id, mock_llm_event.request_id
        end
      end
```

fallwith: > How does this look?

I personally love this, because:

- It's now easier to intuit what is taking place
- We now have an assertion, which I think is nicer than the `expects` we had previously
- We can now remove the `teardown` method and Mocha dependency from this test class entirely

kaylareopelle: Wonderful! Applied in 7bf893aa65ef17d151db76d8e9aaf75957b34c78

fallwith (approved):

test/new_relic/agent/llm_test.rb
hannahramadan: I like this^

hannahramadan (approved):

PR #2536, opened 2024-03-28
simplify tests for AI header setting / not setting

fallwith:
use a variable in combination with a proc to confirm the stubbed method was indeed invoked as expected or not invoked as expected.

this gives us an assertion in the non-invoked path, which is nice.

i believe this fixes (switching from `-> {}` to `proc { |_h| }` an issue to where if the `raise` actually was attempted in the non-invoked path, it would fail.

removed unused variables / mocks


kaylareopelle (approved):
This is wonderful! Thank you, James!

PR #2542, opened 2024-04-04
Update watched gems

hannahramadan:
Update gems to reviece slack notifications about when new version releases are published.

### DIFF
--- a/.github/workflows/slack_notifications.yml
+++ b/.github/workflows/slack_notifications.yml
@@ -28,10 +28,6 @@ jobs:
             httpclient
             mongo
             puma
-            sidekiq
-            sinatra
-            stripe
-            tilt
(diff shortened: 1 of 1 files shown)


.github/workflows/slack_notifications.yml
```diff
@@ -35,7 +36,10 @@ jobs:
             rake
             redis
             resque
-            unicorn"
+            roda
+            ruby-openai
```
kaylareopelle, line 40: Thanks for getting this update started, Hannah! I have two questions:
1. Should we alphabetize this list? (it's so close!)
2. Do we also want to include stripe?

kaylareopelle (approved):

fallwith (approved):

PR #2546, opened 2024-04-10
Configuration: support allowlists

### LINKED ISSUE #2555: Internal feature request: support allowlist type configuration parameter value validation
It has been internally requested that the agent support allowlist style constraints for parameter values so that the common pattern for doing so doesn't need to keep being repeated across all classes that require the functionality.

fallwith:
Configuration hashes defined by `default_source.rb` now support an
additional `:allowlist` key set equal to an array of valid values.
    
If an invalid value is specified for a parameter that has an allowlist,
a warning will be logged that contains the allowlist and the default
value for the parameter will be used.

resolves #2555

### DIFF
--- a/lib/new_relic/agent/configuration/manager.rb
+++ b/lib/new_relic/agent/configuration/manager.rb
@@ -150,7 +150,7 @@ module NewRelic
             begin
               transform.call(value)
             rescue => e
-              ::NewRelic::Agent.logger.error("Error applying transformation for #{key}, pre-transform value was: #{value}.", e)
+              NewRelic::Agent.logger.error("Error applying transformation for #{key}, pre-transform value was: #{value}.", e)
               raise e
             end
           else
@@ -163,7 +163,7 @@ module NewRelic
           return if allowlist.include?(value)
 
           default = default_source.default_for(key)
-          ::NewRelic::Agent.logger.warn "Invalid value '#{value}' for #{key}, applying default value of '#{default}'"
+          NewRelic::Agent.logger.warn "Invalid value '#{value}' for #{key}, applying default value of '#{default}'"
           default
         end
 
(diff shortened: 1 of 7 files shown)


config/database.yml
```diff
@@ -1,5 +0,0 @@
-# This config is used when setting up a basic rails environment for running the
```
kaylareopelle, line 1: Did you mean to include this change in this PR? I'm not opposed to it as long as the tests pass, but it also looks like the sole change in https://github.com/newrelic/newrelic-ruby-agent/pull/2545

lib/new_relic/agent/configuration/manager.rb
```diff
@@ -154,8 +158,21 @@ def apply_transformations(key, value)
           end
         end
 
+        def enforce_allowlist(key, value)
+          return unless allowlist = default_source.allowlist_for(key)
+          return if allowlist.include?(value)
+
+          default = default_source.default_for(key)
+          ::NewRelic::Agent.logger.warn "Invalid value '#{value}' for #{key}, applying default value of '#{default}'"
```
kaylareopelle, line 166: I know you're matching the syntax for the other references in this file, but I don't think we need the `::` before `NewRelic` if we get around to cleaning this up in the future.

kaylareopelle (commented):
Thank you, James! I like this idea. Like we talked about in standup today, it could have more applications, such as:
* Updating our documentation (config, newrelic.yml) to display allowlists so that we don't need to include them in the description text
* Revisiting the `instrumentation.*` settings that permit only `auto`, `prepend`, `chain`, `disabled`

lib/new_relic/agent/configuration/manager.rb
fallwith: Addressed with [642ca83](https://github.com/newrelic/newrelic-ruby-agent/pull/2546/commits/642ca83eaf4b4a6812180eacab39675df7044c72)

config/database.yml
fallwith: Ah, I didn't switch back to dev before creating the branch. Fixed by merging that other branch to dev and then dev to here. Thanks for catching it!

PR #2543, opened 2024-04-05
WIP: Elasticsearch fix

kaylareopelle:
_Before contributing, please read our [contributing guidelines](https://github.com/newrelic/newrelic-ruby-agent/blob/main/CONTRIBUTING.md) and [code of conduct](https://github.com/newrelic/.github/blob/master/CODE_OF_CONDUCT.md)._

# Overview
Describe the changes present in the pull request

Submitter Checklist:
- [ ] Include a link to the related GitHub issue, if applicable
- [ ] Include a security review link, if applicable

# Testing
The agent includes a suite of unit and functional tests which should be used to
verify your changes don't break existing functionality. These tests will run with 
GitHub Actions when a pull request is made. More details on running the tests locally can be found 
[here for our unit tests](https://github.com/newrelic/newrelic-ruby-agent/blob/main/test/README.md), 
and [here for our functional tests](https://github.com/newrelic/newrelic-ruby-agent/blob/main/test/multiverse/README.md).
For most contributions it is strongly recommended to add additional tests which
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label
- [ ] Perform appropriate level of performance testing
- [ ] Confirm all checks passed
- [ ] Add version label prior to acceptance

PR #2547, opened 2024-04-11
Require OpenStruct where missing in tests

### LINKED ISSUE #2548: Fix OpenStruct test failures
Rake 13.2.0 stopped requiring OpenStruct: https://github.com/ruby/rake/pull/545 

In some of our tests, we were not explicitly requiring OpenStruct. Since our tests were run with rake, the need for a require was concealed. 

Add require statements for OpenStruct in files that use it without requiring it.

kaylareopelle:
Rake removed its requirement of OpenStruct in version 13.2.0. This caused our stripe multiverse suite to fail.

Since our tests were executed using Rake, we were able to use OpenStruct without a require in the file.

Now, OpenStruct is explicitly required in all the files it is used.

See: https://github.com/ruby/rake/pull/545 

Resolves #2548

### DIFF
--- a/test/multiverse/suites/curb/curb_test.rb
+++ b/test/multiverse/suites/curb/curb_test.rb
@@ -4,6 +4,7 @@
 
 require 'curb'
 require 'newrelic_rpm'
+require 'ostruct'
 require 'http_client_test_cases'
 require 'new_relic/agent/http_clients/curb_wrappers'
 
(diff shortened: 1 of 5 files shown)


kaylareopelle:
The only failing suite in this PR will be resolved by #2543

fallwith (approved):

hannahramadan (approved):

PR #2559, opened 2024-04-15
Logstasher instrumentation

hannahramadan:
Add support for LogStasher gem versions 1.0.0 and above. 

Full CI run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/9861763118
Test Coverage: 100% new code coverage


lib/new_relic/agent/instrumentation/logstasher/instrumentation.rb
```diff
@@ -0,0 +1,24 @@
...
+module NewRelic::Agent::Instrumentation
+  module LogStasher
+    INSTRUMENTATION_NAME = 'LogStasher'
+
+    def self.enabled?
+      NewRelic::Agent.config[:'instrumentation.logstasher'] != 'disabled'
+    end
+
+    def build_logstash_event_with_new_relic(data, tags)
+      data = yield
```
fallwith, line 14: I think it would be best to use a different variable name here that doesn't collide with the one used for the first method argument.

lib/new_relic/agent/instrumentation/logstasher/instrumentation.rb
```diff
@@ -0,0 +1,24 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic::Agent::Instrumentation
+  module LogStasher
+    INSTRUMENTATION_NAME = 'LogStasher'
```
fallwith, line 7:
```suggestion
    INSTRUMENTATION_NAME = NewRelic::Agent.base_name(name)
```

For instrumentation where the module name is an exact match, you can send `name` over to `base_name`. The intention behind doing so was to keep things dynamic and permit modules to be renamed in future, but granted we don't anticipate doing so often.

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -90,6 +90,42 @@ def record(formatted_message, severity)
...
+        if NewRelic::Agent.config[METRICS_ENABLED_KEY]
+          @counter_lock.synchronize do
+            @seen += 1
+            @seen_by_severity[severity] += 1
+          end
+        end
+
+        return if severity_too_low?(severity)
+        # The message key only exists if users manually add log statements
+        return if log.key?(:message) && (log.key?(:message).nil? || llog.key?(:message).empty?)
```
fallwith, line 107:
```suggestion
        return if log.key?(:message) && (log.key?(:message).nil? || log[:message].empty?)
```

NOTE: 2 changes. `llog` -> `log` and `.key?` -> `[]`

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -90,6 +90,42 @@ def record(formatted_message, severity)
...
+        return unless NewRelic::Agent.config[FORWARDING_ENABLED_KEY]
+        return if @high_security
+
+        txn = NewRelic::Agent::Transaction.tl_current
+        priority = LogPriority.priority_for(txn)
+        message = get_json_message(log)
```
fallwith, line 113: I think given the earlier check for `log[:message].empty?`, you can just do:

```suggestion
        message = log[:message]
```

and then remove the `get_json_message` method.

lib/new_relic/agent/instrumentation/logstasher/instrumentation.rb
hannahramadan: Great call out. [Updated that](https://github.com/newrelic/newrelic-ruby-agent/pull/2559/files#diff-fc9deb947cc1609417b7406320042eccc14b0a10853ab8a876ce0fc1cb4a81ceR14)

lib/new_relic/agent/log_event_aggregator.rb
hannahramadan: New [line](https://github.com/newrelic/newrelic-ruby-agent/pull/2559/files#diff-680b1e6a8547301e66b77f383e273ef141166e5986c9a1180625081dc8e271bcR98)

lib/new_relic/agent/log_event_aggregator.rb
hannahramadan: We still need to send an empty string in place of the message per spec. Since many logstasher logs won't have a message, doing `message = log['message']` [now](https://github.com/newrelic/newrelic-ruby-agent/pull/2559/files#diff-680b1e6a8547301e66b77f383e273ef141166e5986c9a1180625081dc8e271bcR113)

fallwith: My concern is that you can't reach this line with a `nil` message because of the early return you perform earlier when seeing a `nil` message. If `nil` is ok and needs to be converted to `''` to save space over a JSON `null`, that's cool - but I think you need to then drop the early return logic that prevents `nil`.

https://github.com/newrelic/newrelic-ruby-agent/pull/2559/files#diff-680b1e6a8547301e66b77f383e273ef141166e5986c9a1180625081dc8e271bcR98

hannahramadan: I was originally attempting to keep it in line with Logger instrumentation [logic](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/log_event_aggregator.rb#L73), but in LogStasher's case, we first check if there even is a `message` attribute because oftentimes for LogStasher there isn't. 

I suppose this check would only be hit if for some reason a user manually logged something with a `nil` or `''` message, which maybe should be allowed if that is their intent. Thoughts on removing this line entirely?

hannahramadan: By 'this check' and the one I'm thinking to remove, I mean [this](https://github.com/newrelic/newrelic-ruby-agent/pull/2559/files#diff-680b1e6a8547301e66b77f383e273ef141166e5986c9a1180625081dc8e271bcR98)

fallwith: Ah, my fault for misreading things. Sorry, @hannahramadan. I was commenting as if you were returning early by just evaluating `(log['message'].nil? || log['message'].empty?)`, but you're not - you're only applying that check to payloads that have a `'message'` key in the first place with your `if log.key?('message')` check that I skipped over / forgot about.

No changes necessary / great as-is.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+Version <dev> introduces instrumentation support for the LogStasher gem.
+
+- **Feature: Add instrumentation for LogStasher**
+  The agent will now record logs generated by [LogStasher](https://rubygems.org/gems/logstasher). Versions 1.0.0 and above of the LogStasher gem are supported. [#PR 2559](https://github.com/newrelic/newrelic-ruby-agent/pull/2559)
```
fallwith, line 7:
```suggestion
  The agent will now record logs generated by [LogStasher](https://github.com/shadabahmed/logstasher). Versions 1.0.0 and above of the LogStasher gem are supported. [#PR 2559](https://github.com/newrelic/newrelic-ruby-agent/pull/2559)
```

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -90,6 +91,38 @@ def record(formatted_message, severity)
         nil
       end
 
+      def record_logstasher_event(log)
+        return unless logstasher_enabled?
+
+        severity = log['level'] || 'UNKNOWN'
+        return if log.key?('message') && (log['message'].nil? || log['message'].empty?)
```
fallwith, line 98:
```suggestion
        return if log.key?('message') && (log['message'].nil? || log['message'].empty?)

        severity = log['level'] || 'UNKNOWN'
```

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -90,6 +91,38 @@ def record(formatted_message, severity)
...
+
+        if NewRelic::Agent.config[METRICS_ENABLED_KEY]
```
fallwith, line 100: It seems like much of the logic here: incrementing the seen and severity counts, checking for forwarding enabled, checking for high security, getting a transaction and its priority, etc. is shared between the `Logger` and `LogStasher` instrumentations. Perhaps some of it could go into helper methods that are shared by both.

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -166,19 +228,23 @@ def reset!
         super
       end
 
-      def enabled?
+      def logger_enabled?
         @enabled && @instrumentation_logger_enabled
       end
 
+      def logstasher_enabled?
+        @enabled && NewRelic::Agent::Instrumentation::LogStasher.enabled?
```
fallwith, line 236: If you're following the existing `Logger` pattern for `LogStasher`, perhaps you should have a dedicated `@instrumentation_logstasher_enabled` instance var set at initialize time to match `Logger`.

test/multiverse/suites/logstasher/Envfile
```diff
@@ -0,0 +1,18 @@
...
+instrumentation_methods :chain, :prepend
+
+logstasher_versions = [
+  [nil, 2.6]
+]
+
+def gem_list(logstasher_versions = nil)
+  <<~RB
+    gem 'logstasher'#{logstasher_versions}
+    gem 'activesupport', '7.0.8.4'
```
fallwith, line 14: `logstasher` itself depends on `activesupport` and Bundler should automatically bring it in as a dependency. It looks like `logstasher` v1 wants `activesupport` v4+ and `logstasher` v2 wants `activesupport` v5+. What does specifying `'7.0.8.4'` give us? Can we remove this line entirely or do something like `>= 7` if you depend on a major version for your testing?

lib/new_relic/agent/log_event_aggregator.rb
fallwith: (return early before bothering to set the severity)

test/multiverse/suites/logstasher/logstasher_instrumentation_test.rb
```diff
@@ -0,0 +1,123 @@
...
+  end
+
+  def test_level_is_recorded
+    in_transaction do
+      LogStasher.build_logstash_event({'level' => :info, 'message' => 'hi there'}, ['log'])
+    end
+    _, events = @aggregator.harvest!
+
+    assert_equal :info, events[0][1]['level']
+    assert_metrics_recorded(%w[Logging/lines/info])
```
kaylareopelle, line 48: The capitalization of "info" is different from the capitalization of "UNKNOWN" in the following test. I believe this means that we'll get different metrics and counts for "Logging/lines/info" and "Logging/lines/INFO".

I don't see this specifically mentioned in the spec, but the reference always has a capitalized severity. I think we want all the severity levels to be capitalized within the metric name regardless of how they're spelled when we perform the count in order to have an accurate count of that log severity category in general. 

I think we got away with not caring about this for the Ruby logger since it has consistent severity capitalization in its constants, but this issue may exist for that instrumentation too.

test/multiverse/suites/logstasher/logstasher_instrumentation_test.rb
```diff
@@ -0,0 +1,123 @@
...
+
+  def test_attributes_added_to_payload
+    in_transaction do
+      LogStasher.build_logstash_event(json_log_hash, ['log'])
+    end
+    _, events = @aggregator.harvest!
+
+    assert events[0][1]['attributes'].key?(:identifier)
+    assert events[0][1]['attributes'].key?(:name)
+    assert events[0][1]['attributes'].key?('source')
```
kaylareopelle, line 78: Nice job checking for compatibility across strings and symbols! Does New Relic turn them all into strings in the end?

test/multiverse/suites/logstasher/logstasher_instrumentation_test.rb
```diff
@@ -0,0 +1,123 @@
...
+  def test_log_decorating_records_new_relic_metrics
+    with_config(:'application_logging.local_decorating.enabled' => false) do
+      LogStasher.warn('yikes')
+    end
+    logfile = JSON.parse(@written.string)
+
+    refute logfile.key?('entity.name')
+    refute logfile.key?('entity.type')
+    refute logfile.key?('hostname')
+  end
```
kaylareopelle, line 101: I'm not sure the names of these tests make sense to me based on what they're testing. Maybe something including `service_linking_metadata` would make more sense?

test/multiverse/suites/logstasher/logstasher_instrumentation_test.rb
```diff
@@ -0,0 +1,123 @@
...
+
```
kaylareopelle, line 102: It's nice to see that the service linking metadata works with local log decorating. Would you mind adding tests to verify the trace linking metadata works as well?

test/new_relic/agent/log_event_aggregator_test.rb
```diff
@@ -15,6 +15,7 @@ def setup
 
       @enabled_config = {
         :'instrumentation.logger' => 'auto',
+        :'instrumentation.logstasher' => 'auto',
```
kaylareopelle, line 18: These tests are getting me kind of curious about the Ruby Logger / Logstasher instrumentation interaction. Did you experiment with an application that had both Ruby Logger and Logstasher logs recorded?

test/new_relic/agent/log_event_aggregator_test.rb
```diff
@@ -521,5 +527,164 @@ def test_records_log_events_not_within_default_severities
...
+
+      assert_equal 'A trex is near', events[0][1]['message']
+    end
+
+    def test_record_json_returns_empty_string_when_unavaliable
+      @aggregator.record_logstasher_event({'level' => :info})
+      _, events = @aggregator.harvest!
+
+      assert_equal '', events[0][1]['message']
+    end
```
kaylareopelle, line 577: I think this is [a departure from the spec.](https://source.datanerd.us/agents/agent-specs/blob/main/Application-Logging.md#nodejs-exceptions) I believe this attribute is supposed to be omitted if it is nil, empty, missing or simply whitespace.

Why does a missing "message" key create one equal to an empty string, whereas a "message" key equal to an empty string is not sent at all?

The same behavior is [tested here](https://github.com/newrelic/newrelic-ruby-agent/pull/2559/files#diff-b95067722289f3e6cbeef3d275c924f8900cbf9ea650450eecc910ac951e75deR61).

test/new_relic/agent/log_event_aggregator_test.rb
```diff
@@ -521,5 +527,164 @@ def test_records_log_events_not_within_default_severities
...
+
+    def test_add_json_event_attributes_deletes_already_recorded_attributes
+      @aggregator.record_logstasher_event({'message' => 'bye', 'level' => :info, '@timestamp' => 'now'})
+      _, events = @aggregator.harvest!
+
+      refute_includes(events[0][1]['attributes'], 'message')
+      refute_includes(events[0][1]['attributes'], 'level')
+      refute_includes(events[0][1]['attributes'], '@timestamp')
+    end
```
kaylareopelle, line 594: What do you think about adding one attribute that wouldn't fall under the "already recorded" category? This might help us make sure attributes aren't being deleted if they're present alongside the "already recorded" attributes? 

Also, could you add some expectations to verify these attributes are recorded elsewhere in the event?

test/new_relic/agent/log_event_aggregator_test.rb
```diff
@@ -521,5 +527,164 @@ def test_records_log_events_not_within_default_severities
...
+
+    def test_logstasher_forwarding_disabled
```
kaylareopelle, line 596:
```suggestion
    def test_logstasher_forwarding_disabled_and_high_security_enabled
```

test/new_relic/agent/log_event_aggregator_test.rb
```diff
@@ -521,5 +527,164 @@ def test_records_log_events_not_within_default_severities
...
+    end
+
+    def test_nil_when_record_logstasher_errors
+      @aggregator.stub(:severity_too_low?, -> { raise 'kaboom' }) do
+        @aggregator.record_logstasher_event({'level' => :warn, 'message' => 'A trex is near'})
+        _, results = @aggregator.harvest!
+
+        assert_empty results
+      end
+    end
```
kaylareopelle, line 688: This gets me curious about what the logstasher log events look like when an error is raised. The spec has some special attributes to send in those scenarios (ex. `error.stack`), but I don't believe we use them in the logger instrumentation.

IIRC there were some special attributes when logstasher had an error. How does logstasher behave if the log has a related error?

kaylareopelle (changes requested):

lib/new_relic/agent/log_event_aggregator.rb
hannahramadan: I like this! Created some shared methods [here](https://github.com/newrelic/newrelic-ruby-agent/pull/2559/commits/4fabc5bb9191ae478d560eebb9fce9e0d2b03dd3) and [here](https://github.com/newrelic/newrelic-ruby-agent/pull/2559/commits/229c7d1a80992e3d80c52d087cb9add67f5dda3b)

test/multiverse/suites/logstasher/Envfile
hannahramadan: I'm running into testing issues with 7.1, which is why I'm explicity bringing in activesupport. I've changed the pin to look [nicer](https://github.com/newrelic/newrelic-ruby-agent/pull/2559/commits/278c972317055a976cf3a7efc0263ee4c876dd7e) for starters.

test/multiverse/suites/logstasher/logstasher_instrumentation_test.rb
hannahramadan: Thank you! This is a good call out. I've gone ahead and made sure severity levels are always capitalized https://github.com/newrelic/newrelic-ruby-agent/pull/2559/commits/45e6bc4deb266e90fcc683cefe42a1c53cc655af

test/new_relic/agent/log_event_aggregator_test.rb
hannahramadan: Oohh you're absolutely right. On re-reading the spec, we shouldn't be recording a log without a message unless it supports a mapped context functionality. Logstasher logs don't have a message (the only time they do is when a user adds their own log statement). But they do have a lot of other contextual information about events, so I think we could consider it falling into the supporting context attributes. Although this may be worth talking about further live.

For now, I've stopped recording a message when one isn't avaliable. https://github.com/newrelic/newrelic-ruby-agent/pull/2559/commits/f689fb7b0444d05e8a7f1acd284a48e710994fc9

test/new_relic/agent/log_event_aggregator_test.rb
hannahramadan: Great idea! https://github.com/newrelic/newrelic-ruby-agent/pull/2559/commits/84c64b3374ec74d65b9e3fe130128c648d8cf670

test/new_relic/agent/log_event_aggregator_test.rb
hannahramadan: If LogStasher has an error, it adds an `error` key to the hash with the stack trace as the value. I wasn't able to find anything in Logger instrumentation that suggested we do anything special with errors. 

I'm torn between molding the LogStasher data to fit our spec, or simply reporting their data as it is. For example with errors, right now we currently report `error` like any other attribute and it includes the info a LogStasher user might expect. To fit the spec, I'd break the error into two new key/values: `error.stack` and `error.class`.

test/new_relic/agent/log_event_aggregator_test.rb
hannahramadan: When both are enabled, we report both loggers outputs. I could see this being unwanted behavior and a suprise to a user. I wonder if loggers outside of Logger should be an opt in. What do you think?

test/multiverse/suites/logstasher/logstasher_instrumentation_test.rb
hannahramadan: Should've done that to start - thank you! https://github.com/newrelic/newrelic-ruby-agent/pull/2559/commits/32036e5f3e38eef5d362f9e022ba020b57d8b5c8

test/multiverse/suites/logstasher/logstasher_instrumentation_test.rb
hannahramadan: Too much credit 🫣 This is how the LogStasher reports - some strings, some symbols. Do you think it is worth changing everything to a common class?

kaylareopelle: I think that's a good idea, but I don't think you need to tackle it as part of this PR. Maybe open a separate technical debt ticket?

test/new_relic/agent/log_event_aggregator_test.rb
kaylareopelle: That sounds good to me! Happy to talk more about it live if you'd like.

test/new_relic/agent/log_event_aggregator_test.rb
kaylareopelle: We chatted a little more about this one synchronously. I like leaning on reporting the data as logstasher does, but am open to the approach that's closer to the spec as well.

lib/new_relic/agent/log_event_aggregator.rb
hannahramadan: This [commit](https://github.com/newrelic/newrelic-ruby-agent/pull/2559/commits/02d09b3fba028b4dab5a46edc6e95cdce2d84eee) matches Logger. Load order was an issue, so we aren't able to use the dynamic `NewRelic::Agent.base_name(name)` and need to require LogStasher instrumentation early on. I'm leaning more towards not having a LogStasher instance var set at initialize so that we add less code that may have to be adjusted when more loggers are added. But no strong opinion either way

test/new_relic/agent/log_event_aggregator_test.rb
hannahramadan: Chatted live - lets keep closer to how logstasher reports data.

test/new_relic/agent/log_event_aggregator_test.rb
hannahramadan: The look different enough! I suppose if someone is using both loggers, we can report both. It may be a surprise to some users, but also an easy switch to turn off. Lets keep it enbled to keep with precedence of auto enabling!

kaylareopelle: Sounds good to me!

test/multiverse/suites/logstasher/Envfile
hannahramadan: LogStasher has tested up to Rails 6.1.x, I wonder if there is something going on in Rails 7.1 that is incompatiable.

lib/new_relic/agent/log_event_aggregator.rb
hannahramadan: I've reverted this change for now!

test/multiverse/suites/logstasher/Envfile
kaylareopelle: Rails 7.1 introduced a change to their logging system, so that it uses the `ActiveSupport::BroadcastLogger` instead of `ActiveSupport::Logger.broadcast`. I wonder if that's interfering with LogStasher. 

Here's the PR I made a while back to allow our logger instrumentation to work with it: https://github.com/newrelic/newrelic-ruby-agent/pull/2252

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -58,38 +59,68 @@ def capacity
       end
 
       def record(formatted_message, severity)
-        return unless enabled?
+        return unless logger_enabled?
 
         severity = 'UNKNOWN' if severity.nil? || severity.empty?
+        increment_event_counters(severity)
 
-        if NewRelic::Agent.config[METRICS_ENABLED_KEY]
```
hannahramadan, line 65: A few blocks of code were shared by `create_event` and `create_logstasher_event`. This is an example of a block of code moved into a new method (`increment_event_counters` is the name for this)

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -58,38 +59,68 @@ def capacity
...
           end
         end
+      rescue
+        nil
+      end
 
-        return if severity_too_low?(severity)
-        return if formatted_message.nil? || formatted_message.empty?
-        return unless NewRelic::Agent.config[FORWARDING_ENABLED_KEY]
-        return if @high_security
```
hannahramadan, line 75: This block of code lives in a new method, `monitoring_conditions_met?`

test/multiverse/suites/logstasher/logstasher_instrumentation_test.rb
hannahramadan: I chose strings because we define string constants for other attributes. But can easily transform into symbols. I know symbols are usually more performant, but I dont think it matters in this case since everytime we iterate over log attributes, they are technically new strings (or symbols). https://github.com/newrelic/newrelic-ruby-agent/pull/2559/commits/3168617c2f006cebb6b94bdb86d9af42ac65c11f

hannahramadan: This caused some expected Ruby 2.4 errors. I'm reverting for now

kaylareopelle: I think it's fine to leave this as-is. Now that I think about it more, we'll transform everything into strings to send the data as JSON. Sorry for the confusion!

test/multiverse/suites/logstasher/Envfile
hannahramadan: I just tested a demo app using logstasher instrumentation and Rails 7.1. Everything seemed to work fine, so I think this issue might be with out testing environment.

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -58,38 +59,68 @@ def capacity
...
+        log['level'] ? log['level'].to_s.upcase : 'UNKNOWN'
+      end
+
+      def increment_event_counters(severity)
+        if NewRelic::Agent.config[METRICS_ENABLED_KEY]
+          @counter_lock.synchronize do
+            @seen += 1
+            @seen_by_severity[severity] += 1
+          end
+        end
```
fallwith, line 121:
```suggestion
        return unless NewRelic::Agent.config[METRICS_ENABLED_KEY]
        
        @counter_lock.synchronize do
          @seen += 1
          @seen_by_severity[severity] += 1
        end
```

test/multiverse/suites/logstasher/Envfile
```diff
@@ -0,0 +1,18 @@
...
+instrumentation_methods :chain, :prepend
+
+logstasher_versions = [
+  [nil, 2.7]
+]
+
+def gem_list(logstasher_versions = nil)
+  <<~RB
+    gem 'logstasher'#{logstasher_versions}
+    gem 'activesupport', '~> 7.0.0'
```
fallwith, line 14: From your last explanation of why we need to bring in ActiveSupport, I believe you really just want to enforce a minimum version. We should be safe to test with v7.1 that's out now and v8.0 in future. How about adding a comment here that reminds of why we need to bring in ActiveSupport, and how about simply using `>=`?

```suggestion
    gem 'activesupport', '>= 7'
```

hannahramadan: When running a playground locally, there doesn't seem to be an issue. But multiverse tests have an issue with versions above 7.1 that I couldnt figure out/ not sure if it was worth the effort (its a `NoMethodError` coming from activesupport 7.1)

fallwith: Ah, I got it flipped then. You would need `< 7.1`, not `>= 7`. Ok, if it's a ceiling and not a minimum, then perhaps the code can stay as-is but just get the comment reminding us of the problem. Logstasher itself should have its gemspec updated to denote the <7.1 constraint.

fallwith (approved):

kaylareopelle (approved):

PR #2583, opened 2024-04-22
async-http: Use v0.59 for Ruby 2

fallwith:
While async-http v0.60 is itself ok with Ruby 2, it's dependency on a newer async version is not.


kaylareopelle (approved):
Thank you!

PR #2610, opened 2024-04-29
ActiveSupport notifications: allowlist event names

fallwith:
Added a new :'instrumentation.active_support_notifications.active_support_events' configuration parameter to permit a user defined allowlist of Active Support notifications event names for the agent to subscribe to that are related to the Active Support library itself.

Event names that are not included on the list will be ignored by the agent

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1427,6 +1427,11 @@ module NewRelic
             on caching. Any event name not included in this list will be ignored by the agent. Provide complete event
             names such as 'cache_fetch_hit.active_support'. Do not provide asterisks or regex patterns, and do not
             escape any characters with backslashes.
+
+            For a complete list of all possible Active Support event names, see the
+            [list of caching names](https://guides.rubyonrails.org/active_support_instrumentation.html#active-support-—-caching)
+            and the [list of messages names](https://guides.rubyonrails.org/active_support_instrumentation.html#active-support-—-messages)
+            from the official Rails documentation.
           ACTIVE_SUPPORT_EVENTS
         },
         :'instrumentation.active_support_broadcast_logger' => {
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1429,8 +1429,8 @@ module NewRelic
             escape any characters with backslashes.
 
             For a complete list of all possible Active Support event names, see the
-            [list of caching names](https://guides.rubyonrails.org/active_support_instrumentation.html#active-support-—-caching)
-            and the [list of messages names](https://guides.rubyonrails.org/active_support_instrumentation.html#active-support-—-messages)
+            [list of caching names](https://edgeguides.rubyonrails.org/active_support_instrumentation.html#active-support-caching)
+            and the [list of messages names](https://edgeguides.rubyonrails.org/active_support_instrumentation.html#active-support-messages)
             from the official Rails documentation.
           ACTIVE_SUPPORT_EVENTS
         },
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -4,6 +4,7 @@
 
 require 'forwardable'
 require_relative '../../constants'
+require_relative '../instrumentation/active_support_subscriber'
 
 module NewRelic
   module Agent
@@ -1415,6 +1416,19 @@ module NewRelic
           :description => 'Configures the TCP/IP port for the trace observer Host'
         },
         # Instrumentation
+        :'instrumentation.active_support_notifications.active_support_events' => {
+          :default => NewRelic::Agent::Instrumentation::ActiveSupportSubscriber::EVENT_NAME_TO_METHOD_NAME.keys,
+          :public => true,
+          :type => Array,
+          :allowed_from_server => false,
+          :description => <<~ACTIVE_SUPPORT_EVENTS.chomp.tr("\n", ' ')
+            An allowlist array of Active Support notifications events specific to the Active Support library
+            itself that the agent should subscribe to. The Active Support library specific events focus primarily
+            on caching. Any event name not included in this list will be ignored by the agent. Provide complete event
+            names such as 'cache_fetch_hit.active_support'. Do not provide asterisks or regex patterns, and do not
+            escape any characters with backslashes.
+          ACTIVE_SUPPORT_EVENTS
+        },
         :'instrumentation.active_support_broadcast_logger' => {
           :default => instrumentation_value_from_boolean(:'application_logging.enabled'),
           :documentation_default => 'auto',
--- a/lib/new_relic/agent/instrumentation/active_support.rb
+++ b/lib/new_relic/agent/instrumentation/active_support.rb
@@ -18,6 +18,10 @@ DependencyDetection.defer do
       !NewRelic::Agent::Instrumentation::ActiveSupportSubscriber.subscribed?
   end
 
+  depends_on do
+    !NewRelic::Agent.config[EVENT_NAMES_PARAMETER].empty?
+  end
+
   executes do
     NewRelic::Agent.logger.info('Installing ActiveSupport instrumentation')
   end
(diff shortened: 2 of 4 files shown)


lib/new_relic/agent/instrumentation/active_support.rb
```diff
@@ -21,7 +23,8 @@
   end
 
   executes do
-    ActiveSupport::Notifications.subscribe(/\.active_support$/,
+    event_names = NewRelic::Agent.config[EVENT_NAMES_PARAMETER].map { |n| Regexp.escape(n) }.join('|')
```
hannahramadan, line 26: Is this code block saying we won't subscribe to anything if the `EVENT_NAMES_PARAMETER` array is empty aka someone didn't configure it?

fallwith: Hmm... well, it won't be empty by default. It'll be set equal to the list of keys in the `EVENT_NAME_TO_METHOD_NAME` constant in the subscriber class. However, you've pointed out a good edge case to defend against here. It is possible that a user could manually set it `[]` instead of disabling the instrumentation. I'll add a check to guard against that. Thanks!

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1415,6 +1416,19 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :description => 'Configures the TCP/IP port for the trace observer Host'
         },
         # Instrumentation
+        :'instrumentation.active_support_notifications.active_support_events' => {
```
kaylareopelle, line 1419: I like how this is scoped for the specific GTSE. 

I wonder if we want to organize this config to support all active support events in one allowlist/denylist kind of like we have set up for our include/exclude lists with other configs. I'm not sure if a big list or one scoped to the specific instrumentation is more user-friendly.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1415,6 +1416,19 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :default => NewRelic::Agent::Instrumentation::ActiveSupportSubscriber::EVENT_NAME_TO_METHOD_NAME.keys,
+          :public => true,
+          :type => Array,
+          :allowed_from_server => false,
+          :description => <<~ACTIVE_SUPPORT_EVENTS.chomp.tr("\n", ' ')
+            An allowlist array of Active Support notifications events specific to the Active Support library
+            itself that the agent should subscribe to. The Active Support library specific events focus primarily
+            on caching. Any event name not included in this list will be ignored by the agent. Provide complete event
+            names such as 'cache_fetch_hit.active_support'. Do not provide asterisks or regex patterns, and do not
+            escape any characters with backslashes.
```
kaylareopelle, line 1429: Reading this brought me back to the Rails guides on instrumentation. It looks like the anchor links aren't quite working, but I'd like to incorporate a link to the full list or list the events we support somewhere in the description.

https://guides.rubyonrails.org/active_support_instrumentation.html 

I also found it interesting that the documentation separates the notifications into "caching" and "messaging" notifications for active support.

lib/new_relic/agent/instrumentation/active_support_subscriber.rb
```diff
@@ -18,23 +31,9 @@ def add_segment_params(segment, payload)
 
         def metric_name(name, payload)
           store = payload[:store]
-          method = method_from_name(name)
+          method = EVENT_NAME_TO_METHOD_NAME.fetch(name, name.delete_prefix('cache_').delete_suffix('.active_support'))
```
kaylareopelle, line 34: Neat! I don't think I knew about these methods.

lib/new_relic/agent/instrumentation/active_support_subscriber.rb
```diff
@@ -2,12 +2,25 @@
...
+          'cache_fetch_hit.active_support' => 'fetch_hit',
+          'cache_generate.active_support' => 'generate',
+          'cache_read.active_support' => 'read',
+          'cache_write.active_support' => 'write',
+          'cache_delete.active_support' => 'delete',
+          'cache_exist?.active_support' => 'exist?',
+          'cache_read_multi.active_support' => 'read_multi',
+          'cache_write_multi.active_support' => 'write_multi',
+          'cache_delete_multi.active_support' => 'delete_multi',
+          'message_serializer_fallback.active_support' => 'message_serializer_fallback'
```
kaylareopelle, line 21: When I was looking at the docs, in the '[3.8 Active Support — Messages](https://edgeguides.rubyonrails.org/active_support_instrumentation.html#active-support-%E2%80%94-messages)' section, (anchor link might not work) it seems like the payload is pretty different from the payload for the caching events. I wonder if we should create a different instrumentation module for message notifications, even though there's only one.

Instead of key/store/etc. The `message` notification's payload includes serializer/fallback/serialized/deserialized

lib/new_relic/agent/configuration/default_source.rb
fallwith: My initial thought here was to have the `active_support_events` part of the `:'instrumentation.active_support_notifications.active_support_events'` parameter be different for every different notifications based instrumentation option.

But I can see benefits to holistically re-evaluating our entire design. I can see some good pros that could come out of big lists and fewer subscriptions if we could define a good generic central class whose instance would be associated with the subscriptions.

So I think I'd like to either keep a list per instrumentation that we have a dedicated separate `disable_` parameter for, or start fresh looking at notifications without any baggage from the existing codebase and see what new consolidated patterns we could come up with for both speed and ease of use.

lib/new_relic/agent/instrumentation/active_support_subscriber.rb
fallwith: Created #2615

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: Nice! If we see some performance or other user-facing benefits with a consolidated pattern, perhaps we could pitch that as a cornerstone feature of a 10.x major release.

I agree with what you're saying and want to move forward with your current design.

lib/new_relic/agent/instrumentation/active_support_subscriber.rb
kaylareopelle: Thank you!

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: Great insight. I really want this link to work! Here's a PR that fixes the problem: https://github.com/rails/rails/pull/51696

lib/new_relic/agent/instrumentation/active_support_subscriber.rb
fallwith: Apparently neither did Ruby 2.4. 😂

lib/new_relic/agent/configuration/default_source.rb
fallwith: I've added links (that may need to change) with [6a92c5d](https://github.com/newrelic/newrelic-ruby-agent/pull/2610/commits/6a92c5d7238fe0316081a74c8d02dcdaf37f8682)

lib/new_relic/agent/configuration/default_source.rb
fallwith: Created https://github.com/newrelic/newrelic-ruby-agent/issues/2616

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1415,6 +1416,24 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+
+            For a complete list of all possible Active Support event names, see the
+            [list of caching names](https://guides.rubyonrails.org/active_support_instrumentation.html#active-support-—-caching)
+            and the [list of messages names](https://guides.rubyonrails.org/active_support_instrumentation.html#active-support-—-messages)
+            from the official Rails documentation.
```
kaylareopelle, line 1434:
```suggestion
            For a complete list of all possible Active Support event names, see the
            [list of caching names](https://edgeguides.rubyonrails.org/active_support_instrumentation.html#active-support-caching)
            and the [list of messages names](https://edgeguides.rubyonrails.org/active_support_instrumentation.html#active-support-messages)
            from the official Rails documentation.
```

kaylareopelle: ☝️ to incorporate the new URLs. Only the edge guides were updated.

fallwith:
Closing for now. Please leave the branch alone. Tracking future developments with #2685.

PR #2619, opened 2024-05-02
bugfix: expected HTTP codes / error classes / error messages shouldn't impact Apdex

### LINKED ISSUE #2594: Errors that are expected by way of expected_status_codes should not impact Apdex
We have reports that errors with status codes matching specified `expected_status_codes` values are correctly appearing in the Errors Inbox as "expected", but are negatively impacting Apdex.

This relevant specification text confirms that expected errors should not impact Apdex.

> Expected errors MUST create traced errors and error events, but MUST NOT increment any of the Errors/* metrics in order to prevent error rate and Apdex modifications.

We'll need to determine where in the agent or back-end the denotation of an error as expected fails to prevent that error from impacting Apdex.

NOTE: The use of `ignore_status_codes` works as a workaround. Ignored errors are seen as not impacting Apdex. The drawback of the workaround is that ignored errors are not delivered to the Errors Inbox.

fallwith:
If an error is either expected (which still creates an entry in the Errors Inbox) or ignored (which does not), the transaction's Apdex should not factor in the error.

A bug was discovered involving errors that were expected by way of their HTTP response codes that would cause those errors to impact Apdex. Errors that were expected via the agent's Noticed Errors API would correctly not impact Apdex, but HTTP response code based expected errors would incorrectly do so.

Now ignored errors and both kinds of expected errors will correctly not impact Apdex when observed during a transaction.

**UPDATE:** NOTE: the key change introduced here is that we no longer simply check a New Relic error object for an `expected: true` attribute value, but that we also perform an `ErrorCollector#expected?` check. The attribute value can only be set by invoking the `notice_error` API, but the `#expected?` check will examine other things including the HTTP status code (if present), the customer defined expected error class list, and the customer defined expected error message list. So now all ways of configuring the agent to expect an error should result in any matching errors no longer having any impact on Apdex.

resolves #2594


test/agent_helper.rb
```diff
@@ -228,7 +228,9 @@ def assert_metrics_recorded(expected)
   expected.each do |specish, expected_attrs|
     expected_spec = metric_spec_from_specish(specish)
     actual_stats = NewRelic::Agent.instance.stats_engine.to_h[expected_spec]
-    if !actual_stats
+    if actual_stats
+      assert(actual_stats)
```
fallwith, line 232: Previously if `assert_metrics_recorded` was called to simply check that the metric itself was recorded and no expected attributes were specified, no Minitest assertions would actually be performed. Now an `assert` will be called.

florianpilz (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+Version <dev> fixes a bug that would cause an error and a correspondong expected HTTP status code to negatively impact a transaction's Apdex.
```
```suggestion
Version <dev> fixes a bug that would cause an error and a corresponding expected HTTP status code to negatively impact a transaction's Apdex.
```

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
...
+
+- **Bugfix: Expected HTTP status code / error class / error message based errors won't impact Apdex
```
```suggestion
- **Bugfix: Expected errors related to HTTP status code, class, and message won't impact Apdex
```

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
...
+
+  The agent is supposed to prevent observed application errors from negatively impacting Apdex if the errors are to be either ignored or expected. There are 2 ways for the agent to expect an error: via the `notice_error` API receiving an `expected: true` argument or via a matches made against the configuration file based user defined lists for expected HTTP status codes (`:'error_collector.expected_status_codes'`), expected error classes (`:'error_collector.expected_classes'`), or expected error messages (`:'error_collector.expected_messages'`). Previously only errors expected via the `notice_error` API were correctly prevented from impacting Apdex and the configuration parameter based expected errors were incorrectly impacting Apdex. This behavior has been fixed and now both types of expected errors will correctly not impact Apdex. Thanks very much to [@florianpilz](https://github.com/florianpilz) for bringing this issue to our attention. [PR#2619](https://github.com/newrelic/newrelic-ruby-agent/pull/2619)
```
```suggestion
  The agent is supposed to prevent observed application errors from negatively impacting Apdex if the errors are either ignored or expected. There are two ways for the agent to expect an error: via the `notice_error` API receiving an `expected: true` argument or via matches made against user-configured lists for expected HTTP status codes (`:'error_collector.expected_status_codes'`), expected error classes (`:'error_collector.expected_classes'`), or expected error messages (`:'error_collector.expected_messages'`). Previously, only errors expected via the `notice_error` API were correctly prevented from impacting Apdex. Expected errors set by configuration incorrectly impacted Apdex. This behavior has been fixed and now both types of expected errors will correctly not impact Apdex. Thanks very much to [@florianpilz](https://github.com/florianpilz) for bringing this issue to our attention. [PR#2619](https://github.com/newrelic/newrelic-ruby-agent/pull/2619)
```

lib/new_relic/agent/error_collector.rb
```diff
@@ -110,6 +110,29 @@ def error_is_ignored?(error, status_code = nil)
...
+      # Ignored errors are checked via `#error_is_ignored?`
+      # Expected errors are checked in 2 separate ways:
+      #   1. The presence of an `expected: true` attribute key/value pair in the
+      #      options hash, which will be set if that key/value pair was used in
+      #      the `notice_error` public API.
+      #   2. By calling `#expected?` which in turn calls `ErrorFilter#expected?`
+      #      which checks for 3 things:
+      #        - A match for user defined HTTP status codes to expect
+      #        - A match for user defined error classes to expect
+      #        - A match for user defined error messages to expect
```
kaylareopelle, line 124:
```suggestion
      #        - A match for user-defined HTTP status codes to expect
      #        - A match for user-defined error classes to expect
      #        - A match for user-defined error messages to expect
```

test/new_relic/agent/transaction_test.rb
```diff
@@ -339,6 +342,49 @@ def test_apdex_success_with_ignored_error
...
+
+      # this one does impact Apdex
+      in_web_transaction(txn_name) do
+        Transaction.notice_error(RuntimeError.new)
+      end
+
+      assert_metrics_recorded(
+        txn_name => {},
+        'Apdex' => {:apdex_s => 1, :apdex_t => 0, :apdex_f => 1},
+        'Apdex/BigDweebEnergy' => {:apdex_s => 1, :apdex_t => 0, :apdex_f => 1}
```
kaylareopelle, line 360: I'm not very familiar with this part of the codebase. If both incorrectly impacted Apdex, what would these scores look like?

lib/new_relic/agent/transaction.rb
```diff
@@ -816,13 +816,9 @@ def record_queue_time
       end
 
       def had_error_affecting_apdex?
-        @exceptions.each do |exception, options|
-          ignored = NewRelic::Agent.instance.error_collector.error_is_ignored?(exception)
-          expected = options[:expected]
-
-          return true unless ignored || expected
+        @exceptions.each.any? do |exception, options|
```
kaylareopelle, line 819: Do you recall the structure of `@exceptions`? From this file it looks like a Hash, but I'm not sure what's inside the hash. (Can also talk about this sync in standup)

fallwith: Hi @kaylareopelle! Yep. `@exceptions` here is a hash. It looks like this:

```ruby
{ my_error => { expected: false },
  my_other_error => { expected: true } }
```

The key is an instance of a Ruby error class and the value is an options hash related to how the agent is expected to treat that error. In this context we're only concerned with the `:expected` key inside that inner hash.

fallwith: Previously we were using an early `return true` if any key/value pair was found to satisfy the criteria, and then using `return false` if none were found. I refactored that to leverage `Enumerable#any?` instead.

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+Version <dev> fixes a bug that would cause an error and a corresponding expected HTTP status code to negatively impact a transaction's Apdex.
```
tannalynn, line 5: the way this sentence is worded to me sounds like its only about errors related to status codes, rather than all expected errors. My understanding is this is more about expected errors in general than the http status codes config (correct me if i'm wrong of course!). maybe this could be reworded for clarity?
something like this maybe?
```suggestion
Version <dev> fixes a bug that would cause an expected error to negatively impact a transaction's Apdex.
```

fallwith: Great call, @tannalynn. The fix includes routing the "should this error impact apdex?" check through a complete "is the error expected?" check, which will indeed check status, class, and message. But I got tunnel vision when it came to writing the text description.

test/new_relic/agent/transaction_test.rb
fallwith: I borrowed this pattern from the ignored errors tests in which one is ok and one is bad.

Given the 2 transactions, one should good (`apdex_s = 1`) and one should be bad (`apdex_f = 1`). If both errors impacted apdex, none would be good (`apdex_s = 0`) and both would be bad (`apdex_f = 2`).

This particular test which uses (`expected: true`) should pass with what we have now in dev. We just unfortunately didn't previously have any test coverage for the behavior. And then the new tests below this point should fail in dev and require the changes delivered by this PR to pass.

tannalynn (approved):

test/new_relic/agent/transaction_test.rb
kaylareopelle: Thank you!

kaylareopelle (approved):

PR #2623, opened 2024-05-03
Don't start for 'rails runner' and 'rails db'

### LINKED ISSUE #2596: bin/rails subcommands other than "server" continue to involve the agent but shouldn't
PR #2239 was expected make the agent compatible with Rails' standardization around a single `rails` CLI entrypoint and ensure that by default the agent would not be introduced into contexts where the majority of customers would not derive New Relic value from.

However, the submitter of Issue #1313 that was closed by the PR [reports that the agent is still being introduced](https://github.com/newrelic/newrelic-ruby-agent/issues/1313#issuecomment-2074454325).

We should revisit our solution and determine what we can do to improve.

fallwith:
Add `Rails::Command::RunnerCommand` for `bin/rails runner` commands and `Rails::Command::RakeCommand` for `bin/rails db` command to the list of denylisted constants that will prevent the agent from autostarting when seen.

Now when `bin/rails runner` and `bin/rails db` commands are ran, the agent will recognize those contexts and not start
automatically.

Older agent versions that are still supported by New Relic can update to the new list of denylisted constants by having the following line added to the `newrelic.yml` configuration file:

```yaml
autostart.denylisted_constants: "Rails::Command::ConsoleCommand,Rails::Command::CredentialsCommand,Rails::Command::Db::System::ChangeCommand,Rails::Command::DbConsoleCommand,Rails::Command::DestroyCommand,Rails::Command::DevCommand,Rails::Command::EncryptedCommand,Rails::Command::GenerateCommand,Rails::Command::InitializersCommand,Rails::Command::NotesCommand,Rails::Command::RakeCommand,Rails::Command::RoutesCommand,Rails::Command::RunnerCommand,Rails::Command::SecretsCommand,Rails::Console,Rails::DBConsole"
```

resolves #2596

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,21 @@
 
 ## dev
 
-Version <dev> fixes a bug that would cause an expected error to negatively impact a transaction's Apdex.
+Version <dev> fixes a bug that would cause an expected error to negatively impact a transaction's Apdex, and fixes the agent's autostart logic so that by default `rails runner` and `rails db` commands will not cause the agent to start.
 
 - **Bugfix: Expected errors related to HTTP status code, class, and message won't impact Apdex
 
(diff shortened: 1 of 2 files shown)


hannahramadan (approved):
thanks for explaing live!

kaylareopelle (approved):

PR #2635, opened 2024-05-17
added lib/bootstrap.rb

fallwith:
Introduce `lib/bootstrap.rb` to offer bootstrapping functionality for those users who wish to introduce the New Relic agent to a Ruby application without altering the app's `Gemfile` manifest.

Testing instructions:

1. Clone the feature branch

2. Create a brand new Rails app: 
  ```shell 
  rails new railsapp
  ```

4. Use RUBYOPT to point to the bootstrap file (NOTE that the .rb extension is dropped here) and boot the Rails app:
  ```shell
  RUBYOPT='-r /path/to/clone/lib/bootstrap' railsapp/bin/rails server
  ```

6. Now examine `railsapp/log` to confirm that it contains a `newrelic_agent.log` file with indications that Rails library instrumentation took place.

### DIFF
--- a/lib/bootstrap.rb
+++ b/lib/bootstrap.rb
@@ -1,0 +1,97 @@
+# frozen_string_literal: true
+
+# This file is designed to bootstrap a `Bundler.require` based Ruby app (such as
+# a Ruby on Rails app) so that the app can be instrumented and observed by the
+# New Relic Ruby agent without the agent being added to the app as a dependency.
+#
+# Given a production-ready Ruby app that optionally has a pre-packaged "frozen"
+# or "deployment" based gem bundle, the New Relic Ruby agent can be introduced
+# to the app without modifying the app and keeping all of the app's content
+# read-only.
+#
+# Prerequisites:
+#   - Ruby (tested v2.4+)
+#   - Bundler (included with Ruby, tested v1.17+)
+#
+# Instructions:
+#   - First, make sure the New Relic Ruby agent exists on disk. For these
+#     instructions, we'll assume the agent exists at `/newrelic`.
+#     - The agent can be downloaded as the "newrelic_rpm" gem from RubyGems.org
+#       and unpacked with "gem unpack"
+#     - The agent can be cloned from the New Relic public GitHub repo:
+#       https://github.com/newrelic/newrelic-ruby-agent
+#   - Next, use the "RUBYOPT" environment variable to require ("-r") this
+#     file (note that the ".rb" extension is dropped):
+#       ```
+#       export RUBYOPT="-r /newrelic/lib/bootstrap"
+#       ```
+#   - Launch an existing Ruby app as usual. For a Ruby on Rails app, this might
+#     involve running `bin/rails server`.
+#   - In the Ruby app's directory, look for and inspect
+#     `log/newrelic_agent.log`. If this file exists and there are no "WARN" or
+#     "ERROR" entries within it, then the agent was successfully introduced to
+#     the Ruby application.
+
+module NRBundlerPatch
+  NR_AGENT_GEM = 'newrelic_rpm'
+
+  def require(*_groups)
+    super
+
+    require_newrelic
+  end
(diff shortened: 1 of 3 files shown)


lib/bootstrap.rb
```diff
@@ -0,0 +1,99 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+# This file is designed to bootstrap a `Bundler.require` based Ruby app (such as
+# a Ruby on Rails app) so that the app can be instrumented and observed by the
+# New Relic Ruby agent without the agent being added to the app as a dependency.
+#
+# Given a production-ready Ruby app that optionally has a pre-packaged "frozen"
+# or "deployment" based gem bundle, the New Relic Ruby agent can be introduced
```
kaylareopelle, line 10:
```suggestion
# or "deployment" gem bundle, the New Relic Ruby agent can be introduced
```

lib/bootstrap.rb
```diff
@@ -0,0 +1,99 @@
...
+#     file (note that the ".rb" extension is dropped):
+#       ```
+#       export RUBYOPT="-r /newrelic/lib/bootstrap"
+#       ```
+#   - Launch an existing Ruby app as usual. For a Ruby on Rails app, this might
+#     involve running `bin/rails server`.
+#   - In the Ruby app's directory, look for and inspect
+#     `log/newrelic_agent.log`. If this file exists and there are no "WARN" or
+#     "ERROR" entries within it, then the agent was successfully introduced to
+#     the Ruby application.
```
kaylareopelle, line 35: Love these instructions. Does someone also need to pass a license key? If so, should that be called out here?

lib/bootstrap.rb
```diff
@@ -0,0 +1,99 @@
...
+    check_for_rubyopt
+    check_for_bundler
+    Bundler::Runtime.prepend(NRBundlerPatch)
+  end
+
+  private
+
+  def self.check_for_require
+    warn_and_exit "#{__FILE__} is meant to be required, not invoked directly" if $PROGRAM_NAME == __FILE__
+  end
```
kaylareopelle, line 68: nice!

kaylareopelle (commented):
👏 Awesome work! Especially with figuring out how to test this tricky bit of code.

tannalynn (approved):
Great instructions and great work on such a complex issue!

lib/bootstrap.rb
hannahramadan: Simialr to the hypen added above to `'Bundler.require'-based`, do we want to add one here?

lib/bootstrap.rb
hannahramadan: At some point a license key and some required configs are needed... but I'm not sure we want to encourage independent use of introducing the agent in this way quite yet. We actually talked about gating this behind an env var. 

But on the other hand, these instructions get you 90% of the way towards a working agent, so might as well include the final steps. 

I think I'd like to add both a disclaimer: `NOTE: introducing the agent into your application via bootstrap is in beta.`, as well as the final steps for a working agent (license key and config file). 

What do you think?

kreopelle: that sounds great! on the beta line, maybe append something like "use at your own risk"

i'm not sure if you necessarily need a config file to make the agent work since almost everything can be set by env vars. i think it will just fall back to defaults, but i'm not certain.

hannahramadan: Oh you're totally right. You'll still get a `WARN` about a missing config, but things seem to work

hannahramadan: https://github.com/newrelic/newrelic-ruby-agent/pull/2635/commits/2d05f8e381aab52bec1e3f94f13355ab95efd481

kaylareopelle (approved):
Thank you!

PR #2642, opened 2024-05-17
Dynamodb instrumentation

tannalynn:
Adds instrumentation for dynamodb

### DIFF
--- a/lib/new_relic/agent/aws.rb
+++ b/lib/new_relic/agent/aws.rb
@@ -7,6 +7,15 @@ module NewRelic
     module Aws
       CHARACTERS = %w[A B C D E F G H I J K L M N O P Q R S T U V W X Y Z 2 3 4 5 6 7]
 
+      def self.create_arn(service, resource, config)
+        region = config.region
+        account_id = NewRelic::Agent::Aws.convert_access_key_to_account_id(config.credentials.access_key_id)
+
+        "arn:aws:#{service}:#{region}:#{account_id}:#{resource}"
+      rescue => e
+        # log something prbly
+      end
+
       def self.convert_access_key_to_account_id(access_key)
         decoded_key = Integer(decode_to_hex(access_key[4..-1]), 16)
         mask = Integer('7fffffffff80', 16)
--- a/lib/new_relic/agent/aws.rb
+++ b/lib/new_relic/agent/aws.rb
@@ -1,0 +1,50 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    module Aws
+      CHARACTERS = %w[A B C D E F G H I J K L M N O P Q R S T U V W X Y Z 2 3 4 5 6 7]
+
+      def self.convert_access_key_to_account_id(access_key)
+        decoded_key = Integer(decode_to_hex(access_key[4..-1]), 16)
+        mask = Integer('7fffffffff80', 16)
+        (decoded_key & mask) >> 7
+      rescue => e
+        # log something prbly
+      end
+
+      # ############################
+      # helper methods for convert_access_key_to_account_id
+      def self.decode_to_hex(access_key)
+        bytes = access_key.delete('=').each_char.map { |c| CHARACTERS.index(c) }
+
+        bytes.each_slice(8).map do |section|
+          convert_section(section)
+        end.flatten[0...6].join
+      end
+
+      def self.convert_section(section)
+        buffer = 0
+        section.each do |chunk|
+          buffer = (buffer << 5) + chunk
+        end
+
+        chunk_count = (section.length * 5.0 / 8.0).floor
+
+        if section.length < 8
+          buffer >>= (5 - (chunk_count * 8)) % 5
+        end
+
+        decoded = []
+        chunk_count.times do |i|
+          shift = 8 * (chunk_count - 1 - i)
+          decoded << ((buffer >> shift) & 255).to_s(16)
+        end
+
+        decoded
+      end
+    end
+  end
+end
--- a/lib/new_relic/agent/instrumentation/dynamodb.rb
+++ b/lib/new_relic/agent/instrumentation/dynamodb.rb
@@ -1,0 +1,29 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require_relative 'dynamodb/instrumentation'
+require_relative 'dynamodb/chain'
+require_relative 'dynamodb/prepend'
+
+DependencyDetection.defer do
+  named :dynamodb
+
+  depends_on do
+    # The class that needs to be defined to prepend/chain onto. This can be used
+    # to determine whether the library is installed.
+    defined?(::Aws::DynamoDB::Client)
+    # Add any additional requirements to verify whether this instrumentation
+    # should be installed
+  end
+
+  executes do
+    ::NewRelic::Agent.logger.info('Installing dynamodb instrumentation')
+
+    if use_prepend?
+      prepend_instrument ::Aws::DynamoDB::Client, NewRelic::Agent::Instrumentation::Dynamodb::Prepend
+    else
+      chain_instrument NewRelic::Agent::Instrumentation::Dynamodb::Chain
+    end
+  end
+end
--- a/lib/new_relic/agent/instrumentation/dynamodb.rb
+++ b/lib/new_relic/agent/instrumentation/dynamodb.rb
@@ -12,16 +12,16 @@ DependencyDetection.defer do
   depends_on do
     # The class that needs to be defined to prepend/chain onto. This can be used
     # to determine whether the library is installed.
-    defined?(::Aws::DynamoDB::Client)
+    defined?(Aws::DynamoDB::Client)
     # Add any additional requirements to verify whether this instrumentation
     # should be installed
   end
 
   executes do
-    ::NewRelic::Agent.logger.info('Installing dynamodb instrumentation')
+    NewRelic::Agent.logger.info('Installing dynamodb instrumentation')
 
     if use_prepend?
-      prepend_instrument ::Aws::DynamoDB::Client, NewRelic::Agent::Instrumentation::Dynamodb::Prepend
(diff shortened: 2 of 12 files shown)


lib/new_relic/agent/aws.rb
```diff
@@ -0,0 +1,55 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    module Aws
+      CHARACTERS = %w[A B C D E F G H I J K L M N O P Q R S T U V W X Y Z 2 3 4 5 6 7]
```
fallwith, line 8:
```suggestion
      CHARACTERS = %w[A B C D E F G H I J K L M N O P Q R S T U V W X Y Z 2 3 4 5 6 7].freeze
```

lib/new_relic/agent/aws.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def self.create_arn(service, resource, config)
+        region = config.region
+        account_id = NewRelic::Agent::Aws.convert_access_key_to_account_id(config.credentials.access_key_id)
+
+        "arn:aws:#{service}:#{region}:#{account_id}:#{resource}"
+      rescue => e
+        NewRelic::Agent.logger.debug("Failed to create ARN: #{e}")
```
fallwith, line 16: Perhaps `warn` or `error`?

lib/new_relic/agent/aws.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+
+      def self.convert_access_key_to_account_id(access_key)
+        decoded_key = Integer(decode_to_hex(access_key[4..-1]), 16)
+        mask = Integer('7fffffffff80', 16)
```
fallwith, line 21: This never-changing string is a good candidate for a constant.

newrelic.yml
```diff
@@ -403,6 +403,10 @@ common: &default_settings
   # prepend, chain, disabled.
   # instrumentation.bunny: auto
 
+# Controls auto-instrumentation of DynamoDb at start-up.
+# May be one of [auto|prepend|chain|disabled]
+# instrumentation.dynamodb: auto
```
fallwith, line 408: The indentation on these new lines is off. You can either indent a bit further to fix it or revert the changes to this file and let the auto-generated changes to the file handle it later when we release.

test/multiverse/suites/dynamodb/Envfile
```diff
@@ -0,0 +1,10 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+instrumentation_methods :chain, :prepend
+
+gemfile <<~RB
+  gem 'aws-sdk-dynamodb'
+  gem 'nokogiri'
```
fallwith, line 9: What is nokogiri used for?

tannalynn: Before I added nokogiri, the aws sdk failed to install and said "you need to install one of these gems too", and nokogiri was on the list of options, so I just added that one so the aws sdk would work. I'm not using it directly.

newrelic.yml
tannalynn: ahh yeah that's the auto generate command doing that, since this is automated later on now we can probably just remove this from the scaffold command. 
this indentation fixed now though

lib/new_relic/agent/aws.rb
tannalynn: changed to warn

lib/new_relic/agent/instrumentation/dynamodb/chain.rb
```diff
@@ -0,0 +1,34 @@
...
+        include NewRelic::Agent::Instrumentation::Dynamodb
+
+        %w[create_table
+          delete_item
+          delete_table
+          get_item
+          put_item
+          query
+          scan
+          update_item].each do |method_name|
```
kaylareopelle, line 18: Since this array is the same as the one for `prepend`, what do you think about moving it to the `Instrumentation` module and setting it as a constant?

lib/new_relic/agent/instrumentation/dynamodb/instrumentation.rb
```diff
@@ -0,0 +1,47 @@
...
+    def instrument_method_with_new_relic(method_name, *args)
+      return yield unless NewRelic::Agent::Tracer.tracing_enabled?
+
+      NewRelic::Agent.record_instrumentation_invocation(PRODUCT)
+
+      segment = NewRelic::Agent::Tracer.start_datastore_segment(
+        product: PRODUCT,
+        operation: method_name,
+        host: config&.endpoint&.host || DEFAULT_HOST,
+        port_path_or_id: config&.endpoint&.port,
```
kaylareopelle, line 19: Did you run into situations where the endpoint/host/port were nil?

Also, where is `config` coming from?

lib/new_relic/agent/instrumentation/dynamodb.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  executes do
+    NewRelic::Agent.logger.info('Installing dynamodb instrumentation')
+
+    if use_prepend?
+      prepend_instrument Aws::DynamoDB::Client, NewRelic::Agent::Instrumentation::Dynamodb::Prepend
+    else
+      chain_instrument NewRelic::Agent::Instrumentation::Dynamodb::Chain
+    end
```
kaylareopelle, line 23: nit: AWS uses `DynamoDB` and we use `Dynamodb`. Would you be open to updating our classes to use the same capitalization style? (This is probably related to a limitation of the generator I should look into...)

lib/new_relic/agent/instrumentation/dynamodb/instrumentation.rb
```diff
@@ -0,0 +1,47 @@
...
+      segment.add_agent_attribute('cloud.resource_id', arn) if arn
+
+      @nr_captured_request = nil # clear request just in case
+      begin
+        NewRelic::Agent::Tracer.capture_segment_error(segment) { yield }
+      ensure
+        segment.add_agent_attribute('aws.operation', method_name)
+        segment.add_agent_attribute('aws.requestId', @nr_captured_request&.context&.http_response&.headers&.[]('x-amzn-requestid'))
+        segment.add_agent_attribute('aws.region', config&.region)
+        segment&.finish
```
kaylareopelle, line 33:
```suggestion
        segment.finish
```
Since we just called a bunch of methods on `segment` can we feel safe it's not `nil` here and remove the safe operator? Or do we need it on all the methods?

test/multiverse/suites/dynamodb/dynamodb_instrumentation_test.rb
```diff
@@ -0,0 +1,201 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'aws-sdk-dynamodb'
+
+class DynamodbInstrumentationTest < Minitest::Test
+  def setup
+    Aws.config.update(stub_responses: true)
```
kaylareopelle, line 9: This is sooo convenient. I'm glad you found it.

kaylareopelle (approved):
Great as-is, with a few optional recommendations!

lib/new_relic/agent/instrumentation/dynamodb/chain.rb
tannalynn: hmm, yeah i see what you mean. the only thing that gives me pause to move it is that I like it being located here, because it makes it super clear exactly what methods are getting instrumented in the same place we normally would look to see what is getting instrumented.

lib/new_relic/agent/instrumentation/dynamodb/instrumentation.rb
tannalynn: config is an object that conveniently already exists inside of the aws client, giving us access to a bunch of info.

Also, no i didn't, it seems like it's always populate, but i added the default host just in case it's possible.

lib/new_relic/agent/instrumentation/dynamodb/instrumentation.rb
tannalynn: i'd rather add & to the other ones, ill get that updated!

lib/new_relic/agent/instrumentation/dynamodb.rb
tannalynn: yeah, i can't remember exactly what now, but i did end up generating a couple of times bc of weirdness with the capitalization, this one ended up looking most normal so i just did this way, but i can go in and manually change it

lib/new_relic/agent/instrumentation/dynamodb/chain.rb
kaylareopelle: I get that. I think the convenience is really nice and makes this really readable. 

I'm worried about a situation where we need to add a new method in the future and only add it to prepend/forget about chain. Do you think that's likely? Hopefully testing in the CI would catch it.

lib/new_relic/agent/instrumentation/dynamodb/instrumentation.rb
kaylareopelle: Nice! Thanks! I was wondering if we needed the save navigation operators, but since we're relying on an AWS API, I think having the operators will probably be safer for our users if AWS changes things without us knowing.

lib/new_relic/agent/instrumentation/dynamodb/chain.rb
tannalynn: yeah i decided just to change it, if clarity is a problem in the future we can change something then

lib/new_relic/agent/instrumentation/dynamodb/instrumentation.rb
tannalynn: Yeah, for sure, that's totally why i added the & too, just bc it's not ours and we can't for sure know if there are situations it'll be nil, or if something changes in the future

lib/new_relic/agent/instrumentation/dynamodb/chain.rb
kaylareopelle: Nice, thank you!

lib/new_relic/agent/instrumentation/dynamodb.rb
kaylareopelle: Thank you!

PR #2649, opened 2024-05-20
Allow agent attributes on external and datastore segments

tannalynn:
Previously the agent was not adding agent attributes when harvesting extern request segments or datastore segments, other than very specific attributes for each type. There is nothing in the agent specs that says it needs to work this way,  I suspect it functioned this way originally due to a lack of needing any general agent attributes combined with needing to create specific attributes for each type, rather than a specific intention to exclude agent attributes from being added to these segment types. We are going to update the agent to allow agent attributes to be added to both segment types moving forward.
There had previously been some work done to make it possible to add agent attributes in certain situations to external requests, but we have removed that in favor of just always allowing agent attributes to be added instead.


kaylareopelle (approved):
🐕

hannahramadan (approved):
Nice! This makes sense

PR #2663, opened 2024-05-28
CI: test with Ruby v3.4.0-preview1

### LINKED ISSUE #2636: CI: Introduce Ruby 3.4-preview1 as a Ruby instance
With v3.4 on its way for a scheduled 2024-12-25 release and the first public preview now available, we ought to begin testing with v3.4 for compatibility via our CI systems.

### LINKED ISSUE #2665: CI: Need to ensure that base64 and others are present for all Ruby v3.4+ contexts
We previously made the agent itself ready for Ruby v3.4's dropping of base64 as a default library by essentially forking the library and using the fork. That fix, however, only works for Ruby code calls that originate from within the Ruby agent code base. Now that we've introduced Ruby v3.4.0-preview1 to the CI system, we've noticed that the absence of base64 is causing issues for our `test/environments` environments and for certain 3rd party libraries such as `elasticsearch-transport` that may themselves not be Ruby v3.4+ compatible yet.

Definition of done: a full `ci_cron.yml` workflow run works with all Ruby v3.4.0-preview1 contexts passing.

### LINKED ISSUE #2670: ElasticSearch instrumentation is currently incompatible with Ruby v3.4.0
Our ElasticSearch instrumentation currently relies on Ruby caller locations possessing an opening backtick character in order to be able to glean the operation that was performed by the client.

With Ruby 3.4's switch to using single quotes for both the opening and closing of caller location path strings, keying off of the backtick won't work.

### LINKED ISSUE #2672: CI: Ruby v3.4.0-preview1 fails to provide SQL explain plan info for Sequel
For Ruby v3.4.0-preview1, the Sequel multiverse test suite fails to capture explain plans for Sqlite3 based SQL queries which causes fully 4 separate Sequel extension tests to fail.

To reproduce:

```shell
setruby 3.4.0-preview1
bermq sequel
```

fallwith:
- Introduce nightly testing against Ruby v3.4.0-preview1
- Update elasticsearch instrumentation to be Ruby 3.4+ compatible
- Update various test environment `Gemfile` files to include gems that were previously included with Ruby as part of the default installation but are no longer included with Ruby 3.4 or 3.5.

resolves #2636 
resolves #2665 
resolves #2670
resolves #2672

### DIFF
--- a/test/environments/norails/Gemfile
+++ b/test/environments/norails/Gemfile
@@ -22,3 +22,8 @@ gem 'simplecov' if ENV['VERBOSE_TEST_OUTPUT']
 
 gem 'warning'
 gem 'loofah', '~> 2.20.0' if RUBY_VERSION >= '2.4.0' && RUBY_VERSION < '2.5.0'
+
+if RUBY_VERSION.split('.')[0..1].join('.').to_f >= 3.4
+  gem 'mutex_m'
+  gem 'ostruct'
+end
--- a/.github/workflows/ci_cron.yml
+++ b/.github/workflows/ci_cron.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Configure git
         run: 'git config --global init.defaultBranch main'
       - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # tag v4.1.2
-      - uses: ruby/setup-ruby@1198b074305f9356bd56dd4b311757cc0dab2f1c # tag v1.175.1
+      - uses: ruby/setup-ruby@943103cae7d3f1bb1e4951d5fcc7928b40e4b742 # tag v1.177.1
         with:
           ruby-version: '3.3'
       - run: bundle
@@ -36,7 +36,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby-version: [2.4.10, 2.5.9, 2.6.10, 2.7.8, 3.0.7, 3.1.5, 3.2.4, 3.3.1]
+        ruby-version: [2.4.10, 2.5.9, 2.6.10, 2.7.8, 3.0.7, 3.1.5, 3.2.4, 3.3.1, 3.4.0-preview1]
 
     steps:
       - name: Configure git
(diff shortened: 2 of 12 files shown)


kaylareopelle (approved):
Thank you! 

I don't think this needs to be done here, but we should see if there's an easy way to keep all of our ruby/setup-ruby versions in sync across our workflows.

test/environments/norails/Gemfile
```diff
@@ -22,3 +22,8 @@ gem 'simplecov' if ENV['VERBOSE_TEST_OUTPUT']
 
 gem 'warning'
 gem 'loofah', '~> 2.20.0' if RUBY_VERSION >= '2.4.0' && RUBY_VERSION < '2.5.0'
+
+if RUBY_VERSION.split('.')[0..1].join('.').to_f >= 3.4
```
kaylareopelle, line 26: Are these environments able to use the `ruby_verison_float` method?

Also, it seems a little sacrilegious, but what do you think about using the `Gem::Version.new` API to compare Ruby versions? Ruby language versions seem to follow the same process as gem versions. However, it won't allow you to use a float, you'd have to use a string.

`Gem::Version.new(RUBY_VERSION) >= '3.4'`

kaylareopelle (approved):
Fantastic documentation on the Elasticsearch Regex! Thank you!!

hannahramadan (approved):

test/environments/norails/Gemfile
fallwith: > Are these environments able to use the ruby_verison_float method?

Given that these are direct entrypoints (as opposed to the `Envfile` files), I didn't want to drag in any dependencies.

> you'd have to use a string

Yeah, string comparison for math isn't ideal even though we've lucked out with Ruby v1 and v2 never having gone to double digits with the minor version.

Maybe in future we could author a helper method that simply calls `Gem::Version.new` on both `RUBY_VERSION` and the evaluated Ruby version string. I think we've done that in some places previously inline (without using a method).

PR #2666, opened 2024-05-28
Remove metrics helpers that are not called

kaylareopelle:
Testing the waters to remove these methods... will they be missed?


kaylareopelle:
Found where they're called.

PR #2676, opened 2024-06-01
lib/bootstrap.rb -> lib/boot/strap.rb

### LINKED ISSUE #2675: Newrelic 9.10.0 breaks our CI - caused by a collision with the 'bootstrap' gem
## Description
On a dependabot PR for the new version of the agent, all GitHub Action CI steps fail with the following error:

```
New Relic entrypoint at /home/runner/work/human-essentials/human-essentials/vendor/bundle/ruby/3.2.0/gems/newrelic_rpm-9.10.0/lib/bootstrap.rb encountered an issue:
	/home/runner/work/human-essentials/human-essentials/vendor/bundle/ruby/3.2.0/gems/newrelic_rpm-9.10.0/lib/bootstrap.rb is meant to be required via the RUBYOPT env var
```

No previous versions caused these errors.

Sample run: https://github.com/rubyforgood/human-essentials/actions/runs/9297054369/job/25667726353

## Expected Behavior
No crashes!

fallwith:
- Do not put any Ruby files in the root of `lib/` that do not match the
  gem name, or Ruby lib path based issues can arise with other gems.
  In this case, move `lib/bootstrap.rb` to `lib/boot/strap.rb` to avoid
  colliding with `require 'bootstrap'` calls intended to be received by
  the `bootstrap` gem.
- For bootstrapping, do not ever perform `exit`. Instead, simply use
  `raise/rescue` for control flow. As a result, any future unintended
  loading of the bootstrapping will simply `warn` and return instead of
  causing the Ruby runtime to exit.

resolves #2675

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+- **Bugfix: Incompatibility with Bootstrap**
+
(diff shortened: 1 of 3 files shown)


hannahramadan (approved):

kaylareopelle (approved):

Issue #2592, opened 2024-04-24
Aws: Instrument sqs


tannalynn:
Instrument aws-sdk-sqs
For AWS_SQS_QUEUE

PR #2679, opened 2024-06-03
Add SQS instrumentation

tannalynn:
adds instrumentation for the aws-sdk-sqs gem

resolves https://github.com/newrelic/newrelic-ruby-agent/issues/2592

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Version <dev> introduces instrumentation for the aws-sdk-sqs gem, fixes a bug re
 
 - **Feature: Add instrumentation for SQS**
 
-    The agent has added instrumentation for the aws-sdk-sqs gem. The agent will now record message broker spans for SQS client calls made with the aws-sdk-sqs gem. [PR#2679](https://github.com/newrelic/newrelic-ruby-agent/pull/2679)
+    The agent has added instrumentation for the [aws-sdk-sqs gem](https://rubygems.org/gems/aws-sdk-sqs). The agent will now record message broker spans for SQS client calls made with the aws-sdk-sqs gem. [PR#2679](https://github.com/newrelic/newrelic-ruby-agent/pull/2679)
 
 - **Bugfix: HTTP status code based expected errors will now have an "expected" value of "true"**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 ## dev
 
-Version <dev> fixes a bug related to expected errors not bearing a "true" value for the "expected" attribute if expected as a result of an HTTP status code match and changes the way Stripe instrumentation metrics are named to prevent high-cardinality issues.
+Version <dev> introduces instrumentation for the aws-sdk-sqs gem, fixes a bug related to expected errors not bearing a "true" value for the "expected" attribute if expected as a result of an HTTP status code match and changes the way Stripe instrumentation metrics are named to prevent high-cardinality issues.
+
+- **Feature: Add instrumentation for SQS**
+
+    The agent has added instrumentation for the aws-sdk-sqs gem. The agent will now record message broker spans for SQS client calls made with the aws-sdk-sqs gem. [PR#2679](https://github.com/newrelic/newrelic-ruby-agent/pull/2679)
 
 - **Bugfix: HTTP status code based expected errors will now have an "expected" value of "true"**
 
--- a/lib/new_relic/agent/instrumentation/aws_sqs.rb
+++ b/lib/new_relic/agent/instrumentation/aws_sqs.rb
@@ -1,0 +1,25 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require_relative 'awssqs/instrumentation'
+require_relative 'awssqs/chain'
+require_relative 'awssqs/prepend'
+
+DependencyDetection.defer do
+  named :aws_sqs
+
+  depends_on do
+    defined?(::Aws::SQS::Client)
+  end
+
+  executes do
+    ::NewRelic::Agent.logger.info('Installing aws-sdk-sqs instrumentation')
+
+    if use_prepend?
+      prepend_instrument ::Aws::SQS::Client, NewRelic::Agent::Instrumentation::AwsSqs::Prepend
+    else
+      chain_instrument NewRelic::Agent::Instrumentation::AwsSqs::Chain
+    end
+  end
+end
--- a/lib/new_relic/agent/instrumentation/aws_sqs.rb
+++ b/lib/new_relic/agent/instrumentation/aws_sqs.rb
@@ -2,9 +2,9 @@
 # See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
 # frozen_string_literal: true
 
-require_relative 'awssqs/instrumentation'
-require_relative 'awssqs/chain'
-require_relative 'awssqs/prepend'
+require_relative 'aws_sqs/instrumentation'
+require_relative 'aws_sqs/chain'
+require_relative 'aws_sqs/prepend'
 
 DependencyDetection.defer do
   named :aws_sqs
--- a/lib/new_relic/agent/instrumentation/aws_sqs.rb
+++ b/lib/new_relic/agent/instrumentation/aws_sqs.rb
@@ -10,14 +10,14 @@ DependencyDetection.defer do
   named :aws_sqs
 
   depends_on do
-    defined?(::Aws::SQS::Client)
+    defined?(Aws::SQS::Client)
   end
 
   executes do
-    ::NewRelic::Agent.logger.info('Installing aws-sdk-sqs instrumentation')
+    NewRelic::Agent.logger.info('Installing aws-sdk-sqs instrumentation')
 
     if use_prepend?
-      prepend_instrument ::Aws::SQS::Client, NewRelic::Agent::Instrumentation::AwsSqs::Prepend
+      prepend_instrument Aws::SQS::Client, NewRelic::Agent::Instrumentation::AwsSqs::Prepend
     else
       chain_instrument NewRelic::Agent::Instrumentation::AwsSqs::Chain
     end
--- a/lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
@@ -14,6 +14,10 @@ module NewRelic::Agent::Instrumentation
           destination_type: 'Queue',
           destination_name: queue_name
         )
+        segment&.add_agent_attribute('messaging.system', 'aws_sqs')
+        segment&.add_agent_attribute('cloud.region', config&.region)
+        segment&.add_agent_attribute('cloud.account.id', get_account_id(args[0]))
+        # segment&.add_agent_attribute('messaging.destination.name', queue_name)
       rescue => e
         NewRelic::Agent.logger.error('Error starting message broker segment in Aws::SQS::Client#send_message ', e)
       end
@@ -34,6 +38,9 @@ module NewRelic::Agent::Instrumentation
           destination_type: 'Queue',
           destination_name: queue_name
         )
+        segment&.add_agent_attribute('messaging.system', 'aws_sqs')
+        segment&.add_agent_attribute('cloud.region', config&.region)
+        segment&.add_agent_attribute('cloud.account.id', get_account_id(args[0]))
       rescue => e
         NewRelic::Agent.logger.error('Error starting message broker segment in Aws::SQS::Client#send_message ', e)
       end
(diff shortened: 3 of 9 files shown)


lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
```diff
@@ -0,0 +1,87 @@
...
+          action: :produce,
+          library: MESSAGING_LIBRARY,
+          destination_type: :queue,
+          destination_name: queue_name
+        )
+        add_aws_attributes(segment, queue_name, args[0])
+      rescue => e
+        NewRelic::Agent.logger.error('Error starting message broker segment in Aws::SQS::Client#send_message ', e)
+      end
+      NewRelic::Agent::Tracer.capture_segment_error(segment) do
```
fallwith, line 23: Currently, fully 3 operations are wrapped in the initial `begin/rescue` block:

- `get_queue_name`
- `NewRelic::Agent::Tracer.start_message_broker_segment`
- `add_aws_attributes`

Two things:

1. Is there any need to individually wrap each call in its own `begin/rescue` block, possibly by using 3 dedicated methods?

2. This leads me to believe that it is entirely possible to reach this `capture_segment_error` line with a `segment` variable that is `nil`. `capture_segment_error` itself doesn't appear to check `segment` for `nil`, so I think we need to guard for `nil` before calling it.

lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
```diff
@@ -0,0 +1,87 @@
...
+        NewRelic::Agent.logger.error('Error starting message broker segment in Aws::SQS::Client#receive_message ', e)
+      end
+      NewRelic::Agent::Tracer.capture_segment_error(segment) do
+        yield
+      end
+    ensure
+      segment&.finish
+    end
+
+    def add_aws_attributes(segment, queue_name, params)
```
fallwith, line 72: I think you should move all 5 inline defined strings out of this method and into constants. This way you're not creating 5 unique string objects each time the method is called.

lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
```diff
@@ -0,0 +1,87 @@
...
+      segment&.add_agent_attribute('messaging.system', 'aws_sqs')
+      segment&.add_agent_attribute('cloud.region', config&.region)
+      segment&.add_agent_attribute('cloud.account.id', get_account_id(params))
```
fallwith, line 75: Is Ruby smart enough to not even attempt `get_account_id` if `segment` is `nil`? If not, I think you should guard against `segment` being `nil` with an early return.

test/multiverse/suites/awssqs/awssqs_instrumentation_test.rb
```diff
@@ -0,0 +1,90 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+class AwssqsInstrumentationTest < Minitest::Test
```
fallwith, line 5: I think it would be good to either have 100% branch coverage on the new code added, or at least make sure each `rescue` block defined in the new `AwsSqs` module is reached. (Need some unhappy path tests)

fallwith (changes requested):

lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
tannalynn: for #1, The purpose of the begin/rescue is if any issues happen as part of our instrumentation, the gem will still make the original call it intended to make, so anything wrong happening for us won't break the customers application. I'm not sure why we'd need to wrap each line in its own rescue, is there something specific you are thinking of that would make that necessary? 

For #2, we don't need to check for nil. If segment is nil, it basically is just a noop method, since the only time segment is used it first checks to make sure it's the correct class, so if it's nil, it will simply skip it instead.

lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
```diff
@@ -0,0 +1,87 @@
...
+# frozen_string_literal: true
+
+module NewRelic::Agent::Instrumentation
+  module AwsSqs
+    MESSAGING_LIBRARY = 'SQS'
+
+    def send_message_with_new_relic(*args)
+      segment = nil
+      begin
+        queue_name = get_queue_name(args[0])
```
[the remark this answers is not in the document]
tannalynn: It's unlikely that would happen. the sdk itself needs that information in order to make the call, so if it's missing there is something larger wrong going on. If there's no queue name, we don't have the information we need to create the segment anyways. In that case, the rescue would capture the error and prevent the error from bubbling up to the customers code.

lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
tannalynn: yeah i can add that a guard, but jsyk it won't call the things in those args if the &. isn't going to happen

test/multiverse/suites/awssqs/awssqs_instrumentation_test.rb
tannalynn: added a sad test for each method here, checks the log to make sure the message is there https://github.com/newrelic/newrelic-ruby-agent/pull/2679/commits/f251b1498325f0b1ff4e7073227890e03e985b11

lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
```diff
@@ -0,0 +1,87 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic::Agent::Instrumentation
+  module AwsSqs
+    MESSAGING_LIBRARY = 'SQS'
+
+    def send_message_with_new_relic(*args)
```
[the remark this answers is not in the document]
tannalynn: only split once added here w/ a hash https://github.com/newrelic/newrelic-ruby-agent/pull/2679/commits/6e1c09820228c80800eb79021d4a1c79aaa72b95

lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
tannalynn: added a guard here and removed the &. https://github.com/newrelic/newrelic-ruby-agent/pull/2679/commits/fc7f03780c7567ab2bb77b19f6ce9ca8e6e619ac

lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
tannalynn: discussed in slack, this isn't needed because `# frozen_string_literal: true`

fallwith (approved):

CHANGELOG.md
```diff
@@ -2,7 +2,11 @@
 
 ## dev
 
-Version <dev> fixes a bug related to expected errors not bearing a "true" value for the "expected" attribute if expected as a result of an HTTP status code match and changes the way Stripe instrumentation metrics are named to prevent high-cardinality issues.
+Version <dev> introduces instrumentation for the aws-sdk-sqs gem, fixes a bug related to expected errors not bearing a "true" value for the "expected" attribute if expected as a result of an HTTP status code match and changes the way Stripe instrumentation metrics are named to prevent high-cardinality issues.
```
hannahramadan, line 5: Sometimes we link gems and I always kinda liked that
```suggestion
Version <dev> introduces instrumentation for the [aws-sdk-sqs gem](https://rubygems.org/gems/aws-sdk-sqs/versions/1.0.0.rc11), fixes a bug related to expected errors not bearing a "true" value for the "expected" attribute if expected as a result of an HTTP status code match and changes the way Stripe instrumentation metrics are named to prevent high-cardinality issues.
```

CHANGELOG.md
```diff
@@ -2,7 +2,11 @@
...
+
+- **Feature: Add instrumentation for SQS**
+
+    The agent has added instrumentation for the aws-sdk-sqs gem. The agent will now record message broker spans for SQS client calls made with the aws-sdk-sqs gem. [PR#2679](https://github.com/newrelic/newrelic-ruby-agent/pull/2679)
```
hannahramadan, line 9:
```suggestion
    The agent has added instrumentation for the [aws-sdk-sqs gem](https://rubygems.org/gems/aws-sdk-sqs/versions/1.0.0.rc11). The agent will now record message broker spans for SQS client calls made with the aws-sdk-sqs gem. [PR#2679](https://github.com/newrelic/newrelic-ruby-agent/pull/2679)
```

tannalynn: oh thats a good idea, i never thought of that before. how come not linking to the main page though? https://rubygems.org/gems/aws-sdk-sqs

hannahramadan: Oh yes, that is totally better! I read 'gem' and went rubygems lol

tannalynn: haha ok cool! nice ty

lib/new_relic/agent/instrumentation/aws_sqs.rb
```diff
@@ -0,0 +1,25 @@
...
+# frozen_string_literal: true
+
+require_relative 'aws_sqs/instrumentation'
+require_relative 'aws_sqs/chain'
+require_relative 'aws_sqs/prepend'
+
+DependencyDetection.defer do
+  named :aws_sqs
+
+  depends_on do
```
hannahramadan, line 12: Where are we checking the config value `instrumentation.aws_sqs` to know if we should be instrumenting? I feel like I usually see it here

tannalynn: the `named :aws_sqs` is what tells dependency detection to check the `instrumentation.aws_sqs` config value. we don't usually have to check it ourselves (unless we have multiple, or its more complicated for some reason)

hannahramadan: Thank you for going over this!

hannahramadan (approved):

lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
```diff
@@ -0,0 +1,89 @@
...
```
kaylareopelle, line 9: Outside of the `:action` and the error message, the three main methods in this module (`send_message_with_new_relic`, `send_message_batch_with_new_relic` and `receive_message_batch_with_new_relic`) are all looking the same to me. What do you think about breaking the shared code into a separate method?

[The Redis instrumentation takes this approach.](https://github.com/newrelic/newrelic-ruby-agent/blob/89119f566e03487c5eab3e49e9e7211720f166c5/lib/new_relic/agent/instrumentation/redis/instrumentation.rb#L47)

lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
```diff
@@ -0,0 +1,89 @@
...
+      rescue => e
+        NewRelic::Agent.logger.error('Error starting message broker segment in Aws::SQS::Client#receive_message ', e)
+      end
+      NewRelic::Agent::Tracer.capture_segment_error(segment) do
+        yield
+      end
+    ensure
+      segment&.finish
+    end
+
```
kaylareopelle, line 71: What do you think about making the two methods below this line private?

test/multiverse/suites/awssqs/awssqs_instrumentation_test.rb
```diff
@@ -0,0 +1,145 @@
...
+  def setup
+    Aws.config.update(stub_responses: true)
+    @stats_engine = NewRelic::Agent.instance.stats_engine
+  end
+
+  def teardown
+    harvest_span_events!
+    mocha_teardown
+    NewRelic::Agent.instance.stats_engine.clear_stats
```
kaylareopelle, line 14: Does using the instance variable here help? 
```suggestion
    @stats_engine.clear_stats
```

Or if not, do we need the instance var on line 8?

tannalynn: oh yeah i didn't add that, that's something that's done as part of the instrumentation scaffold thor command. this is the template.
```
class <%= @class_name %>InstrumentationTest < Minitest::Test
  def setup
    @stats_engine = NewRelic::Agent.instance.stats_engine
  end

  def teardown
    NewRelic::Agent.instance.stats_engine.clear_stats
  end

  # Add tests here
end
```
im not really sure why it's like that, but i can remove those from my tests since i'm not using them. Perhaps we should update the template if thats not something we want to do every time.

lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
tannalynn: cool, i did that here https://github.com/newrelic/newrelic-ruby-agent/pull/2679/commits/df6d3ca0acb102e7b55e2c0b89ed32c4a46aabdc

lib/new_relic/agent/instrumentation/aws_sqs/instrumentation.rb
tannalynn: https://github.com/newrelic/newrelic-ruby-agent/pull/2679/commits/df6d3ca0acb102e7b55e2c0b89ed32c4a46aabdc

test/multiverse/suites/awssqs/awssqs_instrumentation_test.rb
tannalynn: removed https://github.com/newrelic/newrelic-ruby-agent/pull/2679/commits/278a3b509968facd0aa481d2241325dc37d9da56

kaylareopelle (approved):
Thank you!

PR #2684, opened 2024-06-03
Bugfix: Dynamodb instrumentation access_key_id error on some credential classes

tannalynn:
Not all aws credentials are credentials, but they do all seem to include credentials, so I've updated it to do credentials.credentials instead, which should work. I also added more safety/checks so that it won't cause an error and will just simply not try to make an arn if it doesn't have access to the necessary info.

### DIFF
--- a/lib/new_relic/agent/aws.rb
+++ b/lib/new_relic/agent/aws.rb
@@ -19,6 +19,8 @@ module NewRelic
         return unless access_key_id
 
         NewRelic::Agent::Aws.convert_access_key_to_account_id(access_key_id)
+      rescue => e
+        NewRelic::Agent.logger.debug("Failed to create account id: #{e}")
       end
 
       def self.convert_access_key_to_account_id(access_key)
--- a/lib/new_relic/agent/aws.rb
+++ b/lib/new_relic/agent/aws.rb
@@ -9,8 +9,11 @@ module NewRelic
       HEX_MASK = '7fffffffff80'
 
       def self.create_arn(service, resource, config)
+        access_key_id = config&.credentials&.credentials&.access_key_id if config&.credentials&.credentials&.respond_to?(:access_key_id)
+        return unless access_key_id
+
         region = config.region
-        account_id = NewRelic::Agent::Aws.convert_access_key_to_account_id(config.credentials.access_key_id)
+        account_id = NewRelic::Agent::Aws.convert_access_key_to_account_id(access_key_id)
 
         "arn:aws:#{service}:#{region}:#{account_id}:#{resource}"
       rescue => e
--- a/lib/new_relic/agent/aws.rb
+++ b/lib/new_relic/agent/aws.rb
@@ -8,16 +8,17 @@ module NewRelic
       CHARACTERS = %w[A B C D E F G H I J K L M N O P Q R S T U V W X Y Z 2 3 4 5 6 7].freeze
       HEX_MASK = '7fffffffff80'
 
-      def self.create_arn(service, resource, config)
-        access_key_id = config&.credentials&.credentials&.access_key_id if config&.credentials&.credentials&.respond_to?(:access_key_id)
-        return unless access_key_id
-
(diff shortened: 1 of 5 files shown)


lib/new_relic/agent/instrumentation/dynamodb/instrumentation.rb
```diff
@@ -49,10 +49,14 @@ def build_request_with_new_relic(*args)
       @nr_captured_request = yield
     end
 
+    def nr_account_id
+      @nr_account_id ||= NewRelic::Agent::Aws.get_account_id(config)
```
tannalynn, line 53: the account id will always be the same for a single instance of a client, since it's set up using specific credentials, so we should only need to do the math for that once per client instance

lib/new_relic/agent/instrumentation/dynamodb/instrumentation.rb
```diff
@@ -49,10 +49,14 @@ def build_request_with_new_relic(*args)
...
+    end
+
     def get_arn(params)
-      return unless params[:table_name]
+      return unless params[:table_name] && nr_account_id
```
tannalynn, line 57: table name though could change every time the client makes a call jsyk

lib/new_relic/agent/aws.rb
```diff
@@ -9,8 +9,11 @@ module Aws
       HEX_MASK = '7fffffffff80'
 
       def self.create_arn(service, resource, config)
+        access_key_id = config&.credentials&.credentials&.access_key_id if config&.credentials&.credentials&.respond_to?(:access_key_id)
```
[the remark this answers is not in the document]
tannalynn: true! i'll change that.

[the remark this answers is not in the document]
tannalynn: changed https://github.com/newrelic/newrelic-ruby-agent/pull/2684/commits/251914068c617e03f8c36e4e2b3a59001391db35

lib/new_relic/agent/instrumentation/dynamodb/instrumentation.rb
fallwith: I like the idea of memoizing the value to not have to do the conversion every time. How expensive are the `#respond_to?(:access_key_id)` and `#access_key_id` calls? Is there any value in only checking those once and then essentially memoizing `nil` or are they lightweight enough to keep invoking every time and never having the memoization store anything?

test/new_relic/agent/aws_test.rb
```diff
@@ -6,18 +6,26 @@
 
 class AwsTest < Minitest::Test
   def test_create_arn
```
kaylareopelle, line 8: Do you think it'd be beneficial to add tests that call some of the different AWS classes related to retrieving credentials? I wonder if that'd give us a heads up if a future version breaks our ARN generation.

lib/new_relic/agent/instrumentation/dynamodb/instrumentation.rb
tannalynn: Updated! https://github.com/newrelic/newrelic-ruby-agent/pull/2684/commits/a58d6cba23522f1a43b212f380839c198d62f5ea

test/new_relic/agent/aws_test.rb
tannalynn: Yeah, unfortunately we could only do that in multiverse tests, so we'd have to move these into a multiverse suite. Also, I don't think we could use the one that was broken since, for that one, you have to be running in EC2 to be able to actually use it. Also, since the responses are stubbed for our dynamodb tests, there are no credentials being created anyways. So we'd need to figure out a way around those things. i'd be down to chat about it later and come up with some ideas if you want

kaylareopelle: We chatted about this sync and decided to leave the tests as they are.

kaylareopelle (approved):

Issue #2309, opened 2023-11-03
Investigate: Simplecov branch coverage drop


hannahramadan:
Simplecov branch coverage dropped from above 85% to just ~82% with no clear cause. 

https://github.com/newrelic/newrelic-ruby-agent/pull/2305 was over the threshold and https://github.com/newrelic/newrelic-ruby-agent/pull/2306 was lower, so its possible the change has to do with 2306.

[PR#2308](https://github.com/newrelic/newrelic-ruby-agent/pull/2308) lowered the threshold to get the CI working again. Once the cause is identified and fixed, the threshold should be raised.

**Update:**
Code coverage fell once again. We've [lowered](https://github.com/newrelic/newrelic-ruby-agent/pull/2385) the threshold to unblock PRs.

chynesNR:
Branch coverage is reported too unreliably.

PR #2696, opened 2024-06-05
Remove branch coverage check from PRs

tannalynn:
Branch coverage is wonky and not counting things correctly. We plan to look into it further in the future, but for now lets just not have branch coverage block PRs. I first tried removing the branch coverage but that didnt' seem to do anything somehow, so instead I lowered the value.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -357,3 +357,4 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           resultPath: lib/coverage_results/.last_run.json
           failedThreshold: 93.5
+          failedThresholdBranch: 50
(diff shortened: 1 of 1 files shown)


kaylareopelle (approved):
This works for me. I hope we can get to https://github.com/newrelic/newrelic-ruby-agent/issues/2309 next quarter.

.github/workflows/ci.yml
```diff
@@ -357,4 +357,4 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           resultPath: lib/coverage_results/.last_run.json
           failedThreshold: 93.5
-          failedThresholdBranch: 71.5
```
fallwith, line 360: Real coverage is quite high. The fractured, concurrent testing performed by GHA necessitates this lower number for now until we come up with a solution.

fallwith (approved):

PR #2697, opened 2024-06-06
SecurityInterface: Windows compatibility fix

fallwith:
Some unknown (concurrency or threading?) issue specific to Windows is causing the APM instrumentation logic to not be properly completed prior to the security agent being initialized.

A known fix is poll until `#connected?` responds with `true`, so let's place that check in the `SecurityInterface` as a preflight check.


fallwith:
closing in favor of a security agent repo based fix

PR #2710, opened 2024-06-12
set `expected: true` for matching status codes

### LINKED ISSUE #2708: API driven noticed errors display as `expected: true`, and observed errors should too
**Problem:** Observed exceptions are reported as `expected: false` even when they are treated as expected by the agent.

**Explanation:** There are 2 ways to inform the agent about an error. The public `notice_error` API can be called with an optional boolean based `expected` option, or the agent can observe an exception being raised by the Ruby application it monitors.

One a noticed error is in play, its origin does not matter when it comes to the logic the agent uses to process the error. When it comes to determining if the error is expected, the agent will make use of the `:'error_collector.expected_classes'`, `:'error_collector.expected_messages'`, and `:'error_collector.expected_status_codes'` configuration settings. And once an error is considered expected, that error won't negatively impact apdex. There will be no difference in agent behavior for API driven errors or observed errors.

But when it comes to reporting the error to the New Relic errors inbox, the origin of the error does make a difference when it comes to the `expected` attribute on the noticed error. API driven noticed errors that had the `expected: true` key/value pair present at API method call time will have that key/value pair persist all the way through to the errors inbox UI. But observed errors - despite effectively being treated as expected by the agent - will show up as `expected: false` in the UI.

From my interpretation of our agent specs, particularly the errors guide, I feel that this difference between API driven errors and observed errors is an oversight and therefore a bug. I believe we should set the `expected: false` key/value pair on observed errors once we have conditionally determined the error to be expected and before we report the error to the errors inbox.

**Definition of done:** if an `ActiveRecord::RecordNotFound` error is raised and results in an HTTP 404 status for a web request and the 404 code exists in the `:'error_collector.expected_status_codes'` configuration array, then the instance of the `ActiveRecord::RecordNotFound` error that appears in the errors inbox should display an attribute of "expected" with a value of "true".

fallwith:
previously, errors that were noticed and treated as expected because the HTTP status code matched the list configured via
`:'error_collector.expected_status_codes'` would not actually end up with their `expected` attribute set to a `true` value.

the relevant agent error processing behavior has been updated so that `expected: true` should always be applied to all errors that are treated as expected by the agent.

resolves #2708

### DIFF
--- a/lib/new_relic/agent/error_collector.rb
+++ b/lib/new_relic/agent/error_collector.rb
@@ -237,7 +237,8 @@ module NewRelic
       end
 
       def notice_segment_error(segment, exception, options = {})
-        return if skip_notice_error?(exception)
+        status_code = process_http_status_code(exception, options)
+        return if skip_notice_error?(exception, status_code)
 
         options.merge!(segment.llm_event.error_attributes(exception)) if segment.llm_event
 
@@ -250,15 +251,13 @@ module NewRelic
 
       # See NewRelic::Agent.notice_error for options and commentary
       def notice_error(exception, options = {}, span_id = nil)
-        state = ::NewRelic::Agent::Tracer.state
-        transaction = state.current_transaction
-        status_code = transaction&.http_response_code
-
+        status_code = process_http_status_code(exception, options)
         return if skip_notice_error?(exception, status_code)
 
         tag_exception(exception)
 
-        if options[:expected] || @error_filter.expected?(exception, status_code)
+        state = ::NewRelic::Agent::Tracer.state
+        if options[:expected]
           increment_expected_error_count!(state, exception)
         else
           increment_error_count!(state, exception, options)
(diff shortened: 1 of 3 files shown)


lib/new_relic/agent/error_collector.rb
```diff
@@ -250,26 +251,22 @@ def notice_segment_error(segment, exception, options = {})
 
       # See NewRelic::Agent.notice_error for options and commentary
       def notice_error(exception, options = {}, span_id = nil)
-        state = ::NewRelic::Agent::Tracer.state
-        transaction = state.current_transaction
-        status_code = transaction&.http_response_code
```
hannahramadan, line 255: I really like doing these two things in the new helper method. Clean

hannahramadan (approved):
This looks good to me! The additional cleanup is nice too

lib/new_relic/agent/error_collector.rb
```diff
@@ -250,26 +251,22 @@ def notice_segment_error(segment, exception, options = {})
...
-
+        status_code = process_http_status_code(exception, options)
         return if skip_notice_error?(exception, status_code)
 
         tag_exception(exception)
 
-        if options[:expected] || @error_filter.expected?(exception, status_code)
+        state = ::NewRelic::Agent::Tracer.state
+        if options[:expected]
```
tannalynn, line 260: Whats the reason for removing the || on the if? Is it because if it's expected it should have already returned on line 255?

fallwith: The bit of logic that was ORed (`||`) previously was moved to the new `process_http_status_code` method, which gets called before this line is reached. It doesn't lead to an early return, but it does lead to `expect: true` being set in the `options` hash itself. So now we can simplify the check here and only look for `expect: true` in the options hash.

tannalynn: Ahh I see, that totally makes sense, thanks for clarifying!

tannalynn (commented):
Looks good! Only thing left is the changelog entry.

PR #2716, opened 2024-06-13
Stripe: do not include path ids in metric names

### LINKED ISSUE #2654: Stripe metrics include ids in the group names
## Description

The Stripe integration sends the bare path as the metric name, leading to lots of metrics like this - 

```
Stripe/v1/events/evt_1PJc8exKnUatkhASXasx/get
Stripe/v1/events/evt_3PJZxsadnUat0yas1dea/get
Stripe/v1/customers/cus_DHYhjqasd21w56W4/sources/card_1JGLF9FiV4PKnUatf31fasdas/get
Stripe/v1/customers/cus_IcnFdy321eraSz6B/sources/card_1I1XezFiV4PKnU12s1asd/get
Stripe/v1/customers/cus_Be9k6IbvKHeRor/sources/card_1BGrR0FiV4PKn4casd@/get
Stripe/v1/subscription_items/si_PawAB/get
Stripe/v1/events/evt_3PJjdQFiV4PKnUAsd9/get
```

which can lead to NewRelic dropping metrics via the DENY_NEW_METRICS rule (https://docs.newrelic.com/docs/new-relic-solutions/solve-common-issues/troubleshooting/metric-grouping-issues/)

## Expected Behavior

ids are scrubbed from the path to allow metric grouping - maybe something like-  

```
Stripe/v1/events/*/get
Stripe/v1/customers/*/sources/*/get
Stripe/v1/subscription_items/*/get
```

## Steps to Reproduce

Run an app that uses the Stripe gem, newrelic_rpm, and have the `instrumentation.stripe` config enabled.

### LINKED ISSUE #2709: Stripe integration Causes Cardinality Explosion
Hi,

I noticed when using the stripe integration (added in #2180 ) that metric cardinality explodes with each outgoing http call being traced as separate metrics.

## Description
Specifically we're seeing metrics like `Stripe/v1/customers/cus_XXXXXXXXXXX/get` and `Stripe/v1/payment_methods/pm_XXXXXXXXX/get` etc... 
It caused a metric cardinality error and is preventing us from collecting any new metrics because of an auto-normalization.

## Expected Behavior
We'd expect a metric name like `Stripe/v1/payment_methods/get` 

## Your Environment
```
    newrelic_rpm (9.10.2)
    stripe (11.6.0)
```

## Additional context
Stripe's [example](https://github.com/stripe/stripe-ruby/tree/master?tab=readme-ov-file#example) instrumentation code shows the `path_parts` being split and dropping the middle (high cardinality) item.

I believe this is what's missing.

fallwith:
When naming Stripe metrics, include only the first two (API version and category) portions of the path and not the id. Continue to start the metric name with "Stripe" and end it with the method (ex: "get"). This will prevent metric name exhaustion resulting from regular Stripe usage.

Given a 'get' request performed against a path of
'/v1/payment_methods/pm_8675309', the metric name will now be 'Stripe/v1/payment_methods/get'.

resolves #2654
resolves #2709

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,15 @@
 
 ## dev
 
-Version <dev> fixes a bug related to expected errors not bearing a "true" value for the "expected" attribute if expected as a result of an HTTP status code match.
+Version <dev> fixes a bug related to expected errors not bearing a "true" value for the "expected" attribute if expected as a result of an HTTP status code match and changes the way Stripe instrumentation metrics are named to prevent high-cardinality issues.
 
 - **Bugfix: HTTP status code based expected errors will now have an "expected" value of "true"**
 
     Previously when an error was treated as expected by the agent as a result of a matching HTTP status code being found in the :'error_collector.expected_status_codes' configuration setting, the error would not appear with an "expected" attribute value of "true" in the errors in the errors inbox. [PR#2710](https://github.com/newrelic/newrelic-ruby-agent/pull/2710)
+
+- **Bugfix: Stripe metric names will no longer include full request paths to limit the unique name count**
+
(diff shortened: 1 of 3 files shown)


tannalynn (commented):
Looks great to me, only thing is to add the changelog entry!

hannahramadan (approved):
Nice explanation in the comment!

PR #2720, opened 2024-06-15
Add support for redis-clustering gem

### LINKED ISSUE #2444: New Relic doesn't track metrics for redis-clustering gem
[redis-rb](https://github.com/redis/redis-rb) moved the cluster support to [redis-clustering gem](https://github.com/redis/redis-rb/blob/master/cluster) companion gem which internally uses [redis-cluster-client](https://github.com/redis-rb/redis-cluster-client) gem. I am using ElastiCache cluster and upgrading `redis-rb` gem to latest version.

**Issue:** New Relic is tracking the data for connections created using [redis-client](https://github.com/redis-rb/redis-client) gem but not for [redis-cluster-client](https://github.com/redis-rb/redis-cluster-client) gem. What can I do to make New Relic work with redis cluster & redis-clustering gem.

`Redis::Client` support is there but `Redis::Cluster::Client` is not supported.

kaylareopelle:
The `redis-clustering` gem provides the ability to use Redis in cluster mode for Ruby. Cluster mode is enabled by default for AWS Elasticache Redis clusters.

The `redis-clustering` move was part of the large refactor of the Redis gem in the 5.0 release. It can leverage RedisClient middleware to instrument code.

This PR provides the following:
* An updated `db` method that is compatible with the Redis gem for all our supported versions, renamed to `nr_db`
* A new module, `NewRelic::Agent::Instrumentation::RedisClient::ClusterMiddleware`, that provides instrumentation for the `call` and `connect` methods.
  * We do not use Middleware for our main Redis instrumentation of the `call` and `connect` methods because they are invoked too late to capture certain errors. Keeping module prepending for these methods keeps feature parity for users of newer Redis versions
  * Since our redis-clustering support is new, we forgo the enhanced error capturing ability created by module prepending and choose to use Redis's preferred Middleware approach.
  
Resolves #2444


lib/new_relic/agent/instrumentation/redis.rb
```diff
@@ -33,6 +34,10 @@
     NewRelic::Agent.logger.info('Installing Redis Instrumentation')
     if NewRelic::Agent::Instrumentation::Redis::Constants::HAS_REDIS_CLIENT
       RedisClient.register(NewRelic::Agent::Instrumentation::RedisClient::Middleware)
+
+      if defined?(Redis::Cluster::Client)
+        return RedisClient.register(NewRelic::Agent::Instrumentation::RedisClient::ClusterMiddleware)
```
kaylareopelle, line 39: redis-clustering doesn't need `prepend` nor `chain` instrumentation, those methods are instrumented in the `ClusterMiddleware`.

lib/new_relic/agent/instrumentation/redis/cluster_middleware.rb
```diff
@@ -0,0 +1,26 @@
...
+    module ClusterMiddleware
+      include NewRelic::Agent::Instrumentation::Redis
+
+      # Until we decide to move our Redis instrumentation entirely off patches
+      # keep the middleware instrumentation for the call and connect methods
+      # limited to the redis-clustering instrumentation.
+      #
+      # Redis's middleware option does not capture errors as high in the stack
+      # as our patches. Leaving the patches for call and connect on the main
+      # Redis gem limits the feature disparity our customers experience.
```
kaylareopelle, line 16: I see this as a way to have a clean break for new Redis instrumentation. However, it does provide an inferior experience for error handling. We can look more closely into monkey-patching these methods, but we'll need to patch methods differently than we do for the main Redis gem.

test/multiverse/suites/redis/redis_instrumentation_test.rb
```diff
@@ -59,6 +59,8 @@ def test_records_metrics_for_connect
     end
 
     def test_records_connect_tt_node_within_call_that_triggered_it
+      skip_for_redis_clustering_gem
```
kaylareopelle, line 62: In the end, it seemed the least intrusive to add skips for the six tests that are not compatible with redis-clustering instrumentation:
* We can't test with `Redis::Cluster.new` because we don't have a cluster mode enabled service running for our tests, so we're faking the "cluster" testing regardless
* We don't have a different configuration option for `redis-clustering` than we do for `redis`, so the prepend/chain configurations seemed excessive
* Skipping the tests feels easier to read and understand the feature disparity between the two gems quickly

Open to further discussion to take a different approach!

test/multiverse/suites/redis/Envfile
```diff
@@ -14,8 +14,19 @@ def gem_list(redis_version = nil)
...
+
   RB
 end
 
 create_gemfiles(REDIS_VERSIONS)
+
+# We do not spin up a full Redis cluster in the testing environment, which
+# limits our ability to run unit tests on the redis-clustering behavior.
+# Since the testing capability is limited, only test the latest version of the
+# redis-clustering gem.
```
fallwith, line 26: to explain the Ruby constraint as well as the unpinned gem...

```suggestion
# redis-clustering gem, which itself requires Ruby v2.7+.
```

fallwith (approved):

lib/new_relic/agent/instrumentation/redis/instrumentation.rb
```diff
@@ -94,5 +87,17 @@ def _nr_redis_client_config
...
     end
+
+    def _nr_db
+      # db is a method on the Redis client in versions < 5.x
+      return db if respond_to?(:db)
+
+      begin
+        _nr_redis_client_config.db
+      rescue StandardError => e
+        NewRelic::Agent.logger.error("Failed to determine configured Redis db value: #{e.class} - #{e.message}")
```
tannalynn, line 98: is db something that could be memoized? Like would it be the same all the time for a specific instance, or could the db be different within the same instance? I'm mainly thinking about this in the context of other instrumentation we've done recently (like the dynamodb one), where customers were seeing log messages flood their logs when the client couldnt get the something.

Also, does it make sense that this would be an error level log, or would it be more of a warn or debug situation?

kaylareopelle: Good question! 

It looks like versions below 5.x could have the DB change at any time [(there's a setter and a getter)](https://github.com/redis/redis-rb/blob/v4.8.1/lib/redis/client.rb#L73-L79), so we shouldn't memoize those. 

However, versions 5.x+ don't have a setter for the DB and don't seem to have any expectations where a DB would change for a given config object [(here are the tests)](https://github.com/redis-rb/redis-client/blob/master/test/redis_client/config_test.rb). I think we should expect it to be the same for the given config/client.

Since we're already memoizing the `_nr_redis_client_config`, I'm not sure we'd get extra performance from memoizing the `db` value within that object. 

Despite this, we're still in a situation where lots of errors could be logged. 

I think you're right about the log level. It looks like similar messages are at both the `warn` and `debug` levels. I think we could put it in either, but maybe making it `debug` would limit our chances of causing a log flooding problem. Do you have a preference?

We could also change the code to run `_nr_redis_client_config.db` to only execute if the object responds to `db`. Do you prefer that approach?

tannalynn: That makes sense, thank you!

As for log level, i think based on issues filed relating to a lot of log msgs slowing customers down, I think debug would probably be the best bet, just in case.

I do also like your suggestion of just adding a responds_to instead though, if that seems like a good idea to you. db is allowed to be nil on datastore segments, so it not being available is still ok.

I think either a debug log level or responds_to are both good options to avoid the situation i was thinking about though, so whichever one you prefer is cool with me!

kaylareopelle: Nice! This all sounds good to me. Updated in 337f88842

tannalynn (approved):

PR #2728, opened 2024-06-20
yaml formatting: accommodate for indentation

fallwith:
When updating `newrelic.yml` via `rake newrelic:update_newrelicyml`, take into account the fact that the description text is indented one level (2 spaces) and therefore needs to be truncated at 78 columns instead of 80.


kaylareopelle (approved):
Great thinking! Thank you!

kaylareopelle (approved):

PR #2729, opened 2024-06-21
log decorator: support hashes

fallwith:
if the log decorator is given a hash, merge the metadata into the hash


hannahramadan (approved):
Looks great! Rubocop failures are mostly from the original branch

PR #2734, opened 2024-06-27
Replace `<DoNotTranslate>` with `<DNT>`

hannahramadan:
The docs team is replacing `<DoNotTranslate>` with `<DNT>`. We need to update this file to ensure we remain compatiable with docs parsing.

### DIFF
--- a/lib/tasks/helpers/config.html.erb
+++ b/lib/tasks/helpers/config.html.erb
@@ -76,7 +76,7 @@ When running the Ruby agent in a Rails app, the agent first looks for the `NEW_R
 When you edit the config file, be sure to:
 
 * Indent only with two spaces.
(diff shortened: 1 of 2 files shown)


fallwith (approved):
DNT! https://www.youtube.com/watch?v=NhsK5WExrnE

kaylareopelle (approved):
Thank you! 

I think we might want to make sure the docs team knows we're removing the `<DoNotTranslate>` tag from that `error_collector` line, just in case they know something we don't about its importance.

hannahramadan:
> Thank you!
> 
> I think we might want to make sure the docs team knows we're removing the `<DoNotTranslate>` tag from that `error_collector` line, just in case they know something we don't about its importance.

Good shout! I'll check with them before merging ⚡️

PR #2738, opened 2024-07-02
Add new span attributes for bunny instrumentation

tannalynn:
Adds new attributes to consume and produce spans created in our bunny instrumentation.

### DIFF
--- a/test/multiverse/suites/bunny/bunny_test.rb
+++ b/test/multiverse/suites/bunny/bunny_test.rb
@@ -356,7 +356,7 @@ class BunnyTest < Minitest::Test
     end
   end
 
-  def test_waluigi_publish_adds_attributes_to_span
+  def test_publish_adds_attributes_to_span
     NewRelic::Agent.stubs(:logger).returns(NewRelic::Agent::MemoryLogger.new)
 
     with_queue do |queue|
--- a/test/multiverse/suites/bunny/bunny_test.rb
+++ b/test/multiverse/suites/bunny/bunny_test.rb
@@ -344,7 +344,6 @@ class BunnyTest < Minitest::Test
 
         queue.pop
       end
-      sleep 0.5
       spans = harvest_span_events!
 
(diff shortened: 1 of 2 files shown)


hannahramadan (dismissed):
makes sense to me!

test/multiverse/suites/bunny/bunny_test.rb
```diff
@@ -333,6 +336,42 @@ def test_pop_returning_no_message_doesnt_error
...
 
+  def test_pop_adds_attributes_to_span
+    with_queue do |queue|
+      queue.publish('test_msg', routing_key: queue.name)
+      in_transaction('test_txn') do |txn|
+        txn.stubs(:sampled?).returns(true)
+
+        queue.pop
+      end
+      sleep 0.5
```
fallwith, line 347: How did you arrive at `0.5`? Would the [existing practice](https://github.com/newrelic/newrelic-ruby-agent/blob/7d779876af509c4c59436a5e49057d175964aac7/test/multiverse/suites/bunny/bunny_test.rb#L219) of sleeping `0.1` in a loop for a maximum of 10 times see less time spent sleeping on average? If so, you can move that loop out into a helper for all 3 tests that require sleeping.

kaylareopelle (approved):
This looks good to me! I'm so glad the testing stuff got figured out!

If you wanted to, the SpanEventPrimitive class has constants you could use for the strings 'server.address' and 'server.port', but it might just be easier to read the string within the code. Whatever you think is best!

test/multiverse/suites/bunny/bunny_test.rb
tannalynn: oh thats my bad! I don't actually need them lol. i added them as a random thing to try when i was having those confusing errors yesterday before I knew what was going on. it didn't help and then i just moved on and totally forgot that they were there. i removed them now though https://github.com/newrelic/newrelic-ruby-agent/pull/2738/commits/1827a7aea31a5e53455459dd6b66141f8d254b7e

PR #2745, opened 2024-07-12
include additional AC topics in Rails subscription

fallwith:
When subscribing to Action Controller topics for Rails notifications, include 4 new topics in the subscription regex:

- `exist_fragment?`
- `expire_fragment`
- `read_fragment`
- `write_fragment`

### DIFF
--- a/lib/new_relic/agent/instrumentation/rails_notifications/action_controller.rb
+++ b/lib/new_relic/agent/instrumentation/rails_notifications/action_controller.rb
@@ -32,15 +32,19 @@ DependencyDetection.defer do
     NewRelic::Agent::Instrumentation::ActionControllerSubscriber \
       .subscribe(/^process_action.action_controller$/)
 
-    subs = %w[send_file
-      send_data
-      send_stream
-      redirect_to
+    subs = %w[exist_fragment?
+      expire_fragment
       halted_callback
-      unpermitted_parameters]
+      read_fragment
+      redirect_to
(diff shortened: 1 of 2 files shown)


lib/new_relic/agent/instrumentation/rails_notifications/action_controller.rb
```diff
@@ -32,15 +32,19 @@
...
-      redirect_to
-      halted_callback
-      unpermitted_parameters]
+      write_fragment
+      unpermitted_parameters].map { |s| Regexp.escape(s) }
 
     # have to double escape period because its going from string -> regex
     NewRelic::Agent::Instrumentation::ActionControllerOtherSubscriber \
-      .subscribe(Regexp.new("^(#{subs.join('|')})\\.action_controller$"))
+      .subscribe(Regexp.new("^(?:#{subs.join('|')})\\.action_controller$"))
```
hannahramadan, line 48: What is the addition of `?:` doing as a non-regex pro?

fallwith: By default, wrapping part of the pattern in parens will cause the part of the match to be captured. By using `(?:)` instead of `()`, I'm telling both the parser and the future human reviewer that we don't actually intend to do any capturing with this regex and simply want to use the parens for grouping.

kaylareopelle (approved):

PR #2754, opened 2024-07-17
Re-enable Rails EDGE CI testing, address Rails v8.0-alpha specific issues

fallwith:
- Re-enable CI testing of Rails EDGE
- Apply test fixes for Rails v8.0-alpha
- Drop ActiveRecord tests for gem versions below v4.0


kaylareopelle (approved):
Thanks for the cleanup work too!

PR #2761, opened 2024-07-22
Run url test only on its own

tannalynn:
Due to random URL test issues that happen, let's move it out into it's own run that will notify us when it's messed up, instead of running with the regular CI tests. This will help us know at a quick glance that the failure is URL related, and also won't cause failures when running the ci cron for unrelated reason.


kaylareopelle (approved):
Thank you for taking care of this, Tanna!

Issue #2776, opened 2024-07-25
Add `windows` OS detection in environment report


kaylareopelle:
Expected: `Windows_NT`/`windows`

Even if newrelic_rpm doesn't support Windows env try to provide a generic name for windows env.

Issue #2775, opened 2024-07-25
Update `darwin` environment report output


kaylareopelle:
Expected: `darwin`, Received: `darwin22`/`darwin23`

kaylareopelle:
@prateeksen - Getting `darwin` plus the version (ex. `darwin22`, `darwin23`) seems like valuable context. 

Why would you like the version number removed?

prateeksen:
@kaylareopelle IAST SaaS engine checks for the versions of the OS/Frameworks/Dispatcher and mark the applications on the customer UI whether those are supported by IAST or not. In case of Mac they have check on 'Mac OS X' or 'darwin'. When ruby APM agent sends `darwin` plus the version (ex. `darwin22`, `darwin23`) validation fails and IAST version checker shows that application is not supported by IAST. Adding `darwin22`, `darwin23` is not feasible as every newer versions will need to be added manually in future.
Although, I can ask IAST SaaS team if they can check `darwin*` instead of `darwin` but I find sending OS name only in os name field more better as [OS version](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/environment_report.rb#L69) is already there as a separate field.

kaylareopelle:
Hi @prateeksen, thanks for the clarification, I didn't realize there was an OS version field. This update also gets us closer to the OS attributes for OpenTelemetry. 

I appreciate you catching the issue and opening a PR for the change!

PR #2779, opened 2024-07-26
Update OS detection for environment report

prateeksen:
- Update OS detection for MacOS #2775 
- Update OS detection for Windows #2776

### DIFF
--- a/lib/new_relic/agent/system_info.rb
+++ b/lib/new_relic/agent/system_info.rb
@@ -30,7 +30,7 @@ module NewRelic
       end
 
       def self.windows?
-        !!(ruby_os_identifier =~ /mingw/i || ruby_os_identifier =~ /mswin/i)
+        !!(ruby_os_identifier[/mingw|mswin/i])
       end
 
       def self.darwin?
--- a/lib/new_relic/agent/system_info.rb
+++ b/lib/new_relic/agent/system_info.rb
@@ -19,6 +19,15 @@ module NewRelic
         RbConfig::CONFIG['target_os']
       end
 
+      def self.os_distribution
+        case
+        when darwin? then :darwin
+        when linux? then :linux
+        when bsd? then :bsd
+        else ruby_os_identifier
+        end
+      end
+
       def self.darwin?
         !!(ruby_os_identifier =~ /darwin/i)
       end
--- a/lib/new_relic/agent/system_info.rb
+++ b/lib/new_relic/agent/system_info.rb
@@ -24,8 +24,13 @@ module NewRelic
         when darwin? then :darwin
         when linux? then :linux
         when bsd? then :bsd
+        when windows? then :windows
         else ruby_os_identifier
         end
+      end
+
+      def self.windows?
+        !!(ruby_os_identifier =~ /mingw/i || ruby_os_identifier =~ /mswin/i)
       end
 
       def self.darwin?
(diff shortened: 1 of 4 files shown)


kaylareopelle (changes requested):
@prateeksen - Thank you for this submission! Though the existing tests pass, these methods add some new functionality that we'd like to have test coverage for. 

Could you add tests for your new code?

lib/new_relic/agent/system_info.rb
```diff
@@ -19,6 +19,20 @@ def self.ruby_os_identifier
...
+        when darwin? then :darwin
+        when linux? then :linux
+        when bsd? then :bsd
+        when windows? then :windows
+        else ruby_os_identifier
+        end
+      end
+
+      def self.windows?
+        !!(ruby_os_identifier =~ /mingw/i || ruby_os_identifier =~ /mswin/i)
```
fallwith, line 33: You can have Ruby perform just a single regex match operation by combining the 2 strings with a pipe, and then directly change `=~` to `!~` to prevent having to negate (`!!`) the result:

```suggestion
        ruby_os_identifier !~ /(?:mingw|mswin)/i
```

test/new_relic/agent/system_info_test.rb
```diff
@@ -324,6 +324,46 @@ def test_system_info_bsd_predicate
     refute_predicate NewRelic::Agent::SystemInfo, :bsd?, 'Did not expect OS to match bsd'
   end
 
+  def test_system_info_windows_predicate
+    NewRelic::Agent::SystemInfo.stubs(:ruby_os_identifier).returns('mingw32')
+
+    assert_predicate NewRelic::Agent::SystemInfo, :windows?, 'Expected OS to match windows'
+
+    NewRelic::Agent::SystemInfo.stubs(:ruby_os_identifier).returns('darwin13')
```
fallwith, line 332: I think the 2 halves of this test would be better off living in 2 separate tests. Perhaps:

- test_system_info_windows_is_true
- test_system_info_windows_is_false

prateeksen: @fallwith I have made this test similar to ones already present for linux and mac. If you still want it to be separated out, I can do.

lib/new_relic/agent/system_info.rb
prateeksen: @fallwith I have checked your suggestion it doesn't give the expected output. Instead i have made a different optimisation [here](https://github.com/newrelic/newrelic-ruby-agent/pull/2779/commits/6a0fa469c14e5c3a80ff2181b1098ea4198d7cc9)

fallwith: Ah, sorry about that @prateeksen. I read the line of code too quickly. Given that you are in a method called `windows?`, you wanted `true` for Windows systems, not `false`. I should have suggested one of these:

- ruby_os_identifier =~ /(?:mingw|mswin)/i
- ruby_os_identifier.match?(/(?:mingw|mswin)/i)

Your approach of using `[]` works well too!

test/new_relic/agent/system_info_test.rb
fallwith: Ok. I was only looking at the new code from the GitHub diff view.

fallwith:
I recognize @kaylareopelle's [request](https://github.com/newrelic/newrelic-ruby-agent/pull/2779#pullrequestreview-2202284658) for tests as satisfied now.

fallwith (approved):

PR #2789, opened 2024-07-30
Grape padrino framework detection

prateeksen:
- Grape and Padrino framework detection for environment report. 
  This information is required for newrelic_security to show on UI whether application is supported or not.

### DIFF
--- a/lib/new_relic/control/frameworks/grape.rb
+++ b/lib/new_relic/control/frameworks/grape.rb
@@ -1,0 +1,20 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'new_relic/control/frameworks/ruby'
+module NewRelic
+  class Control
+    module Frameworks
+      # Contains basic control logic for Grape
+      class Grape < NewRelic::Control::Frameworks::Ruby
+        protected
+
+        def install_shim
+          super
+          ::Grape::API::Instance.class_eval { include NewRelic::Agent::Instrumentation::ControllerInstrumentation::Shim }
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/new_relic/control/frameworks/padrino.rb
```diff
@@ -0,0 +1,20 @@
...
+    module Frameworks
+      # Contains basic control logic for Padrino
+      class Padrino < NewRelic::Control::Frameworks::Ruby
+        protected
+
+        def install_shim
+          super
+          ::Padrino::PathRouter::Router.class_eval { include NewRelic::Agent::Instrumentation::ControllerInstrumentation::Shim }
+        end
+      end
```
kaylareopelle, line 17: Rather than install the shim directly into Padrino, I think we should edit this class so that it inherits from the Sinatra frameworks class. That will keep the behavior closer to what existed before.
```suggestion
      class Padrino < NewRelic::Control::Frameworks::Sinatra
      end
```

lib/new_relic/control/frameworks/grape.rb
```diff
@@ -0,0 +1,20 @@
...
+  class Control
+    module Frameworks
+      # Contains basic control logic for Grape
+      class Grape < NewRelic::Control::Frameworks::Ruby
+        protected
+
+        def install_shim
+          super
+          ::Grape::API::Instance.class_eval { include NewRelic::Agent::Instrumentation::ControllerInstrumentation::Shim }
+        end
```
kaylareopelle, line 16: This is adding new functionality I don't think Grape users need. I would prefer to remove this method from the class. Allowing Grape framework applications to inherit from the Ruby framework class will keep the experience consistent with what it was before.

```suggestion
```

lib/new_relic/control/frameworks/padrino.rb
kaylareopelle: If you accept this suggestion, the `require` statement on line 5 most likely needs to be updated as well to require the Sinatra file instead of the Ruby file.

kaylareopelle (changes requested):
Thanks for your patience while I ran some tests with this change! I have a few suggestions to keep the experience more consistent for Padrino and Grape customers.

prateeksen:
Hi @kaylareopelle I have accepted and added your suggestions.

tannalynn (approved):

fallwith (approved):

hannahramadan (approved):

PR #2796, opened 2024-08-07
OpenSearch instrumentation

### LINKED ISSUE #2228: Add instrumentation for OpenSearch
### Feature Description
[OpenSearch](https://opensearch.org) is a fork of Elasticsearch spearheaded by Amazon. Elasticsearch instrumentation was added with #1469. Since this is a fork, much of the work necessary will be highly similar to what is currently present, for example the `perform_request` method is found in [OpenSearch::Transport::Client](https://github.com/elastic/elastic-transport-ruby/blob/5a63daf19f73fc7048d2e80c72c7e8e7bc03a74a/lib/elastic/transport/client.rb#L171-L175).

### Describe Alternatives
None

### Priority
We are currently in the process of migrating to OpenSearch and loosing instrumentation is the only downside. Of course the instrumentation for Elasticsearch was added rather recently and we have been doing fine before that, so I'll put this as "Really Want".

kaylareopelle:
# OpenSearch Instrumentation

Follows the patterns established in the Elasticsearch instrumentation

Todo: 
* Set up CI docker services to run with tests - This will be handled in #2800 

Resolves #2228

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,7 +114,7 @@ jobs:
     runs-on: ubuntu-22.04
     services:
       opensearch:
-        image: opensearchproject/opensearch:1.1.0
+        image: opensearchproject/opensearch:1.3.18
         ports:
           - 9200:9200
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,8 +189,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        multiverse: [rest]
-        ruby-version: [2.4.10, 3.3.4]
+        multiverse: [opensearch]
+        ruby-version: [3.3.4]
 
     steps:
       - name: Configure git
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,8 +140,6 @@ jobs:
         image: opensearchproject/opensearch:1.3.18
         env:
           discovery.type: single-node
-          plugins.security.disabled: true
-          DISABLE_INSTALL_DEMO_CONFIG: true
         ports:
           - 9200:9200
           - 9600:9600
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - name: Configure git
         run: 'git config --global init.defaultBranch main'
-      - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # tag v4.1.2
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # tag v4.1.7
       - uses: ruby/setup-ruby@a6e6f86333f0a2523ece813039b8b4be04560854 # tag v1.190.0
         with:
           ruby-version: '3.3'
@@ -40,7 +40,7 @@ jobs:
     steps:
       - name: Configure git
         run: 'git config --global init.defaultBranch main'
-      - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # tag v4.1.2
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # tag v4.1.7
 
         # - curl is needed for Curb
         # - xslt is needed for older Nokogiris, RUBY_VERSION < 2.5
@@ -74,7 +74,7 @@ jobs:
       - if: matrix.ruby-version == '2.4.10'
         name: Cache mysql55
         id: mysql55-cache
-        uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # tag v3.3.1
+        uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # tag v4.0.2
         with:
           path: /usr/local/mysql55
           key: mysql55-install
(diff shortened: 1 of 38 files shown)


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1463,6 +1463,15 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of bunny at start-up. May be one of: `auto`, `prepend`, `chain`, `disabled`.'
         },
+        :'instrumentation.opensearch' => {
```
kaylareopelle, line 1466: The gem is called `opensearch-ruby`. The class is spelled `OpenSearch`, which is also the spelling used in the documentation. 

I'm torn between whether the lowercase version of the name should be `opensearch`, `open_search`, or `opensearch-ruby`.

.github/workflows/ci.yml
```diff
@@ -118,7 +118,7 @@ jobs:
         env:
           discovery.type: single-node
         ports:
-          - 9200:9200
+          - 9000:9200
```
kaylareopelle, line 121: I'm having trouble getting opensearch to run on a port that isn't 9200, so I'm trying to see if elasticsearch 7.x can accommodate a different port

lib/new_relic/agent/configuration/default_source.rb
fallwith: My order of preference is:

1. `opensearch-ruby` - match the gem name. I like this best.
2. `open_search` - given that the gem is a first party offering from the official OpenSearch project, converting the camel-cased "OpenSearch" into snake-cased "open_search" is reasonable.
3. `opensearch` - I want to not like this one, but they did name the gem 'opensearch-ruby' instead of 'open_search-ruby', so I suppose I can't argue with the lack of an underscore too much.

fallwith (dismissed):

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: Thank you! One thing on `opensearch-ruby`, I think we normally convert `-` to `_` for config options and directories. I'm not sure why, but for consistency with that process, we may want to use `_` if we go with the gem name.

fallwith: Ah, I see. So no matter what no approach will be perfect. Perhaps just merge this as-is then, as it makes plenty of sense!

.github/workflows/ci.yml
fallwith: You can leave OpenSearch running on 9200, and just have the localhost mapping be on another port such as 9275. That way the OpenSearch server instance is happy but there won't be a localhost port use collission.

lib/new_relic/agent/instrumentation/opensearch.rb
```diff
@@ -0,0 +1,25 @@
...
+
+require_relative 'opensearch/instrumentation'
+require_relative 'opensearch/chain'
+require_relative 'opensearch/prepend'
+
+DependencyDetection.defer do
+  named :opensearch
+
+  depends_on do
+    defined?(OpenSearch)
```
hannahramadan, line 13: Were there any version constraints that we know of?

kaylareopelle: Good question! The tests passed on Ruby 3.4 as far back as 1.x, so I don't think we have any version constraints. The 2.x and 1.x versions are compatible with Ruby 2.4, so I could add them to the Envfile, even though we're not currently running that on the CI. What do you think?

kaylareopelle: Added a 2.4.x compatible version in: 0f77f6064

lib/new_relic/agent/instrumentation/opensearch/instrumentation.rb
```diff
@@ -0,0 +1,66 @@
...
+
+module NewRelic::Agent::Instrumentation
+  module OpenSearch
+    PRODUCT_NAME = 'OpenSearch'
+    OPERATION = 'perform_request'
+    OPERATION_PATTERN = %r{/lib/opensearch/api/(?!.+#{OPERATION})}
+    INSTANCE_METHOD_PATTERN = /:in (?:`|')(?:.+#)?([^']+)'\z/
+    INSTRUMENTATION_NAME = NewRelic::Agent.base_name(name)
+
+    def perform_request_with_tracing(_method, _path, params = {}, body = nil, _headers = nil, _opts = {}, &_block)
```
fallwith, line 13: I'm looking at the two Hash instantiations happening per call when the defaults are involved.

- Perhaps the `params` one can be set to `NewRelic::EMPTY_HASH`
- It looks like `_opts` isn't used, so perhaps it can be replaced with `nil`

lib/tasks/instrumentation_generator/templates/dependency_detection.tt
```diff
@@ -12,16 +12,16 @@ DependencyDetection.defer do
   depends_on do
     # The class that needs to be defined to prepend/chain onto. This can be used
     # to determine whether the library is installed.
-    defined?(::<%= @class_name %>)
+    defined?(<%= @class_name %>)
```
fallwith, line 15: Would you please explain the need for the changes here? Perhaps a separate PR if they're not OpenSearch driven?

.github/workflows/ci.yml
kaylareopelle: Thank you! I decided to revert this since we're not adding OpenSearch to the CI at this time.

lib/tasks/instrumentation_generator/templates/dependency_detection.tt
kaylareopelle: I can open a separate PR for this work. Rubocop didn't like the `::` prefix on the generated files, so I figured we should update the template to reflect its current preferences.

kaylareopelle: New PR: https://github.com/newrelic/newrelic-ruby-agent/pull/2805

lib/new_relic/agent/instrumentation/opensearch/instrumentation.rb
kaylareopelle: Fixed in: 35b1090a4

fallwith (approved):

hannahramadan (approved):

PR #2798, opened 2024-08-08
JRuby fix for concurrency issue with manager cache

fallwith:
The configuration manager's alteration of its internal cache hash to preserve previously dynamically defined values has been known to cause problems with JRuby given its concurrent access to the hash object. We have used `synchronize` and `Hash#dup` fixes previously, but still have reports of issues.

When it comes down to it, the alteration of the hash was only added and still only needed for use by the security agent. So let's just not even attempt to alter the hash if JRuby is in play but the security agent is not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## dev
 
-Version <dev> updates framework detection and fixes Falcon dispatcher detection.
+Version <dev> updates framework detection, fixes Falcon dispatcher detection, and addresses a JRuby specific concurrency issue.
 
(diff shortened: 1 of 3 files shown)


kaylareopelle (dismissed):

kaylareopelle:
@fallwith  - Should a changelog entry be added for this?

fallwith:
> @fallwith - Should a changelog entry be added for this?

Sure. Here's one with [cfab1c7](https://github.com/newrelic/newrelic-ruby-agent/pull/2798/commits/cfab1c7d796059b608af453907e3491d6e5405c3)

kaylareopelle (approved):

PR #2799, opened 2024-08-09
Vendor variable-mapper

### LINKED ISSUE #2603: CI: Update `kanga333/variable-mapper` to a version that uses Node.js 20+
The version of `kanga333/variable-mapper` we use in our workflows runs on Node 16. This adds a deprecation warning annotation to our runs, for example:

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: kanga333/variable-mapper@<git-sha>. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

Update the version of `kanga333/variable-mapper` used in all our workflows to a version that uses Node.js 20 or higher. At the time of creating this ticket, Node.js 22 was the most recent version.

hannahramadan:
The action [kanga333/variable-mapper](https://github.com/kanga333/variable-mapper) doesn't seem to be actively maintained. This PR vendors the action and updates it to using Node.js20

Closes #2603


.github/workflows/ci_cron.yml
```diff
@@ -52,10 +52,10 @@ jobs:
       - name: Install Ruby ${{ matrix.ruby-version }}
         uses: ruby/setup-ruby@a6e6f86333f0a2523ece813039b8b4be04560854 # tag v1.190.0
         with:
-          ruby-version: ${{ matrix.ruby-version }}
+          ruby-version: ${{ matrix.ruby-version }}                                                                                                               
 
       - name: Set up mini-envs for ruby version
-        uses: kanga333/variable-mapper@3681b75f5c6c00162721168fb91ab74925eaebcb # tag v0.3.0
+        uses: ./.github/actions/variable-mapper # no tag - locally sourced
```
fallwith, line 58: I think you can discard these comments. They serve as a map between a sha and a tag version. The `./` local path prefix will already convey that the action is locally sourced.

fallwith (dismissed):

.github/workflows/ci_cron.yml
hannahramadan: This tag was added [once upon a time](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/.github/workflows/pr_closed.yml#L17) for another local action so I thought to follow that pattern! But I agree it should be implied

kaylareopelle (approved):

PR #2801, opened 2024-08-09
Vendor workflow-conclusion

### LINKED ISSUE #2790: Create local `technote-space/workflow-conclusion-action` gha
The action we use to send CI results to Slack, [technote-space/workflow-conclusion-action](https://github.com/technote-space/workflow-conclusion-action), has been archived as of Nov 16, 2023.

It currently throws a Node.js 16 deprecated warning. We should consider using a local version of this action similar to .github/actions/issue_closer or .github/actions/simplecov-report so that we can update the Node.js version and keep a copy of this action.

hannahramadan:
The action we use to send CI results to Slack, [technote-space/workflow-conclusion-action](https://github.com/technote-space/workflow-conclusion-action), has been archived as of Nov 16, 2023. Moving this to a local copy and updating the Node.js20

Closes #2790

### DIFF
--- a/.github/actions/workflow-conclusion/package.json
+++ b/.github/actions/workflow-conclusion/package.json
@@ -1,0 +1,73 @@
+{
+  "name": "@technote-space/workflow-conclusion-action",
+  "version": "3.0.3",
+  "description": "GitHub action to get workflow conclusion.",
+  "keywords": [
+    "github",
+    "github actions"
+  ],
+  "homepage": "https://github.com/technote-space/workflow-conclusion-action",
+  "bugs": {
+    "url": "https://github.com/technote-space/workflow-conclusion-action/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/technote-space/workflow-conclusion-action.git"
+  },
+  "license": "MIT",
+  "author": {
+    "name": "Technote",
+    "email": "technote.space@gmail.com",
+    "url": "https://technote.space"
+  },
+  "files": [
+    "lib",
+    "action.yml"
+  ],
+  "scripts": {
+    "build": "rm -rdf lib && rollup -c",
+    "cover": "vitest run --coverage",
+    "postinstall": "[ -n \"$CI\" ] || [ ! -f node_modules/.bin/husky ] || husky install",
+    "lint": "eslint 'src/**/*.ts' --cache",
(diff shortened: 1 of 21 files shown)


.github/actions/workflow-conclusion/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "@rollup/plugin-commonjs": "^23.0.3",
+    "@rollup/plugin-json": "^5.0.2",
+    "@rollup/plugin-node-resolve": "^15.0.1",
+    "@rollup/plugin-typescript": "^10.0.1",
+    "@sindresorhus/tsconfig": "^3.0.1",
+    "@technote-space/github-action-helper": "^5.3.10",
+    "@technote-space/github-action-log-helper": "^0.2.9",
+    "@technote-space/github-action-test-helper": "^0.11.1",
+    "@technote-space/release-github-actions-cli": "^1.9.3",
+    "@types/node": "^18.11.11",
```
fallwith, line 56: Maybe with some follow-up work we ought to use versions that are Node.js v20+ appropriate in case we ever want to repackage.

.github/workflows/ci_special.yml
```diff
@@ -64,7 +64,7 @@ jobs:
     runs-on: ubuntu-22.04
     if: always()
     steps:
-      - uses: technote-space/workflow-conclusion-action@45ce8e0eb155657ab8ccf346ade734257fd196a5 # tag v3.0.3
+      - uses: ./.github/workflows/scripts//workflow-conclusion # no tag - locally sourced
```
fallwith, line 67: Same thought as on the other PR. I'd prefer these comments to be dropped as redundant.

fallwith (dismissed):

.github/workflows/ci_special.yml
```diff
@@ -47,7 +47,7 @@ jobs:
...
```
kaylareopelle, line 50: Doe the double `//` cause a problem? I would've expected to see `./.github/workflows/scripts/workflow-conclusion`

hannahramadan: Oh! This def would've cause an issue. I suppose this action isn't running since this only runs on a merge. I will fix and look into this

kaylareopelle (approved):

PR #2804, opened 2024-08-12
sidekiq test: perms fix for inline bundled redis

fallwith:
We have observed CI errors related to Bundler refusing to clean up its inline bundled instance of `redis`. This fix seeks to proactively prevent the underlying permissions issue that causes the error.


tannalynn (approved):

kaylareopelle (approved):
Thank you!

PR #2811, opened 2024-08-13
Enhanced AWS Lambda serverless functionality

fallwith:
- Introduce support for parsing distributed tracing information for HTTP based trigger invocations of the instrumented Lambda function
- Both AWS API Gateway versions 1.0 and 2.0 are supported
- For web driven invocations of an instrumented function, categorize the invocation as being "web" based, and record relevant HTTP information including the method, URI, and status code post invocation.
- Recognize and report on fully 12 separate AWS resources that are capable of triggering a Lambda function invocation: ALB, API Gateway V1, API Gateway V2, CloudFront, CloudWatch Scheduler, DynamoStreams, Firehose, Kinesis, S3, SES, SNS, and SQS.
- If an AWS resource based trigger is identified, record at least the type of the resource and the relevant arn. For many resources, also record additional context specific information. For example, for an S3 based invocation, record the S3 bucket name.


lib/new_relic/agent/serverless_handler.rb
```diff
@@ -129,14 +144,92 @@ def payload_v1
...
+
+        url_str = "https://#{info[:host]}:#{info[:port]}#{info[:path]}"
+        if info[:query_parameters]
+          qp = info[:query_parameters].map { |k, v| "#{k}=#{v}" }.join('&')
+          url_str += "?#{qp}"
+        end
+
+        URI.parse(url_str)
+      rescue StandardError
+      end
```
kaylareopelle, line 197: Why rescue here, but not elsewhere?

lib/new_relic/agent/serverless_handler.rb
```diff
@@ -146,15 +239,142 @@ def use_named_pipe?
...
+
+      def event_source_event_info
+        return unless @event
+
+        # if every required key for a source is found, consider that source
+        # to be a match
+        EVENT_SOURCES.each do |_type, info|
+          return info unless info['required_keys'].detect { |r| @event.dig(*r).nil? }
+        end
```
kaylareopelle, line 340: What about using `each_value` here instead? [Ruby 2.4 should support it](https://ruby-doc.org/core-2.4.10/Hash.html#method-i-each_value).

 ```ruby
 EVENT_SOURCES.each_value do |info|
          return info unless info['required_keys'].detect { |r| @event.dig(*r).nil? }
        end
 ```
 (Rubocop gave me this rec recently on an OTel repo: https://docs.rubocop.org/rubocop/cops_style.html#stylehasheachmethods; benchmarks show a same-ish result between the two options)

test/new_relic/agent/serverless_handler_test.rb
```diff
@@ -46,6 +87,8 @@ def teardown
       # integration style
 
       def test_the_complete_handoff_from_the_nr_lambda_layer
+        skip 'This serverless test is limited to Ruby v3.2+' unless ruby_version_float >= 3.2
```
kaylareopelle, line 90: Should the whole file be skipped unless Ruby > 3.2? Would this work in setup?

kaylareopelle (commented):
🍛

CHANGELOG.md
```diff
@@ -2,7 +2,11 @@
...
+
+- **Feature: Enhanced AWS Lambda function instrumentation**
+
+When utilized via the latest [New Relic Ruby layer for AWS Lambda](https://layers.newrelic-external.com/), the agent now offers enhanced support for AWS Lambda function instrumentation. The agent's instrumentation for AWS Lambda functions now supports distributed tracing. Web-triggered invocations are now identified as being "web" based when an API Gateway call is involved, with support for both API Gateway versions 1.0 and 2.0. Web-based calls have the HTTP method, URI, and status code recorded. The agent now recognizes and reports on 12 separate AWS resources that are capable of triggering a Lambda function invocation: ALB, API Gateway V1, API Gateway V2, CloudFront, CloudWatch Scheduler, DynamoStreams, Firehose, Kinesis, S3, SES, SNS, and SQS. The type of the triggering resource and its ARN will be recorded for each resource, and for many of them, extra resource specific attributes will be recorded as well. For example, Lambda function invocations triggered by S3 bucket activity will now result in the S3 bucket name being recorded. [PR#2811](https://github.com/newrelic/newrelic-ruby-agent/pull/2811)
```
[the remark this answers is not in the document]
fallwith: I thought about this originally and decided against it but having you also think the same AND reformat it settles it - let's go with the bullets. Thank you.

lib/new_relic/agent/serverless_handler.rb
```diff
@@ -129,14 +144,92 @@ def payload_v1
...
+
+        info = api_v == 2 ? info_for_api_gateway_v2 : info_for_api_gateway_v1
+        info[:query_parameters] = @event.fetch('queryStringParameters', nil)
+
+        @http_method = info[:method]
+        @http_uri = http_uri(info)
+      end
+
+      def http_uri(info)
+        return unless info[:host] && info[:path]
```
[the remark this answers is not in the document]
fallwith: Both the API Gateway v1 and v2 info gathering methods will force a default of `443` for the port value, so we can always rely on it existing.

lib/new_relic/agent/serverless_handler.rb
fallwith: Whoops! Thank you for spotting this. This `rescue` was only added for some debugging and I missed removing it. Removed with [c27f24c](https://github.com/newrelic/newrelic-ruby-agent/pull/2811/commits/c27f24c3d82aa2ddd10376fa224765a68234da98).

lib/new_relic/agent/serverless_handler.rb
fallwith: Good call. In an earlier iteration of this code, I noticed the `_type` silliness and tried a refactor. I might have used `.values.each` instead of your `each_value`, but it was something similar. For some reason, doing so broke the functionality. However, this was in an earlier version of the code than what made it to the PR and it was doing more than this one-liner does. Using `each_value` here with the one-liner has been tested to work great. Thanks! Added with [af52a76](https://github.com/newrelic/newrelic-ruby-agent/pull/2811/commits/af52a76ae743ea8629c6c9138d6077b753e5ab50).

test/new_relic/agent/serverless_handler_test.rb
fallwith: That's a good idea. All tests now skipped via [839bb1b](https://github.com/newrelic/newrelic-ruby-agent/pull/2811/commits/839bb1b0e777c4dbe5d710b99efce296de5cc4d3)

lib/new_relic/agent/serverless_handler.rb
```diff
@@ -22,25 +22,29 @@ class ServerlessHandler
       SUPPORTABILITY_METRIC = 'Supportability/AWSLambda/HandlerInvocation'
       FUNCTION_NAME = 'lambda_function'
       PAYLOAD_VERSION = ENV.fetch('NEW_RELIC_SERVERLESS_PAYLOAD_VERSION', 2)
+      DIGIT = /\d/.freeze
```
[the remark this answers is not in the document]
fallwith: Thank you. Removed in [f3768db](https://github.com/newrelic/newrelic-ruby-agent/pull/2811/commits/f3768db4d5dbcc8d522a36c96928cfc8fec602b6). Even though the object id stays the same, without `#freeze` the pattern object will now return `false` to a `#frozen?` check. I was wistfully thinking that the Ruby VM would somehow give us some performance benefit with a frozen pattern over a non-frozen one, but if it's the same object then that's unlikely.

kaylareopelle (approved):
Thanks, @fallwith!

PR #2812, opened 2024-08-13
CI fixes

fallwith:
- Don't perform the bundler/inline based Sidekiq regression test under GHA
- Prevent "already activated" securerandom Bundler errors when testing with the latest delayed_job gem which will grab activesupport v7.2.0


fallwith:
CI run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/10378331787

kaylareopelle (approved):
Thank you!

PR #2821, opened 2024-08-19
ServerlessHandler: error handling for URI parsing

fallwith:
To address issues with the recent serverless enhancements discovered through local dev testing with the Node.js tool 'serverless':

- Don't append a port value to a URI string that already contains one
- If the URI string construction or parsing fails for any reason, log an error and return `nil` so that further processing of the function invocation is not impacted

These issues are thought to only be reproducible outside of AWS, but it's good to be proactive with a bit of extra caution.

### DIFF
--- a/lib/new_relic/agent/serverless_handler.rb
+++ b/lib/new_relic/agent/serverless_handler.rb
@@ -186,13 +186,18 @@ module NewRelic
       def http_uri(info)
         return unless info[:host] && info[:path]
 
-        url_str = "https://#{info[:host]}:#{info[:port]}#{info[:path]}"
+        url_str = "https://#{info[:host]}"
+        url_str += ":#{info[:port]}" unless info[:host].match?(':')
+        url_str += "#{info[:path]}"
+
         if info[:query_parameters]
           qp = info[:query_parameters].map { |k, v| "#{k}=#{v}" }.join('&')
           url_str += "?#{qp}"
         end
 
         URI.parse(url_str)
+      rescue StandardError => e
+        NewRelic::Agent.logger.error "ServerlessHandler failed to parse the source HTTP URI: #{e}"
       end
 
       def info_for_api_gateway_v2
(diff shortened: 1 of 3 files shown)


test/new_relic/agent/serverless_handler_test.rb
```diff
@@ -423,6 +423,47 @@ def test_metadata_for_payload_v2
         refute_empty metadata[:agent_language]
       end
 
+      # when testing locally with (the Node.js tool) serverless or whatnot, the
```
kaylareopelle, line 426: Is one of the Node tools called "whatnot"? Or are you referring to "whatnot" more generally? (my guess is the latter, but I could see myself combing the internet for a tool called "whatnot")

lib/new_relic/agent/serverless_handler.rb
```diff
@@ -186,13 +186,18 @@ def process_api_gateway_info
...
+        url_str += "#{info[:path]}"
+
         if info[:query_parameters]
           qp = info[:query_parameters].map { |k, v| "#{k}=#{v}" }.join('&')
           url_str += "?#{qp}"
         end
 
         URI.parse(url_str)
+      rescue StandardError => e
+        NewRelic::Agent.logger.error "ServerlessHandler failed to parse the source HTTP URI: #{e}"
```
kaylareopelle, line 200: Do we need to be worried about additional attempts? Or do you think this will only run once per invocation?

test/new_relic/agent/serverless_handler_test.rb
fallwith: Disambiguated with [c869cd9](https://github.com/newrelic/newrelic-ruby-agent/pull/2821/commits/c869cd920660ec65c5b77ab641c34f7ff7cbefc2)

lib/new_relic/agent/serverless_handler.rb
fallwith: The URI is built only once per invocation.

kaylareopelle (approved):

PR #2822, opened 2024-08-20
bring in the latest cross-agent AWS Lambda JSON

fallwith:
test against the cross-agent JSON that defines an AWS API Gateway V2 event payload.


kaylareopelle (approved):
Tak!

PR #2823, opened 2024-08-20
Bundler version conditions

hannahramadan:
Bundler 2+ deprecated `all_specs` in favor of `installed_specs` and emits warnings when `all_specs` is used. We still support Bundler 1, so need to keep `all_specs` around while using `installed_specs` when appropriate. 

Checking for the version proved faster in benchmarking than checking to see if `installed_specs` is define.

### DIFF
--- a/lib/new_relic/agent/instrumentation/grape.rb
+++ b/lib/new_relic/agent/instrumentation/grape.rb
@@ -19,9 +19,14 @@ DependencyDetection.defer do
 
   depends_on do
     begin
-      if defined?(Bundler) && Bundler.rubygems.all_specs.map(&:name).include?('newrelic-grape')
-        NewRelic::Agent.logger.info('Not installing New Relic supported Grape instrumentation because the third party newrelic-grape gem is present')
-        false
+      if defined?(Bundler) &&
+          if Bundler::VERSION >= '2' && Bundler.rubygems.installed_specs.map(&:name).include?('newrelic-grape')
+            NewRelic::Agent.logger.info('Not installing New Relic supported Grape instrumentation because the third party newrelic-grape gem is present')
+            false
+          elsif Bundler.rubygems.all_specs.map(&:name).include?('newrelic-grape')
+            NewRelic::Agent.logger.info('Not installing New Relic supported Grape instrumentation because the third party newrelic-grape gem is present')
+            false
+          end
       else
         true
       end
--- a/lib/new_relic/environment_report.rb
+++ b/lib/new_relic/environment_report.rb
@@ -44,7 +44,11 @@ module NewRelic
     ####################################
     report_on('Gems') do
       begin
-        Bundler.rubygems.all_specs.map { |gem| "#{gem.name}(#{gem.version})" }
+        if Bundler::VERSION >= '2'
+          Bundler.rubygems.installed_specs.map { |gem| "#{gem.name}(#{gem.version})" }
+        else
+          Bundler.rubygems.all_specs.map { |gem| "#{gem.name}(#{gem.version})" }
+        end
       rescue
         # There are certain rubygem, bundler, rails combinations (e.g. gem
         # 1.6.2, rails 2.3, bundler 1.2.3) where the code above throws an error
--- a/lib/new_relic/language_support.rb
+++ b/lib/new_relic/language_support.rb
@@ -88,7 +88,12 @@ module NewRelic
     end
 
     def bundled_gem?(gem_name)
-      defined?(Bundler) && Bundler.rubygems.all_specs.map(&:name).include?(gem_name)
+      defined?(Bundler) &&
+        if Bundler::VERSION >= '2'
+          Bundler.rubygems.installed_specs.map(&:name).include?(gem_name)
+        else
+          Bundler.rubygems.all_specs.map(&:name).include?(gem_name)
+        end
     rescue => e
       ::NewRelic::Agent.logger.info("Could not determine if third party #{gem_name} gem is installed", e)
       false
--- a/lib/new_relic/control/frameworks/rails4.rb
+++ b/lib/new_relic/control/frameworks/rails4.rb
@@ -9,8 +9,10 @@ module NewRelic
     module Frameworks
       class Rails4 < NewRelic::Control::Frameworks::Rails3
         def rails_gem_list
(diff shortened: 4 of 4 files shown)


lib/new_relic/agent/instrumentation/grape.rb
```diff
@@ -19,9 +19,14 @@
...
-        NewRelic::Agent.logger.info('Not installing New Relic supported Grape instrumentation because the third party newrelic-grape gem is present')
-        false
+      if defined?(Bundler) &&
+          if Bundler::VERSION >= '2' && Bundler.rubygems.installed_specs.map(&:name).include?('newrelic-grape')
+            NewRelic::Agent.logger.info('Not installing New Relic supported Grape instrumentation because the third party newrelic-grape gem is present')
+            false
+          elsif Bundler.rubygems.all_specs.map(&:name).include?('newrelic-grape')
+            NewRelic::Agent.logger.info('Not installing New Relic supported Grape instrumentation because the third party newrelic-grape gem is present')
+            false
+          end
```
fallwith, line 29: Two problems here:

1. You don't have an `else` for Bundler 2 being present and `newrelic-grape` being absent. The `else` should yield `true`.
2. Both the `if` and `elsif` do the exact same 2 things (logger statement and yield `false`), so you can just `||` them together and DRY up the code.

Here's a pass at addressing both issues:

```suggestion
          grape_present = if (Bundler::VERSION >= '2' && Bundler.rubygems.installed_specs.map(&:name).include?('newrelic-grape')) ||
                            Bundler.rubygems.all_specs.map(&:name).include?('newrelic-grape')
                            true
                          else
                            false
                          end
          if grape_present
            NewRelic::Agent.logger.info('Not installing New Relic supported Grape instrumentation because the third party newrelic-grape gem is present')
            false
          else
            true
          end
```

lib/new_relic/environment_report.rb
```diff
@@ -44,7 +44,11 @@ def self.registered_reporters=(logic)
     ####################################
     report_on('Gems') do
       begin
-        Bundler.rubygems.all_specs.map { |gem| "#{gem.name}(#{gem.version})" }
+        if Bundler::VERSION >= '2'
+          Bundler.rubygems.installed_specs.map { |gem| "#{gem.name}(#{gem.version})" }
+        else
+          Bundler.rubygems.all_specs.map { |gem| "#{gem.name}(#{gem.version})" }
+        end
```
fallwith, line 51: You could DRY up these ones with metaprogramming, but I completely respect the slight repetition if avoiding metaprogramming is a deliberate goal.

```suggestion
          specs_method = Bundler::VERSION >= '2' ? :installed_specs : :all_specs
          Bundler.rubygems.send(specs_method).map { |gem| "#{gem.name} (#{gem.version})" }
```

hannahramadan: I love this!

lib/new_relic/agent/instrumentation/grape.rb
hannahramadan: Much much cleaner. I don't think I was looking at the code holistically. Great edit.

lib/new_relic/agent/instrumentation/grape.rb
```diff
@@ -19,9 +19,16 @@
...
+      if defined?(Bundler)
+        grape_present = (Bundler::VERSION >= '2' && Bundler.rubygems.installed_specs.map(&:name).include?('newrelic-grape')) ||
+          Bundler.rubygems.all_specs.map(&:name).include?('newrelic-grape')
+
+        if grape_present
+          NewRelic::Agent.logger.info('Not installing New Relic supported Grape instrumentation because the third party newrelic-grape gem is present')
+          false
+        else
+          true
+        end
```
fallwith, line 31:
```suggestion
      if (defined?(Bundler) && (Bundler::VERSION >= '2' && Bundler.rubygems.installed_specs.map(&:name).include?('newrelic-grape')) ||
        Bundler.rubygems.all_specs.map(&:name).include?('newrelic-grape'))
        NewRelic::Agent.logger.info('Not installing New Relic supported Grape instrumentation because the third party newrelic-grape gem is present')
        false
```

fallwith (approved):

lib/new_relic/agent/instrumentation/grape.rb
```diff
@@ -19,7 +19,8 @@
 
   depends_on do
     begin
-      if defined?(Bundler) && Bundler.rubygems.all_specs.map(&:name).include?('newrelic-grape')
+      if defined?(Bundler) && ((Bundler::VERSION >= '2' && Bundler.rubygems.installed_specs.map(&:name).include?('newrelic-grape')) ||
```
kaylareopelle, line 22: What do you think about using `Gem::Version.new` to compare the `Bundler::VERSION` constant with the `2`?

lib/new_relic/language_support.rb
```diff
@@ -88,7 +88,12 @@ def snakeize(string)
     end
 
     def bundled_gem?(gem_name)
-      defined?(Bundler) && Bundler.rubygems.all_specs.map(&:name).include?(gem_name)
+      defined?(Bundler) &&
```
kaylareopelle, line 91: This condition feels a little complex with `defined?(Bundler) &&` at the top. What if you added a guard clause to check for Bundler instead?

Maybe something like: 
```
return false unless defined?(Bundler)

if ...... (rest of the code as-is)
``

kaylareopelle (approved):
Great, as-is! A few small suggestions.

lib/new_relic/agent/instrumentation/grape.rb
hannahramadan: Made this change! https://github.com/newrelic/newrelic-ruby-agent/pull/2823/commits/b33eac4858938c8122354c9d769e6ab2c451cc61

lib/new_relic/language_support.rb
hannahramadan: Less complex is where its at. https://github.com/newrelic/newrelic-ruby-agent/pull/2823/commits/b33eac4858938c8122354c9d769e6ab2c451cc61

kaylareopelle (approved):
thank you!

PR #2841, opened 2024-09-06
Kafka instrumentation: code loading logic fix

fallwith:
Make sure that the `Rdkafka` class based instrumentation is not attempted on systems without it present by relocating the `require_relative` statements into the conditionally executed block.


kaylareopelle (approved):
good find!

tannalynn (approved):

PR #2847, opened 2024-09-11
Lockdown Boolean configs

### LINKED ISSUE #2838: Correctly handle `newrelic.yml` boolean parsing
The agent's config file currently accepts strings where boolean values are expected. This can cause an issue where, for example, setting a boolean-expected config to `'false'` (or any other string) evaluates to `true`. 

The agent's config parsing logic needs to be updated to allow certain strings when a boolean is expected. Some logic currently exists in the agent for the configs `agent_enabled` and `enabled` that gets near the desired behavior, which we may be able to build off:

https://github.com/newrelic/newrelic-ruby-agent/blob/41430b2aac3cdb69c178cd742b4074618e00227c/lib/new_relic/agent/configuration/yaml_source.rb#L135-L148

Following the above logic, the boolean `true` may accept the strings `'true'`, `'yes'`, `'on'` while the boolean `false` may accept `'false'`, `'no'`, `'off'`. We should log an error and use the default config value when unexpected values are given.

hannahramadan:
Enforces configs of type Boolean to contain either a boolean or a string/symbol of `true`, `false`, `yes`, `no`, `on`, or `off`

closes #2838

### DIFF
--- a/lib/new_relic/agent/javascript_instrumentor.rb
+++ b/lib/new_relic/agent/javascript_instrumentor.rb
@@ -164,9 +164,8 @@ module NewRelic
 
       def add_ssl_for_http(data)
         ssl_for_http = NewRelic::Agent.config[:'browser_monitoring.ssl_for_http']
-        unless ssl_for_http.nil?
-          data[SSL_FOR_HTTP_KEY] = ssl_for_http
-        end
+
+        data[SSL_FOR_HTTP_KEY] = ssl_for_http if ssl_for_http
       end
 
       def add_attributes(data, txn)
--- a/test/new_relic/agent/configuration/default_source_test.rb
+++ b/test/new_relic/agent/configuration/default_source_test.rb
@@ -140,12 +140,6 @@ module NewRelic::Agent::Configuration
       end
     end
 
-    def test_application_logging_enabled_default
-      with_config(:'application_logging.enabled' => :foo) do
-        assert_equal :foo, NewRelic::Agent.config['application_logging.enabled']
-      end
-    end
-
     def test_agent_attribute_settings_convert_comma_delimited_strings_into_an_arrays
       types = %w[transaction_tracer. transaction_events. error_collector. browser_monitoring.]
       types << ''
--- a/test/new_relic/agent/configuration/default_source_test.rb
+++ b/test/new_relic/agent/configuration/default_source_test.rb
@@ -285,6 +285,81 @@ module NewRelic::Agent::Configuration
       end
     end
 
+    def test_boolean_configs_accepts_yes_on_and_true_as_strings
+      key = :'send_data_on_exit'
+      default = ::NewRelic::Agent::Configuration::DefaultSource.default_for(key)
+      config_array = %w[yes on true]
+
+      config_array.each do |value|
+        with_config(key => value) do
+          assert NewRelic::Agent.config[key]
+        end
+      end
+    end
+
+    def test_boolean_configs_accepts_yes_on_and_true_as_symbols
+      key = :'send_data_on_exit'
+      default = ::NewRelic::Agent::Configuration::DefaultSource.default_for(key)
+      config_array = %i[yes on true]
+
+      config_array.each do |value|
+        with_config(key => value) do
+          assert NewRelic::Agent.config[key]
+        end
+      end
+    end
+
+    def test_boolean_configs_accepts_no_off_and_false_as_strings
+      key = :'send_data_on_exit'
+      default = ::NewRelic::Agent::Configuration::DefaultSource.default_for(key)
+      config_array = %w[no off false]
+
+      config_array.each do |value|
+        with_config(key => value) do
+          refute NewRelic::Agent.config[key]
+        end
+      end
+    end
+
+    def test_boolean_configs_accepts_no_off_and_false_as_strings_as_symbols
+      key = :'send_data_on_exit'
+      default = ::NewRelic::Agent::Configuration::DefaultSource.default_for(key)
+      config_array = %i[no off false]
+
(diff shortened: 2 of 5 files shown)


test/new_relic/agent/configuration/default_source_test.rb
```diff
@@ -140,12 +140,6 @@ def test_config_search_path_in_warbler
       end
     end
 
-    def test_application_logging_enabled_default
```
hannahramadan, line 143: This is a dead test/ not actually testing with the test intends.

lib/new_relic/agent/javascript_instrumentor.rb
```diff
@@ -164,9 +164,8 @@ def data_for_js_agent(transaction)
 
       def add_ssl_for_http(data)
         ssl_for_http = NewRelic::Agent.config[:'browser_monitoring.ssl_for_http']
-        unless ssl_for_http.nil?
-          data[SSL_FOR_HTTP_KEY] = ssl_for_http
-        end
+
+        data[SSL_FOR_HTTP_KEY] = ssl_for_http if ssl_for_http
```
hannahramadan, line 168: Just a little refactor while in the code ◡̈

test/new_relic/agent/configuration/default_source_test.rb
hannahramadan, line 143: This test conflicts with intention of allowing certain values for Booleans, and defaulting to the default value when an unrecognized value is given. This test is saying if `:foo` is set as the value for `application_logging.enabled`, we should use it. Ripping it out!

test/new_relic/agent/configuration/default_source_test.rb
```diff
@@ -291,6 +285,81 @@ def test_allowlist_blocks_invalid_values_and_uses_a_default
...
     end
 
+    def test_boolean_configs_accepts_yes_on_and_true_as_strings
+      key = :'send_data_on_exit'
+      default = ::NewRelic::Agent::Configuration::DefaultSource.default_for(key)
+      config_array = %w[yes on true]
+
+      config_array.each do |value|
+        with_config(key => value) do
+          assert NewRelic::Agent.config[key]
```
fallwith, line 295: This (nice, DRY) approach to looping over test inputs stored in an array can potentially lead to confusion over which item(s) in the array are actually causing the problem if there are ever any errors in the future.

Sometimes it's worth going against DRY in the tests for this reason. Sometimes it's worth keeping the testing code clean and just adding some extra context to help the tester. I'm thinking that would be a good approach to take here:

```suggestion
          assert NewRelic::Agent.config[key], "The '#{value}' value failed to evaluate as truthy!"
```

test/new_relic/agent/configuration/default_source_test.rb
```diff
@@ -291,6 +285,81 @@ def test_allowlist_blocks_invalid_values_and_uses_a_default
...
+        end
+      end
+    end
+
+    def test_boolean_configs_accepts_yes_on_and_true_as_symbols
+      key = :'send_data_on_exit'
+      default = ::NewRelic::Agent::Configuration::DefaultSource.default_for(key)
+      config_array = %i[yes on true]
+
+      config_array.each do |value|
```
fallwith, line 305: For all of these tests, given that you only ever iterate through the array and don't interact with the array in any other manner, you could optionally forego bothering to store it in a variable:

```suggestion
      %i[yes on true].each do |value|
```

test/new_relic/agent/configuration/default_source_test.rb
```diff
@@ -291,6 +285,81 @@ def test_allowlist_blocks_invalid_values_and_uses_a_default
       end
     end
 
+    def test_boolean_configs_accepts_yes_on_and_true_as_strings
+      key = :'send_data_on_exit'
+      default = ::NewRelic::Agent::Configuration::DefaultSource.default_for(key)
```
fallwith, line 290: I don't see the `default` var being used.

test/new_relic/agent/configuration/default_source_test.rb
```diff
@@ -291,6 +285,81 @@ def test_allowlist_blocks_invalid_values_and_uses_a_default
...
+
+      with_config(key => 'invalid_value') do
+        assert NewRelic::Agent.config[key]
+      end
+    end
+
+    def test_enforce_boolean_logs_warning_on_invalid_value
+      key = :'send_data_on_exit'
+      default = ::NewRelic::Agent::Configuration::DefaultSource.default_for(key)
```
fallwith, line 347: It looks like this is the only test that needs the `default` var.

fallwith (approved):

test/new_relic/agent/configuration/default_source_test.rb
hannahramadan: Nice- great feedback! Made this change https://github.com/newrelic/newrelic-ruby-agent/pull/2847/commits/be01eafceca9862380560b212346bac17f6db246

test/new_relic/agent/configuration/default_source_test.rb
hannahramadan: Love this- accepted! https://github.com/newrelic/newrelic-ruby-agent/pull/2847/commits/be01eafceca9862380560b212346bac17f6db246

test/new_relic/agent/configuration/default_source_test.rb
hannahramadan: Yup yup - copy pasta error. Removed! https://github.com/newrelic/newrelic-ruby-agent/pull/2847/commits/520e9b5c421552188cdfcf4a02af0fd315076196

fallwith (approved):

PR #2851, opened 2024-09-14
Apply method tracers automatically via config

### LINKED ISSUE #2701: Add a configuration option to permit custom method tracers to be defined automatically
Coming out of a k8s operator discussion involving automated injection logic that would allow a customer to leverage New Relic APM functionality without altering their existing code base...

As per the [docs](https://docs.newrelic.com/docs/apm/agents/ruby-agent/api-guides/ruby-custom-instrumentation/), currently a customer wanting to define a custom method tracer must do so by adding a line of code for every method to be traced:

```ruby
add_method_tracer :my_method_to_be_traced
```

It would be ideal if we could leverage a configuration parameter to permit method tracers to by added automatically when the agent starts.

Users could enjoy all of the custom functionality without any alterations being made to their .rb files.

For example, given this Ruby code:

```ruby
module MyCompany
  class Image
    def render_png
      # code to render a PNG
    end
  end

  class User
    def self.notify
      # code to notify users
    end
  end
end
```

We could support a configuration parameter called `:custom_tracer_method_list` like so in `newrelic.yml`:

```yaml
customer_tracer_method_list: MyCompany::Image#render_png,MyCompany::User.notify
```

Method tracers would then automatically be added for the 2 methods on the list.

fallwith:
A new `:automatic_custom_instrumentation_method_list` configuration parameter has been added to permit the user to define a list of namespaced method name strings representing methods that the agent should automatically add custom instrumentation tracers to.

resolves #2701

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,58 @@
 ## dev
 
 
-Version <dev> adds Apache Kafka instrumentation for the rdkafka and ruby-kafka gems and fixes a JRuby bug in the configuration manager.
+Version <dev> adds Apache Kafka instrumentation for the rdkafka and ruby-kafka gems, introduces a configuration based automatic way to add custom instrumentation method tracers, and fixes a JRuby bug in the configuration manager.
 
 - **Feature: Add Apache Kafka instrumentation for the rdkafka and ruby-kafka gems**
 
   The agent now has instrumentation for both the rdkafka and ruby-kafka gems. The agent will record transactions and message broker segments for produce and consume calls made using these gems. [PR#2824](https://github.com/newrelic/newrelic-ruby-agent/pull/2824) [PR#2842](https://github.com/newrelic/newrelic-ruby-agent/pull/2842)
+
+- **Feature: Add a configuration option to permit custom method tracers to be defined automatically**
+
+  A new `:automatic_custom_instrumentation_method_list` configuration parameter has been added to permit the user to define a list of fully qualified (namespaced) Ruby methods for the agent to automatically add custom instrumentation for without requiring any code modifications to be made to the classes that define the methods.
+
+  The list should be an array of `CLASS#METHOD` (for instance methods) and/or `CLASS.METHOD` (for class methods) strings.
+
+  Use fully qualified class names (using the `::` delimiter) that include any module or class namespacing.
+
+  Here is some Ruby source code that defines a `render_png` instance method for an `Image` class and and a `notify` class method for a `User` class, both within a `MyCompany` module namespace:
+
+  ```
+  module MyCompany
+      class Image
+      def render_png
+          # code to render a PNG
+      end
+      end
+      class User
+      def self.notify
+          # code to notify users
+      end
+      end
+  end
+  ```
+
+  Given that source code, the `newrelic.yml` config file might request instrumentation for both of these methods like so:
+
+  ```
+  automatic_custom_instrumentation_method_list:
+    - MyCompany::Image#render_png
+    - MyCompany::User.notify
+  ```
+
+  That configuration example uses YAML array syntax to specify both methods. Alternatively, a comma-delimited string can be used instead:
+
+  ```
+  automatic_custom_instrumentation_method_list: 'MyCompany::Image#render_png, MyCompany::User.notify'
+  ```
+
+  Whitespace around the comma(s) in the list is optional. When configuring the agent with a list of methods via the `NEW_RELIC_AUTOMATIC_CUSTOM_INSTRUMENTATION_METHOD_LIST` environment variable, this comma-delimited string format should be used:
+
+  ```
+  export NEW_RELIC_AUTOMATIC_CUSTOM_INSTRUMENTATION_METHOD_LIST='MyCompany::Image#render_png, MyCompany::User.notify'
+  ```
+
+[PR#2851](https://github.com/newrelic/newrelic-ruby-agent/pull/2851)
 
 - **Bugfix: Jruby not saving configuration values correctly in configuration manager**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## dev
 
 
-Version <dev> adds Apache Kafka instrumentation for the rdkafka and ruby-kafka gems, introduces a configuration based automatic way to add custom instrumentation method tracers, and fixes a JRuby bug in the configuration manager.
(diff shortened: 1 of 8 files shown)


lib/new_relic/agent.rb
```diff
@@ -163,6 +170,86 @@
       end
     end
 
+    def self.add_automatic_method_tracers(arr)
+      return unless arr
+      return arr if arr.respond_to?(:empty?) && arr.empty?
+
+      arr = arr.split(/\s*,\s*/) if arr.is_a?(String)
```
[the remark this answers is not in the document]
fallwith: This `String#split` operation is performed exactly once at agent start-up, and only if a user has opted in to this new functionality which won't engage the regex by default. The pattern in question is also already used for every single comma delimited string for our configuration parameters (permitting the use of env vars to set values), so if we ever want to discuss this concern we should do so more broadly than what is included by the scope of this PR.

CHANGELOG.md
```diff
@@ -3,12 +3,59 @@
 ## dev
 
 
-Version <dev> adds Apache Kafka instrumentation for the rdkafka and ruby-kafka gems and fixes a JRuby bug in the configuration manager.
+Version <dev> adds Apache Kafka instrumentation for the rdkafka and ruby-kafka gems, introduces a configuration based automatic way to add custom instrumentation method tracers, and fixes a JRuby bug in the configuration manager.
```
kaylareopelle, line 6:
```suggestion
Version <dev> adds Apache Kafka instrumentation for the rdkafka and ruby-kafka gems, introduces a configuration-based, automatic way to add custom instrumentation method tracers, and fixes a JRuby bug in the configuration manager.
```

CHANGELOG.md
```diff
@@ -3,12 +3,59 @@
...
 
+- **Feature: Add a configuration option to permit custom method tracers to be defined automatically**
+
+  A new `:automatic_custom_instrumentation_method_list` configuration parameter has been added to permit the user to define a list of fully qualified (namespaced) Ruby methods for the agent to automatically add custom instrumentation for without requiring any code modifications to be made to the classes that define the methods.
+
+  The list should be an array of `CLASS#METHOD` (for instance methods) and/or `CLASS.METHOD` (for class methods) strings.
+
+  Use fully qualified class names (using the `::` delimiter) that include any module or class namespacing.
+
+  Here is some Ruby source code that defines a `render_png` instance method for an `Image` class and and a `notify` class method for a `User` class, both within a `MyCompany` module namespace:
```
kaylareopelle, line 20:
```suggestion
  Here is some Ruby source code that defines a `render_png` instance method for an `Image` class and a `notify` class method for a `User` class, both within a `MyCompany` module namespace:
```

CHANGELOG.md
```diff
@@ -3,12 +3,59 @@
...
+      def render_png
+          # code to render a PNG
+      end
+      end
+      class User
+      def self.notify
+          # code to notify users
+      end
+      end
+  end
```
kaylareopelle, line 34: The indentations in this block feel a little weird. Could you indent the methods within the classes?

lib/new_relic/agent.rb
```diff
@@ -163,6 +170,86 @@ def add_or_defer_method_tracer(receiver, method_name, metric_name, options)
...
+
+    # spawn a thread that will attempt to establish a tracer for each of the
+    # configured methods. the thread will continue to keep trying with each
+    # tracer until one of the following happens:
+    #   - the tracer is successfully established
+    #   - the configured method string couldn't be parsed
+    #   - establishing a tracer for a successfully parsed string failed
+    #   - the maximum number of attempts has been reached
+    # the thread will only be spawned once per agent initialization, to account
+    # for configuration reloading scenarios.
```
kaylareopelle, line 192: I think these comments will show up on Rdoc. What do you think about adding an `# @api private` line to the end?

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1137,6 +1137,56 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
           :description => 'If `false`, custom attributes will not be sent on events.'
         },
+        :automatic_custom_instrumentation_method_list => {
+          :default => NewRelic::EMPTY_ARRAY,
+          :public => true,
+          :type => Array,
+          :allowed_from_server => false,
+          :transform => proc { |arr| NewRelic::Agent.add_automatic_method_tracers(arr) },
+          :description => <<~DESCRIPTION
+            An array of `CLASS#METHOD` (for instance methods) and/or `CLASS.METHOD` (for class methods) strings representing Ruby methods for the agent to add custom instrumentation to automatically without the need for altering any of the source code that defines the methods.
```
kaylareopelle, line 1147:
```suggestion
            An array of `CLASS#METHOD` (for instance methods) and/or `CLASS.METHOD` (for class methods) strings representing Ruby methods for the agent to automatically add custom instrumentation to without the need for altering any of the source code that defines the methods.
```

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1137,6 +1137,56 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+
+            Use fully qualified class names (using the `::` delimiter) that include any module or class namespacing.
+
+            Here is some Ruby source code that defines a `render_png` instance method for an `Image` class and and a `notify` class method for a `User` class, both within a `MyCompany` module namespace:
```
kaylareopelle, line 1151:
```suggestion
            Here is some Ruby source code that defines a `render_png` instance method for an `Image` class and a `notify` class method for a `User` class, both within a `MyCompany` module namespace:
```

CHANGELOG.md
fallwith: Ah, whoops. It looks like the formatting didn't survive the copy and paste from `default_source.rb`. Fixed with [2702347](https://github.com/newrelic/newrelic-ruby-agent/pull/2851/commits/27023470c73c9082c38c574e9963bb5152f8000d)

lib/new_relic/agent.rb
fallwith: Great call here. Added that for all new methods which are truly intended to be private here. Fixed in [87589d7](https://github.com/newrelic/newrelic-ruby-agent/pull/2851/commits/87589d7fe08f8c00c2e1a357b7d3e8a843d1caf6)

kaylareopelle (approved):

tannalynn (approved):

hannahramadan (approved):

Issue #2869, opened 2024-09-28
ActionView::Template::Error: undefined method `identifier'


navidemad:
## Description
Running view_component with recent change https://github.com/ViewComponent/view_component/commit/451543a53f3c7c04036eb6109ac716511f3dcfc3#diff-b69c75ffd85596e0b5fca1327059ecf3b2930f3f77004c78a268fa85e8ec8d9eL601

Makes the following error in our production server:
`ActionView::Template::Error: undefined method `identifier' for class LinkComponent`

## Expected Behavior
Your method metric_path manage already if the identifier is not set to "component" but the problem is coming from in render_in_with_tracing that is doing self.class.identifier

## Your Environment
ruby 3.3.4 (2024-07-09 revision be1089c8ec) +YJIT [arm64-darwin23]
gem view_component (3.15.0)
gem newrelic_rpm (9.13.0)

## For Maintainers Only or Hero Triaging this bug
*Suggested Priority (P1):* We had to pin to pin view_component to 3.14.0 otherwise we were getting a lot of errors

joelhawksley:
👋🏻 this method should never have been used, as it was marked `@private` and not documented. I've re-introduced it in https://github.com/ViewComponent/view_component/releases/tag/v3.15.1 but will remove it in a coming minor release.

fallwith:
Closing, as #2870 has been merged. For anyone coming across this issue in the future:

- The fair point about private API usage is addressed in https://github.com/newrelic/newrelic-ruby-agent/pull/2870#issuecomment-2383943722
- PR #2870 which accompanied this issue has been merged and will be released in the next agent version
- Follow-up work related to https://github.com/newrelic/newrelic-ruby-agent/pull/2870#issuecomment-2383943722 has been captured with #2872 and #2874

ZilvinasKucinskas:
This issue brought the whole production down (every single page). `newrelic_rpm` is disabled by default in development and test environments, so it was impossible to catch. 

How can we enable `newrelic_rpm` in a test environment but not send data to it? It would be helpful to catch such issues in a test environment.

fallwith:
Hi @ZilvinasKucinskas (cc @hasghari). We are sorry for the poor experience that this issue caused for you.

We endeavor to make the agent's instrumentation logic bullet-proof so that even when it crashes it doesn't impact the original Ruby library's functionality, but this error unfortunately did impact the core functionality.

[...]

PR #2870, opened 2024-09-28
Safety guard when the self.class is not responding to method identifier

navidemad:
# Overview

Issue: https://github.com/newrelic/newrelic-ruby-agent/issues/2869

Due to a recent change in the gem view_component 
https://github.com/ViewComponent/view_component/commit/451543a53f3c7c04036eb6109ac716511f3dcfc3#diff-b69c75ffd85596e0b5fca1327059ecf3b2930f3f77004c78a268fa85e8ec8d9eL601

We were getting:
`ActionView::Template::Error: undefined method `identifier'`

The method identifier on their base class has been removed.
This pull request make sure it returns nil if it is not found.

### DIFF
--- a/lib/new_relic/agent/instrumentation/view_component/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/view_component/instrumentation.rb
@@ -11,7 +11,10 @@ module NewRelic::Agent::Instrumentation
 
       begin
         segment = NewRelic::Agent::Tracer.start_segment(
-          name: metric_name(self.class.identifier, self.class.name)
+          name: metric_name(
+            self.class.respond_to?(:identifier) ? self.class.identifier : nil,
+            self.class.name
+          )
         )
         yield
       rescue => e


joelhawksley:
👋🏻 ViewComponent maintainer here! FYI, this method should never have been used as it was marked private (and not documented in the API docs) https://github.com/ViewComponent/view_component/compare/v3.14.0...v3.15.0#diff-b69c75ffd85596e0b5fca1327059ecf3b2930f3f77004c78a268fa85e8ec8d9eL601.

fallwith:
Hi @navidemad, thanks very much for creating the GitHub Issue and submitting this fix. Very much appreciated. Your change looks good and will go out in our next agent release.

Hi @joelhawksley. Point well taken. Our preferred approach for gathering metrics and other information from a library is to leverage public APIs. We also often submit PRs against libraries and/or work with maintainers to deliver new or enhanced APIs whenever we need something that doesn't already exist. We don't always take the preferred approach and fall back on using private APIs or monkeypatching though. This can be for any one of several reasons, including shifting schedules and priorities that see us do something to get some coverage in place - especially when in response to a customer feature request - as quickly as possible. In the particular case of ViewComponent, I think we should revisit our instrumentation at some point and take another look at what is possible by consuming the ActiveSupport notifications the library produces. ViewComponent's own `lib/view_component/instrumentation.rb`, for example, uses `self.class.source_location` for the identifier. We ought to see if we can simply consume those notifications or if not at least make the same `self.class.source_location` call for the identifier. I will create a GitHub issue to remind us to look into doing so. For now, this PR is an improvement upon what we already have so I'll go ahead and merge it.

fallwith (approved):

joelhawksley:
@fallwith makes sense to me. I'd be open to formally exposing the `source_location` as a public API if you'd find it useful. Let me know!

PR #2878, opened 2024-10-01
type coercion for configuration parameters

### LINKED ISSUE #2852: Type coercion behavior inconsistencies and flaws between configuration formats and between APM and CSEC
Existing type coercion behavior is implemented pretty well in the environment variable handling logic, given that all environment variable values are strings on input. But sourcing parameters from elsewhere - especially the YAML configuration file - will bypass this type coercion, leading to the following outstanding issues:

- Behavior from environment variable input (example: `'123'` becomes `123`) can reasonably be expected from other configuration sources, leading to unexpected behavior (or more specifically a lack of coercion behavior in this case).
- The CSEC agent - which needs to deal with numeric parameter values more than the APM agent does - now does interger coercion that the APM agent do. Users may reasonably expect consistency here.
- The lack of integer coercion is particularly problematic because in one way the agent tries to do the right thing by not succeeding with a string value at all (rather than converting a string unexpectedly to `0` or whatnot), but unfortunately has the potential to permit a fatal exception to lay hidden by not performing any type or bounds check on some integers until the time that they are actively used. This lack of config parse time checking in addition to the lack of type coercion can be particularly problematic (buggy).
- Existing non-coercion behavior for the YAML input seems to be pretty well rooted in New Relic's history of only offering a Ruby agent back when the company first started. In those days, knowledge of Ruby and YAML were pretty safely assumed. Nowadays with New Relic offering multiple agents that take advantage of multiple configuration file formats, knowledge of YAML-to-Ruby behavior should not assumed when reasonable coercion can be performed - especially when this coercion already takes place for environment variables. This will be especially important to customers that leverage multiple agents to observe applications written in multiple languages.

While bullet point 4 might make for a good "feature request" candidate, the first 3 bullet points pertaining to inconsistency and potential unexpected behavior swayed us into favoring the "bug" category.

**Definition of Done**
- For all configuration parameters (defined in `default_source.rb`) that make use of a  `:type` attribute, ensure that user supplied values are coerced to that type in a way that is consistent between YAML and environment variables and between the Ruby APM and CSEC agents.
- To address bullet point 3, any problematic invalid values should be identified at agent start time and not at the time the agent goes to use the values that were previously blindly trusted.

fallwith:
Rework the processing of configuration parameters to address #2852.

- nils to produce errors when user-supplied for params that don't support them
- introduce a common shared type coercion layer for all config values so that things like environment variable and YAML values are both consistently coerced
- allow type coercion to Array/Hash to replace all of the equivalent default source  based`:transform` entries
- lay the foundation for increased server side and env var based support for certain params that are currently marked as YAML source only (functionally ready now, updating tests and docs for each affected param to be handled later)
- permit Integer values for Float params (using `#to_f`) and Float values for Integer params (using `#round`)
- prevent fatal interruptions and log warnings when a default value can be found to replace a given bogus input value
- leverage default source's `:allow_nil` setting to enforce non-nil values for params that don't support nil

resolves #2852

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -307,7 +307,7 @@ module NewRelic
           :public => true,
           :type => String,
           :allowed_from_server => false,
-          :transform => proc { |str| str.split(';') },
+          :transform => proc { |v| v.is_a?(String) ? v.split(';') : v },
           :description => 'Specify the [application name](/docs/apm/new-relic-apm/installation-configuration/name-your-application) used to aggregate data in the New Relic UI. To report data to [multiple apps at the same time](/docs/apm/new-relic-apm/installation-configuration/using-multiple-names-app), specify a list of names separated by a semicolon `;`. For example, `MyApp` or `MyStagingApp;Instance1`.'
         },
         :license_key => {
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -35,15 +35,6 @@ module NewRelic
       end
 
       class DefaultSource
-        BOOLEAN_MAP = {
-          'true' => true,
-          'yes' => true,
-          'on' => true,
-          'false' => false,
-          'no' => false,
-          'off' => false
-        }.freeze
-
         attr_reader :defaults
 
         extend Forwardable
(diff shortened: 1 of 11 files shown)


tannalynn (approved):
I think things look good, thanks for going over all of this with us. Could you add a changelog entry for this?

Issue #2879, opened 2024-10-01
9.14 - ArgumentError with rdkafka integration


krisdigital:
## Description
After the update from 9.13 to 9.14 there is an error

```
ArgumentError • vendor/bundle/ruby/3.3.0/gems/newrelic_rpm-9.14.0/lib/new_relic/agent/instrumentation/rdkafka/prepend.rb:58
ArgumentError: wrong number of arguments (given 1, expected 0)
```

## Expected Behavior
No bugs...

## Steps to Reproduce
Just install...

## Your Environment
Using 'karafka', '2.4.7' with ruby 3.3.5 and Rails 7.1.3.4

## Additional context
-

## For Maintainers Only or Hero Triaging this bug
*Suggested Priority (P1,P2,P3,P4,P5):*
*Suggested T-Shirt size (S, M, L, XL, Unknown):*

tannalynn:
Hello @krisdigital 
Thank you for bringing this to our attention! We added instrumentation for the `rdkafka` gem. It looks like this is happening because version number differences between `rdkafka` and `karafka-rdkafka` that introduced a specific change to args on the consumer/producer methods. I was able to reproduce the same error using certain versions of karafka-rdkafka, so I've added an additional check that should allow the correct instrumentation to be installed. This solved the error for me, and you can test this branch with this:
```
gem 'newrelic_rpm', github: 'newrelic/newrelic-ruby-agent', branch: 'rdkafka_karafka_bugfix'
```

krisdigital:
@tannalynn Thank you for your quick reply and fix! We will check with the next version!

mensfeld:
Hey, maintainer of rdkafka and karafka-rdkafka here :wave: 

sorry for the issue. I am working on aligning their versioning as well as unifying them into one repo (rdkafka-ruby). However, this will still take time because of some major differences in error handling between them. Meanwhile do not hesitate if anything is off or if you have any issues :pray:

tannalynn:
The fix for for this has been merged and will be included in the next release. Thank you!

PR #2880, opened 2024-10-01
Karafka rdkafka bugfix

tannalynn:
The rdkafka instrumentation gets installed on both the rdkafka gem and the karafka-rdkafka gem because it is a forked versions or rdkafka. However, they introduced a change in different versions that requires instrumentation to be installed different. 
resolves https://github.com/newrelic/newrelic-ruby-agent/issues/2879

### DIFF
--- a/lib/new_relic/agent/instrumentation/rdkafka/prepend.rb
+++ b/lib/new_relic/agent/instrumentation/rdkafka/prepend.rb
@@ -36,7 +36,8 @@ module NewRelic::Agent::Instrumentation
     module Prepend
       include NewRelic::Agent::Instrumentation::RdkafkaConfig
 
-      if defined?(::Rdkafka) && Gem::Version.new(::Rdkafka::VERSION) >= Gem::Version.new('0.16.0')
+      if (defined?(::Rdkafka) && Gem::Version.new(::Rdkafka::VERSION) >= Gem::Version.new('0.16.0')) ||
+          (Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.7.0'))
         def producer(**kwargs)
           super.tap do |producer|
             set_nr_config(producer)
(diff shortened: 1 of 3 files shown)


lib/new_relic/agent/instrumentation/rdkafka/prepend.rb
```diff
@@ -36,7 +36,8 @@ module RdkafkaConfig
     module Prepend
       include NewRelic::Agent::Instrumentation::RdkafkaConfig
 
-      if defined?(::Rdkafka) && Gem::Version.new(::Rdkafka::VERSION) >= Gem::Version.new('0.16.0')
+      if (defined?(::Rdkafka) && Gem::Version.new(::Rdkafka::VERSION) >= Gem::Version.new('0.16.0')) ||
+          (Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.7.0'))
```
fallwith, line 40: @tannalynn Should the `defined?(::Rdkafka)` check be applied no matter what? With this change, a Ruby version over 2.7.0 will cause the `defined?` check to be skipped.

kaylareopelle: Do we need to check if Rdkafka is defined? 

I thought this file would only load if it's already defined because of the dependency detection check: https://github.com/newrelic/newrelic-ruby-agent/blob/b9c118e1c51b6940a0199416d17c6a9c5b1501bf/lib/new_relic/agent/instrumentation/rdkafka.rb#L5-L18 

If we need to check it, should the `chain` file also have a `defined?(::Rdkafka)` check?

tannalynn: When the agent is running normally it's not a problem, this is something I did in order to fix a test that was failing related to loading instrumentation. https://github.com/newrelic/newrelic-ruby-agent/pull/2849
IIRC, that test will load every single file and make sure no errors are thrown when loading it, so of course this fails because it assumes rdkafka is present already. Adding the defined check was the simplest path forward with no real side effect since it was just an extra check one time on start up.

Also, It's not needed in chain because in the chain file, that code is in a method that gets called later, so it works differently than prepend.

kaylareopelle (approved):

PR #2892, opened 2024-10-05
increase db.statement threshold to 4096

fallwith:
increase the db.statement max from 2000 to 4096 as per the updated cross agent spec


lib/new_relic/agent/span_event_primitive.rb
```diff
@@ -52,6 +52,8 @@ module SpanEventPrimitive
       DATASTORE_CATEGORY = 'datastore'
       CLIENT = 'client'
 
+      DB_STATEMENT_MAX_BYTES = 4096
```
tannalynn, line 55: Should this be 4096 or 4095? Because null terminator?

tannalynn (approved):

PR #2906, opened 2024-10-10
Create helper to access Bundler.rubygems specs

### LINKED ISSUE #2885: Deprecated Bundler::RubygemsIntegration#all_specs method
The bug described in #2733 is still present in 9.14. Namely, the invocation here:

https://github.com/newrelic/newrelic-ruby-agent/blob/9.14.0/lib/new_relic/agent/instrumentation/grape.rb#L23 

```ruby
((Bundler.rubygems.respond_to?(:installed_specs) && Bundler.rubygems.installed_specs.map(&:name).include?('newrelic-grape')) ||
            Bundler.rubygems.all_specs.map(&:name).include?('newrelic-grape'))
```

This expression checks whether the `newrelic-grape` gem is present using the new method and, if it's not, calls the deprecated method.

kaylareopelle:
This PR resolves a bug reported in #2885 to protect against scenarios where the version of Bundler used does not support `Bundler.rubygems.installed_specs` and needs to use `Bundler.rubygems.all_specs` to determine whether the third-party gem, `newrelic-grape` is installed in the environment.

Resolves #2885

### DIFF
--- a/lib/new_relic/agent/instrumentation/grape.rb
+++ b/lib/new_relic/agent/instrumentation/grape.rb
@@ -19,9 +19,13 @@ DependencyDetection.defer do
 
   depends_on do
     begin
-      if defined?(Bundler) &&
-          ((Bundler.rubygems.respond_to?(:installed_specs) && Bundler.rubygems.installed_specs.map(&:name).include?('newrelic-grape')) ||
-            Bundler.rubygems.all_specs.map(&:name).include?('newrelic-grape'))
+      specs = if Bundler.rubygems.respond_to?(:installed_specs)
+        Bundler.rubygems.installed_specs
+      else
+        Bundler.rubygems.all_specs
+      end
+
+      if specs.map(&:name).include?('newrelic-grape')
         NewRelic::Agent.logger.info('Not installing New Relic supported Grape instrumentation because the third party newrelic-grape gem is present')
         false
       else
--- a/lib/new_relic/agent/instrumentation/grape.rb
+++ b/lib/new_relic/agent/instrumentation/grape.rb
@@ -19,13 +19,7 @@ DependencyDetection.defer do
 
   depends_on do
     begin
-      specs = if Bundler.rubygems.respond_to?(:installed_specs)
-        Bundler.rubygems.installed_specs
-      else
-        Bundler.rubygems.all_specs
-      end
-
-      if specs.map(&:name).include?('newrelic-grape')
+      if NewRelic::Helper.rubygems_specs.map(&:name).include?('newrelic-grape')
         NewRelic::Agent.logger.info('Not installing New Relic supported Grape instrumentation because the third party newrelic-grape gem is present')
         false
       else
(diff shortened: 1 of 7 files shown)


lib/new_relic/agent/instrumentation/grape.rb
```diff
@@ -19,9 +19,13 @@
...
-      if defined?(Bundler) &&
-          ((Bundler.rubygems.respond_to?(:installed_specs) && Bundler.rubygems.installed_specs.map(&:name).include?('newrelic-grape')) ||
-            Bundler.rubygems.all_specs.map(&:name).include?('newrelic-grape'))
+      specs = if Bundler.rubygems.respond_to?(:installed_specs)
+        Bundler.rubygems.installed_specs
+      else
+        Bundler.rubygems.all_specs
+      end
+
+      if specs.map(&:name).include?('newrelic-grape')
```
kaylareopelle, line 28: Maybe a safe navigation operator before map JIC specs is nil? 

I took out the defined?(Bundler) check because if it raises an error, the error should be rescued within this block, but happy to add it back in if that's preferable.

fallwith: I think this is good as is, given that all known Bundler versions will return an array that can handle `#map`, empty or not. If you really want to try to harden and futureproof the logic, we could:

- Use `if/elsif/else` to check for `#all_specs` to be defined with an `elsif` and `else` into an early `false` return if neither are defined.
- Move the logic out into a helper shared by all similar Bundler specs checks

But both of those ideas seem like overkill and out of place stylistically for the norm for dependency detection checks which like to be simple and self-contained.

kaylareopelle: Since this isn't tested, I like the idea of moving this out into a helper shared by all similar Bundler specs checks. The lack of tests here feels wrong, but the mechanics with multiverse or mocks seemed heavy/unreliable.

We have similar spec gathering logic in four separate locations. Getting the logic to create the specs array in one place may make it easier to test. I'll give that a try tomorrow. Thank you!

kaylareopelle: Updated in:  9cdcd2c33c312acbcf72161ca9dbafe728b495a9

test/new_relic/helper_test.rb
```diff
@@ -77,4 +77,28 @@ def test_run_command_sad_exception
...
+    stub(:defined?, nil, ['Bundler']) do
+      result = NewRelic::Helper.rubygems_specs
+
+      assert_instance_of Array, result
+      assert_empty Array, result
+    end
+  end
+
+  def test_rubygems_specs_works_with_all_specs_when_installed_specs_is_absent
+    Bundler.rubygems.stub(:respond_to?, nil) do
```
kaylareopelle, line 94: I don't understand why, but Rubies 2.6 and below don't like the third argument for a stub in this scenario. However, they're fine with it in the test above. 🤷

fallwith: My guess is that the older Rubies are given an older Minitest. Wanting to use 3 arguments was the primary reason for the introduction of our `skip_unless_minitest5_or_above` helper.

fallwith (approved):

PR #2907, opened 2024-10-10
Update View Component metric building

### LINKED ISSUE #2874: ViewComponent: evaluate the way that we obtain the component class identifier
**NOTE: This is designed to be a complement to #2872 but if that issue has already been addressed by the time this one is worked on, the scope or need for this work may be impacted.**

PR #2870 fixed a bug to allow the agent to support view_component v3.15.0+, which stemmed from the agent's use of private `identifier` method that was removed in v3.15.0. The bugfix works by not attempting to call `identifier` on newer view_component versions and simply using `nil` instead. This works great - and certainly is a big improvement over erroring out - but may potentially cause cardinality issues going forward given that all components will now produce similar metric names.

As mentioned in https://github.com/newrelic/newrelic-ruby-agent/pull/2870#issuecomment-2383943722, view_component itself makes use of `self.class.source_location` for the identifier value it creates an ActiveSupport notification with. Could we leverage a `self.class.source_location` call and massage the resulting value into a format suitable for a metric name?

hannahramadan:
View Component removed private method `.identifier` in v3.16.0, which is how the library used to identify components for instrumentation, and is what our was based off of as well. View Component now uses `source_location`. This PR updates our instrumentation to do the same and adds error handling in the case that the methods we rely on to build metric names are changed. `source_location` is a confirmed compatible call with the earliest version of View Component available on RubyGems, v1.16.0.

closes #2874

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+Version <dev> updates View Componment instrumentation to use a default metric name when one is unavaliable.  
+
+- **Feature: Use default `View/component` metric name for unidentified View Components**
+  Previously, when a View Component metric name could not be identified, the agent would set the name as `nil`. Now, the agent defaults to using `View/component` as the metric name when one can not be identified. [PR#2907](https://github.com/newrelic/newrelic-ruby-agent/pull/2907)
+
+
 ## v9.14.0
 
 Version 9.14.0 adds Apache Kafka instrumentation for the rdkafka and ruby-kafka gems, introduces a configuration-based, automatic way to add custom instrumentation method tracers, correctly captures MIME type for ActionDispatch 7.0+ requests, properly handles Boolean coercion for `newrelic.yml` configuration, fixes a JRuby bug in the configuration manager, fixes a bug related to `Bundler.rubygems.installed_specs`, and fixes a bug to make the agent compatible with ViewComponent v3.15.0+.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 Version <dev> updates View Componment instrumentation to use a default metric name when one is unavaliable.  
 
 - **Feature: Use default `View/component` metric name for unidentified View Components**
+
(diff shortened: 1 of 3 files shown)


test/multiverse/suites/view_component/view_component_instrumentation_test.rb
```diff
@@ -60,12 +60,12 @@ def test_error_raised
 
   # Test metric name being built when the controller class doesn't respond to :identifier
   # https://github.com/newrelic/newrelic-ruby-agent/pull/2870
-  def test_the_metric_name_omits_the_identifier_when_absent
+  def test_the_metric_name_records_default_name_on_error
```
kaylareopelle, line 63: Should the comment on line 61 be updated?

kaylareopelle (commented):
Thank you! Would you mind adding a changelog entry?

test/multiverse/suites/view_component/view_component_instrumentation_test.rb
hannahramadan: Yes! Great catch

hannahramadan: Added https://github.com/newrelic/newrelic-ruby-agent/pull/2907/commits/b7f97152908d993867a9ee2000853893deb090f8

fallwith (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+Version <dev> updates View Componment instrumentation to use a default metric name when one is unavaliable.  
+
+- **Feature: Use default `View/component` metric name for unidentified View Components**
+  Previously, when a View Component metric name could not be identified, the agent would set the name as `nil`. Now, the agent defaults to using `View/component` as the metric name when one can not be identified. [PR#2907](https://github.com/newrelic/newrelic-ruby-agent/pull/2907)
```
kaylareopelle, line 8: We may need a newline between the title line and the description? Not sure though.

```suggestion

  Previously, when a View Component metric name could not be identified, the agent would set the name as `nil`. Now, the agent defaults to using `View/component` as the metric name when one can not be identified. [PR#2907](https://github.com/newrelic/newrelic-ruby-agent/pull/2907)
```

kaylareopelle (dismissed):

tannalynn (approved):

tannalynn (approved):

Issue #2912, opened 2024-10-12
Devise a way to report "unsatisfied" for Padrino to the csec agent


fallwith:
The Padrino gem's relationship to Sinatra led to an instrumentation design in the New Relic Ruby agent that sees the `:'instrumentation.sinatra'` configuration option being inspected and no independent option with the word "padrino" used.

As a result, there is currently no way to convey to the security agent that the Padrino gem's instrumentation dependencies were unsatisfied independently of the Sinatra gem.

A similar issue exists between the "memcache" and "dalli" gems, with our Dalli instrumentation effectively clobbering the "memcache" gem's configuration option, but given that the "memcache" gem is considered defunct, it's only Padrino that it is of concern currently.

PR #2914, opened 2024-10-15
prevent false 'unsatisfied' dependency status

### LINKED ISSUE #2893: Instrumentation configuration values not being properly preserved
**Updated 2024-10-11** As per Tanna's commented note below, this issue seems be related to an existing `chain` or `prepend` configuration setting being changed to `unsatisfied` incorrectly. This can convey incorrect information to the csec agent about what the APM agent did with a given instrumentation, but it does not impact regular APM usage of the instrumentation.

At this time it is hypothesized that only Padrino and the two forms of Dalli instrumentation are impacted.

~~The agent has offered support for module chaining based instrumentation installation dating back to a time before Ruby itself supported module prepending. Support for chaining has been retained primarily as a means to avoid colliding with other software that may itself be using module prepending.~~

~~Unfortunately it has been reported (internally) that agent's chain functionality is currently unreachable due to an apparent bug in the processing of the related configuration options.~~

~~Setting `instrumentation.<library>: chain` in the config file will see the agent not respect the `chain` value and use `prepend` instead.~~

~~**developer notes:** the issue seems to stem from DependencyDetection's `use_prepend?` method making calls to `prepend_configured?` - a method that does not exist.~~

fallwith:
address 2 issues causing dependency detection to falsely label an instrumentation's dependencies as "unsatisfied".

- handle Padrino with a special case pending resolution of https://github.com/newrelic/newrelic-ruby-agent/issues/2912
- don't mark an item as unsatisfied the second (or higher) time it comes around through the `detect!` loop

resolves #2893

### DIFF
--- a/lib/new_relic/dependency_detection.rb
+++ b/lib/new_relic/dependency_detection.rb
@@ -25,11 +25,9 @@ module DependencyDetection
 
   def detect!
     @items.each do |item|
-      if item.dependencies_satisfied?
-        item.execute
-      else
-        item.configure_as_unsatisfied unless item.disabled_configured?
-      end
+      next if item.executed || item.disabled_configured?
+
+      item.dependencies_satisfied? ? item.execute : item.configure_as_unsatisfied
     end
   end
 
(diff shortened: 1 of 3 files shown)


lib/new_relic/dependency_detection.rb
```diff
@@ -65,6 +63,13 @@ def dependencies_satisfied?
     end
 
     def configure_as_unsatisfied
+      # TODO: currently using :unsatisfied for Padrino will clobber the value
+      #       already set for Sinatra, so skip Padrino and circle back with a
+      #       new Padrino specific solution in the future.
+      #
+      #       https://github.com/newrelic/newrelic-ruby-agent/issues/2912
```
tannalynn, line 70: Really appreciate the comment here about what's going on with this. thank you!

tannalynn (approved):

kaylareopelle (approved):

PR #2915, opened 2024-10-17
Use double splat for Ruby >= 3.0

patrickarnett:
# Overview
`Kafka::Consumer#each_message` takes keyword arguments. In Ruby >= 3.0, prepending it with a method that takes only positional arguments causes an ArgumentError when calling `super`.

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label


tannalynn (approved):
Thank you for making this update! 
We did something similar for our rdkafka instrumentation but this one slid under the radar and I didn't realize this would happen. We appreciate the contribution!

gfmurphy:
@tannalynn any ideas on when this will land in a patch release? We recently ran into this when updating to 9.14.0. Thanks!

tannalynn:
@gfmurphy Thanks for checking in on this! We plan on releasing this fix later this week, as 9.15.0. Hopefully that helps!

Issue #2844, opened 2024-09-10
Optimisation to shrink the agent's footprint: don't require disabled or absent gem based instrumentation code


fallwith:
Most every gem instrumentation's detection code has some top-level lines like so:

```ruby
require_relative 'widget/instrumentation'
require_relative 'widget/chain'
require_relative 'widget/prepend'
```

These lines are evaluated by Ruby and cause all of instrumentation logic for the gem to be loaded into the Ruby VM even if the gem does not exist in the customer's tech stack and even if the customer's configuration has explicitly disabled the instrumentation. Essentially some Ruby methods will sit defined but orphaned in the customer's Ruby VM instance.

We can shrink the relevant footprint by moving these `require_relative` lines either into the existing `executes do` block within each instrumentation file, or within their own dedicated `executes do` block. By having the lines live within a block, they will only being executed by Ruby conditionally, based on the instrumentation's `depends_on` block(s).

We did this recently with the `rdkafka` gem instrumentation and we should consider doing the same for all gems.

PR #2917, opened 2024-10-21
DD templating: conditionally invoke requires

fallwith:
as per #2844, don't perform `require_relative` on content that won't be used unless all dependency detection checks succeed

### DIFF
--- a/lib/tasks/instrumentation_generator/templates/dependency_detection.tt
+++ b/lib/tasks/instrumentation_generator/templates/dependency_detection.tt
@@ -15,17 +15,17 @@ DependencyDetection.defer do
 
   executes do
     require_relative '<%= @snake_name.downcase %>/instrumentation'
-    require_relative '<%= @snake_name.downcase %>/chain'
-    require_relative '<%= @snake_name.downcase %>/prepend'
 
     # prepend_instrument and chain_instrument call extract_supportability_name
-    # to get the library name for supportability metrics and info-level logging. 
+    # to get the library name for supportability metrics and info-level logging.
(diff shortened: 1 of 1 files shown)


lib/tasks/instrumentation_generator/templates/dependency_detection.tt
```diff
@@ -18,6 +14,10 @@ DependencyDetection.defer do
   end
 
   executes do
+    require_relative '<%= @snake_name.downcase %>/instrumentation'
+    require_relative '<%= @snake_name.downcase %>/chain'
+    require_relative '<%= @snake_name.downcase %>/prepend'
+
```
kaylareopelle, line 20: Thoughts on going a step further... to require the instrumentation file here, and the chain and/or prepend files within the condition below? 

```
require_relative '<%= @snake_name.downcase %>/instrumentation'

  if use_prepend? 
    require_relative '<%= @snake_name.downcase %>/prepend'
    prepend_instrument <%= @class_name %>, NewRelic::Agent::Instrumentation::<%= @class_name %>::Prepend
 else
   require_relative '<%= @snake_name.downcase %>/chain'
   chain_instrument NewRelic::Agent::Instrumentation::<%= @class_name %>::Chain
 end
 ```

fallwith: Sure, addressed with [1c6d8a3](https://github.com/newrelic/newrelic-ruby-agent/pull/2917/commits/1c6d8a35f40240168f043379b4d9f9947fa89877).

kaylareopelle (approved):
Thank you!

PR #2918, opened 2024-10-22
DD: update the rendering of the 'named' line

fallwith:
the template's rendering of a `named` line was previously attempting to
wrap symbol values with single quotes when necessary.

but given that everything is routed through the `snake_name` method,
nothing will ever need the quotes if we simply use the output of that
method, held in `@snake_name`.

### DIFF
--- a/lib/tasks/instrumentation_generator/templates/dependency_detection.tt
+++ b/lib/tasks/instrumentation_generator/templates/dependency_detection.tt
@@ -3,7 +3,7 @@
 # frozen_string_literal: true
 
 DependencyDetection.defer do
-  named :<%= @name.match?(/\-|\_/) ? "'#{@snake_name}'" : @name.downcase %>
+  named :<%= @snake_name %>
 
   depends_on do
     # The class that needs to be defined to prepend/chain onto. This can be used


lib/tasks/instrumentation_generator/templates/dependency_detection.tt
```diff
@@ -3,7 +3,7 @@
 # frozen_string_literal: true
 
 DependencyDetection.defer do
-  named :<%= @name.match?(/\-|\_/) ? "'#{@snake_name}'" : @name.downcase %>
+  named :<%= (@name == @snake_name ? @name : @snake_name).downcase %>
```
kaylareopelle, line 6: the more I look at this, maybe it just needs to be `<%= @snake_name %>`? 
* we'll have an `@snake_name` assigned every time, that will either be equal to `@name` or be the `@name` with hyphens transformed into underscores
* `snake_name()` calls `downcase`, so we shouldn't have to here

fallwith: Great call. Once we're out of the business of using quotes or not, we can just use `@snake_name`. Updated [the commit]([2507052](https://github.com/newrelic/newrelic-ruby-agent/pull/2918/commits/2507052f091da56e21ecea71ec6d918d4b942584)). Will update the PR description accordingly.

kaylareopelle (approved):

PR #2923, opened 2024-10-28
add .vscode/ to gitignore

tannalynn:
My spellcheck stores saved words in a .vscode/ directory, and apparently thats not something we've had happen in our repo before, so I added it to the gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ lib/new_relic/build.rb
 .tm_properties
 .bundle
 .yardoc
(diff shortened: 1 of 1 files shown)


fallwith (approved):
I think something like `.vscode` is a better fit for a `~/.gitignore` file, but we've already got lots of things like `.idea` already in the project file that this isn't any worse than.

kaylareopelle (approved):

PR #2926, opened 2024-10-29
instrumentation for aws-sdk-lambda

fallwith:
Instrument the aws-sdk-lambda gem and upon function invocations performed by the client, create external request segments that include AWS resource information.

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -1529,6 +1529,15 @@ module NewRelic
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of bunny at start-up. May be one of: `auto`, `prepend`, `chain`, `disabled`.'
         },
+        :'instrumentation.aws_sdk_lambda' => {
+          :default => 'auto',
+          :documentation_default => 'auto',
+          :public => true,
+          :type => String,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => 'Controls auto-instrumentation of the aws_sdk_lambda library at start-up. May be one of `auto`, `prepend`, `chain`, `disabled`.'
+        },
         :'instrumentation.ruby_kafka' => {
           :default => 'auto',
           :public => true,
--- a/lib/new_relic/agent/instrumentation/aws_sdk_lambda.rb
+++ b/lib/new_relic/agent/instrumentation/aws_sdk_lambda.rb
@@ -1,0 +1,28 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+DependencyDetection.defer do
+  named :aws_sdk_lambda
+
+  depends_on do
+    defined?(Aws::Lambda::Client)
+  end
+
+  executes do
+    require_relative 'aws_sdk_lambda/instrumentation'
+
+    # prepend_instrument and chain_instrument call extract_supportability_name
+    # to get the library name for supportability metrics and info-level logging.
+    # This is done by spliting on the 2nd to last spot of the instrumented
+    # module. If this isn't how we want the name to appear, pass in the desired
+    # name as a third argument.
+    if use_prepend?
+      require_relative 'aws_sdk_lambda/prepend'
+      prepend_instrument Aws::Lambda::Client, NewRelic::Agent::Instrumentation::AwsSdkLambda::Prepend
+    else
+      require_relative 'aws_sdk_lambda/chain'
+      chain_instrument NewRelic::Agent::Instrumentation::AwsSdkLambda::Chain
+    end
+  end
+end
--- a/lib/new_relic/agent/instrumentation/aws_sdk_lambda.rb
+++ b/lib/new_relic/agent/instrumentation/aws_sdk_lambda.rb
@@ -12,11 +12,6 @@ DependencyDetection.defer do
   executes do
     require_relative 'aws_sdk_lambda/instrumentation'
 
-    # prepend_instrument and chain_instrument call extract_supportability_name
-    # to get the library name for supportability metrics and info-level logging.
-    # This is done by spliting on the 2nd to last spot of the instrumented
-    # module. If this isn't how we want the name to appear, pass in the desired
-    # name as a third argument.
     if use_prepend?
       require_relative 'aws_sdk_lambda/prepend'
       prepend_instrument Aws::Lambda::Client, NewRelic::Agent::Instrumentation::AwsSdkLambda::Prepend
--- a/lib/new_relic/agent/instrumentation/aws_sdk_lambda/instrumentation.rb
+++ b/lib/new_relic/agent/instrumentation/aws_sdk_lambda/instrumentation.rb
@@ -1,0 +1,117 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic::Agent::Instrumentation
+  module AwsSdkLambda
+    INSTRUMENTATION_NAME = 'aws_sdk_lambda'
+    AWS_SERVICE = 'lambda'
+    CLOUD_PLATFORM = 'aws_lambda'
+    ERROR_STATUS_REGEX = /^(?:4|5)/
+    WRAPPED_RESPONSE = Struct.new(:status_code, :has_status_code?)
+
+    def invoke_with_new_relic(*args)
+      with_tracing(:invoke, *args) { yield }
+    end
+
+    def invoke_async_with_new_relic(*args)
+      with_tracing(:invoke_async, *args) { yield }
+    end
+
+    def invoke_with_response_stream_with_new_relic(*args)
+      with_tracing(:invoke_with_response_stream, *args) { yield }
+    end
+
+    private
+
+    def with_tracing(action, *args)
+      segment = generate_segment(action, *args)
+
+      # prevent additional instrumentation for things like Net::HTTP from
+      # creating any segments that may appear as redundant / confusing
+      NewRelic::Agent.disable_all_tracing do
+        begin
+          response = yield
+          process_response(response, segment)
+          response
+        rescue => e
+          # notice error that was unhandled by the AWS SDK Lambda client
+          NewRelic::Agent.notice_error(e)
+          raise
+        end
+      end
+    ensure
+      segment&.finish
+    end
+
+    def process_response(response, segment)
+      process_status_code(response, segment) if response.respond_to?(:status_code)
+      process_function_error(response.function_error, segment) if response.respond_to?(:function_error)
+    end
+
+    def process_status_code(response, segment)
+      status_code = response.status_code
+      return unless status_code
+
+      segment.process_response_headers(WRAPPED_RESPONSE.new(status_code, true))
+
+      # notice error that was handled by the function
+      if status_code.to_s.match?(ERROR_STATUS_REGEX)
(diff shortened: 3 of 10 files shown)


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1537,6 +1537,15 @@ def self.notify
...
           :description => 'Controls auto-instrumentation of bunny at start-up. May be one of: `auto`, `prepend`, `chain`, `disabled`.'
         },
+        :'instrumentation.aws_sdk_lambda' => {
+          :default => 'auto',
+          :documentation_default => 'auto',
+          :public => true,
+          :type => String,
+          :dynamic_name => true,
+          :allowed_from_server => false,
+          :description => 'Controls auto-instrumentation of the aws_sdk_lambda library at start-up. May be one of `auto`, `prepend`, `chain`, `disabled`.'
```
kaylareopelle, line 1547: This seems like a quirk of the instrumentation generator, but should this be `aws-sdk-lambda` to match the gem name?

lib/new_relic/agent/instrumentation/aws_sdk_lambda.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  executes do
+    require_relative 'aws_sdk_lambda/instrumentation'
+
+    # prepend_instrument and chain_instrument call extract_supportability_name
+    # to get the library name for supportability metrics and info-level logging.
+    # This is done by spliting on the 2nd to last spot of the instrumented
+    # module. If this isn't how we want the name to appear, pass in the desired
+    # name as a third argument.
```
kaylareopelle, line 19: Do we need this comment? It doesn't look like a third argument is being passed.

lib/new_relic/agent/instrumentation/aws_sdk_lambda/instrumentation.rb
```diff
@@ -0,0 +1,112 @@
...
+      # prevent additional instrumentation for things like Net::HTTP from
+      # creating any segments that may appear as redundant / confusing
+      NewRelic::Agent.disable_all_tracing do
+        begin
+          response = yield
+          process_response(response, segment)
+          response
+        rescue => e
+          NewRelic::Agent.notice_error(e)
+          raise
```
kaylareopelle, line 40: Is this intended to capture errors both raised by the agent and raised by the customer's invocation of the AWS lambda gem?

lib/new_relic/agent/instrumentation/aws_sdk_lambda/instrumentation.rb
```diff
@@ -0,0 +1,112 @@
...
+
+    def generate_segment(action, options = {})
+      function = function_name(options)
+      region = aws_region
+      account_id = aws_account_id
+      arn = aws_arn(function, region)
+
+      segment = NewRelic::Agent::Tracer.start_external_request_segment(
+        library: INSTRUMENTATION_NAME,
+        uri: "https://lambda.#{aws_region || 'unknown-region'}.amazonaws.com",
```
kaylareopelle, line 82: 1. Is the `'unknown-region'` string part of a spec or an AWS thing? 
2. Would it be valuable to set `region` as `'unknown-region'` as a fallback value whenever region can't be derived?
3. Does it make sense to use `region` here instead of calling `aws_region` again?

lib/new_relic/agent/instrumentation/aws_sdk_lambda/instrumentation.rb
```diff
@@ -0,0 +1,112 @@
...
+        procedure: action
+      )
+      segment.name = "External/Lambda/#{action}/#{function}"
+      segment.add_agent_attribute('cloud.account.id', account_id)
+      segment.add_agent_attribute('cloud.platform', CLOUD_PLATFORM)
+      segment.add_agent_attribute('cloud.region', region)
+      segment.add_agent_attribute('cloud.resource_id', arn) if arn
```
kaylareopelle, line 89: There's a guard for `arn`, but not for `account_id` or `region`. Is it okay to have nil/empty values for the other attributes?

lib/new_relic/agent/instrumentation/aws_sdk_lambda/instrumentation.rb
```diff
@@ -0,0 +1,112 @@
...
+    def process_function_error(response)
+      function_error = response.function_error
+      return unless function_error
+
+      msg = "[#{function_error}]"
+      payload = response.payload&.string if response.respond_to?(:payload)
+      payload_hash = JSON.parse(payload) if payload
+      msg = "#{msg} #{payload_hash['errorType']} - #{payload_hash['errorMessage']}" if payload_hash
+      e = StandardError.new(msg)
+      e.set_backtrace(payload_hash['stackTrace']) if payload_hash
```
kaylareopelle, line 69: I recently learned about this feature, neat to see it used here!

test/multiverse/suites/aws_sdk_lambda/aws_sdk_lambda_instrumentation_test.rb
```diff
@@ -0,0 +1,118 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require_relative 'aws_sdk_monkeypatch'
+require 'json'
+
+class AwsSdkLambdaInstrumentationTest < Minitest::Test
+  REGION = 'us-east-2'
+  AWS_ACCOUNT_ID = '8675309'
```
kaylareopelle, line 10: Jenny!

test/multiverse/suites/aws_sdk_lambda/aws_sdk_lambda_instrumentation_test.rb
```diff
@@ -0,0 +1,118 @@
...
+        client.config.account_id = AWS_ACCOUNT_ID
+
+        client.send(method, {function_name: function_name}.merge(extra_args))
+
+        segment = lambda_segment(txn)
+
+        assert_equal("External/Lambda/#{method}/#{function_name}", segment.name)
+        assert_equal("lambda.#{REGION}.amazonaws.com", segment.host)
+        assert_equal(200, segment.http_status_code) unless method == :invoke_async
+        assert_equal('aws_sdk_lambda', segment.library) # rubocop:disable Minitest/EmptyLineBeforeAssertionMethods
```
kaylareopelle, line 97: Is the `rubocop:disable` needed here because of the `unless` on the previous line?

kaylareopelle (commented):
This looks great, James! It's been a while since I've thought about lambdas, so I have a few clarification questions in the PR.

lib/new_relic/agent/instrumentation/aws_sdk_lambda/instrumentation.rb
fallwith: Yeah, my interpretation of the spec and existing aws-sdk related instrumentation is that `arn` is the only one that gets a guard. Further spec refinement might end up clarifying that the others need guards too, but for now `arn` was the only one with a specific instruction to guard.

test/multiverse/suites/aws_sdk_lambda/aws_sdk_lambda_instrumentation_test.rb
fallwith: Yeah, precisely. I disagreed with the need for the blank line in this context.

lib/new_relic/agent/instrumentation/aws_sdk_lambda/instrumentation.rb
fallwith: #3 DRYed up with [7d64822](https://github.com/newrelic/newrelic-ruby-agent/pull/2926/commits/7d648221edd4419e8fb53b3d00f6edb633cef411)

lib/new_relic/agent/instrumentation/aws_sdk_lambda.rb
kaylareopelle:
```suggestion
```
Got it. Mind if we remove it? Or do you prefer we keep it JIC a third arg is needed in the future?

lib/new_relic/agent/instrumentation/aws_sdk_lambda/instrumentation.rb
kaylareopelle: Ah interesting. I'm a little concerned about stopping the customer's application if the agent raises an error. 

Would a `NewRelic::Agent::Tracer.capture_segment_error` block work around the yield to catch customer-raised errors? Totally understand if not, this seems like a pretty unique scenario compared to our other instrumentation.

lib/new_relic/agent/instrumentation/aws_sdk_lambda.rb
fallwith: Boilerplate comment removed with [375db67](https://github.com/newrelic/newrelic-ruby-agent/pull/2926/commits/375db6736890c175e18f402a3239345310f03fec)

lib/new_relic/agent/instrumentation/aws_sdk_lambda/instrumentation.rb
fallwith: Great call. I was thinking of it only from the perspective of both sources of exceptions sharing a `rescue`. They certainly should not share the `raise` which was intended only for client sdk raised exceptions. Will refactor!

lib/new_relic/agent/instrumentation/aws_sdk_lambda.rb
fallwith: Argh! Sorry, @kaylareopelle. I quickly skimmed the email notification saying "Mind if we remove it?" without noticing that you'd offered a suggested commit. Now in the web UI I see it clearly.

kaylareopelle: That's okay, I don't feel offended! 😄  Good to know about how things render on email!

lib/new_relic/agent/instrumentation/aws_sdk_lambda/instrumentation.rb
kaylareopelle: Wow, even different languages?! 🙃 
Nice accommodation!

lib/new_relic/agent/instrumentation/aws_sdk_lambda/instrumentation.rb
fallwith: Refactored with [0a1527d](https://github.com/newrelic/newrelic-ruby-agent/pull/2926/commits/0a1527d8012cfb34c3950d34180d0974b51ab9ed) (with co-author!)

lib/new_relic/agent/instrumentation/aws_sdk_lambda/instrumentation.rb
```diff
@@ -0,0 +1,112 @@
...
+
+    private
+
+    def with_tracing(action, *args)
+      segment = generate_segment(action, *args)
+
+      # prevent additional instrumentation for things like Net::HTTP from
+      # creating any segments that may appear as redundant / confusing
+      NewRelic::Agent.disable_all_tracing do
+        response = NewRelic::Agent::Tracer.capture_segment_error(segment) { yield }
```
tannalynn, line 34: Does `capture_segment_error` work normally inside a `disable_all_tracing`?  I assume it works normally but not totally sure since I don't think I've thought about it before, so I wanted so check

fallwith: Good question, @tannalynn. Yes, it will work normally as desired. It looks like the only layer of context that would possibly have an impact here would be high security mode, which would prevent noticed errors from being sent up.

tannalynn (approved):

PR #2944, opened 2024-11-12
CI: introduce Rails v8.0

fallwith:
* start testing with Rails v8.0
* update Rails EDGE testing to reference v8.1
* simplify `Gemfile` inclusion of `pry`
* update multiverse suites with newer library versions
* misc cleanup


kaylareopelle (dismissed):

test/environments/railsedge/config/application.rb
```diff
@@ -13,7 +13,7 @@
 module RpmTestApp
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults(8.0) # TODO: keep this number up to date (and keep this TODO)
+    config.load_defaults(8.1) # TODO: keep this number up to date (and keep this TODO)
```
kaylareopelle, line 16: There may not be Rails 8.1 defaults yet... 
https://github.com/newrelic/newrelic-ruby-agent/actions/runs/11806613018/job/32891658488#step:12:6141 
in active_record_pg suite running on Rails 8.0

test/multiverse/suites/active_support_logger/Envfile
```diff
@@ -5,6 +5,9 @@
 instrumentation_methods :chain, :prepend
 
 ACTIVE_SUPPORT_VERSIONS = [
+  ['8.0.0', 3.2],
+  ['7.2.2', 3.1],
+  ['7.1.5', 2.7],
```
kaylareopelle, line 10: ActiveSupport::Logger.broadcast is not defined on Rails versions above 7.0.
The tests fail because we call `ActiveSupport::Logger.broadcast` in the setup, regardless of the version running the test. 

We can either test on newer versions and update the setup to gracefully fail when we're using a version where it isn't supported, or just skip testing on Rails versions above 7.0. 

```suggestion
# ActiveSupport::Logger.broadcast is only defined on Rails 7.0 and below
# Do not test newer Rails versions on this suite.
ACTIVE_SUPPORT_VERSIONS = [
```

kaylareopelle: @tannalynn or @hannahramadan, do you have a preference on the approach?

tannalynn: Is there any reason why we'd need to run these tests on versions that don't do the broadcast? Because if there isn't a specific reason, it makes more sense to just not run them on those versions

tannalynn: oh also, Do we install the instrumentation only on the versions that support broadcast?

tannalynn (approved):

Issue #2947, opened 2024-11-13
Ruby 3.0+ can't find run_id for thread profiler


kaylareopelle:
Ruby 3.0, 3.1, 3.2, and 3.3 all fail the following test in the `agent_only` multiverse suite:
```
ThreadProfilingTest#test_thread_profiling_can_stop:
NoMethodError: undefined method `run_id' for nil
    /home/runner/work/newrelic-ruby-agent/newrelic-ruby-agent/test/multiverse/suites/agent_only/thread_profiling_test.rb:88:in `test_thread_profiling_can_stop'
```

Figure out where `run_id` went and how we can get it back.

PR #2961, opened 2024-11-19
Skip test_thread_profiling on 3.0+

kaylareopelle:
Intermittent failures occur on the other test, `test_thread_profiling`, so skip those until we can address with #2947


hannahramadan (approved):
Groovy 😎 Updated the issue description in #2947 to keep track!

PR #2964, opened 2024-11-19
Resurrect aws account id decoding

tannalynn:
Now that we have the official approval from AWS to get the aws account ID by decoding, i have now added that logic back to the AWS module and updated the instrumentation modules using that.

### DIFF
--- a/lib/new_relic/agent/aws.rb
+++ b/lib/new_relic/agent/aws.rb
@@ -9,6 +9,9 @@ module NewRelic
       HEX_MASK = '7fffffffff80'
 
       def self.create_arn(service, resource, region, account_id)
+        # if any of the values are nil, we can't create an ARN
+        return unless service && resource && region && account_id
+
         "arn:aws:#{service}:#{region}:#{account_id}:#{resource}"
       rescue => e
         NewRelic::Agent.logger.warn("Failed to create ARN: #{e}")
--- a/lib/new_relic/agent/aws.rb
+++ b/lib/new_relic/agent/aws.rb
@@ -5,12 +5,60 @@
 module NewRelic
   module Agent
     module Aws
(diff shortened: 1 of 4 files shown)


lib/new_relic/agent/aws.rb
```diff
@@ -5,13 +5,64 @@
...
     module Aws
-      def self.create_arn(service, resource, region)
-        return unless NewRelic::Agent.config[:'cloud.aws.account_id']
+      CHARACTERS = %w[A B C D E F G H I J K L M N O P Q R S T U V W X Y Z 2 3 4 5 6 7].freeze
+      HEX_MASK = '7fffffffff80'
 
-        "arn:aws:#{service}:#{region}:#{NewRelic::Agent.config[:'cloud.aws.account_id']}:#{resource}"
+      def self.create_arn(service, resource, region, account_id)
+        # if any of the values are nil, we can't create an ARN
+        return unless service && resource && region && account_id
```
hannahramadan, line 13: Do we want to log if we can't create the arn because of a `nil`

hannahramadan (approved):

kaylareopelle (approved):

lib/new_relic/agent/aws.rb
kaylareopelle: Those are great points. The customer also wouldn't have anything they could do to fix the problem and start creating an ARN (outside of adding it via the config option, but that's easy for us to diagnose). Most of this data either comes from our instrumentation or from the AWS gem. I'm hesitant to give a customer logs that they can't do much about.

What if we keep this place in mind if we encounter problems with ARNs not getting created, and add logging here on a per-customer basis for GTSEs?

hannahramadan: Perfect - all sounds good to me too!

PR #2967, opened 2024-11-20
Remove Trivy scans

kaylareopelle:
They are only required for teams that ship containers. We do not ship containers.


hannahramadan (approved):
✌️

PR #2973, opened 2024-11-26
Firehose instrumentation

hannahramadan:
Add instrumentation support for [aws-sdk-firehose](https://rubygems.org/gems/aws-sdk-firehose).


lib/new_relic/agent/instrumentation/aws_sdk_firehose/instrumentation.rb
```diff
@@ -0,0 +1,66 @@
...
+      begin
+        NewRelic::Agent::Tracer.capture_segment_error(segment) { yield }
+      ensure
+        segment&.add_agent_attribute('cloud.platform', AWS_KINESIS_DELIVERY_STREAMS)
+        segment&.finish
+      end
+    end
+
+    def get_segment_name(method_name, params)
+      stream_name = params&.dig(:delivery_stream_name)
```
kaylareopelle, line 44: Just for parallelism with the other PR, if you're removing the digs there, and `:delivery_stream_name` is at the top level, please remove them from this file too!

lib/new_relic/agent/instrumentation/aws_sdk_firehose/instrumentation.rb
```diff
@@ -0,0 +1,66 @@
...
+module NewRelic::Agent::Instrumentation
+  module Firehose
+    INSTRUMENTED_METHODS = %w[
+      create_delivery_stream
+      delete_delivery_stream
+      describe_delivery_stream
+      list_delivery_streams
+      list_tags_for_delivery_stream
+      put_record
+      put_record_batch
```
kaylareopelle, line 14: Is it worth asking if these should be message broker segments? They seem very similar to the ones we switched in Kinesis.

hannahramadan: Definitely worth asking! I will do this

lib/new_relic/agent/instrumentation/aws_sdk_firehose/instrumentation.rb
hannahramadan: I decided to keep in the digs for both files. I liked the benefit of the safe navigation and while I think stream_name will be top level for both cases, I thought the extra safety would be nice. Open to other thoughts. You mentioned pros/cons of `dig` but Im not fully sure of the cons. Do you think those outweigh pros here?

kaylareopelle: Ah, my bad. I misunderstood the conversation you and Tanna had on the other PR. I guess the other side of the story is performance. I think we might have to microbenchmark and/or understand the likelihood of the params hash having a structure we didn't expect to understand which is preferable. If you and Tanna landed in the other direction, I support y'all.

hannahramadan: No no def my bad. I did go back and forth on this - sorry for adding confusion! Did some benchmarking. Dig is somewhere between `same-ish` and about 1.2x slower than the bracket method.

kaylareopelle: Okay, I'm fine with either approach! We can just leave things as-is.

lib/new_relic/agent/instrumentation/aws_sdk_firehose/instrumentation.rb
hannahramadan: Per spec, Firehose has no message broker segments. So we're good here!

kaylareopelle (approved):
🔥

kaylareopelle (approved):
🚒

Issue #2898, opened 2024-10-08
Add instrumentation and entity relationship support for AWS_KINESIS_DELIVERY_STREAM


kford-newrelic:
### Description
Add core instrumentation support for the AWS_KINESIS_DELIVERY_STREAM service to the Ruby Agent, along with necessary linking data for entity relationships.

### Acceptance Criteria
When APM monitored applications are using the AWS_KINESIS_DELIVERY_STREAM service, entity relationships are created.

### Design Consideration/Limitations
Assume the data to be emitted for linking will need to rely on ARN-like matching, requiring the AWS account ID to be available in some manner.

### Dependencies
Should follow the general guidelines (agent spec?) established by the Java Agent.

### Additional context
N/A

PR #2974, opened 2024-11-26
Kinesis instrumentation

hannahramadan:
Add support for [aws-sdk-kinesis](https://rubygems.org/gems/aws-sdk-kinesis). 

closes [#2898](https://github.com/newrelic/newrelic-ruby-agent/issues/2898)


lib/new_relic/agent/instrumentation/kinesis/instrumentation.rb
```diff
@@ -0,0 +1,71 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic::Agent::Instrumentation
+  module Kinesis
+    INSTRUMENTED_METHODS = %w[
```
kaylareopelle, line 7: We need to update the instrumentation to create message broker spans for: 
`get_records`
`put_record`
`put_records`

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1561,6 +1561,15 @@ def self.notify
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of the aws_sdk_lambda library at start-up. May be one of `auto`, `prepend`, `chain`, `disabled`.'
         },
+        :'instrumentation.kinesis' => {
```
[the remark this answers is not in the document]
hannahramadan: This is a great call out! Consistency is important. I like the idea of using the gem name. I think in most places that is what we do that. 

Although I dont think we use hypens in config names, so this would be `instrumentation.aws_sdk_kinesis`. What do you think? cc @tannalynn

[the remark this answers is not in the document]
kaylareopelle: `instrumentation.aws_sdk_kinesis` looks good to me. Curious to hear @tannalynn's thoughts on the naming approach overall.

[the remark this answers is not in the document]
hannahramadan: The name has been udpated.

lib/new_relic/agent/instrumentation/aws_sdk_kinesis/instrumentation.rb
```diff
@@ -0,0 +1,88 @@
...
+    def get_segment_name(method_name, params)
+      return "#{KINESIS}/#{method_name}/#{params[:stream_name]}" if params&.dig(:stream_name)
+
+      "#{KINESIS}/#{method_name}"
+    rescue => e
+      NewRelic::Agent.logger.warn("Failed to create segment name: #{e}")
+    end
+
+    def get_stream_name(params, arn)
+      params[:stream_name] || arn.split('/').last || 'unknown'
```
kaylareopelle, line 71: Does it make sense to use `dig` here, like on L63?

hannahramadan: I was thinking we don't need it here because if `params[:stream_name]` is `nil` then we will just move on to getting the stream name from the arn, but what do you think? Def open to adding it.

lib/new_relic/agent/instrumentation/aws_sdk_kinesis/chain.rb
```diff
@@ -0,0 +1,21 @@
...
+module NewRelic::Agent::Instrumentation
+  module Kinesis::Chain
+    def self.instrument!
+      ::Aws::Kinesis::Client.class_eval do
+        include NewRelic::Agent::Instrumentation::Kinesis
+
+        NewRelic::Agent::Instrumentation::Kinesis::INSTRUMENTED_METHODS.each do |method_name|
+          alias_method("#{method_name}_without_new_relic".to_sym, method_name.to_sym)
+
+          define_method(method_name) do |*args|
```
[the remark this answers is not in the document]
hannahramadan: None of the SDK methods use kwargs or blocks! Do you want to add something to safe guard if that changes in the future?

lib/new_relic/agent/instrumentation/aws_sdk_kinesis/instrumentation.rb
kaylareopelle: Oh, interesting. I think I might be a bit confused. I think of using `dig` to access a key in a deeply nested hash. Will `:stream_name` always be at the top level?

hannahramadan: Oh fair point! I've seen `:stream_name` at top level so far, but I am not confident this will always be the case. Lets `dig`

lib/new_relic/agent/instrumentation/aws_sdk_kinesis/chain.rb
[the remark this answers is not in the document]
kaylareopelle: Thank you for checking! I leave that up to you. Either way has its pros/cons.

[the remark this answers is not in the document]
hannahramadan: Took another look at the SDK methods - all instrumented methods take a `params` hash, and when `:stream_name` is expected, it looks like it should actually be top-level. So on second look I think we should be safe to not need to use `dig` when looking for `:stream_name`

lib/new_relic/agent/instrumentation/aws_sdk_kinesis/instrumentation.rb
hannahramadan: Updated for dig! https://github.com/newrelic/newrelic-ruby-agent/pull/2974/commits/39dfac2ec4005a4538d9c23aae81244d292e6742. In testing, I realized there is a chance that `params` can be empty in which calling `[]` will raise an error. Dig works to prevent that which is a nice benefit

kaylareopelle (approved):
Looks good to me! Thanks for all your hard work on this!

lib/new_relic/agent/instrumentation/aws_sdk_kinesis/instrumentation.rb
```diff
@@ -0,0 +1,88 @@
...
+    def nr_account_id
+      return @nr_account_id if defined?(@nr_account_id)
+
+      @nr_account_id = NewRelic::Agent::Aws.get_account_id(config)
+    end
+
+    def get_arn(params)
+      return params[:stream_arn] if params&.dig(:stream_arn)
+
+      NewRelic::Agent::Aws.create_arn(KINESIS.downcase, "stream/#{params[:stream_name]}", config&.region, nr_account_id) if params&.dig(:stream_name)
```
tannalynn, line 85: how come the if is checking for `params&.dig(:stream_name)`, but method is using just `params[:stream_name]`?

lib/new_relic/agent/instrumentation/aws_sdk_kinesis/instrumentation.rb
```diff
@@ -0,0 +1,88 @@
...
+      begin
+        NewRelic::Agent::Tracer.capture_segment_error(segment) { yield }
+      ensure
+        segment&.add_agent_attribute('cloud.platform', AWS_KINESIS_DATA_STREAMS)
+        segment&.finish
+      end
+    end
+
+    def get_segment_name(method_name, params)
+      return "#{KINESIS}/#{method_name}/#{params[:stream_name]}" if params&.dig(:stream_name)
```
tannalynn, line 63: Same question on the dig VS direct hash on `stream_name` here

lib/new_relic/agent/configuration/default_source.rb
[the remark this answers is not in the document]
tannalynn: oh i didnt see this till now, oops! but yeah that sounds good to me too. As for the inconsistency on the aws namings, that's my bad, sorry about that! I didn't think about that until now.

lib/new_relic/agent/instrumentation/aws_sdk_kinesis/instrumentation.rb
hannahramadan: Ah I was using the params&.dig in part for safe navigation, but to your point this doesnt really make sense to dig and then assume the stream_name is top level. An oversight. Updating!

lib/new_relic/agent/instrumentation/aws_sdk_kinesis/instrumentation.rb
hannahramadan: Updated!

kaylareopelle (approved):
💖

tannalynn (approved):

PR #2987, opened 2024-12-06
Do not rename final Grape segment

### LINKED ISSUE #2982: Grape instrumentation renames last segment
## Description
Last tracked event (database query, external api request or custom instrumentation anything that newrelic_rpm support) always renamed to "Middleware/Grape/<...>/call" where <...> is class name that implements grape api.

## Expected Behavior
Grape instrumentation doesn't rename any custom instrumentation or instrumentation included in newrelic_rpm gem.

## Steps to Reproduce
It is always reproducible with setup using: rackup + grape + newrelic_rpm

I created very simple example app https://github.com/seriousdev-gh/grape_new_relic_example

image
In screenshot you can see recieved segments. Example app create two segments with names `Custom/LongProcess/1`, `Custom/LongProcess/2` and durations 1 and 2 seconds respectively. 2 second event must be named `Custom/LongProcess/2` instead of `Middleware/Grape/Main/call`.

## Your Environment
ruby 3.3.2 (2024-05-30 revision e5a195edf6) [x86_64-linux]
newrelic_rpm (9.16.1)
Other gem versions in example app.

## Additional context
Found this line of code that i think causes the issue https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/grape/instrumentation.rb#L62

For a workaround in my production app i wrote a monkeypatch:

```ruby
module NewRelic::Agent::Instrumentation # rubocop:disable Style/ClassAndModuleChildren
  module Grape
    module Instrumentation
      extend self # rubocop:disable Style/ModuleFunction

      def name_transaction(route, class_name, version)
        txn_name = name_for_transaction(route, class_name, version)
        NewRelic::Agent::Transaction.set_default_transaction_name(txn_name, :grape)
      end
    end
  end
end
```

hannahramadan:
Previously, the agent renamed the last segment in a Grape transaction to `"Middleware/Grape/#{class_name}/call"`. From what we can tell, this was helpful in an older version of the agent's architecture, but it is no longer relevant today.

closes #2982

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## dev
 
-- **Bugfix: Do not rename final Grape segment**
+- **Bugfix: Stop renaming final Grape segment**
 
   Previously, the agent renamed the final segment in Grape transactions to `"Middleware/Grape/#{class_name}/call"`. This was a part of an old instrumentation pattern that is no longer relevant. Many thanks to [@seriousdev-gh](https://github.com/seriousdev-gh) for bringing this issue to our attention and along with a great reproduction and suggested fix. [PR#2987](https://github.com/newrelic/newrelic-ruby-agent/pull/2987).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+- **Bugfix: Do not rename final Grape segment**
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+- **Bugfix: Do not rename final Grape segment**
```
kaylareopelle, line 5: Thinking about the agent update UI, I wonder if this matches better?
```suggestion
- **Bugfix: Stop renaming final Grape segment**
```

kaylareopelle (dismissed):

kaylareopelle (approved):

PR #2995, opened 2024-12-12
AC/FC integration

### LINKED ISSUE #2859: AC/FC Integration Support
### Description
We are pivoting to making Kubernetes the primary experience for NR Control Public Preview in Jan. This means we must deliver remote health, config, install and upgrade for Ruby.

The main body of work will be implementing support for health and any additional work to support the GA of the Agent operator that will manage install and upgrade.

### Acceptance Criteria

- [ ] Implementation of Agent Health Reporting:  [spec](https://source.datanerd.us/agents/agent-specs/blob/main/super_agent/Health-Check.md)
- [ ] Testing/Confirmation of behavior when there is conflicting APM instrumentation (i.e. the agent is already deployed, and the operator injects an agent ‘on top’)

### Design Consideration/Limitations
N/A

### Dependencies

- K8s Agent Operator is GA’ed

### Additional context
- DACI: [K8s Monitor APM Agents Health](https://newrelic.atlassian.net/wiki/spaces/INST/pages/3526296055/DACI+K8s+Monitor+APM+Agents+Health)
- [Super Agent - APM k8s-based CDD](https://newrelic.atlassian.net/wiki/spaces/INST/pages/3395715073/Super+Agent+-+APM+k8s-based+CDD)

kaylareopelle:
**Feature: Add health checks when the agent runs within Agent Control**

When the agent is started within an [Agent Control](https://docs-preview.newrelic.com/docs/new-relic-agent-control) environment, a health check file will be created at the configured location for every agent process. By default, this location is: '/newrelic/apm/health'. The health check files will be updated at the configured frequency, which defaults to every five seconds.

Resolves #2859


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
   The agent now supports Ruby 3.4.0. We've made incremental changes throughout the preview stage to reach compatibility. This release includes an update to the Thread Profiler for compatibility with Ruby 3.4.0's new backtrace format. [Issue#2992](https://github.com/newrelic/newrelic-ruby-agent/issues/2992) [PR#2997](https://github.com/newrelic/newrelic-ruby-agent/pull/2997)
 
+- **Feature: Add health checks when the agent runs within Agent Control**
+
+  When the agent is started with a within an agent control environment, automatic health check files will be created within the configured file destination at the configured frequency. [PR#2995](https://github.com/newrelic/newrelic-ruby-agent/pull/2995)
```
kaylareopelle, line 11: Waiting to hear if there's a product-approved changelog entry, so this may need to be updated.

lib/new_relic/agent/agent_helpers/connect.rb
```diff
@@ -198,6 +198,7 @@ def connect(options = {})
         rescue NewRelic::Agent::UnrecoverableAgentException => e
           handle_unrecoverable_agent_error(e)
         rescue StandardError, Timeout::Error, NewRelic::Agent::ServerConnectionException => e
+          NewRelic::Agent.agent.health_check.update_status(NewRelic::Agent::HealthCheck::FAILED_TO_CONNECT)
```
kaylareopelle, line 201: I've struggled to write tests to verify the status is updated when we expect it to be updated. If we think tests are valuable for these updates, I could use some help!

lib/new_relic/agent/health_check.rb
```diff
@@ -0,0 +1,125 @@
...
+module NewRelic
+  module Agent
+    class HealthCheck
+      def initialize
+        @start_time = nano_time
+        @fleet_id = ENV['NEW_RELIC_AGENT_CONTROL_FLEET_ID']
+        # The spec states file paths for the delivery location will begin with file://
+        # This does not create a valid path in Ruby, so remove the prefix when present
+        @delivery_location = ENV['NEW_RELIC_AGENT_CONTROL_HEALTH_DELIVERY_LOCATION']&.gsub('file://', '')
+        @frequency = ENV['NEW_RELIC_AGENT_CONTROL_HEALTH_FREQUENCY'].to_i
```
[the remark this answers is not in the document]
tannalynn: Yeah, i think doing the default like this makes sense since we can't use our normal config stuff to handle the default value

lib/new_relic/agent/health_check.rb
```diff
@@ -0,0 +1,125 @@
...
+      end
+
+      HEALTHY = {healthy: true, last_error: 'NR-APM-000', message: 'Healthy'}
+      INVALID_LICENSE_KEY = {healthy: false, last_error: 'NR-APM-001', message: 'Invalid liense key (HTTP status code 401)'}
+      MISSING_LICENSE_KEY = {healthy: false, last_error: 'NR-APM-002', message: 'License key missing in configuration'}
+      FORCED_DISCONNECT = {healthy: false, last_error: 'NR-APM-003', message: 'Forced disconnect received from New Relic (HTTP status code 410)'}
+      HTTP_ERROR = {healthy: false, last_error: 'NR-APM-004', message: 'HTTP error response code [%s] recevied from New Relic while sending data type [%s]'}
+      MISSING_APP_NAME = {healthy: false, last_error: 'NR-APM-005', message: 'Missing application name in agent configuration'}
+      APP_NAME_EXCEEDED = {healthy: false, last_error: 'NR-APM-006', message: 'The maximum number of configured app names (3) exceeded'}
+      PROXY_CONFIG_ERROR = {healthy: false, last_error: 'NR-APM-007', message: 'HTTP Proxy configuration error; response code [%s]'}
```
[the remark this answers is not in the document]
tannalynn: I think it makes sense to leave them, to be complete. Plus, if we ever update the agent to handle those things, then these will already be here ready to use. And if we never do make those updates, at least in the future when we're comparing the code to the spec to look into something we won't be confused by why some would be missing.

lib/new_relic/agent/health_check.rb
```diff
@@ -0,0 +1,125 @@
...
+        # The spec states file paths for the delivery location will begin with file://
+        # This does not create a valid path in Ruby, so remove the prefix when present
+        @delivery_location = ENV['NEW_RELIC_AGENT_CONTROL_HEALTH_DELIVERY_LOCATION']&.gsub('file://', '')
+        @frequency = ENV['NEW_RELIC_AGENT_CONTROL_HEALTH_FREQUENCY'].to_i
+        @continue = true
+        @status = HEALTHY
+      end
+
+      HEALTHY = {healthy: true, last_error: 'NR-APM-000', message: 'Healthy'}
+      INVALID_LICENSE_KEY = {healthy: false, last_error: 'NR-APM-001', message: 'Invalid liense key (HTTP status code 401)'}
```
chynesNR, line 20: Are any of these errors recoverable for the Ruby Agent? i.e. where we could potentially need to set the status back to `HEALTHY`?

kaylareopelle: Great question! Some of these errors are recoverable. From what I'm seeing in the code, a successful HTTP request made by the `NewRelicService` class to an external endpoint would erase the error state. As of now, this is the only place where I'll update the status to `HEALTHY`: 
https://github.com/newrelic/newrelic-ruby-agent/pull/2995/files#diff-ecdbe8dde33106ec28a206dcdbfca4ea2c647caff89504a82c58dbceaf749e9aR644 

However, I may have missed some scenarios and there could be other places where explicitly setting a `HEALTHY` status would demonstrate recovery.

lib/new_relic/agent/health_check.rb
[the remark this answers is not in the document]
kaylareopelle: It turns out that this was a requirement to get the health checks to work in the test sidecar environment, so it's been updated. Thank you!

lib/new_relic/agent/health_check.rb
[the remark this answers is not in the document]
kaylareopelle: Great! Do you think we would benefit from a comment stating they're not called anywhere?

lib/new_relic/agent/health_check.rb
```diff
@@ -0,0 +1,125 @@
...
+      PROXY_CONFIG_ERROR = {healthy: false, last_error: 'NR-APM-007', message: 'HTTP Proxy configuration error; response code [%s]'}.freeze
+      AGENT_DISABLED = {healthy: false, last_error: 'NR-APM-008', message: 'Agent is disabled via configuration'}.freeze
+      FAILED_TO_CONNECT = {healthy: false, last_error: 'NR-APM-009', message: 'Failed to connect to New Relic data collector'}.freeze
+      FAILED_TO_PARSE_CONFIG = {healthy: false, last_error: 'NR-APM-010', message: 'Agent config file is not able to be parsed'}.freeze
+      SHUTDOWN = {healthy: true, last_error: 'NR-APM-099', message: 'Agent has shutdown'}.freeze
+
+      def create_and_run_health_check_loop
+        unless health_check_enabled?
+          @continue = false
+        end
```
kaylareopelle, line 35: Some of the tests benefit from `@continue = false` being set only when the health check loop starts. I'm not sure if we should turn off the checks here, or on initialization. Thoughts?

We could also add the logger messages directly to initialize if we move `@continue = false`.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2188,6 +2188,30 @@ def self.notify
           :transform => DefaultSource.method(:convert_to_constant_list),
           :description => 'Specify a list of exceptions you do not want the agent to strip when [strip_exception_messages](#strip_exception_messages-enabled) is `true`. Separate exceptions with a comma. For example, `"ImportantException,PreserveMessageException"`.'
         },
+        # Agent Control
+        :'agent_control.fleet_id' => {
+          :default => nil,
+          :allow_nil => true,
+          :public => true,
```
tannalynn, line 2195: if these new configs are something that will be set by fleet control, we don't want customers setting it themselves right? Should they all be set to not public then?

tannalynn: Also it looks like these are all env variable only, so maybe we should add that to the description?

lib/new_relic/agent/health_check.rb
tannalynn: so health_check_enabled? checks to make sure fleet_id, delivery_location_and frequency are all valid. but on the lines after this we check each on individually and return if it's not. So setting continue here would be more for the benefit of other methods that are outside of this one. from what i can tell, this is only update_status, correct? 
Is there a situation outside of the tests that would be affected by continue being false at the start and then set to true for this method? Is it possible for it to start out as false, some update_status thing call happens which just returns and doesn't save the info, and then this loop starts but we've lost the info that we dropped when update_status was called? 
Not sure if that situation is possible, but i hope my question makes sense.

lib/new_relic/agent/health_check.rb
```diff
@@ -0,0 +1,127 @@
...
+            rescue StandardError => e
+              NewRelic::Agent.logger.error("Aborting agent control health check. Error raised: #{e}")
+              @continue = false
+            end
+          end
+        end
+      end
+
+      def update_status(status, options = [])
+        puts '***** update_status AGENT_DISABLED *******' if status == AGENT_DISABLED
```
tannalynn, line 59: leftover puts

lib/new_relic/agent/health_check.rb
```diff
@@ -0,0 +1,127 @@
...
+        return unless @continue
+
+        puts '***** update_status @continue = true *****' if status == AGENT_DISABLED
```
tannalynn, line 62: leftover puts

test/new_relic/agent/agent/start_test.rb
```diff
@@ -43,25 +44,57 @@ def test_disabled_negative
 
   def test_check_config_and_start_agent_disabled
     self.expects(:monitoring?).returns(false)
+    @health_check.expects(:create_and_run_health_check_loop)
     check_config_and_start_agent
   end
 
+  def test_monitoring_false_updates_health_status
+    puts '*************** monitoring false test ***************'
+    puts '*************** TEST START ******************'
```
tannalynn, line 53: leftover puts

test/new_relic/agent/agent/start_test.rb
```diff
@@ -43,25 +44,57 @@ def test_disabled_negative
...
+    with_config(:monitoring => false) do
+      puts NewRelic::Agent.agent.health_check.inspect
+      # make sure the health checks are set up to run
+      NewRelic::Agent.agent.health_check.instance_variable_set(:@continue, true)
+
+      NewRelic::Agent.agent.health_check.expects(:update_status).with(NewRelic::Agent::HealthCheck::AGENT_DISABLED)
+      monitoring?
+      puts '************ TEST END *********************'
```
tannalynn, line 61: leftover puts

tannalynn (commented):
couple of questions and noticed some puts, wanted to comment on them just to make sure they dont get forgotten in the future.

lib/new_relic/agent/health_check.rb
```diff
@@ -0,0 +1,127 @@
...
+        # The spec states file paths for the delivery location will begin with file://
+        # This does not create a valid path in Ruby, so remove the prefix when present
+        @delivery_location = ENV['NEW_RELIC_AGENT_CONTROL_HEALTH_DELIVERY_LOCATION']&.gsub('file://', '')
+        @frequency = ENV['NEW_RELIC_AGENT_CONTROL_HEALTH_FREQUENCY'] ? ENV['NEW_RELIC_AGENT_CONTROL_HEALTH_FREQUENCY'].to_i : 5
+        @continue = true
+        @status = HEALTHY
+      end
+
+      HEALTHY = {healthy: true, last_error: 'NR-APM-000', message: 'Healthy'}.freeze
+      INVALID_LICENSE_KEY = {healthy: false, last_error: 'NR-APM-001', message: 'Invalid liense key (HTTP status code 401)'}.freeze
```
hannahramadan, line 20:
```suggestion
      INVALID_LICENSE_KEY = {healthy: false, last_error: 'NR-APM-001', message: 'Invalid license key (HTTP status code 401)'}.freeze
```

lib/new_relic/agent/health_check.rb
kaylareopelle: > so health_check_enabled? checks to make sure fleet_id, delivery_location_and frequency are all valid. but on the lines after this we check each on individually and return if it's not. So setting continue here would be more for the benefit of other methods that are outside of this one. from what i can tell, this is only update_status, correct?

correct!

> Is there a situation outside of the tests that would be affected by continue being false at the start and then set to true for this method? Is it possible for it to start out as false, some update_status thing call happens which just returns and doesn't save the info, and then this loop starts but we've lost the info that we dropped when update_status was called?
Not sure if that situation is possible, but i hope my question makes sense.

I think I understand what you're asking. I believe I've moved the call to `create_and_run_health_check_loop` to a place high enough that this shouldn't happen, but since that isn't necessarily clear, it seems safer to set `@continue = false` during initialization if the attributes that enable health checks aren't valid.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
   The agent now supports Ruby 3.4.0. We've made incremental changes throughout the preview stage to reach compatibility. This release includes an update to the Thread Profiler for compatibility with Ruby 3.4.0's new backtrace format. [Issue#2992](https://github.com/newrelic/newrelic-ruby-agent/issues/2992) [PR#2997](https://github.com/newrelic/newrelic-ruby-agent/pull/2997)
 
+- **Feature: Add health checks when the agent runs within Agent Control**
+
+  When the agent is started within an agent control environment, a health check file will be created at the configured file destination for every agent process. The health check files will be updated at the configured frequency. [PR#2995](https://github.com/newrelic/newrelic-ruby-agent/pull/2995)
```
kaylareopelle, line 11: Returning comment: validate this is okay with product; see if there's a doc we can link to about agent control

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: @tannalynn - How does this rewording sound for the last sentence of each of these configs? 

`This configuration is set by agent control, or one of its components, as an environment variable prior to agent startup.`

lib/new_relic/agent/health_check.rb
kaylareopelle: @tannalynn I updated the code to set `@continue = false` on initialization if the env vars are missing or invalid. 

a7e87cb753d1475830629ad5d0aca1dff2b1e147

lib/new_relic/agent/health_check.rb
```diff
@@ -0,0 +1,153 @@
...
+      def write_file
+        @file ||= "#{create_file_path}/#{file_name}"
+
+        File.write(@file, contents)
+      rescue StandardError => e
+        NewRelic::Agent.logger.error("Agent control health check raised an error while writing a file: #{e}")
+        @continue = false
+      end
+
+      def create_file_path
```
kaylareopelle, line 124: Keep an eye on the spec for this; we may not need to create the path if it doesn't exist.

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: I've updated the configs to reference the environment variable. I'm going to resolve this, but feel free to reopen if you'd like them worded in a different way.

CHANGELOG.md
kaylareopelle: @hannahramadan - Updated in a1bcd2a39ea8ca50dd026cf4cf92bddb85cb0753

CHANGELOG.md
kaylareopelle: ### @hannahramadan - updated in a1bcd2a39ea8ca50dd026cf4cf92bddb85cb0753

kaylareopelle: 😂 missed the ### there

lib/new_relic/agent/health_check.rb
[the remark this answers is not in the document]
kaylareopelle: @tannalynn - I'm going to resolve this without a comment, but lmk if you want it changed!

lib/new_relic/agent/agent_helpers/shutdown.rb
```diff
@@ -19,6 +19,9 @@ def shutdown
           revert_to_default_configuration
 
           @started = nil
+
+          NewRelic::Agent.agent.health_check.update_status(NewRelic::Agent::HealthCheck::SHUTDOWN) if NewRelic::Agent.agent.health_check.healthy?
```
hannahramadan, line 23: I know this all works so I imagine I'm reading it wrong, but I'm seeing "Update status to SHUTDOWN if the status is HEALTHY". I would've thought it would shutdown unless healthy?

kreopelle: Good question! I believe the spec outlines this behavior, and that SHUTDOWN should only be set if the status is healthy and the agent is shutting down. Otherwise, the final status should be whatever unhealthy status was last reported. Would you mind checking out the spec section on the error codes to see if you come to the same conclusion?

chynesNR: Yep, that's correct. The idea is to indicate that even though the agent isn't running right now, that's fine, because it shut down cleanly with the application and didn't crash.

hannahramadan: Yup @kaylareopelle that is totally outlined in the spec! Thank you both for the explanation here.

lib/new_relic/agent/health_check.rb
```diff
@@ -0,0 +1,136 @@
...
+        end
+      end
+
+      def set_delivery_location
+        @delivery_location = if ENV['NEW_RELIC_AGENT_CONTROL_HEALTH_DELIVERY_LOCATION']
+          # The spec states file paths for the delivery location will begin with file://
+          # This does not create a valid path in Ruby, so remove the prefix when present
+          ENV['NEW_RELIC_AGENT_CONTROL_HEALTH_DELIVERY_LOCATION']&.gsub('file://', '')
+        else
+          # The spec default is 'file:///newrelic/apm/health', but since we're just going to remove it anyway...
```
kaylareopelle, line 80:
```suggestion
          # The spec default is 'file:///newrelic/apm/health', but since we're just going to remove the prefix anyway...
```

hannahramadan (approved):
🚀

tannalynn (approved):

kaylareopelle:
I'm going to keep the branch around for AC testing until we've released the agent.

PR #2997, opened 2024-12-17
Thread Profile / Backtrace 3.4.0 fixes

### LINKED ISSUE #2992: Update `BacktraceNode` and `ThreadProfile` classes for Ruby 3.4
The `BacktraceNode` class has a regex that relies on the pre-Ruby 3.4 backtrace syntax: 

https://github.com/newrelic/newrelic-ruby-agent/blob/92a012836440ea93ed11ccad93807f26099a9184/lib/new_relic/agent/threading/backtrace_node.rb#L126-L129

In Ruby 3.4, this syntax will remove the backtick character and replace it with a single quotation mark. The content within the single quotation mark will become the Class name combined with the method name (ex. `BacktraceNode#parse_backtrace_frame`). 

The new syntax is not compatible with our existing regex.

Update the regex for Ruby 3.4+ compatibility and conditionally apply the regex based on Ruby version.

kaylareopelle:
Closes #2992

CI, passing on the relevant runs: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/12403980066

### DIFF
--- a/test/new_relic/agent/threading/backtrace_node_test.rb
+++ b/test/new_relic/agent/threading/backtrace_node_test.rb
@@ -228,7 +228,6 @@ module NewRelic::Agent::Threading
     end
 
     def test_aggregate_limits_recorded_depth
-
       deep_backtrace = (0..2000).to_a.map { |i| ruby_3_4_0_or_above? ? "foo.rb:#{i}:in 'Foo#foo'" : "foo.rb:#{i}:in `foo'" }
 
       root = BacktraceRoot.new
--- a/test/new_relic/agent/threading/backtrace_node_test.rb
+++ b/test/new_relic/agent/threading/backtrace_node_test.rb
@@ -7,15 +7,22 @@ require 'new_relic/agent/threading/backtrace_node'
 
 module NewRelic::Agent::Threading
   class BacktraceNodeTest < Minitest::Test
-    SINGLE_LINE = "irb.rb:69:in `catch'"
-
     def setup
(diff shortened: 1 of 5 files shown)


test/new_relic/agent/threading/backtrace_node_test.rb
```diff
@@ -7,15 +7,22 @@
 
 module NewRelic::Agent::Threading
   class BacktraceNodeTest < Minitest::Test
-    SINGLE_LINE = "irb.rb:69:in `catch'"
-
     def setup
+      @single_line = ruby_3_4_0_or_above? ? "irb.rb:69:in 'Kernel#catch'" : "irb.rb:69:in `catch'"
```
hannahramadan, line 11: fun instance variable setup!

hannahramadan (approved):
Looks great thank you! For tests that use the new methods `ruby_3_4_0_or_above?`, do you think we need to add TODO to remove when 3.3 is dropped similar to the code comment or should the code TODO be enough of a remind to go check out the tests

kaylareopelle:
> Looks great thank you! For tests that use the new methods `ruby_3_4_0_or_above?`, do you think we need to add TODO to remove when 3.3 is dropped similar to the code comment or should the code TODO be enough of a remind to go check out the tests

Great question! If it seems like a TODO would be helpful, I think it's worth including. They're tests that we don't look at often, so I don't think it'd be very bothersome.

hannahramadan (approved):

tannalynn (approved):

PR #3003, opened 2024-12-23
Remove version lock on haml in Rails multiverse

kaylareopelle:
CI with the other Rails versions passing also: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/12519740439 
(timeouts though for some of the `rest` suite)


hannahramadan (approved):
Thank you!

PR #3006, opened 2025-01-03
🎁 Run tests on Ruby 3.4.1

kaylareopelle:
Ruby 3.4.1 is out! This PR updates the CI to test the newest version and replaces the areas where we used Ruby 3.3 as our default version to Ruby 3.4.

Workflow dispatches to check: 
* Perf tests: ~~https://github.com/newrelic/newrelic-ruby-agent/actions/runs/12590713430~~ 
  * ~~https://github.com/newrelic/newrelic-ruby-agent/actions/runs/12590853852~~ (failing with base64 error)
  * https://github.com/newrelic/newrelic-ruby-agent/actions/runs/12601064376 => failing with the same error, but this seems to be because it's trying to use 3.4 on `main`, which is still pointing to the old Rails version. Perf tests ran locally just fine on the current branch. I think we're safe to test.
* 🟢 URL test CI: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/12590710130 (as an example of a generic workflow with an overall Ruby version update)

Though other workflows were also updated, I don't want to trigger them. We'll have to keep an eye out for funny business the next time they run.

### DIFF
--- a/test/multiverse/suites/grpc/Envfile
+++ b/test/multiverse/suites/grpc/Envfile
@@ -8,9 +8,9 @@ end
 
 instrumentation_methods(:chain, :prepend)
 
-# TODO: permit testing of the nil (latest) version against Ruby 3.3+
+# TODO: permit testing of the nil (latest) version against Ruby 3.4+
 GRPC_VERSIONS = [
-  [nil, 2.6, 3.2],
+  [nil, 2.6],
   ['1.48.0', 2.5, 3.1]
 ]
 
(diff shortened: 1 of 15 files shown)


test/multiverse/suites/grpc/Envfile
```diff
@@ -8,9 +8,9 @@ end
 
 instrumentation_methods(:chain, :prepend)
 
-# TODO: permit testing of the nil (latest) version against Ruby 3.3+
+# TODO: permit testing of the nil (latest) version against Ruby 3.4+
 GRPC_VERSIONS = [
-  [nil, 2.6, 3.2],
+  [nil, 2.6],
```
kaylareopelle, line 13: Will all new Rubies work? Let's see...

kaylareopelle: nope, they timeout. need to revert.

PR #3013, opened 2025-01-13
Use permalink for Logger::Severity

kaylareopelle:
Ruby moved the logger code out of the ruby/ruby repository. This change uses the ruby/logger repository and also uses a permalink. 

I imagine the newrelic.yml changes will be updated again whenever we release, so I just added the permalink to the existing setup to get the URL CI to pass.

Passing URL CI: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/12755313559 🟢


tannalynn (approved):
thank you for fixing this!

Issue #3004, opened 2024-12-27
Scheduled CI times out on various multiverse groups intermittently


kaylareopelle:
Our Scheduled Continuous Integration workflow fails various multiverse groups intermittently. When a job times out, you are unable to see the log output, which makes it difficult to decode what suites may be causing the problem.

This is the error:
```
The hosted runner: GitHub Actions #{action_number} lost communication with the server. 
Anything in your workflow that terminates the runner process, starves it for CPU/Memory, 
or blocks its network access can cause this error.
```

The jobs seem to time out after 45-50 minutes. 

Failed groups/Ruby verisons include:
* [`database`, 2.7.8](https://github.com/newrelic/newrelic-ruby-agent/actions/runs/12514123458/job/34909773097)
* [`background_2`, 3.0.6](https://github.com/newrelic/newrelic-ruby-agent/actions/runs/12514123458/job/34909772045)
* [`rest`, 3.2.6](https://github.com/newrelic/newrelic-ruby-agent/actions/runs/12443890548/job/34743617855) -- this suite also failed recently without [a timeout on the `bunny` suite](https://github.com/newrelic/newrelic-ruby-agent/actions/runs/12479546397/job/34828937537#step:12:972)
* [`rest`, 3.0.7](https://github.com/newrelic/newrelic-ruby-agent/actions/runs/12443890548/job/34743617574)

kaylareopelle:
This may be a solution: https://github.com/actions/runner/issues/3539#issuecomment-2553720294

PR #3016, opened 2025-01-14
Update multiverse jobs in CI to avoid resource issues in github actions

tannalynn:
So, this does seem to work to prevent the github communication errors/running out of resources thing from happening, yay!

scheduled CI run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/12772900369
I saw a couple of random actual intermittent test failures, like a timeout or openai complaining about not having a token, BUT I haven't seen ANY github communication errors come up on there and I tried rerunning it several times to give it as many chances to fail as I could. 

I don't love how much bigger the files are now and all the duplication I had to add. 
But it's better to have an annoying file than have github actions run out of resources all the time though, so ¯\\_(ツ)_/¯ oh well

Here is what changed for multiverse:

- multiverse: removed all services and runs most our multiverse like normal
- multiverse_services_1: elasticsearch, mongo, bunny
- multiverse_services_2: redis, sidekiq (bc sidekiq uses redis), memcache
- multiverse_services_kafka: rdkafka, ruby_kafka
- multiverse_services_mysql_pg: active_record, active_record_pg

I wanted mysql running with very few other things since that was mentioned in some of the comments we saw as being a problem. 
I also didn't want too many services running in the same one just because I wanted to minimize the number of services each group was running to avoid resource issues.
Also I wanted kafka on its own because we've seen some weird issues with rdkafka in the past so just wanted those alone in case those things continue happening.

resolves https://github.com/newrelic/newrelic-ruby-agent/issues/3004

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -201,17 +201,6 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 10
-      zookeeper:
-        image: bitnami/zookeeper
-        ports:
-          - 2181:2181
-        env:
-          ALLOW_ANONYMOUS_LOGIN: yes
-        options: >-
-          --health-cmd "echo mntr | nc -w 2 -q 2 localhost 2181"
-          --health-interval 10s
-          --health-timeout 5s
-          --health-retries 5
       mongodb:
         image: ${{ contains(fromJson('["2.4.10"]'), matrix.ruby-version) && 'mongo:5.0.11' || 'mongo:latest' }}
         ports:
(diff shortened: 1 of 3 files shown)


kaylareopelle (commented):
I'm so glad you fixed this problem! 🎉 🧶 

Seeing all the duplication, I wonder if we could use [a composite action](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action) to reduce some of the steps we write again and again? 

Here's the GHA doc on [reducing duplication](https://docs.github.com/en/actions/sharing-automations/avoiding-duplication) that taught me about this concept. 

Rather than push things up into a separate repo, I imagine we could just use our `actions` directory. 

I also understand if that refactor seems like too much for now. We could create a ticket and add it to the backlog if you prefer to leave things as-is.

tannalynn:
> I'm so glad you fixed this problem! 🎉 🧶
> 
> Seeing all the duplication, I wonder if we could use [a composite action](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action) to reduce some of the steps we write again and again?
> 
> Here's the GHA doc on [reducing duplication](https://docs.github.com/en/actions/sharing-automations/avoiding-duplication) that taught me about this concept.
> 
> Rather than push things up into a separate repo, I imagine we could just use our `actions` directory.
> 
> I also understand if that refactor seems like too much for now. We could create a ticket and add it to the backlog if you prefer to leave things as-is.

I think looking into that would be a good idea, but I do think it would be better as a future improvement. It would be my preference to leave the duplication for now, just so things are working again at least. That way I can pick redshift back up again to make sure we don't fall behind there, since that's been tricky. 
But i do like that idea, i definitely don't love the duplication! lol.

kaylareopelle (approved):

kaylareopelle:
Sounds good! Let's create an issue after this PR is merged.

PR #3020, opened 2025-01-16
Explicitly require logger in suite.rb

kaylareopelle:
Suites related to Rails 6.0, 6.1 and 7.0 started to fail with an uninitialized constant error related to the Logger::Severity constant. This only fixes failures in the multiverse suite.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -15,6 +15,12 @@ $: << File.expand_path('../../lib', __FILE__)
 $: << File.expand_path('../../test', __FILE__)
 $:.uniq!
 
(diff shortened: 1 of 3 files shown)


tannalynn (approved):

test/test_helper.rb
```diff
@@ -49,6 +49,7 @@ module NewRelic; TEST = true; end unless defined? NewRelic::TEST
 
 # This is the public method recommended for plugin developers to share our
 # agent helpers. Use it so we don't accidentally break it.
+require 'logger'
```
kaylareopelle, line 52:
```suggestion
```

hannahramadan (approved):

Issue #3021, opened 2025-01-16
Race condition with net_http instrumentation on JRuby causing intermittent errors


NC-piercej:
## Description
We are seeing occasional errors logged in NewRelic that appear to be originating in the NewRelic gem code. Here is a sample backtrace:

```
NoMethodError: undefined method `add_request_headers' for nil:NilClass
…ms/newrelic_rpm-9.16.1/lib/new_relic/agent/instrumentation/net_http/instrumentation.rb:24:in `request_with_tracing'
…undle/gems/newrelic_rpm-9.16.1/lib/new_relic/agent/instrumentation/net_http/prepend.rb:15:in `request'
…local/bundle/gems/aws-sdk-core-3.214.0/lib/seahorse/client/net_http/connection_pool.rb:348:in `request'
    /usr/local/bundle/gems/aws-sdk-core-3.214.0/lib/seahorse/client/net_http/handler.rb:80:in `block in transmit'
    /usr/local/bundle/gems/aws-sdk-core-3.214.0/lib/seahorse/client/net_http/handler.rb:134:in `block in session'
…local/bundle/gems/aws-sdk-core-3.214.0/lib/seahorse/client/net_http/connection_pool.rb:106:in `session_for'
    /usr/local/bundle/gems/aws-sdk-core-3.214.0/lib/seahorse/client/net_http/handler.rb:129:in `session'
    /usr/local/bundle/gems/aws-sdk-core-3.214.0/lib/seahorse/client/net_http/handler.rb:77:in `transmit'
    /usr/local/bundle/gems/aws-sdk-core-3.214.0/lib/seahorse/client/net_http/handler.rb:46:in `block in call'
    /usr/local/bundle/gems/aws-sdk-core-3.214.0/lib/seahorse/client/net_http/handler.rb:206:in `block in span_wrapper'
        /usr/local/bundle/gems/aws-sdk-core-3.214.0/lib/aws-sdk-core/telemetry/no_op.rb:29:in `in_span'
    /usr/local/bundle/gems/aws-sdk-core-3.214.0/lib/seahorse/client/net_http/handler.rb:202:in `span_wrapper'
    /usr/local/bundle/gems/aws-sdk-core-3.214.0/lib/seahorse/client/net_http/handler.rb:45:in `call'
…r/local/bundle/gems/aws-sdk-core-3.214.0/lib/seahorse/client/plugins/content_length.rb:24:in `call'
[...]

PR #3046, opened 2025-01-30
Add safe navigation in case segment is nil in net http instrumentation

tannalynn:
related to https://github.com/newrelic/newrelic-ruby-agent/issues/3021

This isn't addressing the root cause of the issue yet, as that's still being investigated, but this change should prevent the primary reported error from occurring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,9 @@
 
 ## dev
 
-
 - **Feature: Add elasticsearch.capture_cluster_name configuration option**
 
-    A new configuration option, `elasticsearch.capture_cluster_name`, has been added to control capturing Elasticsearch cluster names. Cluster names are captured by default, but can now be disabled as needed. [PR#3038](https://github.com/newrelic/newrelic-ruby-agent/pull/3038)
+  A new configuration option, `elasticsearch.capture_cluster_name`, has been added to control capturing Elasticsearch cluster names. Cluster names are captured by default, but can now be disabled as needed. [PR#3038](https://github.com/newrelic/newrelic-ruby-agent/pull/3038)
 
 - **Bugfix: Prevent a `nil` segment from causing errors in Net::HTTP instrumentation**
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+- **Bugfix: Prevent a nil segment from causing errors in Net HTTP instrumentation**
```
hannahramadan, line 5:
```suggestion
- **Bugfix: Prevent a `nil` segment from causing errors in Net::HTTP instrumentation**
```

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+
+  When using JRuby, a race condition can happen that causes the segment creation to fail and returns nil. This would cause an error to occur when methods were later called on the nil segment. These methods will not be called now if it is nil, preventing that error from occurring. [PR#3046](https://github.com/newrelic/newrelic-ruby-agent/pull/3046)
```
hannahramadan, line 7:
```suggestion
  When using JRuby, a race condition can happen that causes the segment creation to fail and return `nil`. This would cause an error to occur when methods were later called on the `nil` segment. These methods will no longer be called if the segment is `nil`, preventing that error from occurring. [PR#3046](https://github.com/newrelic/newrelic-ruby-agent/pull/3046)
```

hannahramadan (dismissed):
yeehaw

hannahramadan (approved):

CHANGELOG.md
```diff
@@ -2,10 +2,16 @@
 
 ## dev
 
+
```
kaylareopelle, line 5: Do we normally have two lines below `##dev`?

kaylareopelle (approved):
Small thing, but good to merge w/o addressing.

CHANGELOG.md
tannalynn: im not sure lol. it happened when i was resolving a merge conflict and i just didnt go back and remove it. I have changed it now though lol

kaylareopelle (approved):

PR #3048, opened 2025-02-07
Prepend NR middleware to Sidekiq middleware chain

### LINKED ISSUE #3037: Spurious Sidekiq::OverLimit exceptions
## Description
This is really the same issue that I raised years back with Appsignal: https://github.com/appsignal/appsignal-ruby/issues/526

The concurrent rate limiters in `sidekiq-enterprise` raise `Sidekiq::OverLimit` when a lock cannot be acquired within a certain amount of time. (See https://github.com/sidekiq/sidekiq/wiki/Ent-Rate-Limiting#concurrent for details.) Sidekq's own middleware rescues this exception and re-enqueues the job to be run later -- unless the job has exceeded its maximum number of retries, in which case the `OverLimit` exception is allowed to escape the middleware.

When Sidekiq re-enqueues the job in response to an `OverLimit`, this isn't an error, and we'd prefer it not show up as one in newrelic. On the other hand, if the exception escapes Sidekiq's middleware (i.e., the number of retries has been exceeded), we'd like to see it.

As with appsignal, the behavior would seem to be a function of the order of sidekiq's middleware vis-a-vis newrelic's. We could reorder the middleware chain on our end, but it would be a good idea for newrelic to ensure that its middleware sits earlier in the chain than sidekiq's own.

## Expected Behavior
When `Sidekiq::OverLimit` is raised by a rate limiter, unless Sidekiq's own middleware re-raises the exception, it should not show up in newrelic as an error.

## Steps to Reproduce
To reproduce, you need to create contention for a sidekiq concurrent rate limiter. You could create a concurrent `Sidekiq::Limiter` with concurrency 1 (i.e., a mutex) and a short `wait_timeout`. Have two processes/threads run concurrently, trying to acquire the lock and, with it, sleep for longer than the timeout. The one that loses the race to acquire the lock should raise `Sidekiq::OverLimit`.

## Your Environment
newrelic_rpm 9.16.1
ruby 3.3.6
rails 8.0.1
sidekiq-ent 7.3.4

hannahramadan:
Prepend the agent to Sidekiq's middleware chain. This stops Sidekiq-resolved exceptions from being reported to New Relic. 

We have a playground app that reproduces this issue. The introduction of this code successfully stops the error from being reported.

Closes #3037

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@
   When using JRuby, a race condition can happen that causes the segment creation to fail and return `nil`. This would cause an error to occur when methods were later called on the `nil` segment. These methods will no longer be called if the segment is `nil`, preventing that error from occurring. [PR#3046](https://github.com/newrelic/newrelic-ruby-agent/pull/3046)
 
 - **Bugfix: Stop reporting rescued Sidekiq::OverLimit exceptions**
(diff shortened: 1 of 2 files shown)


kaylareopelle (commented):
Thanks, @hannahramadan!

tannalynn (approved):

PR #3059, opened 2025-02-27
Move Slack notifications up 2 hours

hannahramadan:
Let's have slack notifications for new gem releases run an hour before our CI. These should now run at about midnight and the CI kicks off at 1am PST.


kaylareopelle (approved):
💡

PR #3064, opened 2025-03-04
Add supportability metric for agent version

### LINKED ISSUE #2627: Add supportability metric for agent version
When a customer reports an issue in a support case, a troubleshooting step I take when supporting the .NET agent is to go the application's Metric Explorer view in order to check if the customer's issue might be related to an agent version change in the past. 

For example, the .NET agent has a supportability metric that looks like this: 
`Supportability/AgentVersion/[AGENT VERSION]`

Currently an application's environment page only displays the current agent version being reported, and there is not an easy way to see past agent versions and when an agent version was changed. 

It is helpful to have evidence of an agent version change around the time an issue is described as happening, because sometimes customers are not aware that an agent change happened. 

### Feature Description
Add a supportability metric for the agent version, which can be viewed in the Metric Explorer and used for as a resource when troubleshooting customer issues. 

### Priority
Nice to Have

kaylareopelle:
Proposed new metric name: `Supportability/AgentVersion/newrelic_rpm/#{NewRelic::VERSION::STRING}`

Closes #2627

### DIFF
--- a/lib/new_relic/control/instance_methods.rb
+++ b/lib/new_relic/control/instance_methods.rb
@@ -74,6 +74,7 @@ module NewRelic
         NewRelic::Agent.agent = NewRelic::Agent::Agent.instance
         init_instrumentation
         init_security_agent
+        report_agent_version_metric
       end
 
       def determine_env(options)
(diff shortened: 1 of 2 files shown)


hannahramadan (approved):
love it + the metric name is perf

tannalynn (approved):

PR #3082, opened 2025-03-12
fix the regex used for deciding to require rackup/handler

tannalynn:
so the rack problem we had was because there was simply a 2 in the version number.
For deciding when to install rackup/handler, we were using the regex `/^1|2/` which checks to see if it starts with a 1 OR 2, BUT what it's actually checking is if (starts with a 1) OR 2. So just a 2 hanging out anywhere in the version will be like, yeah that's matching. 
So i changed it to `/^1|^2/` which is just starts with a 1 OR starts with a 2. which is what we actually intended originally.

### DIFF
--- a/test/new_relic/fake_server.rb
+++ b/test/new_relic/fake_server.rb
@@ -5,7 +5,7 @@
 require 'webrick'
 require 'webrick/https'
 require 'rack'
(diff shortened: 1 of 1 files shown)


kaylareopelle (approved):
Great find! Thank you!

PR #3094, opened 2025-03-25
Parallelize vendor metadata requests

### LINKED ISSUE #2845: Optimisation: agent start-up time for cloud environment detection could be improved
The agent performs 3 linear checks - each with a 1 second timeout - to test for the contexts of AWS, GCP, and Azure. For agent instances outside of those contexts, start-up time could be waiting for 3 seconds.

We should investigate how to optimise this start-up delay.

Perhaps:

- We could multithread the 3 checks
- We could better enable customer control, through config and/or docs, over specifying their cloud environment to skip the others

tannalynn:
This updated our checks for vendor metadata to be run in parallel to decrease how much time it takes to get this information.
It will wait until one successful request OR until all finish unsuccessfully 

closes #2845 (the other idea mentioned in the issue is something we are already doing)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
   Sidekiq delay extensions were removed from Sidekiq in 7.x and are now avaliable through the [sidekiq-delay_extensions](https://rubygems.org/gems/sidekiq-delay_extensions) gem. Thanks to [@sobrinho](https://github.com/sobrinho), the agent now has continued support for delay extensions.[PR#3056](https://github.com/newrelic/newrelic-ruby-agent/pull/3056)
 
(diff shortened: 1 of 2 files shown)


kaylareopelle (commented):
Thank you, @tannalynn! One question -- Do you think the existing tests cover this new logic?

hannahramadan (approved):

tannalynn:
> Thank you, @tannalynn! One question -- Do you think the existing tests cover this new logic?

Yeah it does. When i had something mispelled and had issues when i was working on it, it made some tests fail so I think we're good with the coverage from the existing tests

kaylareopelle (approved):

PR #3096, opened 2025-03-25
Set hostname for logs using config

### LINKED ISSUE #1696: Logger does not use `process_host.display_name` when setting `hostname` in the logs
The logger does not use `process_host.display_name` when setting `hostname` in the logs. Instead it uses `Socket.gethostname` via a call to `NewRelic::Agent::Hostname.get`. 

I believe it is a bug. (In another issue, a New Relic developer assumes the feature already acts this way when she says  "Alternatively, we have the ability to set the app name (entity.name in the default attributes) and hostname by configuration in either the newrelic.yml file or by environment variable"). https://github.com/newrelic/newrelic-ruby-agent/issues/1141

You can see the offending code here:
https://github.com/newrelic/newrelic-ruby-agent/blob/74737a46a141d1d7e251c06f8db157743b02f3d8/lib/new_relic/agent/logging.rb#L51

I think the fix is to just replace with:
`Agent.config[:'process_host.display_name']`

I can create a PR if that would be helpful. 

Thanks!

kaylareopelle:
The default value of 'process_host.display_name' is the same as NewRelic::Agent::Hostname.get, so this change will only impact users who have set a unique value on the config.

Resolves #1696

**PENDING MERGE: Need to finish cross-agent discussion**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,16 @@
 
   Sidekiq delay extensions were removed from Sidekiq in 7.x and are now avaliable through the [sidekiq-delay_extensions](https://rubygems.org/gems/sidekiq-delay_extensions) gem. Thanks to [@sobrinho](https://github.com/sobrinho), the agent now has continued support for delay extensions.[PR#3056](https://github.com/newrelic/newrelic-ruby-agent/pull/3056)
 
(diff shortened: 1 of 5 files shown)


kaylareopelle:
Closing, as we're taking a different approach to introduce this value.

Issue #3077, opened 2025-03-10
W3C Sampling Flag Support


chynesNR:
"We have a request to play nicer with OTel agents in regards to sampling. To that end, we'd like to respect the inbound traceparent's sampled flag."

Details are in [this Agent spec](https://source.datanerd.us/agents/agent-specs/blob/main/distributed_tracing/Trace-Context-Payload.md).

PR #3135, opened 2025-04-16
W3C Sampling Flag Support

tannalynn:
Adds logic to support the new configuration options to modify how we decide sampling and priority based on incoming traceparent value. When in default mode, the existing logic of the agent remains unchanged. 

resolves https://github.com/newrelic/newrelic-ruby-agent/issues/3077

related agent specs
[Original PR](https://source.datanerd.us/agents/agent-specs/pull/726)
[Clarifying Configs](https://source.datanerd.us/agents/agent-specs/pull/729)

### DIFF
--- a/lib/new_relic/agent/transaction/trace_context.rb
+++ b/lib/new_relic/agent/transaction/trace_context.rb
@@ -136,16 +136,21 @@ module NewRelic
 
           transaction.distributed_tracer.parent_transaction_id = payload.transaction_id
 
-          unless payload.sampled.nil?
-            transaction.sampled = payload.sampled
-            transaction.priority = payload.priority if payload.priority
-          end
+          trace_context_sampled_priority(payload)
+
           NewRelic::Agent.increment_metric(ACCEPT_SUCCESS_METRIC)
           true
         rescue => e
           NewRelic::Agent.increment_metric(ACCEPT_EXCEPTION_METRIC)
           NewRelic::Agent.logger.warn('Failed to accept trace context payload', e)
           false
+        end
+
+        def trace_context_sampled_priority(payload)
+          unless payload.sampled.nil?
+            transaction.sampled = payload.sampled
+            transaction.priority = payload.priority if payload.priority
+          end
         end
 
         def ignore_trace_context?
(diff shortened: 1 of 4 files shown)


lib/new_relic/agent/transaction/trace_context.rb
```diff
@@ -148,6 +146,29 @@ def accept_trace_context(header_data)
           false
         end
 
+        def determine_sampling_decision(payload, header_data)
+          unless payload.sampled.nil?
+            if header_data.trace_parent['trace_flags'] == '01'
```
chynesNR, line 151: I believe [trace-flags is a bitmask](https://www.w3.org/TR/trace-context/#trace-flags), even though `01` is the only value that's currently valid. Do we want to future-proof it now, or not worry about it yet?

tannalynn: Yeah, I think that's a good question. The reason I decided to check if it's 01 and then just else was because if it's anything else (like some weird value or even some future unknown value), I figured falling back to the not sampled logic makes the most sense. Since any new value would require us to implement whatever logic is required for the new behavior tied to that value, we'll need to make changes anyways though. 

I'm open to swapping the logic and defaulting to assuming everything is sampled if it's not 00 if we think that would be better, I was just thinking that assuming not sampled would work out better, for max samples stored reasons, in most situations if we don't know what it is.

chynesNR:
The customer issue is resolved and no changes should be necessary for us. I’ve been asked to make sure we honor this bit from the spec:

> If there is no valid traceparent, then the legacy priority sampling logic will be maintained.

lib/new_relic/agent/transaction/trace_context.rb
tannalynn: Yeah, i think based on the latest thing from chris about the customer issue, with the "If there is no valid traceparent, then the legacy priority sampling logic will be maintained." thing, keeping it the way it is currently is the right call.

tannalynn: hmm or actually. i think i might make a change now that i'm thinking about it again

kaylareopelle: Is there a scenario where we'd receive a value in `header_data.trace_parent['trace_flags']` that wouldn't `==` to `01`, but should be marked as sampled? Would we ever receive `0x01` or something else?

PR #3172, opened 2025-05-02
Add support for ECS Fargate

tannalynn:
Adds support for ECS fargate and adds a new ecs vendor to record the ecsDockerId at startup.


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+- **Feature: Add ECS Docker ID for Fargate**
+
+  Previously, the ruby agent did not record the docker id when in an AWS ECS Fargate situation. The docker id will now be recorded correctly. [PR#3172](https://github.com/newrelic/newrelic-ruby-agent/pull/3172)
```
```suggestion
  Previously, the Ruby agent did not record the Docker ID when running in an AWS ECS Fargate environment. The Docker ID will now be recorded correctly. [PR#3172](https://github.com/newrelic/newrelic-ruby-agent/pull/3172)
```

lib/new_relic/agent/utilization/ecs.rb
```diff
@@ -0,0 +1,22 @@
...
+  module Agent
+    module Utilization
+      class ECS < Vendor
+        vendor_name 'ecs'
+        endpoint ENV['ECS_CONTAINER_METADATA_URI'] || ''
+        headers 'Metadata' => 'true'
+        keys %w[DockerId]
+
+        def transform_key(key)
+          'ecs' + key
```
kaylareopelle, line 17: Does this need to pass back up to the `transform_key` method in the parent class?

kaylareopelle (commented):
Thanks, Tanna! I have a few questions in the review. I haven't worked with the Utilization/Vendor part of the codebase, so they might be way off! :)

lib/new_relic/agent/utilization_data.rb
```diff
@@ -84,10 +86,32 @@ def to_collector_hash
         result
       end
 
+      def append_ecs_info(collector_hash)
+        return unless Agent.config[:'utilization.detect_aws']
```
kaylareopelle, line 90: I'm not sure I understand why we want to return early here. Is it because ECS is a service of AWS, so if someone specifies they're running in AWS, we won't get more specific?

lib/new_relic/agent/utilization_data.rb
```diff
@@ -84,10 +86,32 @@ def to_collector_hash
...
+
+        Thread.new do
+          # try v4 first, and only try unversioned endpoint if v4 fails
+          ecs = Utilization::ECSV4.new
+          if ecs.detect
+            collector_hash[:vendors] ||= {}
+            collector_hash[:vendors][:ecs] = ecs.metadata
+          else
+            ecs = Utilization::ECS.new
```
kaylareopelle, line 99: Is the unversioned endpoint for everything below V4 or above V4 or for something else entirely?

test/fixtures/utilization/aws/ecs.json
```diff
@@ -0,0 +1 @@
+{"DockerId":"1234567890987654321"}
```
kaylareopelle, line 1: Was this pulled in from the cross-agent tests?

test/multiverse/suites/agent_only/utilization_data_collection_test.rb
```diff
@@ -79,6 +79,38 @@ def test_omits_sending_vendor_data_on_connect_when_not_available
     assert_equal expected, single_connect_posted.utilization
   end
 
+  def test_ecs_v4_vendor_data
+    expected = {'ecsDockerId' => '1234567890987654321'}
+
+    aws_fixture_path = File.expand_path('../../../../fixtures/utilization/aws', __FILE__)
+    fixture = File.read(File.join(aws_fixture_path, 'ecs.json'))
+
+    NewRelic::Agent::Utilization::ECSV4.endpoint('http://169.254.169.254:51117/v4/metadata')
```
kaylareopelle, line 88: Where did this endpoint come from?

kaylareopelle: Ah, I see. Thank you!

test/fixtures/utilization/aws/ecs.json
tannalynn: I don't believe there are any cross agent tests for this. I originally was using the example from the new agent spec for this, but it was long and almost everything was not even looked at, so I removed the unused keys in the hash and gave the id value a fake number.

lib/new_relic/agent/utilization_data.rb
tannalynn: V4 is the current one. the unversioned one is only if the V4 one fails for some reason, we'll use unversioned as a backup. It's defined that way in the agent specs, so I didn't look too much futher into the different versions other than that though.

lib/new_relic/agent/utilization_data.rb
tannalynn: If someone has detect_aws turned off, we want to skip the ecs check as well since they are both controlled by the aws config.

lib/new_relic/agent/utilization/ecs.rb
tannalynn: No, this class is the only place would define any key transformations that the parent would make (but we aren't defining any). I had to overwrite the transform_key method because it needed to work slightly differently overall for ECS

lib/new_relic/agent/utilization_data.rb
hannahramadan: This is all new to me! So basically because Fargate is an AWS service, it falls in the category of 'detect aws'?

lib/new_relic/agent/utilization_data.rb
hannahramadan: Is there an easy way to future-proof this for when a newer version comes out?

lib/new_relic/agent/utilization_data.rb
hannahramadan: Nevermind - I see this is all explained in the spec!

kaylareopelle: 🤦 - I should've read the spec first. Thank you!

lib/new_relic/agent/utilization_data.rb
tannalynn: Beyond the existing logic for the unversioned check we do if v4 fails, I don't think so. It's good that unversioned check is there as a backup in case v4 no longer works, but if something changes with the structure of the data, or the unversioned on changes also, then we don't know what we'd need to support, so we'll have to make changes.

hannahramadan (approved):

PR #3180, opened 2025-05-14
Enable "Starting transaction tests" in hybrid agent CAT

### LINKED ISSUE #3145: HA CAT - Starting transaction tests
Get the "Starting transaction tests" `testDescription` to pass. See the `hybrid_agent.json` fixture for details.

kaylareopelle:
This PR changes the strategy used to create transactions and segments when OpenTelemetry APIs are called to cue off the parent_context rather than the span kind.

## Summary
* Add `finishable` to OpenTelemetry Span class to track the
corresponding New Relic segment/transaction
* Add `Tracer#start_span` patch for `:server` span kinds
* Refactor `Tracer#in_span` to use `#start_span`
* Start new transactions based on parent context rather than span kind
* Do not start a transaction if span kind is internal
* Update arguments for OpenTelemetry Span class to accept
only `span_context`, and create the span context in a
a separate method
* Update agent output tests to look at individual keys rather
than a full hash match

Closes #3145

### DIFF
--- a/lib/new_relic/agent/opentelemetry/transaction_patch.rb
+++ b/lib/new_relic/agent/opentelemetry/transaction_patch.rb
@@ -15,42 +15,18 @@ module NewRelic
 
         def set_current_segment(new_segment)
           @current_segment_lock.synchronize do
-            # detach the current token, if one is present
             unless opentelemetry_context.empty?
               ::OpenTelemetry::Context.detach(opentelemetry_context[otel_current_span_key])
             end
 
-            span = nil
-
-            # create an otel span for the new segment
-            if new_segment.instance_variable_defined?(:@otel_span)
-              span = new_segment.instance_variable_get(:@otel_span)
-            else
-              span = Trace::Span.new(span_context: span_context_from_segment(new_segment))
-              new_segment.instance_variable_set(:@otel_span, span)
-            end
-
-            # set otel's current span to the newly created otel span
+            span = find_or_create_span(new_segment)
             ctx = ::OpenTelemetry::Context.current.set_value(otel_current_span_key, span)
-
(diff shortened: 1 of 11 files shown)


lib/new_relic/agent/opentelemetry/transaction_patch.rb
```diff
@@ -26,7 +21,7 @@ def set_current_segment(new_segment)
             end
 
             # create an otel span for the new segment
-            span = Trace::Span.new(segment: new_segment, transaction: new_segment.transaction)
+            span = Trace::Span.new(span_context: span_context_from_segment(new_segment))
```
kaylareopelle, line 24: This is spec compliant with OTel. The only argument in initialize for the span API is `span_context`. Since we're not creating real OTel spans, we just need the context to make sure the OpenTelemetry context calls match our New Relic context calls (ex. current_transaction, current_segment).

test/multiverse/suites/hybrid_agent/parsing_helpers.rb
```diff
@@ -123,47 +115,48 @@ def matches_assertion(parameters)
...
+    end
+  end
 
-            assert_equal output['parent_name'], result, msg
-            output.delete('parent_name')
-          end
+  def verify_span_output(output)
+    spans = harvest_span_events![1].flatten.reject { |h| h.empty? }
 
-          assert event >= output, msg
```
kaylareopelle, line 148: Previously, the agent output was checked in its entirety to make sure every element matched the cross agent tests. However, that's not the intention of the tests, there may be more data in the harvest. Since that's the case, now we assert on every attribute.

tannalynn (dismissed):

PR #3189, opened 2025-05-30
Enable injecting outbound trace context tests

kaylareopelle:
Enables the "injecting outbound trace context tests" for the hybrid agent cross-agent tests.

* Adds a new TracePropagator to manage context propagation for OpenTelemetry
* Defines the inject method within the propagator, which is a close equivalent to our `DistributedTracing.insert_distributed_trace_headers` API method
* See [hybrid_agent.json L296-L531](https://github.com/newrelic/newrelic-ruby-agent/blob/f7a989066d88818fce0a0b599ad036bb918d67ae/test/fixtures/cross_agent_tests/hybrid_agent.json#L296-L531) for what this test covers


test/multiverse/suites/hybrid_agent/commands.rb
```diff
@@ -57,21 +57,23 @@ def record_exception_on_span(error_message:, &block)
...
+    @headers = {url: url}
     yield if block
   end
 
   def o_tel_inject_headers(&block)
-    # TODO: figure out how to pass the request
-    request = {}
-    OpenTelemetry.propagation.inject(request)
-    yield if block
+    NewRelic::Agent.instance.stub(:connected?, true) do
```
kaylareopelle, line 67: The agent has to be connected in order to attach DT headers

lib/new_relic/agent/opentelemetry/context/propagation/trace_propagator.rb
```diff
@@ -0,0 +1,29 @@
...
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    module OpenTelemetry
+      module Context
+        module Propagation
+          class TracePropagator
+            # The carrier is the object carrying the headers
```
hannahramadan, line 11: These are helpful comments! Great to include for otel stuff. This was also helpful https://opentelemetry.io/docs/concepts/context-propagation/

lib/new_relic/agent/opentelemetry_bridge.rb
```diff
@@ -19,9 +19,13 @@ def self.install
         require 'opentelemetry' # requires the opentelemetry-api gem
         require_relative 'opentelemetry/trace'
         require_relative 'opentelemetry/transaction_patch'
+        require_relative 'opentelemetry/context'
+
+        # TODO: Add a warning if SDK gem is installed
```
hannahramadan, line 24: What is behind the warning? Should the SDK gem not be installed

test/multiverse/suites/hybrid_agent/trace_propagator_test.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module NewRelic
+  module Agent
+    module OpenTelemetry
+      module Context
+        module Propagation
+          class TracePropagatorTest < Minitest::Test
+            class FakePropError < StandardError; end
+
+            def test_inject_calls_distributed_tracing_api
```
hannahramadan, line 13: Stubbing our API to raise an error is a fun way to ensure our API is called

test/multiverse/suites/hybrid_agent/hybrid_agent_test.rb
```diff
@@ -14,6 +14,14 @@ class HybridAgentTest < Minitest::Test
 
   def setup
     @tracer = OpenTelemetry.tracer_provider.tracer
+
+    # in order to inject headers, there must be a parent account ID and a
+    # parent application ID
+    @config = {
```
hannahramadan, line 20: Where/how are these values being used. I didn't see anything in the hybrid_agent.json file that is looking for these values (probably the wrong spot to be looking anyways!) Down to review live if easier.

lib/new_relic/agent/opentelemetry/context/propagation/trace_propagator.rb
kaylareopelle: That link's a great find!

lib/new_relic/agent/opentelemetry_bridge.rb
kaylareopelle: This isn't specifically related to this PR, but it's a thought I had while writing it. 

In the lines below the todo we set global default for the tracer provider and propagator to the New Relic SDK implementation. Whenever someone calls an OpenTelemetry API method, it'll look to that implementation to call the related methods. These same lines run when someone calls `OpenTelemetry::SDK.configure`, but set the tracer provider and propagator to the OTel SDK classes.

Only one provider can be set to be globally recognized by the API. Subsequent calls to `OpenTelemetry.tracer_provider =` will overwrite the value.

If someone has the OTel SDK installed and calls configure after we run this file, I think the New Relic classes would be lost/ignored.

There may be another way to handle this, but it's out of scope for this particular issue.

test/multiverse/suites/hybrid_agent/hybrid_agent_test.rb
kaylareopelle: Great question! New Relic has some rules about when it'll try to inject headers. Some of the conditions are making sure the agent is connected and that there's an account id and application id. These things aren't necessarily true all the time when we run our tests.

The code snippet below runs when trying to inject headers:

https://github.com/newrelic/newrelic-ruby-agent/blob/dfadcfdd49e40e9035cad2cd78581b4f4a7e78be/lib/new_relic/agent/distributed_tracing/trace_context_payload.rb#L113-L124

If we don't set the account id or app id here, lines 116 and 117 will fail, causing the injection attempt to return nil. 

The other part, about making sure the agent is connected, is taken care of in the Commands module in the injection-related methods. 

Since we have so much parsing going on to get the cross-agent tests to work, I thought it was easier to put the config in the global startup. However, I'm happy to move this to another place if we want it to only run for the injection-related tests.

lib/new_relic/agent/opentelemetry_bridge.rb
hannahramadan: Thank you for explaining ◡̈

test/multiverse/suites/hybrid_agent/hybrid_agent_test.rb
hannahramadan: Ah okay it makes more sense what is happening, and your comment totally covers this. I'm not immediately familiar with injecting header rules so I read this as an otel specific need. Thank you!

hannahramadan (approved):
🌻

tannalynn (approved):

PR #3195, opened 2025-06-03
Hybrid Agent Cross-Agent Tests: Inbound distributed tracing tests

kaylareopelle:
This PR adds an `extract` method to the TracePropagator to handle inbound distributed tracing context.

When spans have remote parents, the information is updated after the transaction/segment is started, but before yielding [(example)](https://github.com/newrelic/newrelic-ruby-agent/blob/5ca88b7a2f46d1b94e9540543c7b72e8aaa2ccf0/lib/new_relic/agent/instrumentation/rdkafka/instrumentation.rb#L28). The OpenTelemetry bridge follows this approach. 

There's also some renaming of variables to make things more clear.

These changes allow the ["Inbound distributed tracing tests"](https://github.com/newrelic/newrelic-ruby-agent/blob/f7a989066d88818fce0a0b599ad036bb918d67ae/test/fixtures/cross_agent_tests/hybrid_agent.json#L524-L629) described in the hybrid_agent.json file to pass.

### DIFF
--- a/lib/new_relic/agent/opentelemetry/context/propagation/trace_propagator.rb
+++ b/lib/new_relic/agent/opentelemetry/context/propagation/trace_propagator.rb
@@ -16,10 +16,24 @@ module NewRelic
               NewRelic::Agent::DistributedTracing.insert_distributed_trace_headers(carrier)
             end
 
-            # TODO: Implement in full for inbound distributed tracing test
-            # The getter argument is a no-op, added for consistency with the OpenTelemetry API
-            def extract(carrier, context: ::OpenTelemetry::Context.current, getter: nil)
-              NewRelic::Agent::DistributedTracing.accept_distributed_trace_headers
+            def extract(carrier, context: ::OpenTelemetry::Context.current, getter: ::OpenTelemetry::Context::Propagation.text_map_getter)
+              trace_parent_value = carrier[TRACEPARENT_KEY]
+              return context unless trace_parent_value
+
+              # There's an optional format key that we may need to address eventually
+              # If the format is rack, then the key is HTTP_TRACEPARENT
+              trace_context = NewRelic::Agent::DistributedTracing::TraceContext.parse(carrier: carrier)
+
+              tp = trace_context.trace_parent
+
+              # TODO: Add tracestate parsing
+              span_context = ::OpenTelemetry::Trace::SpanContext.new(trace_id: tp['trace_id'], span_id: tp['parent_id'], trace_flags: tp['trace_flags'], tracestate: nil, remote: true)
+
+              span = ::OpenTelemetry::Trace.non_recording_span(span_context)
+              ::OpenTelemetry::Trace.context_with_span(span, parent_context: context)
+            rescue StandardError => e
+              NewRelic::Agent.logger.error("Unable to extract context: #{e.message}")
+              context
             end
           end
         end
--- a/lib/new_relic/agent/opentelemetry/context/propagation/trace_propagator.rb
+++ b/lib/new_relic/agent/opentelemetry/context/propagation/trace_propagator.rb
@@ -17,23 +17,32 @@ module NewRelic
             end
 
             def extract(carrier, context: ::OpenTelemetry::Context.current, getter: ::OpenTelemetry::Context::Propagation.text_map_getter)
-              trace_parent_value = carrier[TRACEPARENT_KEY]
-              return context unless trace_parent_value
-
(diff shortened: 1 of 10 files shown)


kaylareopelle:
We met to discuss this PR. Two improvements need to be made before reviews continue: 
1. Incorporate use of the `getter` to allow for Rack headers to be parsed appropriately
2. Update the `do_work_with_inbound_context` method to use tracestate

lib/new_relic/agent/opentelemetry/context/propagation/trace_propagator.rb
```diff
@@ -16,10 +16,39 @@ def inject(carrier, context: ::OpenTelemetry::Context.current, setter: nil)
...
+            # The return value for this method should be an instance of the
+            # OpenTelemetry Context class. The return value of
+            # #accept_distributed_trace_headers is a transaction, so we cannot
+            # use it to extract the context.
+            def extract(carrier, context: ::OpenTelemetry::Context.current, getter: ::OpenTelemetry::Context::Propagation.text_map_getter)
+              carrier_format = determine_format(getter)
+              trace_state_entry_key = "#{NewRelic::Agent.config[:account_id]}@nr"
+              trace_context = NewRelic::Agent::DistributedTracing::TraceContext.parse(carrier: carrier, format: carrier_format, trace_state_entry_key: trace_state_entry_key)
+              tp = trace_context.trace_parent
+              # TODO: Not much happens with tracestate at this time, revisit before GA
```
tannalynn: For revist before GA, is there anything specific that you think will be changed, or is it more of a TBD situation?

test/multiverse/suites/hybrid_agent/commands.rb
```diff
@@ -28,12 +28,26 @@ def do_work_in_span_with_remote_parent(span_name:, span_kind:, &block)
 
   def do_work_in_span_with_inbound_context(span_name:, span_kind:, trace_id_in_header:,
     span_id_in_header:, sampled_flag_in_header:, &block)
-    # TODO
+    sampled = sampled_flag_in_header == '0' ? '00' : '01'
```
tannalynn, line 31: Did you find out why the test was using just '0'?

kaylareopelle: Yes! It was just a choice made by the person who wrote the tests. `0` should map to `00` in the header, and `1` maps to `01`. The header needs two chars to be valid.

lib/new_relic/agent/opentelemetry/context/propagation/trace_propagator.rb
kaylareopelle: This is a great question. I wrote this because I noticed the trace context was only giving the trace state payload to the SpanContext, which isn't used by any of the NR objects. At the time, I thought it was a TBD situation, but now that I look again there's work I can do to pass the tracestate back to the NR transaction in the OTel tracer so that it more closely resembles the accept headers method. I'll fix that up and resubmit.

lib/new_relic/agent/transaction/trace_context.rb
```diff
@@ -136,7 +136,7 @@ def accept_trace_context(header_data)
 
           transaction.distributed_tracer.parent_transaction_id = payload.transaction_id
 
-          determine_sampling_decision(payload, header_data)
+          determine_sampling_decision(payload, header_data.trace_parent['trace_flags'])
```
kaylareopelle, line 139: @tannalynn - I switched things up a little bit here so that this method could be used by OTel as well. However, I'm not sure if I'm missing anything. It seemed like this is the only place it was called besides now in the OTel tracer.

test/multiverse/suites/hybrid_agent/hybrid_agent_test.rb
```diff
@@ -35,19 +35,8 @@ def teardown
...
-    %w[
-      creates_opentelemetry_segment_in_a_transaction
-      creates_new_relic_span_as_child_of_opentelemetry_span
-      does_not_create_segment_without_a_transaction
-      starting_transaction_tests
-      opentelemetry_api_and_new_relic_api_can_inject_outbound_trace_context
-      opentelemetry_api_can_record_errors
-      opentelemetry_api_can_add_custom_attributes_to_spans
-    ]
+    %w[]
```
kaylareopelle, line 39: Now that all the tests are passing, we can leave the focus empty and they all should run automatically.

tannalynn (approved):

hannahramadan (approved):
🌻

Issue #3153, opened 2025-04-24
Support Elasticsearch 9.0.0


kaylareopelle:
Elasticsearch 9.0.1 was released, and now we're running into some issues on the CI. Address the issues and get the suites passing again. Also, look into the changelog and see if there are any new features/code paths we should support.

PR #3205, opened 2025-06-09
Set up elasticsearch 9 in CI

tannalynn:
Adds elasticsearch 9 service to the CI and allows those tests to start running again\
Ran into some docker memory usage issues so I took out elasticsearch 8 from the CI since its the same as 9. Also moved elasticsearch into its own group.

closes https://github.com/newrelic/newrelic-ruby-agent/issues/3153

schedule ci https://github.com/newrelic/newrelic-ruby-agent/actions/runs/15542099551

### DIFF
--- a/test/multiverse/suites/elasticsearch/Envfile
+++ b/test/multiverse/suites/elasticsearch/Envfile
@@ -6,7 +6,7 @@ instrumentation_methods :chain, :prepend
 
 ELASTICSEARCH_VERSIONS = [
   # only testing 7 and 9 to prevent memory issues in the CI (8 and 9 are the same)
-  [nil, 2.5],
+  [nil, 2.6],
   ['7.17.1', 2.4]
 ]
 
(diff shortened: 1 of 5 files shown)


test/multiverse/suites/elasticsearch/Envfile
```diff
@@ -5,8 +5,7 @@
 instrumentation_methods :chain, :prepend
 
 ELASTICSEARCH_VERSIONS = [
-  # [nil, 2.5], # temporarily stop testing elasticsearch 9
-  ['8.12.0', 2.5],
+  [nil, 2.5],
```
hannahramadan: I coulddd see a comment being useful here about elastic 8 and 9 being the same and CI having mem issues so not testing both. Sometimes I wonder about decisions made in why we test certain versions in these files

hannahramadan (approved):

test/multiverse/suites/elasticsearch/Envfile
kaylareopelle: I would find that helpful too

tannalynn: for sure, added it here https://github.com/newrelic/newrelic-ruby-agent/pull/3205/commits/2aa4eb9977b0450b5b52086911a267f84068a9fb

PR #3206, opened 2025-06-11
Hybrid agent span output test fix

kaylareopelle:
Two changes here:
1. Improvements made to the `verify_span_output` parsing helper to make it easier to debug and fix the attributes assertion
2. Updated to the latest version of the `hybrid_agent.json` file

### DIFF
--- a/test/multiverse/suites/hybrid_agent/parsing_helpers.rb
+++ b/test/multiverse/suites/hybrid_agent/parsing_helpers.rb
@@ -143,20 +143,21 @@ module ParsingHelpers
       # span[1] is for custom attributes
       # span[2] is for agent attributes (which aren't tested here)
       actual = spans.find { |s| s[0]['name'] == expected['name'] }
+
       assert actual, "Span output could not be verified. Span was not found.\n" \
         "Expected: #{expected}\n" \
         "Harvested spans: #{spans}\n"
 
       if expected['category']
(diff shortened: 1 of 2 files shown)


test/multiverse/suites/hybrid_agent/parsing_helpers.rb
```diff
@@ -131,28 +131,46 @@ def verify_transaction_output(output)
...
 
       if expected['entryPoint']
-        assert_equal expected['entryPoint'], actual['nr.entryPoint'], 'Unexpected entryPoint'
+        assert_equal expected['entryPoint'], actual[0]['nr.entryPoint'], 'Span was not an entrypoint.'
       end
 
       expected['attributes']&.each do |expected_key, expected_value|
-        puts 'SKIPPING ASSERTION - NEED TO IMPLEMENT ATTRIBUTES AND ERROR HANDLING'
-        # assert_equal expected_value, actual['attributes'][expected_key], 'Unexpected attribute'
+        assert_equal expected_value, actual[1] & [expected_key],
```
kreopelle, line 160: we don't need to keep this commented out anymore because the test that was failing changed upstream, so attributes are no longer needed for the tests on dev

tannalynn (approved):

hannahramadan (approved):

PR #3207, opened 2025-06-11
OpenTelemetry API can add custom attributes to spans

### LINKED ISSUE #3090: HA CAT - OpenTelemetry API can add custom attributes to spans
Update the New Relic Ruby agent to add custom attributes set by the OpenTelemetry method to the current New Relic Span and pass the "OpenTelemetry API can add custom attributes to spans" hybrid agent cross agent test.

Relevant OpenTelemetry APIs:
* [OpenTelemetry::Trace#current_span](https://open-telemetry.github.io/opentelemetry-ruby/opentelemetry-api/v1.5.0/OpenTelemetry/Trace.html#current_span-instance_method) 
* [OpenTelemetry::Trace::Span#add_attributes](https://open-telemetry.github.io/opentelemetry-ruby/opentelemetry-api/v1.5.0/OpenTelemetry/Trace/Span.html#add_attributes-instance_method)

hannahramadan:
Updates test and adds ability to add attributes to spans

closes #3090

### DIFF
--- a/lib/new_relic/agent/opentelemetry/trace/tracer.rb
+++ b/lib/new_relic/agent/opentelemetry/trace/tracer.rb
@@ -27,6 +27,7 @@ module NewRelic
 
             span = get_span_from_finishable(finishable)
             span.finishable = finishable
+            span.add_attributes(attributes) if attributes
             span
           end
 
--- a/test/fixtures/cross_agent_tests/hybrid_agent.json
+++ b/test/fixtures/cross_agent_tests/hybrid_agent.json
@@ -212,7 +212,7 @@
                 ]
               },
               {
-                "command": "AddOTelAttributeHash",
+                "command": "AddOTelAttribute",
                 "parameters": {
                   "name": "spanNumber",
                   "value": 1
--- a/test/fixtures/cross_agent_tests/hybrid_agent.json
+++ b/test/fixtures/cross_agent_tests/hybrid_agent.json
@@ -212,7 +212,7 @@
                 ]
               },
               {
-                "command": "AddOTelAttribute",
+                "command": "AddOTelAttributeHash",
                 "parameters": {
                   "name": "spanNumber",
                   "value": 1
--- a/test/multiverse/suites/hybrid_agent/parsing_helpers.rb
+++ b/test/multiverse/suites/hybrid_agent/parsing_helpers.rb
@@ -157,7 +157,8 @@ module ParsingHelpers
       end
 
       expected['attributes']&.each do |expected_key, expected_value|
-        assert_equal expected_value, actual[1] & [expected_key],
+        actual_value = actual[1][expected_key]
+        assert_equal expected_value, actual_value,
           "Expected attributes not found.\n" \
           "Expected attribute: {'#{expected_key}' => '#{expected_value}'}\n" \
           "Actual span: #{actual}\n"
--- a/test/multiverse/suites/hybrid_agent/parsing_helpers.rb
+++ b/test/multiverse/suites/hybrid_agent/parsing_helpers.rb
@@ -158,6 +158,7 @@ module ParsingHelpers
 
       expected['attributes']&.each do |expected_key, expected_value|
         actual_value = actual[1][expected_key]
+
         assert_equal expected_value, actual_value,
           "Expected attributes not found.\n" \
           "Expected attribute: {'#{expected_key}' => '#{expected_value}'}\n" \
--- a/test/multiverse/suites/hybrid_agent/parsing_helpers.rb
+++ b/test/multiverse/suites/hybrid_agent/parsing_helpers.rb
@@ -157,9 +157,14 @@ module ParsingHelpers
       end
 
       expected['attributes']&.each do |expected_key, expected_value|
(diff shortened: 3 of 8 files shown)


test/multiverse/suites/hybrid_agent/parsing_helpers.rb
```diff
@@ -157,7 +157,8 @@ def verify_span_output(output)
       end
 
       expected['attributes']&.each do |expected_key, expected_value|
-        assert_equal expected_value, actual[1] & [expected_key],
```
hannahramadan: the `assert_equal` method didn't like the `&`. I hope the spirit of the test is the same, but happy to rework if not.

lib/new_relic/agent/opentelemetry/trace/tracer.rb
```diff
@@ -27,6 +27,7 @@ def start_span(name, with_parent: nil, attributes: nil, links: nil, start_timest
 
             span = get_span_from_finishable(finishable)
             span.finishable = finishable
+            span.add_attributes(attributes) if attributes
```
hannahramadan, line 30: I'm not 100% sure about this, but it seems if a span is create with attributes, they should be added

test/fixtures/cross_agent_tests/hybrid_agent.json
```diff
@@ -212,7 +212,7 @@
                 ]
               },
               {
-                "command": "AddOTelAttribute",
+                "command": "AddOTelAttributeHash",
```
kaylareopelle: Unfortunately, this JSON file needs to stay in sync with what's available upstream so that all the language agents are running the same kinds of tests.

Instead, I recommend adding a test related to adding OTel attributes using a hash to the `test/multiverse/suites/hybrid_agent/span_test.rb` file.

test/multiverse/suites/hybrid_agent/parsing_helpers.rb
```diff
@@ -157,7 +157,8 @@ def verify_span_output(output)
...
+        actual_value = actual[1][expected_key]
```
kaylareopelle: What do you think about adding the safe navigation operator to this? My thinking is that we'll get a better error message. Without it, I believe we end up with the `Undefined method [] for nil` message if the attribute isn't on the span.

test/multiverse/suites/hybrid_agent/parsing_helpers.rb
kaylareopelle: whoops, missed this! does the `&` work in the variable assignment?

lib/new_relic/agent/opentelemetry/trace/tracer.rb
kaylareopelle: I think this is a good call! My only recommendation would be to add some tests to the `tracer_test.rb` file that make sure the attributes get assigned. I don't think this will be hit by any of the cross-agent tests.

test/multiverse/suites/hybrid_agent/parsing_helpers.rb
hannahramadan: I dont think I've seen `&` used in this way before - what were we trying to do?

test/fixtures/cross_agent_tests/hybrid_agent.json
hannahramadan: Ahh okay I'll revert this and check out that other test file ◡̈

lib/new_relic/agent/opentelemetry/trace/tracer.rb
kaylareopelle: This should be the equivalent line in the OTel SDK: https://github.com/open-telemetry/opentelemetry-ruby/blob/a1f1629abb7c87ae316cf0f4541111e48e433d67/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb#L149

test/multiverse/suites/hybrid_agent/parsing_helpers.rb
kaylareopelle: Me neither -- it was a Rubocop suggestion. The goal is to make sure that `actual[1]` isn't `nil`. However, maybe that's getting taken care of with the `assert actual` at the top of the method. If a span is found, it's pretty safe that it will have that hash in the first index. 

After thinking about it more, I'm fine with leaving it out! Thanks for asking.

lib/new_relic/agent/opentelemetry/trace/tracer.rb
hannahramadan: Test added!: https://github.com/newrelic/newrelic-ruby-agent/pull/3207/commits/edd4aa14d02b92f249198d29ea728f572cac6742

test/fixtures/cross_agent_tests/hybrid_agent.json
hannahramadan: Fixed here ◡̈ https://github.com/newrelic/newrelic-ruby-agent/pull/3207/commits/b8af3a00f15308df9b13c2d5209fa70366f4ddab

tannalynn (approved):

Issue #3143, opened 2025-04-21
HA CAT - OpenTelemetry API can add custom attributes to spans


kaylareopelle:
Get the "OpenTelemetry API can add custom attributes to spans" `testDescription` to pass. See the `hybrid_agent.json` fixture for details.

PR #3209, opened 2025-06-12
Add errors api to hybrid agent

tannalynn:
closes https://github.com/newrelic/newrelic-ruby-agent/issues/3143

### DIFF
--- a/test/multiverse/suites/hybrid_agent/parsing_helpers.rb
+++ b/test/multiverse/suites/hybrid_agent/parsing_helpers.rb
@@ -157,7 +157,14 @@ module ParsingHelpers
       end
 
       expected['attributes']&.each do |expected_key, expected_value|
-        assert_equal expected_value, actual[1] & [expected_key],
+        # 0 is intrinsic attributes, 1 is custom attributes, 2 is agent attributes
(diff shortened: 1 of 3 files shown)


test/multiverse/suites/hybrid_agent/parsing_helpers.rb
```diff
@@ -157,7 +157,14 @@ def verify_span_output(output)
       end
 
       expected['attributes']&.each do |expected_key, expected_value|
-        assert_equal expected_value, actual[1] & [expected_key],
+        # 0 is intrinsic attributes, 1 is custom attributes, 2 is agent attributes
```
tannalynn, line 160: Alternatively, we could flatten the hash or something instead, which might make more sense (as long as that doesnt cause any issues for what the hybrid agent stuff expects)

tannalynn: if we have to mess with these verify things again, then we can think about flattening it or something, not needed for now.

kaylareopelle: We chatted about this sync, but I think this is good for now. We can always revisit later on.

PR #3210, opened 2025-06-16
Composite actions for CI and Scheduled CI

tannalynn:
Makes our ci and ci_cron workflows easier to read and modify.
Also, adds more gem versions to the bundler setup that have been making certain test runs fail the first time each time.

ci cron run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/15715092685

### DIFF
--- a/.github/actions/run_multiverse/action.yml
+++ b/.github/actions/run_multiverse/action.yml
@@ -7,6 +7,10 @@ inputs:
   group:
     description: 'What multiverse group to run'
     required: true
+  pr-ci:
+    description: 'If this is run on PR'
+    required: true
+    default: true
 
 runs:
   using: "composite"
@@ -45,7 +49,7 @@ runs:
         VERBOSE_TEST_OUTPUT: true
         SERIALIZE: 1
         COVERAGE: true
-        CI_FOR_PR: true
+        CI_FOR_PR: ${{ inputs.pr-ci}}
         POSTGRES_USERNAME: postgres
         POSTGRES_PASSWORD: password
         MYSQL_PASSWORD: root
@@ -59,6 +63,7 @@ runs:
       uses: ./.github/actions/annotate
 
     - name: Save coverage results
+      if: ${{ inputs.pr-ci }} == 'true'
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # tag v4.6.2
       with:
         name: coverage-report-multiverse-${{ inputs.ruby-version }}-${{ inputs.group }}
--- a/.github/actions/run_multiverse/action.yml
+++ b/.github/actions/run_multiverse/action.yml
@@ -37,7 +37,7 @@ runs:
       uses: nick-fields/retry@ce71cc2ab81d554ebbe88c79ab5975992d79ba08 # tag v3.0.2
       with:
         timeout_minutes: 60
-        max_attempts: 2
+        max_attempts: 1
         command: bundle exec rake test:multiverse[group="${{ inputs.group }}"]
       env:
         VERBOSE_TEST_OUTPUT: true
--- a/.github/actions/run_multiverse/action.yml
+++ b/.github/actions/run_multiverse/action.yml
@@ -11,10 +11,6 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Configure git
-      run: 'git config --global init.defaultBranch main'
-    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # tag v4.2.2
-
       # - curl is needed for Curb
       # - xslt is needed for older Nokogiris, RUBY_VERSION < 2.5
       # - sasl is needed for memcached
--- a/.github/actions/run_multiverse/action.yml
+++ b/.github/actions/run_multiverse/action.yml
@@ -19,6 +19,7 @@ runs:
       # - xslt is needed for older Nokogiris, RUBY_VERSION < 2.5
       # - sasl is needed for memcached
     - name: Install OS packages
+      shell: bash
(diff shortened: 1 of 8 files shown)


.github/workflows/scripts/setup_bundler
```diff
@@ -74,23 +74,23 @@ function install_ruby_version_specific_gems {
...
-  fi
+    echo "DEBUG: running 'gem install' on specific gem versions"
 
-  # for some reason, ruby 2.7+ fails unless strscan 3.0.4 is installed
-  if [[ $RUBY_VERSION =~ ^3\. || $RUBY_VERSION =~ ^2\.7 ]]; then 
-    echo "DEBUG: running 'gem install strscan'"
+    gem install --default cgi:0.5.0
+    gem install --default date:3.4.1
+    gem install --default digest:3.1.0
+    gem install --default digest:3.2.0
```
tannalynn: I changed the way this works to that all ruby versions will do this because it was annoying to read the way it was before. Ones that have multiple versions like this is because the newer versions wont install on older ruby versions, so I install both bc it doesn't really take much extra time and overall its a lot easier to understand, and the second one just overwrites the first one if it needs to.

.github/actions/run_multiverse/action.yml
```diff
@@ -0,0 +1,74 @@
...
+  steps:
+      # - curl is needed for Curb
+      # - xslt is needed for older Nokogiris, RUBY_VERSION < 2.5
+      # - sasl is needed for memcached
+    - name: Install OS packages
+      shell: bash
+      run: sudo apt-get update; sudo apt-get install -y --no-install-recommends libcurl4-nss-dev libsasl2-dev libxslt1-dev
+
+    - name: Install Ruby ${{ inputs.ruby-version }}
+      uses: ruby/setup-ruby@13e7a03dc3ac6c3798f4570bfead2aed4d96abfb # tag v1.244.0
```
kaylareopelle, line 26: One thing to note -- Dependabot can't access actions you define yourself to update PRs, so we may want to go back to using the task James made to bump versions. Alternately, we could do what OTel does and just add a commit to the dependabot PRs to bump the versions in our composite actions.

.github/workflows/scripts/setup_bundler
kaylareopelle: thank you for explaining! this list is so much easier to read. would you mind adding a comment that explains the double versions?

.github/workflows/ci.yml
```diff
@@ -197,61 +90,16 @@ jobs:
...
-      - name: Save gem manifest
-        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # tag v4.6.2
-        with:
-          name: gem_manifest_${{ matrix.ruby-version }}_services_1.json
-          path: gem_manifest_${{ matrix.ruby-version }}_services_1.json
-          retention-days: 2
+          group: services_1
 
 
   multiverse_services_2:
```
kaylareopelle, line 105: It looks like `multiverse_services_2` starts a Kafka service. Do we still need that, now that there's a separate Kafka job?

Also, I see it spins up a postgres services too -- is that needed? or is it handled by the MySQL/Postgres job now?

.github/actions/run_multiverse/action.yml
```diff
@@ -0,0 +1,74 @@
...
+        SERIALIZE: 1
+        COVERAGE: true
+        CI_FOR_PR: ${{ inputs.pr-ci}}
+        POSTGRES_USERNAME: postgres
+        POSTGRES_PASSWORD: password
+        MYSQL_PASSWORD: root
+        DB_PASSWORD: root
+        DB_PORT: ${{ job.services.mysql.ports[3306] }}
+        MYSQL_PORT: ${{ job.services.mysql.ports[3306] }}
+        MYSQL_HOST: 127.0.0.1
```
kaylareopelle, line 58: Is the DB stuff used outside the DB jobs?

tannalynn: No, I just combined all the env vars we used in the different multiverse runs here so I didn't have to mess around with passing env vars or anything since having them present if we aren't using them doesn't cause any issues.

.github/workflows/ci.yml
tannalynn: oh huh, i'll have to check, I don't think those are needed there since i split things out previously, I must have just missed that the services were still defined because of how awful the file was to deal with lol. ill take a look.

.github/actions/run_multiverse/action.yml
tannalynn: Yeah, i think adding a commit to depandabot prs is prbly a good idea, because otherwise I'd never think about updating those and that pr being created would be a good way to remember!

.github/workflows/scripts/setup_bundler
tannalynn: sure! added https://github.com/newrelic/newrelic-ruby-agent/pull/3210/commits/1d437329cc761c2111bfc06675643d64300ec537

.github/actions/run_multiverse/action.yml
```diff
@@ -0,0 +1,74 @@
+name: 'Run Multiverse'
```
kaylareopelle, line 1: Not a big deal, but I noticed that "Run Multiverse" now has some parens after the name: 
Screenshot 2025-06-18 at 9 41 35 AM

I wonder if you take out the quotes if it would go away?

tannalynn: oh oops, i didnt notice that, its from copy pasting lol 
fixed https://github.com/newrelic/newrelic-ruby-agent/pull/3210/commits/c2128ea0e3476c05b0cda3da32466959189fc71e

.github/workflows/ci.yml
tannalynn: the postgres was real so i put it back lol

kaylareopelle: haha thanks for trying!

PR #3215, opened 2025-06-25
remove unnecessary cat configs from http client tests

tannalynn:


hannahramadan (dismissed):

test/new_relic/http_client_test_cases.rb
```diff
@@ -320,8 +318,7 @@ def test_agent_doesnt_add_a_request_header_to_outgoing_requests_if_xp_disabled
   end
 
   def test_agent_doesnt_add_a_request_header_if_empty_cross_process_id
-    with_config(:'cross_application_tracer.enabled' => true,
-      :'distributed_tracing.enabled' => false,
+    with_config(:'distributed_tracing.enabled' => false,
```
kaylareopelle: Why is DT false here?

test/new_relic/http_client_test_cases.rb
```diff
@@ -331,7 +328,6 @@ def test_agent_doesnt_add_a_request_header_if_empty_cross_process_id
 
   def test_agent_doesnt_add_a_request_header_if_empty_encoding_key
     with_config(
-      :'cross_application_tracer.enabled' => true,
       :'distributed_tracing.enabled' => false,
```
kaylareopelle: Also confused about DT here... I don't think I understand what's happening.

test/new_relic/http_client_test_cases.rb
```diff
@@ -629,7 +595,6 @@ def test_raw_synthetics_header_is_passed_along_when_cat_disabled
       config = {
         :encoding_key => test['settings']['agentEncodingKey'],
         :trusted_account_ids => test['settings']['trustedAccountIds'],
-        :'cross_application_tracer.enabled' => true,
         :'distributed_tracing.enabled' => false
```
kaylareopelle: Does it make sense to turn DT on now/just return to our defaults for the synthetics tests? (everything is passing, so it's probably not too important; mostly IDK if DT needs to be off for synthetics to run)

kaylareopelle (commented):
Thanks, @tannalynn! Overall, this looks good but I'm a little confused about some of the DT configs and could use your help understanding what's happening.

test/new_relic/http_client_test_cases.rb
tannalynn: oh yeah all of these are just like, leftover or whatever and don't matter, I just didn't notice before bc i was focused on the CAT setting, so i didn't really look at other settings. I went through and removed all of these since it doesn't matter if DT is on or not in any of these situations.

kaylareopelle: Cool, thank you!

kaylareopelle (approved):

Issue #2357, opened 2023-12-11
Spec: Add stack trace attribute to datastore segments


tannalynn:
If a datastore segment time passes the configured threshold, an attribute `code.stacktrace` should be added to the datastore segment.

[Spec PR](https://source.datanerd.us/agents/agent-specs/pull/651)

PR #3220, opened 2025-07-09
add code.stacktrace attribute to datastore span

tannalynn:
When the duration is over the threshold, we should capture the stacktrace as an attribute on spans. Previously, the stacktrace was only on slowsql statements and not added to spans.

resolves https://github.com/newrelic/newrelic-ruby-agent/issues/2357

### DIFF
--- a/lib/new_relic/agent/span_event_primitive.rb
+++ b/lib/new_relic/agent/span_event_primitive.rb
@@ -123,7 +123,7 @@ module NewRelic
           agent_attributes[DB_STATEMENT_KEY] = truncate(segment.nosql_statement, DB_STATEMENT_MAX_BYTES)
         end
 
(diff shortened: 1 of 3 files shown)


lib/new_relic/agent/span_event_primitive.rb
```diff
@@ -122,6 +123,10 @@ def for_datastore_segment(segment) # rubocop:disable Metrics/AbcSize
           agent_attributes[DB_STATEMENT_KEY] = truncate(segment.nosql_statement, DB_STATEMENT_MAX_BYTES)
         end
 
+        if segment.params[:backtrace]
```
tannalynn, line 126: the backtrace is only ever added to the params hash when the duration is exceeded. That check is already done by the time we get here, so no need to check the duration against the threshold again

Issue #2516, opened 2024-03-19
Dependency detection: duplicated log entries, potentially duplicated invocations


fallwith:
**Issues:**
- dependency detection logic is double reported for every instrumented gem
- dependency detection logic may possibly also be involving doubled, redundant invocations

**To reproduce:**
- set the log level to debug and init the agent
- observe the log to see duplicated detection lines such as these:

```
DEBUG : Using auto configuration value for bunny to configure instrumentation
DEBUG : Using auto configuration value for bunny to configure instrumentation
DEBUG : Using auto configuration value for concurrent_ruby to configure instrumentation
DEBUG : Using auto configuration value for concurrent_ruby to configure instrumentation
DEBUG : Using auto configuration value for curb to configure instrumentation
DEBUG : Using auto configuration value for curb to configure instrumentation
DEBUG : Using auto configuration value for delayed_job to configure instrumentation
DEBUG : Using auto configuration value for delayed_job to configure instrumentation
```

**Source:**
The log output is generated by the `fetch_config_value` method defined in `lib/new_relic/dependency_detection.rb`.

**Pathways:**
The doubled log output is produced by 2 different call stacks:

``` 
# pathway 1
"lib/new_relic/dependency_detection.rb:183:in `fetch_config_value'",
"lib/new_relic/dependency_detection.rb:202:in `config_value'",
"lib/new_relic/dependency_detection.rb:170:in `block (2 levels) in <class:Dependent>'",
"lib/new_relic/dependency_detection.rb:139:in `allowed_by_config?'",
"lib/new_relic/dependency_detection.rb:122:in `check_dependencies'",
"lib/new_relic/dependency_detection.rb:64:in `dependencies_satisfied?'",
"lib/new_relic/dependency_detection.rb:28:in `block in detect!'",
"lib/new_relic/dependency_detection.rb:27:in `each'",
"lib/new_relic/dependency_detection.rb:27:in `detect!'",
"lib/new_relic/control/instrumentation.rb:64:in `_install_instrumentation'",
"lib/new_relic/control/instrumentation.rb:47:in `install_instrumentation'",
"lib/new_relic/control/private_instance_methods.rb:39:in `init_instrumentation'",
"lib/new_relic/control/instance_methods.rb:75:in `init_plugin'",

# pathway 2
"lib/new_relic/dependency_detection.rb:183:in `fetch_config_value'",
"lib/new_relic/dependency_detection.rb:202:in `config_value'",
[...]

Issue #3029, opened 2025-01-22
Remove double debug-level instrumentation logs


hannahramadan:
Some work was done in https://github.com/newrelic/newrelic-ruby-agent/pull/2881 to remove duplicated `info`-level instrumentation logs. However it appears `debug`-level log are still duplicating library instrumentation configurations. We should only be reporting this value once.

```log
[2025-01-22 13:41:13 -0800 LOANER-C02FT0HWMD6P (99660)] DEBUG : Using auto configuration value for ruby_openai to configure instrumentation
[2025-01-22 13:41:13 -0800 LOANER-C02FT0HWMD6P (99660)] DEBUG : Using auto configuration value for ruby_openai to configure instrumentation
[2025-01-22 13:41:13 -0800 LOANER-C02FT0HWMD6P (99660)] DEBUG : Using auto configuration value for sequel to configure instrumentation
[2025-01-22 13:41:13 -0800 LOANER-C02FT0HWMD6P (99660)] DEBUG : Using auto configuration value for sequel to configure instrumentation
[2025-01-22 13:41:13 -0800 LOANER-C02FT0HWMD6P (99660)] DEBUG : Using auto configuration value for sidekiq to configure instrumentation
[2025-01-22 13:41:13 -0800 LOANER-C02FT0HWMD6P (99660)] DEBUG : Using auto configuration value for sidekiq to configure instrumentation
[2025-01-22 13:41:13 -0800 LOANER-C02FT0HWMD6P (99660)] DEBUG : Using auto configuration value for stripe to configure instrumentation
[2025-01-22 13:41:13 -0800 LOANER-C02FT0HWMD6P (99660)] DEBUG : Using auto configuration value for stripe to configure instrumentation
[2025-01-22 13:41:13 -0800 LOANER-C02FT0HWMD6P (99660)] DEBUG : Using auto configuration value for thread to configure instrumentation
[2025-01-22 13:41:13 -0800 LOANER-C02FT0HWMD6P (99660)] DEBUG : Using auto configuration value for thread to configure instrumentation
[2025-01-22 13:41:13 -0800 LOANER-C02FT0HWMD6P (99660)] DEBUG : Using auto configuration value for thread to configure instrumentation
[2025-01-22 13:41:13 -0800 LOANER-C02FT0HWMD6P (99660)] DEBUG : Using auto configuration value for thread to configure instrumentation
[...]

PR #3221, opened 2025-07-09
Improving debug logs for configs

tannalynn:
So instead of doing what https://github.com/newrelic/newrelic-ruby-agent/issues/2516 actually was talking about, i decided to finally figure out how the config logging works and just make it work in a more helpful way, so we dont even need that extra logging we are doing, i dont think those statements were very helpful on their own anyways.

So basically, before, the way the config logging worked was we'd read in a config (like from yaml file or server response), and then we would log "hey we added this, here is now what the entire config looks like after looking at all precedence of config sources". which, while helpful in general, did not help us know _exactly_ what configs each source reads in. 
So now it'll log both. 

Also env source never logged anything bc its weird, but i fixed that.

Here is an example with a fresh baked config option i created for this example (i removed all the other configs so its easier to read lol)

BEFORE
```
[2025-07-09 13:22:34 -0500 NJMVXY1JF4 (98350)] DEBUG : Updating config (add) from NewRelic::Agent::Configuration::ManualSource. Results: {waluigi: "env var", .....} 

[2025-07-09 13:22:34 -0500 NJMVXY1JF4 (98350)] DEBUG : Updating config (add) from NewRelic::Agent::Configuration::YamlSource. Results: {waluigi: "env var", .....}

[2025-07-09 13:22:37 -0500 NJMVXY1JF4 (98350)] DEBUG : Updating config (add) from NewRelic::Agent::Configuration::ServerSource. Results: {waluigi: "env var", ....}
```

AFTER
```
[2025-07-09 13:05:19 -0500 NJMVXY1JF4 (87354)] DEBUG : Updating config (add) from NewRelic::Agent::Configuration::EnvironmentSource with values: {waluigi: "env var"}. 
Config Stack Results: {waluigi: "env var", ......... }

[2025-07-09 13:05:19 -0500 NJMVXY1JF4 (87354)] DEBUG : Updating config (add) from NewRelic::Agent::Configuration::ManualSource with values: {config: #<Rails::Application::Configuration:0x000000000027c0>}. 
Config Stack Results: {waluigi: "env var", ......... }

[2025-07-09 13:05:19 -0500 NJMVXY1JF4 (87354)] DEBUG : Updating config (add) from NewRelic::Agent::Configuration::YamlSource with values: {waluigi: "yaml file", .......... }. 
Config Stack Results: {waluigi: "env var",  ......... }

[2025-07-09 13:05:21 -0500 NJMVXY1JF4 (87354)] DEBUG : Updating config (add) from NewRelic::Agent::Configuration::ServerSource with values: {..., waluigi: "server src"}. 
Config Stack Results: {waluigi: "env var",  ......... }
```

In the manual source in the after, the weird ` {config: #<Rails::Application::Configuration:0x000000000027c0>}. ` is fine, that actually makes sense bc of lib/newrelic_rpm.rb, where we pass that in to the init_plugin.

Also all those "using auto config to install whatever instrumentation" messages are all just gone, because we can look at configs better now.
resolves https://github.com/newrelic/newrelic-ruby-agent/issues/2516 https://github.com/newrelic/newrelic-ruby-agent/issues/3029

### DIFF
--- a/test/new_relic/agent/configuration/manager_test.rb
+++ b/test/new_relic/agent/configuration/manager_test.rb
@@ -300,7 +300,7 @@ module NewRelic::Agent::Configuration
 
       log_lines = log.array
 
-      refute_match(/DEBUG.*asdf/, log_lines[0])
+      refute_match(/DEBUG.*Config Stack.*asdf/, log_lines[0])
     end
 
     def test_config_is_correctly_initialized
@@ -533,7 +533,7 @@ module NewRelic::Agent::Configuration
     end
 
     def test_logger_does_not_receive_excluded_settings
-      log = with_array_logger(:debug) { @manager.log_config('direction', 'source') }.array.join('')
+      log = with_array_logger(:debug) { @manager.log_config('direction', ManualSource.new({})) }.array.join('')
 
       assert_includes(log, 'app_name')
       refute_includes(log, 'license_key')


hannahramadan (approved):
Nice yay! And I think this ticket may also solve https://github.com/newrelic/newrelic-ruby-agent/issues/3029 ?

PR #3237, opened 2025-07-24
Ruby 3.5 support

hannahramadan:
### Code Changes
**Code-Level Metrics**
- In Rubies < 3.5, Method#source_location would return a two-element array: `[file_path, line_number]`.
- In Ruby 3.5, Method#source_location is expected to return a five-element array: `[file_path, start_line, start_column, end_line, end_column]` [source](https://github.com/ruby/ruby/commit/073c4e1cc712064e626914fa4a5a8061f903a637)
- Code-level metrics was previously getting the line number by looking at the last element in the array. Now, looking at the second element should work for all Rubies.


Gemfile
```diff
@@ -3,4 +3,7 @@
 # frozen_string_literal: true
 
 source 'https://rubygems.org'
+
 gemspec name: 'newrelic_rpm'
+
+gem 'logger'
```
kaylareopelle: I don't think we need to add the dependency in both places. The gemspec should be enough.

PR #3257, opened 2025-09-09
Update method tracer for Ruby 3.5 compatibility

hannahramadan:
In Rubies < 3.5, Method#source_location would return a two-element array: `[file_path, line_number]`.

In Ruby 3.5, Method#source_location is expected to return a five-element array: `[file_path, start_line, start_column, end_line, end_column]` [PR](https://github.com/ruby/ruby/pull/12539)

Code-level metrics was previously getting the line number by looking at the last element in the array. Now, looking at the second element should work for all Rubies.

This is a backwards compatible change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+- **Feature: Ensure compatibility with Ruby 3.5 change to Method#source_location**
+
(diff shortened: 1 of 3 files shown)


tannalynn (approved):

kaylareopelle (dismissed):
Thank you! There's a few other references to `source_location.last` in the codebase that we should look at before the release. Love getting this in now.

kaylareopelle (commented):
One more thing -- What do you think about adding a changelog entry for this? It's backwards-compatible, but could be a good way to announce that we're working on Ruby 3.5 stuff.

kaylareopelle (approved):
🦖

PR #3258, opened 2025-09-11
Add `logger` as a dependency

hannahramadan:
This PR adds `logger` as a dependency of the agent. 

It also updates tests to account for the removal of `logger` and `benchmark` from Ruby 3.5.0.

Full CI run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/17749090770

### DIFF
--- a/.github/workflows/scripts/setup_bundler
+++ b/.github/workflows/scripts/setup_bundler
@@ -82,12 +82,16 @@ function install_ruby_version_specific_gems {
     gem install --default digest:3.2.0
     gem install --default erb:4.0.4
     gem install --default erb:5.0.2
-    gem install --default logger:1.7.0 if RUBY_VERSION > '3.5.0'
+
     gem install --default securerandom:0.3.2
     gem install --default securerandom:0.4.1
     gem install --default set:1.1.2
     gem install --default strscan:3.0.4
     gem install --default strscan:3.1.5
+
+    if [[ $RUBY_VERSION =~ ^(2\.|3\.[0-4]) ]]; then
+    gem install --default logger:1.7.0
+    fi
 
     echo "DEBUG: Completed installing gems"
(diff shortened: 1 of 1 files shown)


.github/workflows/scripts/setup_bundler
```diff
@@ -81,8 +81,7 @@ function install_ruby_version_specific_gems {
     gem install --default digest:3.1.0
     gem install --default digest:3.2.0
     gem install --default erb:4.0.4
-    gem install --default erb:5.0.1
-    gem install --default logger:1.7.0
+    gem install --default erb:5.0.2
```
tannalynn, line 84: how come this one changed from 5.0.1 to 5.0.2?

.github/workflows/scripts/setup_bundler
```diff
@@ -81,8 +81,7 @@ function install_ruby_version_specific_gems {
     gem install --default digest:3.1.0
     gem install --default digest:3.2.0
     gem install --default erb:4.0.4
-    gem install --default erb:5.0.1
-    gem install --default logger:1.7.0
```
tannalynn, line 85: Instead of totally removing it, we should put this behind an if statement that checks the ruby version first, otherwise we'll see the issue we saw before with "i already activated a different version" the first time we run a test suite on other ruby versions

.github/workflows/scripts/setup_bundler
hannahramadan: We were seeing `You have already activated erb 5.0.1, but your Gemfile requires erb 5.0.2.` for some suites!

.github/workflows/scripts/setup_bundler
hannahramadan: I've made this edit ◡̈

PR #3264, opened 2025-09-22
Remove asserts in some tests for RubyGems

### LINKED ISSUE #3231: Investigate "nil versions are discouraged and will be deprecated in Rubygems 4"
I noticed this deprecation warning when looking at some failed tests: 

`nil versions are discouraged and will be deprecated in Rubygems 4`

We leverage `nil` versions to run our tests on the latest versions of gems without needing to update our code. Figure out if we need to stop this practice, and any other implications this change may have for our agent.

hannahramadan:
Delate some tests that were causing a RubyGems deprecation notice:
`nil versions are discouraged and will be deprecated in Rubygems 4`

In these tests, the agent called Gem::Version.new, which on [initialization](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/version.rb#L222) will [check](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/version.rb#L174)  to see if a version is nil. An empty string and nil evaluate to 0, so the tests below the removed ones should cover the scenario

closes [#3231](https://github.com/newrelic/newrelic-ruby-agent/issues/3231)

An original concern is how the multiverse uses `nil` to grab the latest version of a gem. In the agent, we never actually ended up building a gemfile to read `gem 'NAME', -> nil`. Instead, we [skip adding](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/test/multiverse/lib/multiverse/envfile.rb#L127-L137) a version if it is `nil`.

### DIFF
--- a/test/new_relic/helper_test.rb
+++ b/test/new_relic/helper_test.rb
@@ -87,7 +87,7 @@ class HelperTest < Minitest::Test
     assert(NewRelic::Helper.version_satisfied?(1, '<', 2))
     assert(NewRelic::Helper.version_satisfied?(1.2, '<', 1.3))
     assert(NewRelic::Helper.version_satisfied?(Gem::Version.new('1.2'), '<', Gem::Version.new('1.3')))
-    assert(NewRelic::Helper.version_satisfied?(nil, '<', 1))
+    assert(NewRelic::Helper.version_satisfied?('', '<', 1))
     assert(NewRelic::Helper.version_satisfied?('', '<', 1))
     assert(NewRelic::Helper.version_satisfied?('1.2', '<', 1.3))
   end
@@ -98,7 +98,7 @@ class HelperTest < Minitest::Test
     assert(NewRelic::Helper.version_satisfied?(1, '<=', 2))
     assert(NewRelic::Helper.version_satisfied?(1.2, '<=', 1.3))
     assert(NewRelic::Helper.version_satisfied?(Gem::Version.new('1.2'), '<=', Gem::Version.new('1.3')))
-    assert(NewRelic::Helper.version_satisfied?(nil, '<=', 1))
+    assert(NewRelic::Helper.version_satisfied?('', '<=', 1))
     assert(NewRelic::Helper.version_satisfied?('', '<=', 1))
     assert(NewRelic::Helper.version_satisfied?('1.2', '<=', 1.3))
   end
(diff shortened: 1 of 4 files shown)


test/new_relic/helper_test.rb
```diff
@@ -87,7 +87,7 @@ def test_version_satisfied_greater_than
     assert(NewRelic::Helper.version_satisfied?(1, '<', 2))
     assert(NewRelic::Helper.version_satisfied?(1.2, '<', 1.3))
     assert(NewRelic::Helper.version_satisfied?(Gem::Version.new('1.2'), '<', Gem::Version.new('1.3')))
-    assert(NewRelic::Helper.version_satisfied?(nil, '<', 1))
+    assert(NewRelic::Helper.version_satisfied?('', '<', 1))
```
kaylareopelle, line 90: I might be missing something, but this looks like the same test as the one below it? If so, what do you think about removing the assertion? 

```suggestion
```

test/new_relic/helper_test.rb
```diff
@@ -98,7 +98,7 @@ def test_version_satisfied_greater_than_or_equal_to
     assert(NewRelic::Helper.version_satisfied?(1, '<=', 2))
     assert(NewRelic::Helper.version_satisfied?(1.2, '<=', 1.3))
     assert(NewRelic::Helper.version_satisfied?(Gem::Version.new('1.2'), '<=', Gem::Version.new('1.3')))
-    assert(NewRelic::Helper.version_satisfied?(nil, '<=', 1))
+    assert(NewRelic::Helper.version_satisfied?('', '<=', 1))
```
kaylareopelle, line 100:
```suggestion
```

test/new_relic/helper_test.rb
```diff
@@ -109,7 +109,7 @@ def test_version_satisfied_less_than
     assert(NewRelic::Helper.version_satisfied?(2, '>', 1))
     assert(NewRelic::Helper.version_satisfied?(1.3, '>', 1.2))
     assert(NewRelic::Helper.version_satisfied?(Gem::Version.new('1.3'), '>', Gem::Version.new('1.2')))
-    assert(NewRelic::Helper.version_satisfied?(1, '>', nil))
+    assert(NewRelic::Helper.version_satisfied?(1, '>', ''))
```
kaylareopelle, line 110:
```suggestion
```

test/new_relic/helper_test.rb
```diff
@@ -120,7 +120,7 @@ def test_version_satisfied_less_than_or_equal_to
     assert(NewRelic::Helper.version_satisfied?(2, '>=', 1))
     assert(NewRelic::Helper.version_satisfied?(1.3, '>=', 1.2))
     assert(NewRelic::Helper.version_satisfied?(Gem::Version.new('1.3'), '>=', Gem::Version.new('1.2')))
-    assert(NewRelic::Helper.version_satisfied?(1, '>=', nil))
+    assert(NewRelic::Helper.version_satisfied?(1, '>=', ''))
```
kaylareopelle, line 120:
```suggestion
```

test/new_relic/helper_test.rb
hannahramadan: Oh wow I didn't even see that. You are right! delete delete delete

kaylareopelle (approved):

Issue #3244, opened 2025-08-08
Automate adding banner to major version release notes


kaylareopelle:
As part of our new EOL / Major version release process, we need to include the following header on our release notes for a major version update:

> Info Major Version Update: This version of the X agent is a SemVer MAJOR update and contains the following breaking changes. MAJOR versions may drop support for language runtimes that have reached End-of-Life according to the maintainer. Additionally, MAJOR versions may drop support for and remove certain instrumentation. For more details on these changes please see the migration guide <here>. 
[(internal source)](https://newrelic.atlassian.net/wiki/spaces/INST/pages/3955949842/APM+Agents+Major+Version+Release+Process)

To make things easier on us, let's just add this to our automation when we generate release notes for major versions so we don't need to remember.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-449731

PR #3267, opened 2025-09-25
Add major version banner to release notes

kaylareopelle:
Now, when a new major version is detected, the required banner will be added to the generated release notes file.

I slightly updated the text from `contains the following breaking changes` to `contains breaking changes`, assuming that we will make breaking changes clear in our changelog entries. If that isn't enough, let me know.

~~This also takes care of a little technical debt. Our `slack_notifications` tests weren't running because the file ended in `tests` instead of `test`. There are a few tests failing in each of the files. They're skipped for now, and the tests now live in a new directory that's closer to the `.github/workflows` directory where the tested files live.~~ This technical debt will be addressed in #3269 

✅ Full CI: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/18023776803

Closes: #3244 

Here is an example generated release note: 
```mdx
---
subject: Ruby agent
releaseDate: '2025-09-25'
version: 9.21.0
downloadLink: https://rubygems.org/downloads/newrelic_rpm-9.21.0.gem
features: ["Ensure compatibility with Ruby 3.5 change to Method#source_location"]
bugs: []
security: []
---
<Callout variant='important'>
  **Major Version Update:** This version of the Ruby agent is a SemVer MAJOR update and contains breaking changes. MAJOR versions may drop support for language runtimes that have reached End-of-Life according to the maintainer. Additionally, MAJOR versions may drop support for and remove certain instrumentation. For more details on these changes please see the migration guide [here](/docs/apm/agents/ruby-agent/getting-started/migration-10x-guide/).
</Callout>

<Callout variant="important">
  We recommend updating to the latest agent version as soon as it's available. If you can't upgrade to the latest version, update your agents to a version no more than 90 days old. Read more about [keeping agents up to date](/docs/new-relic-solutions/new-relic-one/install-configure/update-new-relic-agent/).

  See the New Relic Ruby agent [EOL policy](https://docs.newrelic.com/docs/apm/agents/ruby-agent/getting-started/ruby-agent-eol-policy/) for information about agent releases and support dates.
</Callout>

## dev

- **Feature: Ensure compatibility with Ruby 3.5 change to Method#source_location**

  Updated the agent to correctly parse the return value of Method#source_location, which is changing in Ruby 3.5 from a two-element to a five-element array. This change maintains support for older Ruby versions while adding support for the future release.
```


.github/workflows/scripts/generate_release_notes.rb
```diff
@@ -49,12 +53,28 @@ def build_release_content
...
+  def major_version_banner
+    <<~BANNER
+      <Callout variant='important'>
+        **Major Version Update:** This version of the Ruby agent is a SemVer MAJOR update and contains breaking changes. MAJOR versions may drop support for language runtimes that have reached End-of-Life according to the maintainer. Additionally, MAJOR versions may drop support for and remove certain instrumentation. For more details on these changes please see the migration guide.
+      </Callout>
+    BANNER
+  end
+
+  def major_bump?
+    previous_release_version = @split_changelog[2][/^ v(\d+\.\d+\.\d+)$/, 1]
```
kaylareopelle, line 71: TIL: https://stackoverflow.com/questions/519572/return-first-match-of-ruby-regex

kaylareopelle:
Issues authenticating on Docker Hub: https://www.dockerstatus.com/pages/incident/533c6539221ae15e3f000031/68d47a2f93c09e05486d93a9

newrelic_rpm.gemspec
```diff
@@ -52,7 +52,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency 'bundler'
   s.add_development_dependency 'feedjira', '3.2.1' unless ENV['CI'] || RUBY_VERSION < '2.5' # for Gabby
-  s.add_development_dependency 'httparty' unless ENV['CI'] # for perf tests and Gabby
+  s.add_development_dependency 'httparty'
```
kaylareopelle: Darn, we'll have to add HTTParty as a dependency to every Rails env if we want to run the Slack Notifier tests with the main unit_tests CI. We could eventually hit a problem with `feedjira` too since it has a similar condition and is also used for Gabby. 

Is there a way to exclude a test from the Rails environments without moving it to multiverse? 

Regardless, this is getting into territory unrelated to generating the release notes, so maybe it's better to revert the changes related to Slack Notifications and leave them for a different ticket. wdyt?

hannahramadan: I agree that we're entering a different territory! Maybe revert and I'll include this note in a new ticket

.github/workflows/scripts/generate_release_notes.rb
```diff
@@ -12,13 +12,23 @@ class GenerateReleaseNotes
   <Callout variant="important">
     We recommend updating to the latest agent version as soon as it's available. If you can't upgrade to the latest version, update your agents to a version no more than 90 days old. Read more about [keeping agents up to date](/docs/new-relic-solutions/new-relic-one/install-configure/update-new-relic-agent/).
 
-    See the New Relic Ruby agent [EOL policy](https://docs.newrelic.com/docs/apm/agents/ruby-agent/getting-started/ruby-agent-eol-policy/) for information about agent releases and support dates.
+    See the New Relic Ruby agent [EOL policy](/docs/apm/agents/ruby-agent/getting-started/ruby-agent-eol-policy/) for information about agent releases and support dates.
```
kaylareopelle, line 15: Just for consistency's sake, I removed the https prefix and went for a relative link to the file

.github/workflows/scripts/generate_release_notes.rb
```diff
@@ -49,12 +59,21 @@ def build_release_content
       bugs: #{metadata[:bugs]}
       security: #{metadata[:security]}
       #{DIVIDER}
-
+      #{MAJOR_VERSION_BANNER % NewRelic::VERSION::MAJOR if major_bump?}
```
kaylareopelle, line 62: Waiting to interpolate the constant until the metadata is built allows the test to use the stubbed constant instead of the real value.

.github/workflows/scripts/generate_release_notes.rb
```diff
@@ -12,13 +12,23 @@ class GenerateReleaseNotes
...
+  MAJOR_VERSION_BANNER = <<~BANNER
+  <Callout variant='important'>
+    **Major Version Update:** This version of the Ruby agent is a SemVer MAJOR update and contains breaking changes. MAJOR versions may drop support for language runtimes that have reached End-of-Life according to the maintainer. Additionally, MAJOR versions may drop support for and remove certain instrumentation. For more details on these changes please see the migration guide [here](/docs/apm/agents/ruby-agent/getting-started/migration-%sx-guide/).
+  </Callout>
+  BANNER
+
+  # pass the filename as an arg to simplify testing
+  def initialize(changelog_filename = 'CHANGELOG.md')
+    changelog = File.read(changelog_filename)
+    @split_changelog = changelog.split('##')
```
hannahramadan, line 27: nice clean up!

hannahramadan (approved):

PR #3270, opened 2025-09-29
Update pull_request_template.md

kaylareopelle:
Update the pull request template based on current expectations for community PRs.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,7 +5,7 @@ Describe the changes present in the pull request
 
 Submitter Checklist:
 - [ ] Include a link to the related GitHub issue, if applicable
-- [ ] Include a security review link, if applicable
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -5,7 +5,7 @@ Describe the changes present in the pull request
 
 Submitter Checklist:
 - [ ] Include a link to the related GitHub issue, if applicable
-- [ ] Include a security review link, if applicable
+- [ ] Add new tests for your change, if applicable
```
kaylareopelle, line 8: Idk if we need this. There's something similar referenced on line 16 of this file so it might be redundant.

tannalynn: i think it seems good to have, just in case

hannahramadan: I agree! I like the redundancy

PR #3293, opened 2025-10-09
Ruby 4.0 support - Add `logger` as a dependency

hannahramadan:
This PR adds the `logger` gem as an agent dependency. `logger` is no longer a part of Ruby default gems starting from Ruby 4.0.0, but agent code is quite intertwined with it and all other Ruby versions we support include it as a default, so adding it as a dependency makes sense. 

This PR also updates tests to account for the removal of `logger` and `benchmark` from Ruby 4.0.0.

Full CI run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/17749090770

(This PR was created when Ruby 3.5.0 was the planned release for Christmas 2025 instead of 4.0.0)

### DIFF
--- a/test/multiverse/suites/activemerchant/Envfile
+++ b/test/multiverse/suites/activemerchant/Envfile
@@ -16,6 +16,7 @@ def gem_list(activemerchant_version = nil)
     gem 'activesupport'
     gem 'nokogiri'
     gem 'minitest', '~> 5.1.0'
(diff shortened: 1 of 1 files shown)


kaylareopelle (approved):

hannahramadan:
Updated the changelog to reflect the eventual change of ruby 3.5 to ruby 4.0

PR #1461, opened 2022-09-23
Update deployment rake task

tannalynn:
_Before contributing, please read our [contributing guidelines](https://github.com/newrelic/newrelic-ruby-agent/blob/main/CONTRIBUTING.md) and [code of conduct](https://github.com/newrelic/.github/blob/master/CODE_OF_CONDUCT.md)._

# Overview
Describe the changes present in the pull request

Submitter Checklist:
- [ ] Include a link to the related GitHub issue, if applicable
- [ ] Include a security review link, if applicable

# Testing
The agent includes a suite of unit and functional tests which should be used to
verify your changes don't break existing functionality. These tests will run with 
Github Actions when a pull request is made. More details on running the tests locally can be found 
[here for our unit tests](https://github.com/newrelic/newrelic-ruby-agent/blob/main/test/README.md), 
and [here for our functional tests](https://github.com/newrelic/newrelic-ruby-agent/blob/main/test/multiverse/README.md).
For most contributions it is strongly recommended to add additional tests which
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Add performance label
- [ ] Perform appropriate level of performance testing
- [ ] Confirm all checks passed
- [ ] Add version label prior to acceptance

PR #3299, opened 2025-10-21
v10.0.0 - Remove `newrelic deployments` and Capistrano support

hannahramadan:
Remove support for `newrelic deployments` and Capistrano

### DIFF
--- a/bin/newrelic
+++ b/bin/newrelic
@@ -1,0 +1,8 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# This command has been renamed "newrelic_rpm"
+# executes one of the commands in the new_relic/commands directory
+# pass the name of the command as an argument
+
+load File.dirname(__FILE__) + '/newrelic_rpm'
--- a/lib/new_relic/cli/command.rb
+++ b/lib/new_relic/cli/command.rb
@@ -60,6 +60,12 @@ module NewRelic
         options = ARGV.options do |opts|
           script_name = File.basename($0)
 
+          # TODO: MAJOR VERSION - remove newrelic, deprecated since version x.xx
+          if /newrelic$/.match?(script_name)
+            $stdout.puts "warning: the 'newrelic' script has been renamed 'newrelic_rpm'"
+            script_name = 'newrelic_rpm'
+          end
+
           opts.banner = "Usage: #{script_name} [ #{@command_names.join(' | ')} ] [options]"
           opts.separator("use '#{script_name} <command> -h' to see detailed command options")
           opts


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -214,16 +214,10 @@ def self.host
         end
 
         def self.api_host
-          # only used for deployment task
```
hannahramadan, line 217: I'm unsure on this. We use `api_host` in tests and in the ServerMethods module. Instead of removing this method, its been refactored to not use `api_key`

README.md
```diff
@@ -77,13 +77,6 @@ For complete documentation on installing the New Relic Ruby agent, see the follo
 * [Update the agent](https://docs.newrelic.com/docs/agents/ruby-agent/installation/update-ruby-agent)
 * [Uninstall the Ruby agent](https://docs.newrelic.com/docs/agents/ruby-agent/installation/uninstall-ruby-agent)
 
-### Recording deploys
-
-The Ruby agent supports recording deployments in New Relic via a command line
-tool or Capistrano recipes. For more information on these features, see
-[our deployment documentation](http://docs.newrelic.com/docs/ruby/recording-deployments-with-the-ruby-agent)
-for more information.
```
kaylareopelle, line 85: What do you think about leaving this section in for a bit, but adding some sort of warning that we're removing the feature in 10.0 and the API will stop working at some point? I'm worried about people who aren't actively updating coming to the README looking for help when their deployments command randomly stops working.

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: I wonder what this looked like before we made the changes for API key? Do we still need the `'api'` condition for the ServerModule?

test/environments/norails/Rakefile
```diff
@@ -3,7 +3,6 @@
 # frozen_string_literal: true
 
 # Add your own tasks in files placed in lib/tasks ending in .rake,
-# for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
```
kaylareopelle, line 6: I don't think this has to be removed because it's part of the Rails boilerplate, but taking it out isn't hurting anything either. Just wanted to call it out in case it comes back if we generate Rails 8.1 apps before this gets merged.

lib/new_relic/agent/configuration/default_source.rb
tannalynn: This is the change when we added api key https://github.com/newrelic/newrelic-ruby-agent/pull/1461/files#diff-d116d3c8c7bebe204a18ee3983ac59c5d15bd70da7123f2a2c4a12f144bca181
B

Also, i think we can remove it from server module, because that is only used by the  api_server method which is just used by the deployments. I think we can just remove this method entirely and anything that's using it tbh. I think the only api we have ever used it for is deployments.

README.md
tannalynn: I think removing it is fine for when we release 10.0, we'll have the pinned issue. We wont' have it anymore as part of the agent, Plus we might forget about it and leave it there too long which could be confusing.

kaylareopelle: That makes sense to me. Thanks @tannalynn!

test/environments/norails/Rakefile
hannahramadan: Ohhh didn't realize this was boilerplate rails. Adding back in!

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: Nice! I've removed this method and it's references.

kaylareopelle: @hannahramadan - I found a few more references to `api_host` in the `newrelic.yml` files in multiverse suites and in the instrumentation generator template. Could you take a look and remove those too? Also understand if it's easier to take care of that cleanup in a separate PR.

hannahramadan: Thank you @kaylareopelle! cleaned those guys up

PR #3300, opened 2025-10-21
Deprecation notice - Capistrano

hannahramadan:
We will be removing support for Capistrano. Adding a deprecation notice.


kaylareopelle (approved):
🚀

Issue #3201, opened 2025-06-05
Perf testing: Create reports for docker monitoring info


tannalynn:
Create reports based on the data recorded from monitoring the docker containers. We should be able to do something very similar to php.

references: [slides](https://docs.google.com/presentation/d/1v9TkwJaka_S1LRb7nVbqgel6x0YPajsjUymjL02UuVg/edit?slide=id.p#slide=id.p), [php example](https://github.com/newrelic/php-agent-testing/tree/main/soak-tests/tools/soakmon-reports)

Blocked by: https://github.com/newrelic/newrelic-ruby-agent/issues/3200

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-422646

PR #3302, opened 2025-10-24
Add Perfverse

tannalynn:
Adds the Perfverse for performance testing the agent. Will run the specified agent tags and upload graphs comparing each tag to the workflow as artifacts.

Can be run [here](https://github.com/newrelic/newrelic-ruby-agent/actions/workflows/run_perfverse.yml). 

Currently only set up to run ruby 3.3.3 and rails 7 (because I reused the existing agent 2 query app). 
We could expand ruby and rails versions in the future and select which to use by adding more workflow inputs.

Currently can run up to 4 agent git tags plus disabled agent for comparison.

Added a readme for where to find it in GHA, and also how to run each step locally. 
resolves https://github.com/newrelic/newrelic-ruby-agent/issues/3201

Successful run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/18786625140

### DIFF
--- a/.github/actions/run_perfverse/action.yml
+++ b/.github/actions/run_perfverse/action.yml
@@ -2,7 +2,7 @@ name: 'Run Perfverse'
 description: 'Runs performance testing on the specified agent tag'
 inputs:
   agent_tag:
-    description: 'What ruby version to install'
+    description: 'What agent version to install'
     required: true
   agent_env_vars:
     description: '' 


.github/actions/run_perfverse/action.yml
```diff
@@ -0,0 +1,69 @@
+name: 'Run Perfverse'
+description: 'Runs performance testing on the specified agent tag'
+inputs:
+  agent_tag:
+    description: 'What ruby version to install'
```
kaylareopelle: ~~Wait, can we run multiple Ruby versions?~~

Ohh, I think this confused me initially. When I saw "what ruby version to install", I was thinking something like `3.3`, `3.4.5`, etc. Maybe "what agent tag to use"? or "what agent version to install"?

.github/workflows/run_perfverse.yml
```diff
@@ -14,11 +47,118 @@ jobs:
...
+          iterations: ${{ inputs.iterations }}
+          run_label: run_one
 
+  run_perf_tests_2:
+    runs-on: ubuntu-latest
+    if: ${{ inputs.run_2 != '' }}
+    steps:
+      - name: Configure git
+        run: 'git config --global init.defaultBranch main'
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # tag v5.0.0
```
kaylareopelle, line 65: Does git have to be reconfigured every time even though it was configured at the top level?

.rubocop.yml
```diff
@@ -14,6 +14,7 @@ AllCops:
   Exclude:
     - 'test/multiverse/suites/active_record/db/schema.rb'
     - 'test/multiverse/suites/active_record_pg/db/schema.rb'
+    - 'test/perfverse/rails7/**/*'
```
kaylareopelle, line 17: Was the rubocop for the test app pretty bad? Wondering if it's worth running once now.

.github/workflows/scripts/run_perf_tests.rb
```diff
@@ -0,0 +1,103 @@
+
```
kaylareopelle: do we want frozen string literal magic comment here?

.github/workflows/scripts/run_perf_tests.rb
```diff
@@ -0,0 +1,103 @@
...
+  app_name = "ruby_perf_app_#{ENV['TEST_TAG']}_#{agent_tag}_#{iteration}"
+  output_line("Running ruby app in background. Name: #{app_name}")
+  cpu_mem = '--cpus 4 --memory 2G'
+
+  Thread.new do
+    run_command("cd ./test/perfverse/ && docker run --rm --name #{app_name} #{cpu_mem} #{env_str} -e NEW_RELIC_LICENSE_KEY=$NR_LICENSE_KEY -e NEW_RELIC_APP_NAME=#{app_name} -e NEW_RELIC_HOST=staging-collector.newrelic.com -e s -p 3000:3000 ruby_perf_app:local")
+  end
+  sleep 2
+  thread = run_docker_report(agent_tag, app_name, iteration)
+  sleep 1
```
kaylareopelle, line 59: are the sleeps just to make sure the app is fully running before the report is generated?

test/perfverse/docker_monitor/docker-monitor-threading.py
```diff
@@ -0,0 +1,113 @@
+import os
```
kaylareopelle, line 1: Could the python files be translated into Ruby? (Or does it even make any sense to?)

test/perfverse/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+# start and daemonize a Rails server process
+bundle exec rails server -b 0.0.0.0
+
+# generate load
+# ./tester
```
kaylareopelle: Is this just if you're running it locally, without GHA?

test/perfverse/newrelic.yml
```diff
@@ -0,0 +1,15 @@
+common: &default_settings
+  app_name: Ruby Agent Perf testing
+  log_level: info
+  log_file_path: STDOUT
+  code_level_metrics:
+    enabled: true
```
kaylareopelle: isn't this on by default?
```suggestion
```

test/perfverse/newrelic.yml
```diff
@@ -0,0 +1,15 @@
...
+  audit_log:
+    enabled: false
+  heroku:
+    use_dyno_names: true
```
kaylareopelle: these are also set to these values by default. do we need to specify them here?
```suggestion
```

kaylareopelle (commented):
This is so awesome, @tannalynn! Thanks for your work here.

I think being able to test different Ruby versions would be cool down the line, but probably a pretty rare thing we'd need to test.

.github/workflows/scripts/run_perf_tests.rb
tannalynn: Yeah, the first sleep is to make sure the container has time to start before dockermon is started, and then the other sleep is to make sure everything is ready for the traffic generator to begin

test/perfverse/docker_monitor/docker-monitor-threading.py
tannalynn: I'm not sure they if they could be changed to ruby, but maybe. It's using a python library to access the docker stats, so we'd need to find one that does the same thing in Ruby. I don't think we really need to though. This file didn't change a whole lot from the php version, most of the changes are just changing the naming of things or removing attributes we aren't using. Also, I think this file is something that is less likely to need to be updated in the future since it's just getting the stats.

test/perfverse/entrypoint.sh
tannalynn: to be totally honest, i don't know, that's something that was part of the test app when it was agent 2 query. i can just like remove it though bc i haven't used that

test/perfverse/newrelic.yml
tannalynn: oh yeah I just copied this from the agent 2 query app, and didn't really think twice about it tbh

.rubocop.yml
tannalynn: yeah it was pretty mad about a lot of stuff, that's why i added it in the exclude here. Even some stuff where it was like "that files actually a syntax error" because it's assuming it needs to work on older ruby versions (because thats what we have it configured for for the agent code), but it's only running on 3.3, so there is not syntax error. 
The syntax error things are what made me not even try to get rubocop happy with it. Plus it's not like, real code we need to maintain anyways.

.github/workflows/scripts/run_perf_tests.rb
tannalynn: sure, why not lol https://github.com/newrelic/newrelic-ruby-agent/pull/3302/commits/890d01a77a5960fc9f452e228ed6e3e314086adb

.github/actions/run_perfverse/action.yml
tannalynn: oh yeah, that was originally ruby version and then i changed it and forgot to change the description my bad 
https://github.com/newrelic/newrelic-ruby-agent/pull/3302/commits/be7994abf7d11459fbb846a21fe63bb019952ddc

.github/workflows/run_perfverse.yml
tannalynn: i'm not sure what you mean by configured at the top level?
but i dont really know, i just know this is how we do it in all of our workflows, so I am doing it this way also

test/perfverse/docker_monitor/docker-monitor-threading.py
kaylareopelle: Nice, sounds good!

.rubocop.yml
kaylareopelle: ooh, hadn't considered the older ruby version issues. that makes sense!

.github/workflows/run_perfverse.yml
kaylareopelle: ah, I didn't realize it was in every single job! just saw duplication and felt puzzled. now that i've looked into it more, seems like it's something the checkouts action is aware of: 
https://github.com/actions/checkout/pull/465 
https://github.com/actions/checkout/issues/2281

PR #3304, opened 2025-10-24
Remove railsedge from CI

kaylareopelle:
This is just until we have time to update the test suites for Rails 8.1 support


kaylareopelle:
Issue #3305 will re-enable these tests.

PR #3306, opened 2025-10-24
Add new cross-agent tests, update remote parent sampling decisions

kaylareopelle:
The new cross-agent tests expect us to use the remote parent sampling configuration options in additional cases:
* when the tracestate header is missing, but the traceparent is present
* when only the newrelic header is present

This PR accounts for those cases.

There's a lot of repetition here. I plan to do a larger refactoring of the sampling decision logic when we introduce the root sampler and trace ratio based sampler option into the mix.

### DIFF
--- a/lib/new_relic/agent/distributed_tracing/trace_context_payload.rb
+++ b/lib/new_relic/agent/distributed_tracing/trace_context_payload.rb
@@ -96,6 +96,14 @@ module NewRelic
         @timestamp = timestamp
       end
 
+      # This is an invalid instance of the TraceContextPayload that is used to
+      # help us determine the sampling decision for an incoming trace when the
+      # tracestate header is missing. Creating an invalid instance allows us
+      # to call methods like .sampled on it without throwing errors.
+      # The timestamp value will represent the time it was created/when this
+      # class was initialized.
+      #
+      # @api private
       INVALID = create
 
       def parent_type


lib/new_relic/agent/transaction/distributed_tracing.rb
```diff
@@ -161,6 +161,25 @@ def assign_payload_and_sampling_params(payload)
           transaction.parent_span_id = payload.id
 
           unless payload.sampled.nil?
+            if payload.sampled == true
```
tannalynn, line 164: So to make sure I understand, this is adding the logic we did for the trace parent originally into the NR DT section, correct?

lib/new_relic/agent/transaction/trace_context.rb
```diff
@@ -132,11 +132,17 @@ def accept_trace_context(header_data)
...
 
-          return false unless payload = assign_trace_state_payload
+          trace_flags = header_data.trace_parent['trace_flags']
+          payload = assign_trace_state_payload
 
-          transaction.distributed_tracer.parent_transaction_id = payload.transaction_id
+          if payload
+            determine_sampling_decision(payload, trace_flags)
+          else
+            determine_sampling_decision({}, trace_flags)
```
tannalynn: For passing `{}`, that's for an empty tracestate header, right? 
I ask because in the `use_nr_tracestate_sampled` method, we call `payload.sampled.nil?`, which I think could give an error if payload is just `{}`. That could happen I think if the config is just set to default and using the adaptive sampler, right?

lib/new_relic/agent/transaction/distributed_tracing.rb
kaylareopelle: Yes, that's correct. So in cases where only the `newrelic` header is present, we'll now use the remote sampling configs to make the sampling decision.

It's based on [this line in the spec](https://source.datanerd.us/agents/agent-specs/blob/main/distributed_tracing/Distributed-Tracing.md?plain=1#L31), and [these cross-agent tests](https://source.datanerd.us/agents/cross_agent_tests/blob/f9a93cfb579bcbbc17f29adb860ef5d5da6b9d87/distributed_tracing/trace_context.json#L178-L263).

lib/new_relic/agent/transaction/trace_context.rb
kaylareopelle: Yes, that's for getting the sampling decision with an empty tracestate header. 

Good question! I'm not sure how that would behave and don't think there's a case for that right now. I'll give it a try.

kaylareopelle: @tannalynn - Good catch! You're right! An error was silently raised by using the empty hash when `use_nr_tracestate_sampled` was called. All the tests still passed and the error never reached the surface.

As an alternative, I made an invalid instance of the `TraceContextPayload` and saved it in a constant. That invalid instance will be used, so calls to `payload.sampled.nil?` behave accordingly. See [2750102](https://github.com/newrelic/newrelic-ruby-agent/pull/3306/commits/27501028bb6afdb43948516a942a72fc8db9370c). How does that adjustment look to you?

kaylareopelle: Haha well, I guess something's off between my local environment and the CI because everything's failing now. I'll take another look.

kaylareopelle: Okay, same premise is still true, but the `INVALID` constant needed to be defined after the initialize method. See: 4b05da9

tannalynn: Looks good, thank you for checking on this!

tannalynn (dismissed):

tannalynn (approved):

lib/new_relic/agent/transaction/distributed_tracing.rb
```diff
@@ -161,6 +161,25 @@ def assign_payload_and_sampling_params(payload)
...
+        end
+
+        def set_priority_and_sampled_newrelic_header(config, payload)
+          if config == 'always_on'
+            transaction.sampled = true
+            transaction.priority = 2.0
+          elsif config == 'always_off'
+            transaction.sampled = false
+            transaction.priority = 0
+          else # case for 'default' value
```
hannahramadan, line 182: helpful comment thank you ◡̈

lib/new_relic/agent/transaction/trace_context.rb
```diff
@@ -132,11 +132,17 @@ def accept_trace_context(header_data)
...
 
-          return false unless payload = assign_trace_state_payload
+          trace_flags = header_data.trace_parent['trace_flags']
+          payload = assign_trace_state_payload
 
-          transaction.distributed_tracer.parent_transaction_id = payload.transaction_id
+          if payload
+            determine_sampling_decision(payload, trace_flags)
+          else
+            determine_sampling_decision(TraceContextPayload::INVALID, trace_flags)
```
hannahramadan, line 141: Is this understanding true - the constant `TraceContextPayload::INVALID` is a new instance of TraceContextPayload, which is the result of the `create` method? I think the name `INVALID` is tripping me up.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
   Sending application deployment information using a Capistrano recipe is deprecated and will be removed in agent version 10.0.0. For recording deployments, please see our guide to [Change Tracking](https://docs.newrelic.com/docs/change-tracking/change-tracking-introduction/) for a list of available options.
 
+- **Feature: Use remote parent sampling configurations for decisions in more scenarios**
+
+  Previously, the `distributed_tracing.sampler.remote_parent_sampled` and `distributed_tracing.sampler.remote_parent_not_sampled` configuration options were used for the sampling decision only when the `traceparent` and `tracestate` headers were present. Now, these configuration options are applied in cases when the `tracestate` header is missing and when only the `newrelic` header is available. [PR#3306](https://github.com/newrelic/newrelic-ruby-agent/pull/3306)
```
hannahramadan: Thoughts on adding a line for what this change means for users? (I'm not set on this suggested change/maybe there is better wording)
```suggestion
  Previously, the `distributed_tracing.sampler.remote_parent_sampled` and `distributed_tracing.sampler.remote_parent_not_sampled` configuration options were used for the sampling decision only when the `traceparent` and `tracestate` headers were present. Now, these configuration options are applied in cases when the `tracestate` header is missing and when only the `newrelic` header is available. This change makes distributed trace sampling more consistent and predictable. [PR#3306](https://github.com/newrelic/newrelic-ruby-agent/pull/3306)
```

lib/new_relic/agent/transaction/trace_context.rb
kaylareopelle: Somewhat! `INVALID` is a constant that's the result of a `TraceContextPayload.create` call without any arguments passed to the method call. The object ID should be the same every time, so we're not creating new instances whenever this is called. So it is a TraceContextPayload, but an invalid one. 

This results in a TraceContextPayload that looks like this: 
```
#<NewRelic::Agent::TraceContextPayload:0x000000011bc83fe0
 @id=nil,
 @parent_account_id=nil,
 @parent_app_id=nil,
 @parent_type_id=0,
 @priority=nil,
 @sampled=nil,
 @timestamp=1761761672348,
 @transaction_id=nil,
 @version=0>
 ```

The `create` method has some default values it provides for `parent_type_id`, `version`, and it also creates a `timestamp`. 

There's an instance method in this class called `valid?`, which returns `false` when it's called on `INVALID`. 

I went back and forth on whether `create` or `new` was the better approach to create this constant. The object created if we call `new` with all `nil` arguments looks like: 
```
#<NewRelic::Agent::TraceContextPayload:0x000000011dd35fb8
 @id=nil,
 @parent_account_id=nil,
 @parent_app_id=nil,
 @parent_type_id=nil,
 @priority=nil,
 @sampled=nil,
 @timestamp=nil,
 @transaction_id=nil,
 @version=nil>
 ```
 
 When we call `valid?` on this object, it returns `nil`.
 
 They should both behave the same way in the context of this method. If we took the `new` approach, `EMPTY` may be a more appropriate name.
 
 I'm also open to other ideas on naming or finding a solution here! The problem it's trying to solve is allowing `payload.sampled.nil?` calls in the `use_nr_tracestate_sampled` method to behave appropriately without erroring out.

CHANGELOG.md
kaylareopelle: This is a great add, thank you!

lib/new_relic/agent/transaction/trace_context.rb
hannahramadan: Thanks for chatting live about this!

hannahramadan (approved):

PR #3307, opened 2025-10-24
Stepping stone: Accept default, adaptive, always_on and always off for three samplers

kaylareopelle:
This is just a step toward implementing the trace ID ratio based sampler. 

This PR creates a consistent pattern for the root sampler, as well as when the sampling decision is made in trace context, or using the newrelic header (in `lib/new_relic/agent/transaction/distributed_tracing.rb`). 

It also removes old code that isn't being used, the `enforce_fallback` method. The configs that used this option no longer exist. The `allowlist` functionality is much more performant and is used here instead. 

Case statements are evaluated in order, so I ordered the when statements by my best judgment of what is most and least likely to occur.

### DIFF
--- a/test/fixtures/cross_agent_tests/distributed_tracing/trace_context.json
+++ b/test/fixtures/cross_agent_tests/distributed_tracing/trace_context.json
@@ -263,7 +263,7 @@
     }
   },
   {
-    "test_name": "skip_no_headers_root_always_off",
+    "test_name": "no_headers_root_always_off",
     "comment": "Traces originating from current service are never sampled",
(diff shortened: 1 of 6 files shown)


test/fixtures/cross_agent_tests/distributed_tracing/trace_context.json
```diff
@@ -289,6 +289,33 @@
       }
     }
   },
+  {
```
kaylareopelle, line 292: This isn't part of the cross-agent tests right now, but I think I'll open a PR for it.

PR #3308, opened 2025-10-31
Rename OpenTelemetry configs

kaylareopelle:
These configs are private so this doesn't need to be a breaking change. The name we've agreed on across the agents is `opentelemetry.*`


tannalynn (approved):
lgtm. since they are private we also dont need a changelog entry for it 👍

Issue #3274, opened 2025-09-30
Remove support for Ruby 2.4 and 2.5


kaylareopelle:
Raise the minimum Ruby version for newrelic_rpm to 2.6.

- [x] Add changelog entry
- [ ] Update NR docs 
- [x] Remove 2.4 and 2.5 from CI
- [x] Update the min ruby version in the gemspec
- [x] Look at setup-bundler script and remove anything specific to 2.4 or 2.5
- [ ] ~Move Rails 4.0 and 4.1 tests to 2.6 [(should be compatible)](https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html)~
- [x] Address all `TODO: OLD RUBIES` comments related to the removal

PR #3314, opened 2025-11-04
Remove support for Ruby 2.4 and 2.5

tannalynn:
- Removes 2.4 and 2.5 from all CI test runs and setup steps
- Update min ruby in gemspec
- Addressed `TODO: OLD RUBIES` for ruby < 2.6
- Adds changelog entry for removing 2.4 and 2.5
- Updates version bump rake task and generating release notes to work with new Major Version indicator in changelog

[full ci run](https://github.com/newrelic/newrelic-ruby-agent/actions/runs/19081113172)
resolves https://github.com/newrelic/newrelic-ruby-agent/issues/3274

### DIFF
--- a/lib/tasks/helpers/version_bump.rb
+++ b/lib/tasks/helpers/version_bump.rb
@@ -46,7 +46,7 @@ module VersionBump
   def self.determine_bump_type
     file = read_file('CHANGELOG.md')
     lines = file.split('## ')[1].split('- **')
-    return MAJOR if lines.any? { |line| line.include?('Refactor:') }
+    return MAJOR if lines.any? { |line| line.include?('Breaking Change:') }
     return MINOR if lines.any? { |line| line.include?('Feature:') }
 
     TINY
--- a/.github/workflows/scripts/generate_release_notes.rb
+++ b/.github/workflows/scripts/generate_release_notes.rb
@@ -35,8 +35,8 @@ class GenerateReleaseNotes
     titles.each do |t|
       category = t.split(':').first
       case category
-      when 'Refactor'
-        metadata[:features] << t.delete_prefix('Refactor: ')
+      when 'Breaking Change'
+        metadata[:features] << t.delete_prefix('Breaking Change: ')
       when 'Feature'
         metadata[:features] << t.delete_prefix('Feature: ')
       when 'Bugfix'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## dev
 
-- **Refactor: Removed support for Ruby 2.4 and 2.5**
+- **Breaking Change: Removed support for Ruby 2.4 and 2.5**
   Support for Ruby versions 2.4 and 2.5 has been removed. The new minimum required Ruby version is now 2.6. [PR#3314](https://github.com/newrelic/newrelic-ruby-agent/pull/3314)
 
 ## v9.23.0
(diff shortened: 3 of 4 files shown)


.github/actions/run_unit_tests/action.yml
```diff
@@ -30,12 +29,6 @@ runs:
         key: ${{ inputs.ruby-version }}
         map: |
           {
-            "2.4.10": {
-              "rails": "norails,rails52,rails51,rails50,rails42,rails41,rails40"
-            },
-            "2.5.9": {
-              "rails": "norails,rails61,rails60,rails52,rails51,rails50,rails42"
-            },
```
kaylareopelle: Does this mean we need to mention we're dropping support for 4.1 / 4.0? Or just leave it untested?

newrelic_rpm.gemspec
kaylareopelle, line 1: There's a Ruby version condition in this file for the feedjira dependency that can be cleaned up too

infinite_tracing/lib/infinite_tracing/transformer.rb
kaylareopelle, line 1: speaking of infinite tracing, the min version is 2.5; what do you think about bumping the gemspec?

kaylareopelle, line 1: also see a 2.5 transform_keys comment in connection.rb; not sure why that wasn't picked up by rubocop?

kaylareopelle (commented):
Some other files to look at: 
* lib/new_relic/agent/instrumentation/mongo_db_command_subscriber.rb - operation method
* lib/new_relic/agent/new_relic_service.rb - has 3 mentions of 2.5

infinite_tracing/lib/infinite_tracing/transformer.rb
tannalynn: I'm confused by what rubocop would pickup? It's just a comment explaining that transform_keys is being used there because it's infinite tracing and doesn't work on older ruby versions anyways. 
In any case, i removed the comment. https://github.com/newrelic/newrelic-ruby-agent/pull/3314/commits/49b65777d6055b04504ba3ced7fc3709232981f1

tannalynn: sure https://github.com/newrelic/newrelic-ruby-agent/pull/3314/commits/7100ed7e71b148251da87a582d505dcb9e46a34a

newrelic_rpm.gemspec
tannalynn: 👍 https://github.com/newrelic/newrelic-ruby-agent/pull/3314/commits/486f539a6cafa5b5945146257306b6d722361560

tannalynn:
> Requesting changes on here to update the prefix from "Refactor:" to "Breaking Change:", other than that this looks good! Thank you!

updated https://github.com/newrelic/newrelic-ruby-agent/pull/3314/commits/cb849e5503f91e8c50014ce1b1cbf5b13d734554

lib/tasks/helpers/version_bump.rb
```diff
@@ -46,7 +46,7 @@ def self.write_file(path, file)
   def self.determine_bump_type
     file = read_file('CHANGELOG.md')
     lines = file.split('## ')[1].split('- **')
-    return MAJOR if lines.first.include?('Major version')
+    return MAJOR if lines.any? { |line| line.include?('Breaking Change:') }
```
hannahramadan, line 49: This makes sense! I guess we would never do a major if there wasn't a breaking change, so presumably all Majors would have this. Nice

hannahramadan (approved):
Looks good! Addressing other Major TODOs in another PR makes sense.

PR #3315, opened 2025-11-04
Update samplers to include root and trace id ratio based options

### LINKED ISSUE #3284: Implement Trace ID Ratio-Based Sampler
In accordance with the agent spec and OTel spec, implement the Trace ID ratio-based sampler in our agent.

Agent Spec PR: https://source.datanerd.us/agents/agent-specs/pull/740/
OTel Spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md

kaylareopelle:
This PR: 
* Adds the trace id ratio based sampling options, along with ratio configs for each of those options
  * This sampling matches the OpenTelemetry Ruby SDK sampler and not the recently defined algorithm that will be used for the Probability Sampler
* Adds the `adaptive` option to all DT samplers
* Adds the ability to configure the root sampling decision to the same options as the remote parent samplers: `default`, `adaptive`, `trace_id_ratio_based`, `always_on`, `always_off`
  * If you're interested in the more minimal addition of the root sampler, you can review #3307

Closes #3284 

**Questions**
- [ ] Do we want to include this in our 10.0 release? Or a 9.x release? 

**Todos** 
- [ ] Add changelog entry


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -281,16 +282,6 @@ def self.convert_to_constant_list(raw_value)
           constants.compact!
           constants
         end
-
-        def self.enforce_fallback(allowed_values: nil, fallback: nil)
```
kaylareopelle: This wasn't being used. We have the allowlist functionality now that is more performant

test/fixtures/cross_agent_tests/distributed_tracing/trace_context.json
```diff
@@ -290,7 +290,34 @@
     }
   },
   {
-    "test_name": "skip_no_headers_root_uses_ratio_sampler",
+    "test_name": "no_headers_root_always_on",
```
kaylareopelle, line 293: This isn't currently part of the cross-agent tests. If we think it's valuable, I'll open a PR.

test/new_relic/agent/distributed_tracing/trace_context_cross_agent_test.rb
```diff
@@ -51,9 +51,17 @@ def self.focus_tests
...
                 :'span_events.enabled' => test_case.fetch('span_events_enabled', true),
+                :'distributed_tracing.sampler.root' => test_case.fetch('root', 'default'),
                 :'distributed_tracing.sampler.remote_parent_sampled' => test_case.fetch('remote_parent_sampled', 'default'),
                 :'distributed_tracing.sampler.remote_parent_not_sampled' => test_case.fetch('remote_parent_not_sampled', 'default')
               }
+
+              if test_case.key?('ratio')
+                strategies = %w[root remote_parent_sampled remote_parent_not_sampled]
+                key = strategies.find { |f| test_case[f] == 'trace_id_ratio_based' }
+                config["distributed_tracing.sampler.#{key}.trace_id_ratio_based.ratio".to_sym] = test_case['ratio']
```
kaylareopelle, line 62: This expects only one ratio to ever be configured in a test case, which is how the fixture is currently set up. I can adjust it to be more robust if you'd like to account for the possibility of multiple configs having ratios.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1463,19 +1454,68 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => true,
           :description => 'Distributed tracing lets you see the path that a request takes through your distributed system. Enabling distributed tracing changes the behavior of some New Relic features, so carefully consult the [transition guide](/docs/transition-guide-distributed-tracing) before you enable this feature.'
         },
+        :'distributed_tracing.sampler.root' => {
+          :default => 'default',
+          :public => true,
+          :type => String,
+          :allowed_from_server => false,
```
kaylareopelle, line 1461: I don't think we want to allow these from the server at this time, but happy to change.

lib/new_relic/agent/transaction/sampling_decision.rb
```diff
@@ -0,0 +1,125 @@
...
+        #
+        # @param transaction [Transaction] the transaction to sample
+        # @return [Hash] with keys :sampled (Boolean) and :priority (Float)
+        #
+        # @api private
+        def determine_root_sampling(transaction)
+          trace_id = transaction.trace_id
+          config_value = NewRelic::Agent.config[:'distributed_tracing.sampler.root']
+
+          case config_value
```
kaylareopelle, line 27: We could consider refactoring this to memoize the config value and skip the case statement evaluation every time. The downside is that we would have a harder time adapting if we get a config value change from server side config while the agent is running. I don't know if that's intended as a future feature.

kaylareopelle:
The refactor went too far. Closing in favor of #3330

PR #3319, opened 2025-11-07
Validate `record_custom_event` type

### LINKED ISSUE #3126: `Agent#record_custom_event` should raise on invalid `event_type`
## Description

While the constraints of valid `eveny_type`s are well documented:

>     # @param [Symbol or String] event_type The name of the event type to record. Event
>     #                           types must consist of only alphanumeric
>     #                           characters, '_', ':', or ' '.

(https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent.rb#L468-L470)

It's still easier than it should be to send an invalid `event_type` to New Relic, never see that data in New Relic, and then just presume the event never fires (might happen if you are instrumenting events in rare paths).

I believe a simple developer experience improvement would be to raise an `ArgumentError` if `event_type` does not meet the documented constraints using a regular expression.

Happy to help implement this if the maintainers agree it would be useful / don't foresee pitfalls here.

hannahramadan:
The `NewRelic::Agent#record_custom_event` API will now raise an `ArgumentError` on an invalid `event_type`. 

A valid `event_type` must consist of only alphanumeric characters, `_`, `:`, or ` `.

closes #3126


lib/new_relic/agent.rb
```diff
@@ -117,6 +117,9 @@ class SerializationError < StandardError; end
     # been configured for
     AUTOMATIC_TRACER_MAX_ATTEMPTS = 60 # 60 = try about twice a second for 30 seconds
 
+    # Event types must consist of only alphanumeric characters, '_', ':', or ' '.
+    VALID_CUSTOM_EVENT_TYPE = /\A[\w: ]+\z/
```
kaylareopelle: I took a look at the [agent spec (internal)](https://source.datanerd.us/agents/agent-specs/blob/main/Custom-Events.md) for custom event names, and I don't see anything about the alphanumeric/character limitations. Where did you find these?

kaylareopelle, line 121: I see in our API documentation that ` ` (spaces) should be allowed in the event names. I looked at the agent spec for custom events and didn't see anything about naming requirements. The [public New Relic doc for custom events](https://docs.newrelic.com/docs/data-apis/custom-data/custom-events/data-requirements-limits-custom-event-data/#general:~:text=Naming%20syntax%20for%20events%20and%20attributes) doesn't include spaces. Have you tested out whether spaces are valid? Wondering if the docs/specs need to be updated.

hannahramadan: Here is a [new test](https://github.com/newrelic/newrelic-ruby-agent/pull/3319/files#diff-a55158c03830ae80f6d01bbf38fff4adbbbc098553b4107ecc5e18cfbda1b3c1R43) that ensures spaces are okay.

Great research on this. I wouldn't have thought to double check this API's requirements outside of our agent. This current regex would fail `.` and `/`, which docs say is okay. I will look into this.

hannahramadan: Looks like forward slashes [were added](https://github.com/newrelic/docs-website/commit/7654b2b2f522a4274cffaa2ce2d39bd957026b79) to NR docs 2 years ago. I don't see anything in specs. What do you think about allowing `.` and `/`, but also keeping spaces. I tested spaces, and it does send/show up in the UI

hannahramadan: Commit: https://github.com/newrelic/newrelic-ruby-agent/pull/3319/commits/7b0c1abbe59a226a63e3811b72479f448330ba3e

kaylareopelle: The text is a little tricky to read, but I think it says only attribute names (not event names) can include `.` or `/`. If that's how you're interpreting it too, I think we shouldn't include those in the name validation.

Thanks for running these UI tests!

hannahramadan: Yes you are correct! Event name vs attribute name. I am reverting this change lol.

tannalynn (commented):
We should probably add a changelog entry for this since customers could see different behavior now if they were using it wrong previously. That way they'll know why it is acting differently for them.

kaylareopelle (changes requested):
I'm requesting changes for the changelog entry, just so that I don't keep opening up the PR wondering why I haven't approved it yet! 😅

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
 - **Breaking Change: Rename `bin/newrelic` command to `bin/newrelic_rpm`**
   The executable file for the agent's CLI has been renamed from `bin/newrelic` to `bin/newrelic_rpm`. This change resolves a name collision with the standalone New Relic CLI tool. [PR#3323](https://github.com/newrelic/newrelic-ruby-agent/pull/3323)
 
+- **Feature: Add argument validation for the `Agent#record_custom_event` API**
```
kaylareopelle: I think customers have to put `NewRelic::` in front of agent for the API call to work.

```suggestion
- **Feature: Add argument validation for the `NewRelic::Agent#record_custom_event` API**
```

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
...
+  The `Agent#record_custom_event` API now raises an `ArgumentError` when an invalid `event_type` is provided. A valid event type must consist only of alphanumeric characters, underscores (`_`), colons (`:`), or spaces (` `). [PR#3319](https://github.com/newrelic/newrelic-ruby-agent/pull/3319)
```
kaylareopelle: We'll have to go back and fix this for the other 10.0.0 entries, but we normally have an empty line between the headings and the descriptions:

```suggestion

  The `NewRelic::Agent#record_custom_event` API now raises an `ArgumentError` when an invalid `event_type` is provided. A valid event type must consist only of alphanumeric characters, underscores (`_`), colons (`:`), or spaces (` `). [PR#3319](https://github.com/newrelic/newrelic-ruby-agent/pull/3319)
```

kaylareopelle (approved):

PR #3320, opened 2025-11-07
Rename ActiveJob segments

hannahramadan:
Active Job segments have been renamed to include their job class.

`Ruby/ActiveJob/#{queue}/#{method}` is now `Ruby/ActiveJob/#{job_class}/#{queue}/#{method}`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,16 @@
 
 ## dev
 
-- **Breaking Change: Removed support for Ruby 2.4 and 2.5**
+- **Breaking Change: Remove support for Ruby 2.4 and 2.5**
   Support for Ruby versions 2.4 and 2.5 has been removed. The new minimum required Ruby version is now 2.6. [PR#3314](https://github.com/newrelic/newrelic-ruby-agent/pull/3314)
 
 - **Breaking Change: Rename ActiveJob metrics**
   ActiveJob metrics have been updated to include the job's class name for more specific reporting. This is a breaking change and may require updating custom dashboards or alerts. [PR#3320](https://github.com/newrelic/newrelic-ruby-agent/pull/3320)
     - Old format: `Ruby/ActiveJob/<QueueName>/<Method>`
     - New format: `Ruby/ActiveJob/<ClassName>/<QueueName>/<Method>`
+
+- **Breaking Change: Rename `bin/newrelic` command to `bin/newrelic_rpm`**
+  The executable file for the agent's CLI has been renamed from `bin/newrelic` to `bin/newrelic_rpm`. This change resolves a name collision with the standalone New Relic CLI tool. [PR#3323](https://github.com/newrelic/newrelic-ruby-agent/pull/3323)
 
 ## v9.23.0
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 - **Breaking Change: Removed support for Ruby 2.4 and 2.5**
   Support for Ruby versions 2.4 and 2.5 has been removed. The new minimum required Ruby version is now 2.6. [PR#3314](https://github.com/newrelic/newrelic-ruby-agent/pull/3314)
 
+- **Breaking Change: Renamed ActiveJob metrics**
```
kaylareopelle: Shoot, just noticed that the heading for this entry and the one before are both past tense. We normally use the imperative.

lib/new_relic/agent/instrumentation/active_job_subscriber.rb
```diff
@@ -22,8 +22,9 @@ def metric_name(name, payload)
           job = payload[:job] || payload[:jobs].first
 
           queue = job.queue_name
+          job_class = job.class.name.split('::').last
```
kaylareopelle: I did a little performance testing (with help from Claude) and I think we can make this faster and with fewer allocations:
```suggestion
          job_class = class_name.include?('::') ?
  class_name[class_name.rindex('::')+2..-1] : class_name
```

Essentially, `split` creates an array every time with new strings for every split in the array. For deeply nested classes, `rindex` + `slice` (the `+2..-1` part) is faster than calling `split` + `last`. 

Since this will run for every call, I think the more obscure approach is worth it.

hannahramadan: Nice kayla! I didn't realized `split` was creating a new object. Great idea to performance test difference approaches and this is a great add.

CHANGELOG.md
```suggestion
- **Breaking Change: Rename ActiveJob metrics**
```

hannahramadan: Has been fixed ◡̈

test/new_relic/agent/instrumentation/rails/active_job_subscriber.rb
```diff
@@ -31,12 +31,12 @@ class ActiveJobSubscriberTest < Minitest::Test
     SUBSCRIBER = NewRelic::Agent::Instrumentation::ActiveJobSubscriber.new
 
     def test_segment_naming_with_unknown_method
-      assert_equal 'Ruby/ActiveJob/default/Unknown',
+      assert_equal 'Ruby/ActiveJob/TestJob/default/Unknown',
```
kaylareopelle, line 44: Do we have tests somewhere with more deeply nested names? something like `MyApp::Jobs::EmailNotificationJob`? The code behaves correctly with names like this, but I'm worried about potential regressions.

hannahramadan: I added a test that is more nested, just stealing from your idea `MyApp::Jobs::EmailNotificationJob`. There is now another job class and a test to grab the name!

tannalynn (approved):

Issue #2009, opened 2023-05-12
Address Major Version TODO's


hannahramadan:
We have a few comments floating around: `# TODO: MAJOR VERSION`

Let's address all of these. We should also search for other variations of these types of comments in case they're not formatted in the same way as above.

PR #3326, opened 2025-11-14
Address TODO: Major Version

tannalynn:
Addresses the TODO: Major version comments we have in the code base that can be addressed with this major version. The remaining major version todos that I did not address are for versions of things we will still be supporting with 10.0.0. 

For the excon one, I just updated the comment. The min version there is simply the first version that could have instrumentation installed the way we do. We don't typically prevent instrumentation from installing on older than supported versions.
resolves https://github.com/newrelic/newrelic-ruby-agent/issues/2009

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@
 
 - **Breaking Change: Remove support for Puma versions < 3.9.0**
 
-  The minimum version of puma now supported is 3.9.0 or higher. [PR#3326](https://github.com/newrelic/newrelic-ruby-agent/pull/3326)
+  The minimum version of Puma now supported is 3.9.0 or higher. [PR#3326](https://github.com/newrelic/newrelic-ruby-agent/pull/3326)
 
 ## v9.23.0
 


CHANGELOG.md
```diff
@@ -27,6 +27,10 @@
 
   The experimental feature, Configurable Security Policies (CSP), is no longer supported and has been removed. [PR#3292](https://github.com/newrelic/newrelic-ruby-agent/pull/3292)
 
+- **Breaking Change: Remove support for Puma versions < 3.9.0**
+
+  The minimum version of puma now supported is 3.9.0 or higher. [PR#3326](https://github.com/newrelic/newrelic-ruby-agent/pull/3326)
```
```suggestion
  The minimum version of Puma now supported is 3.9.0 or higher. [PR#3326](https://github.com/newrelic/newrelic-ruby-agent/pull/3326)
```

kaylareopelle: Thoughts on adding a todo in our documentation ticket to make sure requirements and supported frameworks is updated?

tannalynn: Added it to the doc issue

PR #3330, opened 2025-11-14
Add Root sampler, Trace ID ratio based sampler options

### LINKED ISSUE #3284: Implement Trace ID Ratio-Based Sampler

kaylareopelle:
* Adds the trace id ratio based sampling options, along with ratio configs for each of those options
  * This sampling matches the OpenTelemetry Ruby SDK sampler and not the recently defined algorithm that will be used for the Probability Sampler
* Adds the `adaptive` option to all DT samplers
* Adds the ability to configure the root sampling decision to the same options as the remote parent samplers: `default`, `adaptive`, `trace_id_ratio_based`, `always_on`, `always_off`

Closes #3284 
Extends #3307


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -281,16 +282,6 @@ def self.convert_to_constant_list(raw_value)
           constants.compact!
           constants
         end
-
-        def self.enforce_fallback(allowed_values: nil, fallback: nil)
```
kaylareopelle, line 285: This is unused code. We have an allowlist feature now with a fallback function that's more performant.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1463,19 +1454,68 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :allowed_from_server => true,
           :description => 'Distributed tracing lets you see the path that a request takes through your distributed system. Enabling distributed tracing changes the behavior of some New Relic features, so carefully consult the [transition guide](/docs/transition-guide-distributed-tracing) before you enable this feature.'
         },
+        :'distributed_tracing.sampler.root' => {
+          :default => 'default',
+          :public => true,
+          :type => String,
+          :allowed_from_server => false,
```
kaylareopelle, line 1461: Don't know if these should or shouldn't be allowed

lib/new_relic/agent/transaction/trace_context.rb
```diff
@@ -171,14 +179,20 @@ def use_nr_tracestate_sampled(payload)
           end
         end
 
-        def set_priority_and_sampled(config, payload)
-          if config == 'always_on'
+        def set_priority_and_sampled(sampler, ratio, payload)
+          case sampler
+          when 'default'
```
kaylareopelle, line 192: These case statements are a little weird. Ruby evaluates case statements top to bottom. Listing the statements in the order of most likely to be assigned keeps things performant. If we put more than one string in a condition, it'll check to see if the case matches any one of those conditions before moving on to the next one.

Even though `adaptive` is the same as `default`, I hypothesize `adaptive` will be used so infrequently, it should just be listed last rather than doubling up with it on the first line. If it's too weird though, down to refactor and put them both on the same `when` clause.

CHANGELOG.md
```diff
@@ -1,10 +1,36 @@
 # New Relic Ruby Agent Release Notes
 
-## v9.23.0
+## dev
 
-- **Feature: Add sidekiq.ignore_retry_errors configuration option**
+- **Feature: Add Trace ID Ratio Based sampling options**
```
kaylareopelle, line 13: Not sold on these entries. Would love feedback!

lib/new_relic/agent/configuration/default_source.rb
tannalynn: We aren't expecting these to come from the server ever? I know DT can be turned on server side, so I would think it might be in the plans to have server side send these down as well. If that's not something that is planned to happen though, leaving it as false is fine with me

lib/new_relic/agent/transaction/trace_context.rb
tannalynn: I think that's totally reasonable, and a good reason for having it this way.

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: Good point. I can't tell from the specs, so I asked in #agent-dev

lib/new_relic/agent/transaction/trace_context.rb
kaylareopelle: Thanks! Do you think it needs a comment in the code?

tannalynn: It couldn't hurt, but i don't really think it needs it. dealers choice? lol.

CHANGELOG.md
tannalynn: I think it sounds good to me.

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: @tannalynn - 📰 None of these configs are expected to be allowed from the server.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1463,19 +1454,68 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :allowlist => %w[default adaptive always_on always_off trace_id_ratio_based],
```
hannahramadan, line 1462: What is behind the `default` value being `default` instead of just `adaptive`? Did we need to have `default` as a kept option per like the spec or something

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1463,19 +1454,68 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
...
+          :description => 'This setting controls the behavior of transaction sampling when a remote parent is sampled. Available values are `default`, `adaptive`, `always_on`, `always_off`, and `trace_id_ratio_based`. At this time `default` and `adaptive` are the same.'
         },
         :'distributed_tracing.sampler.remote_parent_not_sampled' => {
           :default => 'default',
           :public => true,
           :type => String,
-          :allowed_from_server => true,
-          :description => 'This setting controls the behavior of transaction sampling when a remote parent is not sampled and the trace flag is not set in the traceparent. Available values are `default`, `always_on`, and `always_off`.'
+          :allowed_from_server => false,
+          :allowlist => %w[default adaptive always_on always_off trace_id_ratio_based],
```
hannahramadan, line 1486: I like this allowlist feature

lib/new_relic/agent/configuration/sampler_config_validator.rb
```diff
@@ -0,0 +1,54 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+module NewRelic
+  module Agent
+    module Configuration
+      # Handles validation for the distributed tracing sampler configs
```
hannahramadan:
```suggestion
      # Handles validation for the `distributed_tracing.sampler.*` configs
```

lib/new_relic/agent/transaction.rb
```diff
@@ -307,8 +328,24 @@ def trace_id=(value)
         @trace_id = value
       end
 
+      # This may be better in the adaptive sampler?
```
hannahramadan: I think I agree! It feels more intuitive to live in the adaptive sampler class

lib/new_relic/agent/transaction/trace_context.rb
hannahramadan: I would take a comment! In the context of this entire PR it makes sense, but I could see looking at this in isolation more confusing

hannahramadan (dismissed):

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: Yes, it's a spec thing. I considered having some config shenanigans happen behind the scenes to have any `adaptive` configs get re-applied as `default` or vise-versa to simplify the way the configs are called in the code. However, looking at the spec today, it seems like things have changed and now the `adaptive` option needs a suboption, `sampling_target`, which can be an integer between 1 and 120. `default` doesn't have the same suboption. I think that's a feature for a different PR at this point.

lib/new_relic/agent/configuration/default_source.rb
kaylareopelle: me too! was happy to find it. a wonderful james feature 🍵 🦊

lib/new_relic/agent/transaction.rb
kaylareopelle: The AdaptiveSampler is a class. Currently, this priority is being used for both the `default`/`adaptive` options as well as the `trace_id_ratio_based` option. If we want to move it out of transaction, maybe we make it a class method? I don't think turning it into an instance method makes sense since it spans multiple sampling types. 

Another option could be to rename this priority method to something without adaptive in the name.

What would you prefer?

lib/new_relic/agent/transaction/trace_context.rb
kaylareopelle: Sounds good! Added in 85cfd5dcc

lib/new_relic/agent/transaction.rb
kaylareopelle: Discussed sync. We decided to rename the method to `default_priority` and leave it in the `Transaction` class. 229ac1156

hannahramadan (approved):

PR #3333, opened 2025-11-18
Force Disable CAT

tannalynn:
This removes the ability for CAT to be used and deletes the API methods that were specific to CAT. 

- removed config
- removed 4 public api methods related to CAT (and the file some of them were in)
- Deleted tests testing for CAT behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@
 
   The executable file for the agent's CLI has been renamed from `bin/newrelic` to `bin/newrelic_rpm`. This change resolves a name collision with the standalone New Relic CLI tool. [PR#3323](https://github.com/newrelic/newrelic-ruby-agent/pull/3323)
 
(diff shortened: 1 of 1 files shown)


lib/new_relic/agent/distributed_tracing/cross_app_tracing.rb
```diff
@@ -164,8 +164,7 @@ def valid_encoding_key?
       end
 
       def cross_application_tracer_enabled?
```
kaylareopelle, line 166: What do the other methods in this module give us if a customer can't enable CAT?

tannalynn: we won't be keeping these long term, but I'm not ripping out everything in this PR, just force disabling CAT and removing publicly documented api methods.

kaylareopelle (approved):

Issue #3335, opened 2025-11-18
Update minimum ruby version - multiverse test suite


hannahramadan:
Update the minimum ruby version we are testing for multiverse test suites

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-489422

PR #3323, opened 2025-11-12
Rename `bin/newrelic` to `bin/newrelic_rpm`

hannahramadan:
Remove `bin/newrelic` command in favor of `bin/newrelic_rpm`

closes #2240

PR #3337, opened 2025-11-19
Update multiverse suites to Ruby 2.6+

hannahramadan:
Remove the testing of and conditions for Ruby < 2.6 from the test suites. 

Its also removes a straggler `newrelic` executable line that should have been removed in deprecating the `newrelic` executable in favor of `newrelic_rpm`

closes [#3335](https://github.com/newrelic/newrelic-ruby-agent/issues/3335)

### DIFF
--- a/test/multiverse/suites/config_file_loading/Envfile
+++ b/test/multiverse/suites/config_file_loading/Envfile
@@ -10,7 +10,7 @@ PSYCH_VERSIONS = [
   # Until those errors are resolved, limit testing for the newest versions of
   # Psych to 2.7+
   # 
-  [nil],
+  [nil, 2.7, 3.4],
   ['5.2.0', 2.7, 3.4],
   ['5.1.0', 2.6, 3.3],
   ['5.0.0', 2.6, 3.3],


newrelic_rpm.gemspec
```diff
@@ -21,8 +21,7 @@ Gem::Specification.new do |s|
     https://github.com/newrelic/newrelic-ruby-agent/
   EOS
   s.email = 'support@newrelic.com'
-  # TODO: MAJOR VERSION - remove newrelic, deprecated since version xxx.
```
hannahramadan, line 24: Removing this was a bonus, it should've been removed in deprecating `newrelic` in favor of `newrelic_rpm` https://github.com/newrelic/newrelic-ruby-agent/pull/3323

test/multiverse/suites/active_record_pg/Envfile
```diff
@@ -64,7 +64,7 @@ create_gemfiles(ACTIVERECORD_VERSIONS)
 #   RB
 # end
 
-# if RUBY_VERSION >= '2.5.0' && RUBY_PLATFORM == 'java'
+# if RUBY_VERSION >= '2.6.0' && RUBY_PLATFORM == 'java'
```
kaylareopelle:
```suggestion
# if RUBY_PLATFORM == 'java'
```

test/multiverse/suites/config_file_loading/Envfile
```diff
@@ -5,17 +5,17 @@
 omit_collector!
 
 PSYCH_VERSIONS = [
-  # TODO: Psych 5.2 supports Ruby 2.5+, but we began to get some Bundler errors
-  # On May 1, 2025 for Ruby 2.5 and 2.6 for Psych 5.2.4.
+  # TODO: Psych 5.2 supports Ruby 2.6+, but we began to get some Bundler errors
+  # On May 1, 2025 for Ruby 2.6 for Psych 5.2.4.
```
kaylareopelle, line 9: Is it worth checking to see if these errors have been fixed?

test/multiverse/suites/delayed_job/Envfile
```diff
@@ -7,28 +7,15 @@ instrumentation_methods :chain, :prepend
...
-    "gem 'sqlite3', '~> 1.4.2'"
   else
-    "gem 'sqlite3', '~> 1.3.13'"
-  end
-end
-
-def bigdecimal
-  unless RUBY_PLATFORM == 'java' || RUBY_VERSION < '2.5.0'
-    "gem 'bigdecimal', '~> 3.0.0'"
-  end
```
kaylareopelle, line 20: Do we still need this for the Java condition?

newrelic_rpm.gemspec
tannalynn: Oh interesting, I missed this in my major version Todo PR as well even though its marked with it. I think it's because I have my search function only check .rb files usually, so .gemspec didn't show up. I'll have to keep that in mind

test/new_relic/github/workflows/scripts/generate_release_notes_test.rb
```diff
@@ -9,7 +9,7 @@
 module NewRelic
   class GenerateReleaseNotesTest < Minitest::Test
     def setup
-      skip unless NewRelic::Helper.version_satisfied?(RUBY_VERSION, '>=', '2.5.0')
+      skip unless NewRelic::Helper.version_satisfied?(RUBY_VERSION, '>=', '2.6.0')
```
kaylareopelle: I don't think we need this condition anymore since the test should never run on an incompatible version.

```suggestion
```

test/multiverse/suites/delayed_job/Envfile
tannalynn: Interesting, I just checked and it looks like we don't run these tests on jruby actually, so removing this shouldn't have an impact on the CI. Likely this is left over from a long time ago when everything was run on jruby

hannahramadan: I see that now! _Should_ we be running this on JRuby? It looks like a few good of our suites are not. If we took it out for a reason, it is probably a good one, but I can't think of why

tannalynn: If we took it out of jruby CI, it was very likely intentional. We have a lot of issues of intermittent failures when running on jruby, so that could be one reason (and it's probably the most likely reason).

hannahramadan: Good thinking. I'll keep this out!

test/multiverse/suites/config_file_loading/Envfile
hannahramadan: Tested, this error remains!

Issue #3130, opened 2025-04-14
Evaluate API logging


hannahramadan:
As raised in https://github.com/newrelic/newrelic-ruby-agent/issues/3126, the agent will, in some cases, not validate an argument's data type and fail to both alert and send the data along. We should look at all our APIs and ensure we aren't silently failing when an unexpected data type is passed.

PR #3339, opened 2025-11-19
Misc 10.0 cleanup

kaylareopelle:
While working on #3130, I noticed we were missing documentation for the second argument in the `NewRelic::Agent.increment_metric` method.

Also, we forgot to remove the `newrelic` executable from the gemspec.

### DIFF
--- a/lib/new_relic/agent.rb
+++ b/lib/new_relic/agent.rb
@@ -335,7 +335,7 @@ module NewRelic
     # +metric_name+ should be a String that follows a slash separated
     # path convention. Application specific metrics should begin with
     # "Custom/".
-    # +amount+ should be a Numeric value to increment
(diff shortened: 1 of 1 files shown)


lib/new_relic/agent.rb
```diff
@@ -332,8 +332,10 @@ def base_name(klass_name)
 
     # Increment a simple counter metric.
     #
-    # +metric_name+ should follow a slash separated path convention. Application
-    # specific metrics should begin with "Custom/".
+    # +metric_name+ should be a String that follows a slash separated
+    # path convention. Application specific metrics should begin with
+    # "Custom/".
+    # +amount+ should be a Numeric value to increment
```
tannalynn: Maybe we should say it is an optional arg with the default being 1, for clarity.

kaylareopelle: Great point! b67bfae2b

PR #3341, opened 2025-11-19
Update config type coercion

### LINKED ISSUE #2935: Config: apply type coercion, at least for integer based values
We have some configuration parameter value type coercion logic parked in the `lab` branch which is a) passing all CI tests, b) received dev approval, and c) passed csec tests for known integer handling issues.

Our CI suites involve lots and lots of iterations that completely flush and entirely reinitialize all agent configuration and layering in type coercion does have a performance impact that is easily seen in a CI context. Before merging the type coercion logic in, we'd like to be absolutely sure that the performance hit is expected to be slight outside of a CI context.

As a fallback plan, we should at least incorporate integer focused coercion for the csec agent's benefit.

kaylareopelle:
From `lab` branch and `presto_chango`
Closes #2935
Originally #2878 : 
> Rework the processing of configuration parameters to address https://github.com/newrelic/newrelic-ruby-agent/issues/2852.
>
> * nils to produce errors when user-supplied for params that don't support them
> * introduce a common shared type coercion layer for all config values so that things like environment variable and YAML values are both consistently coerced
> * allow type coercion to Array/Hash to replace all of the equivalent default source based:transform entries
> * lay the foundation for increased server side and env var based support for certain params that are currently marked as YAML source only (functionally ready now, updating tests and docs for each affected param to be handled later)
> * permit Integer values for Float params (using #to_f) and Float values for Integer params (using #round)
> * prevent fatal interruptions and log warnings when a default value can be found to replace a given bogus input value
> * leverage default source's :allow_nil setting to enforce non-nil values for params that don't support nil
>
> resolves https://github.com/newrelic/newrelic-ruby-agent/issues/2852

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,10 +23,6 @@
 
     The deprecated `newrelic deployments` CLI command has been removed. To track changes and deployments in New Relic, please see our guide to [Change Tracking](https://docs.newrelic.com/docs/change-tracking/change-tracking-introduction/) for a list of available options. [PR#3299](https://github.com/newrelic/newrelic-ruby-agent/pull/3299)
 
-- **Feature: Add argument validation for the `NewRelic::Agent#record_custom_event` API**
-
-  The `NewRelic::Agent#record_custom_event` API now raises an `ArgumentError` when an invalid `event_type` is provided. A valid event type must consist only of alphanumeric characters, underscores (`_`), colons (`:`), or spaces (` `). [PR#3319](https://github.com/newrelic/newrelic-ruby-agent/pull/3319)
-
 - **Breaking Change: Remove experimental feature Configurable Security Policies (CSP)**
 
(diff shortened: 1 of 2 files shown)


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2196,11 +2131,12 @@ def self.enforce_fallback(allowed_values: nil, fallback: nil)
           :description => 'If true, the agent strips messages from all exceptions except those in the [allowed classes list](#strip_exception_messages-allowed_classes). Enabled automatically in [high security mode](/docs/accounts-partnerships/accounts/security/high-security).'
         },
         :'strip_exception_messages.allowed_classes' => {
-          :default => '',
+          :default => NewRelic::EMPTY_ARRAY,
```
hannahramadan, line 2134: we should make sure that our newrelic.yml generator transforms this into `[]`

kaylareopelle: It does! This is what it looks like when I run the generator 
```
  # strip_exception_messages.allowed_classes: []
```

Issue #3272, opened 2025-09-30
Remove Cross-Application Tracing


kaylareopelle:
Cross-Application Tracing has been deprecated since September 2021 with the release of version 8.0.0 of the agent. 

We will remove its functionality entirely as part of our 10.0 release. 

Users should migrate to distributed tracing to get similar functionality. 

- [ ] Remove code related to CAT
- [ ] Remove tests related to CAT
- [ ] Update docs to state last version with CAT
- [ ] Remove configs related to CAT (update trusted_account_keys description to remove CAT reference, this is still used for adaptive sampling)

PR #3344, opened 2025-11-20
remove CAT

tannalynn:
this pr actually removes all CAT related functionality. 

closes https://github.com/newrelic/newrelic-ruby-agent/issues/3272

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,10 +14,10 @@ jobs:
       - name: Configure git
         run: 'git config --global init.defaultBranch main'
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # tag v5.0.0
-      - uses: ruby/setup-ruby@d5126b9b3579e429dd52e51e68624dda2e05be25 # tag v1.267.0
+      - uses: ruby/setup-ruby@8aeb6ff8030dd539317f8e1769a044873b56ea71 # tag v1.268.0
         with:
-          ruby-version: 3.5.0-preview1
-      - run: bundle
+          ruby-version: 4.0.0-preview2
+      - run: bundle install
       - run: rubocop
 
 
@@ -36,7 +36,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby-version: [2.6.10, 3.4.4, 3.5.0-preview1]
+        ruby-version: [2.6.10, 3.4.4, 4.0.0-preview2]
     steps:
       - name: Configure git
         run: 'git config --global init.defaultBranch main'
@@ -56,7 +56,7 @@ jobs:
       fail-fast: false
       matrix:
         multiverse: [agent, ai, background, background_2, frameworks, httpclients, httpclients_2, hybrid_agent, rails, rest]
-        ruby-version: [2.6.10, 3.4.4, 3.5.0-preview1]
+        ruby-version: [2.6.10, 3.4.4, 4.0.0-preview2]
     steps:
       - name: Configure git
         run: 'git config --global init.defaultBranch main'
@@ -89,7 +89,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby-version: [2.6.10, 3.4.4, 3.5.0-preview1]
+        ruby-version: [2.6.10, 3.4.4, 4.0.0-preview2]
     steps:
       - name: Configure git
         run: 'git config --global init.defaultBranch main'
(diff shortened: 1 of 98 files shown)


lib/new_relic/agent/messaging.rb
```diff
@@ -364,6 +362,15 @@ def transaction_name(library, destination_type, destination_name, action = nil)
 
         transaction_name
       end
+
+      # Filter out internal New Relic headers from message headers
+      # Reject: newrelic, traceparent, tracestate, NewRelicID, NewRelicTransaction, NewRelicSynthetics
+      def reject_messaging_internal_headers(headers)
```
tannalynn, line 369: For headers we read in that were inserted by old agent versions, this prevents those cat headers from being added as regular attributes.

lib/new_relic/agent/instrumentation/bunny/instrumentation.rb
```diff
@@ -34,9 +34,7 @@ def publish_with_tracing(payload, opts = {})
             begin
               destination = exchange_name(name)
 
-              tracing_enabled =
-                NewRelic::Agent::CrossAppTracing.cross_app_enabled? ||
-                NewRelic::Agent.config[:'distributed_tracing.enabled']
+              tracing_enabled = NewRelic::Agent.config[:'distributed_tracing.enabled']
               opts[:headers] ||= {} if tracing_enabled
```
tannalynn:
```suggestion
              opts[:headers] ||= {} if NewRelic::Agent.config[:'distributed_tracing.enabled']
```

lib/new_relic/agent/instrumentation/excon.rb
```diff
@@ -46,7 +45,7 @@ def install_middleware_excon_instrumentation
     defaults = Excon.defaults
 
     if defaults[:middlewares]
-      defaults[:middlewares] << Excon::Middleware::NewRelicCrossAppTracing
+      defaults[:middlewares] << Excon::Middleware::NewRelicTracing
```
tannalynn, line 48: It renamed this because its not actually related to CAT

.github/workflows/ci.yml
```diff
@@ -287,7 +287,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby-version: [2.7.8, 3.4.4, 3.5.0-preview1]
+        ruby-version: [2.7.8, 3.4.4]
```
tannalynn, line 290: infinite tracing tests seems to be a little slow on the preview, so i wanted to take it out of the pr ci

test/new_relic/agent/transaction/external_request_segment_test.rb
```diff
@@ -100,110 +98,35 @@ def test_segment_records_expected_metrics_for_non_cat_txn
...
+        expected_metrics = [
+          'External/remotehost.com/Net::HTTP/GET',
+          'External/all',
+          'External/remotehost.com/all',
+          'External/allWeb',
+          ['External/remotehost.com/Net::HTTP/GET', 'test']
+        ]
 
-          assert_metrics_recorded expected_metrics
+        if Agent.config[:'distributed_tracing.enabled']
```
kaylareopelle: Do we still need this condition?

lib/new_relic/agent/stats_engine.rb
```diff
@@ -4,6 +4,7 @@
 
 require 'new_relic/agent/stats_engine/gc_profiler'
 require 'new_relic/agent/stats_engine/stats_hash'
+require 'new_relic/agent/tracer'
```
kaylareopelle, line 7: why are we requiring this here now?

kaylareopelle (commented):
phew! good work! a few questions.

lib/new_relic/agent/stats_engine.rb
tannalynn: We are using the Tracer class in this file. it must not have needed the require before because somehow some file related to CAT was bringing it in, but after we removed those files, some tests were failing saying it didn't know about tracer, so it had to be added

test/new_relic/agent/transaction/external_request_segment_test.rb
tannalynn: i tried removing it and it still works fine, so no, we do not lol.

kaylareopelle (approved):

.github/workflows/ci.yml
hannahramadan: is not testing infinite tracing on PR CIs something we wanna keep? So only the scheduled CI

tannalynn: Oh i was just changing it so that the preview wasn't running on prs, not stopping all 8t tests from running on prs

PR #3354, opened 2025-11-27
Add missing Span APIs

### LINKED ISSUE #3275: Hybrid Agent: Implement missing Span APIs
The `opentelemetry-api` gem has a few APIs we have not implemented for the Span class. Since the New Relic implementation functions as an SDK, we should implement these APIs so they behave as expected if users choose to call them. 

- [x] [recording?](https://github.com/open-telemetry/opentelemetry-ruby/blob/4b275071a107026e37c2f0f3a3c51d70115a1284/api/lib/opentelemetry/trace/span.rb#L36-L43) - this should be updated by the agent, as instrumentation often uses the API to determine whether additional attributes should be applied
- [x] [status=](https://github.com/open-telemetry/opentelemetry-ruby/blob/4b275071a107026e37c2f0f3a3c51d70115a1284/api/lib/opentelemetry/trace/span.rb#L138-L149) - this should be available for users to set
- [x] [name=](https://github.com/open-telemetry/opentelemetry-ruby/blob/4b275071a107026e37c2f0f3a3c51d70115a1284/api/lib/opentelemetry/trace/span.rb#L151-L160) - this should be available for users to set

There are two other APIs we're waiting on more guidance from NR to implement: `add_link` and `add_event`. These will be added in a separate issue.

kaylareopelle:
There are three Span APIs commonly used by OpenTelemetry Ruby instrumentation that weren't specifically needed to pass the Hybrid Agent cross agent tests.

* `status=` => Adds new custom attributes to NR spans: `status.code` and `status.description`. By default, segments created by OTel will now have the `status.code` set to "unset". This API can be called to update the status to `ok` or `error`. If the status is `error`, there may be a description.
* `name=` => Allows the Span name to be updated. We may want to mark this as private, since we don't have a public API for this behavior.
* `recording?` => Checks to see if the span is still recording/hasn't finished. NR agent's equivalent API is `finished?`

Closes #3275

### DIFF
--- a/test/multiverse/suites/hybrid_agent/span_test.rb
+++ b/test/multiverse/suites/hybrid_agent/span_test.rb
@@ -63,8 +63,8 @@ module NewRelic
             spans = harvest_span_events![1]
             span_attributes = spans[0][1]
 
-            assert_equal span_attributes['yosemite'], 'california'
-            assert_equal span_attributes['yellowstone'], 'wyoming'
+            assert_equal('california', span_attributes['yosemite'])
+            assert_equal('wyoming', span_attributes['yellowstone'])
           end
 
           def test_recording_works_with_finishable_transactions_when_finished


lib/new_relic/agent/opentelemetry/trace/span.rb
```diff
@@ -24,6 +30,30 @@ def add_attributes(attributes)
           def record_exception(exception, attributes: nil)
             NewRelic::Agent.notice_error(exception, attributes: attributes)
           end
+
+          def recording?
```
kaylareopelle, line 30: while writing the PR description, I wondered if it would be helpful to have better documentation for these APIs, even if they're private. wdyt?

test/multiverse/suites/hybrid_agent/span_test.rb
```diff
@@ -61,7 +61,129 @@ def test_add_attributes_patch_for_spans
             spans = harvest_span_events![1]
             span_attributes = spans[0][1]
 
-            assert_equal span_attributes, attributes
+            assert_includes span_attributes.keys, 'yosemite'
+            assert_includes span_attributes.keys, 'yellowstone'
```
kaylareopelle: Since the status attributes are now added by default, these aren't the only attrs on the span.

lib/new_relic/agent/opentelemetry/trace/span.rb
hannahramadan: I err on the documentation side, so would love to see it ◡̈

test/multiverse/suites/hybrid_agent/span_test.rb
```diff
@@ -7,8 +7,9 @@ module Agent
     module OpenTelemetry
       module Trace
         class SpanTest < Minitest::Test
-          Segment = Struct.new(:guid, :transaction)
-          Transaction = Struct.new(:trace_id)
+          # Tests in this file can easily become flaky and influence other files
```
hannahramadan: nice love this comment

lib/new_relic/agent/opentelemetry/trace/span.rb
```diff
@@ -24,6 +30,30 @@ def add_attributes(attributes)
...
+            !finishable&.finished?
+          end
+
+          def name=(name)
+            if recording?
+              if finishable.is_a?(NewRelic::Agent::Transaction)
+                finishable.overridden_name = name
```
hannahramadan, line 45: Learn something new all the time - didn't know about overriden_name

lib/new_relic/agent/opentelemetry/trace/span.rb
```diff
@@ -24,6 +30,30 @@ def add_attributes(attributes)
...
+              elsif finishable.is_a?(NewRelic::Agent::Transaction::Segment)
+                finishable.instance_variable_set(:@name, name)
+              end
+            else
+              NewRelic::Agent.logger.warn('Calling name= on an ended OpenTelemetry Span') unless recording?
```
hannahramadan: is the word 'ended' an otel language thing? I feel like I usually see the word finished but could be a NR leaning verbiage

```suggestion
              NewRelic::Agent.logger.warn('Calling name= on a finished OpenTelemetry Span') unless recording?
```

kaylareopelle: whoops! yep, that's a straight copy of the otel message.

lib/new_relic/agent/opentelemetry/trace/span.rb
kaylareopelle: updated in 81d1705fe7dbc7cc03601317bdab7d0a50db06ca

hannahramadan (approved):
I LOVE all the documentation that exists in this PR. Especially because its so Otel/outside this codebase tied.

test/multiverse/suites/hybrid_agent/tracer_test.rb
```diff
@@ -119,6 +123,8 @@ def test_set_otel_tracestate_with_newrelic_entry
           end
 
           def test_set_otel_tracestate_without_newrelic_entry
+            skip 'need to make some trace context changes that are bigger than this PR'
```
hannahramadan: Maybe in case this gets accidentally overlooked, but not needed
```suggestion
            TODO: skip 'need to make some trace context changes that are bigger than this PR'
```

kaylareopelle: Like the Todo idea! `skip` is a method, so I can't put `TODO:` straight in front of it. I added a comment in 3a4d728fb 

Does that look good to you?

hannahramadan: oopsies! yes that looks good. can't tell if that is passing or if it needs to be a comment above the `skip` or something ◡̈

PR #3368, opened 2025-12-08
Extend sampling decision logic to OTel

kaylareopelle:
When updating the sampling decision logic to handle cases without tracestate payloads, I forgot to update the OTel case without a tracestate payload. The `test_set_otel_tracestate_without_newrelic_entry` test failed intermittently without these changes.

The logic has been updated to use the trace flags to fall back to adaptive samplers for each sampling type (ex. a new adaptive sampler used for all `remote_parent_sampled`/trace flag `01` decisions).

We might run into some merge conflicts since the work to remove the `default` option for samplers in favor of `adaptive` was merged into v10.


lib/new_relic/agent/transaction/trace_context.rb
```diff
@@ -157,29 +157,39 @@ def determine_sampling_decision(payload, trace_flags)
...
         end
 
-        def use_nr_tracestate_sampled(payload)
-          unless payload.sampled.nil?
+        def use_nr_tracestate_sampled(payload, trace_flags)
+          if payload.sampled.nil?
+            if trace_flags == '01'
+              transaction.sampled = NewRelic::Agent.instance.adaptive_sampler_remote_parent_sampled.sampled?
+            elsif trace_flags == '00'
+              transaction.sampled = NewRelic::Agent.instance.adaptive_sampler_remote_parent_not_sampled.sampled?
```
kaylareopelle, line 182: These cases should only be called when the sampler is `adaptive`/`default`, so we use the adaptive sampler specific to the trace flags (ex. remote_parent_sampled for flag 01)

lib/new_relic/agent/transaction/trace_context.rb
```diff
@@ -157,29 +157,39 @@ def determine_sampling_decision(payload, trace_flags)
...
           end
         rescue
-          use_nr_tracestate_sampled(payload)
+          use_nr_tracestate_sampled(payload, trace_flags)
         end
 
-        def use_nr_tracestate_sampled(payload)
-          unless payload.sampled.nil?
+        def use_nr_tracestate_sampled(payload, trace_flags)
+          if payload.sampled.nil?
```
tannalynn, line 178: So just want to make sure i understand right, but payload.sampled with be nil when it's the INVALID payload object, right?

kaylareopelle: Correct! This is what the invalid payload object looks like: 
```ruby
#<NewRelic::Agent::TraceContextPayload:0x00000001250555c0
 @id=nil,
 @parent_account_id=nil,
 @parent_app_id=nil,
 @parent_type_id=0,
 @priority=nil,
 @sampled=nil,
 @timestamp=1765396576302,
 @transaction_id=nil,
 @version=0>
```

And the method should only be called with payload objects, so the `sampled` method should always be callable.

PR #3372, opened 2025-12-10
10.0.0 changes

tannalynn:
https://github.com/newrelic/newrelic-ruby-agent/milestone/119


hannahramadan (approved):
Will re-review before launching to `main` but looks good!

PR #3378, opened 2025-12-18
CI changes

tannalynn:
passing jruby run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/20349177631

### DIFF
--- a/test/multiverse/suites/sidekiq/Envfile
+++ b/test/multiverse/suites/sidekiq/Envfile
@@ -6,7 +6,7 @@ SIDEKIQ_VERSIONS = [
   [nil, 3.2],
   ['7.3.9', 2.7],
   ['6.4.0'],
-  ['5.0.3']
+  ['5.0.3', 2.6, 2.6]
 ]
 
 def gem_list(sidekiq_version = nil)


test/multiverse/suites/sidekiq/Envfile
```diff
@@ -6,7 +6,7 @@ SIDEKIQ_VERSIONS = [
   [nil, 3.2],
   ['7.3.9', 2.7],
   ['6.4.0'],
-  ['5.0.3']
+  ['5.0.3', 2.6, 2.6]
```
tannalynn, line 9: when we removed the min ruby versions when we dropped 2.4 and 2.5, here we accidentally removed the max ruby version of 2.6 to run the tests on, so that's why jruby started having problems.

.github/workflows/performance_tests.yml
```diff
@@ -1,9 +1,6 @@
 name: Performance Tests
 
 on:
-  pull_request:
-    branches:
-      - main
```
tannalynn, line 6: removing this because now we have perfverse, and also one of the github actions we run to upload the results to slack is having problems.

test/new_relic/agent/configuration/manager_test.rb
```diff
@@ -693,8 +693,19 @@ def test_validate_nil_does_not_warn_for_nr_internal_category
     def test_enforce_allowlist_only_operates_on_params_with_allowlists
       key = :unguarded
 
+      # stubbing it weirdly bc jruby
       default_source = Object.new
-      default_source.stubs(:allowlist_for).returns(nil)
+      default_source.define_singleton_method(:allowlist_for) do |k|
```
tannalynn, line 698: the changes in this file are all related to one of the 10.0 changes. these tests are new and were never stubbed to run correctly on jruby.

PR #3379, opened 2025-12-18
Try to fix intermittent rdkafka failures.

tannalynn:
so those weird kafka failures where no tests fails. maybe this will help?

### DIFF
--- a/.github/workflows/scripts/setup_bundler
+++ b/.github/workflows/scripts/setup_bundler
@@ -55,7 +55,9 @@ function install_ruby_version_specific_gems {
     echo "DEBUG: running 'gem install' on specific gem versions"
     # The doubles below are needed for older Ruby versions, since the newer one won't install on the older Ruby versions
     gem install --default cgi:0.5.0
+    gem install --default cgi:0.5.1
     gem install --default date:3.4.1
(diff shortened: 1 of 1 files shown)


kaylareopelle:
I ran this on the ci cron and found an error related to Ruby 4.0. It's strange, because the same test presumably ran in this context and passed? 

https://github.com/newrelic/newrelic-ruby-agent/actions/runs/20759323209/job/59609800149 

I think the problem may be resolved if `dev` is merged into this branch to run on the 🎅 4.0 rather than the preview version. wdyt?

.github/workflows/scripts/setup_bundler
```diff
@@ -55,7 +55,9 @@ function install_ruby_version_specific_gems {
     echo "DEBUG: running 'gem install' on specific gem versions"
     # The doubles below are needed for older Ruby versions, since the newer one won't install on the older Ruby versions
     gem install --default cgi:0.5.0
+    gem install --default cgi:0.5.1
     gem install --default date:3.4.1
+    gem install --default date:3.5.1
```
tannalynn, line 60: these changes aren't for rdkafka, just snuck them in here to help the CI that's failing the first time every time on some rails versions

PR #3388, opened 2025-12-31
Guard non-string values for utilization detection

hannahramadan:
The agent is assuming that any GCP environment will provide a machine type string, which isn't really applicable for a Serverless environment. This PR adds a string guard in the agent's detection logic.


lib/new_relic/agent/utilization/gcp.rb
```diff
@@ -25,7 +25,7 @@ def prepare_response(response)
         end
 
         def trim_leading(value)
-          value.split(NewRelic::SLASH).last
+          value&.split(NewRelic::SLASH)&.last
```
kaylareopelle: I'm curious about the second safe navigation operator. What cases are we protecting for there?

hannahramadan: Is just a defensive move, since originally I don't think we ever expected a `nil` value, but we got one now! I would be worried about any other unexpected value, like maybe something that doesn't like `.last` so I don't want to see a crash again.

kaylareopelle: Ah okay. I have another strategy to propose in that case. 

`split` is only a String method, so if `value` is anything but a String or `nil`, we'll raise an error after the safe navigation operator. 

If `split` can be called successfully, it will return an Array, which can call `last`. If `split` is called on an empty string `last` will return `nil`.

Since we expect a String, we might want the return value to be a String. 

If we want to be super defensive against other datatypes, we could try: 
```
return NewRelic::EMPTY_STR unless value.is_a?(String)
```

And leave the rest of the method in its original form. 

What do you think?

hannahramadan: I think this works well! In this case, we are reporting an empty string instead of a nil. I will make this change.

kaylareopelle (dismissed):
Thank you!

lib/new_relic/agent/utilization/gcp.rb
kaylareopelle: @hannahramadan - whoops, approved before I realized there were some GCP test failures. Maybe the empty string won't work?

kaylareopelle (approved):
🦖

PR #3396, opened 2026-01-07
only run oldest and newest ruby versions on PR

tannalynn:
Since 4.0.0 is officially out now and we dont have to use the previews, I'm removing the PR ci middle ruby version so we're only running oldest supported and newest now on PRs.


.github/workflows/scripts/setup_bundler
```diff
@@ -60,6 +60,7 @@ function install_ruby_version_specific_gems {
     gem install --default digest:3.2.0
     gem install --default erb:4.0.4
     gem install --default erb:5.0.2
+    gem install --default erb:6.0.1
```
tannalynn, line 63: oh oops i meant to do this on the other PR where i was already doing this, but oh well

PR #3399, opened 2026-01-07
Don't notice errors when `sidekiq_ignore_errors` is `true`

hannahramadan:
When `sidekiq_ignore_errors` is `true`, instrumentation should create transaction traces for retry jobs, but not errors on those jobs. We should only report errors that make it to Sidekiq's Dead Set.

### DIFF
--- a/test/multiverse/suites/sidekiq_ignore_retry_errors_enabled/sidekiq_ignore_retry_errors_enabled_test.rb
+++ b/test/multiverse/suites/sidekiq_ignore_retry_errors_enabled/sidekiq_ignore_retry_errors_enabled_test.rb
@@ -76,6 +76,38 @@ class SidekiqIgnoreRetryErrorEnabledTest < Minitest::Test
       'Expected middleware to NOT report errors when sidekiq.ignore_retry_errors is true'
   end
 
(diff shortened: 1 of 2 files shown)


test/multiverse/suites/sidekiq_ignore_retry_errors_enabled/sidekiq_ignore_retry_errors_enabled_test.rb
```diff
@@ -58,6 +58,56 @@ def test_death_handlers_registered_when_sidekiq_ignore_retry_errors_is_true
       'Expected NewRelic death_handler to be registered when sidekiq.ignore_retry_errors is true'
   end
 
+  def test_middleware_does_not_report_errors_when_sidekiq_ignore_retry_errors_is_true
+    # TODO: MAJOR VERSION - remove this when Sidekiq v5 is no longer supported
+    skip 'Test requires Sidekiq v6+' unless Sidekiq::VERSION.split('.').first.to_i >= 6
+
+    noticed_errors = []
+
+    NewRelic::Agent.stub :notice_error, proc { |error| noticed_errors.push(error) } do
```
kaylareopelle, line 67: love this approach!

kaylareopelle (approved):

PR #3400, opened 2026-01-08
Fix warnings for Otel bridge record_metric

tannalynn:
Comments out the disabled metric for the otel bridge, and the related test. We will change how this works more in depth in the future.


kaylareopelle (approved):

PR #3403, opened 2026-01-09
Update hybrid agent transaction starting rules

### LINKED ISSUE #3277: Hybrid Agent: Update Transaction starting rules
The rules for when a transaction or a segment should start have changed slightly since our initial implementation. 

See the [OTel Span Rules](https://source.datanerd.us/agents/agent-specs/blob/main/otel_bridge/Tracing-API.md#otel-span-rules)(internal) section of the agent spec for details. 

The biggest change is related to behavior dependent on span kind when the span does not have a remote parent. Some of these changes are not included in the hybrid_agent.json cross-agent test file.

kaylareopelle:
Since our initial implementation, the rules used to start a transaction have changed to account for span kind and transaction type.

See [the agent spec](https://source.datanerd.us/agents/agent-specs/blob/main/otel_bridge/Tracing-API.md#otel-span-rules) (internal) on span rules for details. 

Segment creation is outside the scope of this ticket and will be handled in a separate PR.

Closes #3277

### DIFF
--- a/lib/new_relic/agent/opentelemetry/trace/tracer.rb
+++ b/lib/new_relic/agent/opentelemetry/trace/tracer.rb
@@ -97,9 +97,6 @@ module NewRelic
             # and has a kind of :client, :producer, :internal or nil
             return false if !parent_otel_context.valid? &&
               KINDS_THAT_DO_NOT_START_TXNS_WITHOUT_REMOTE_PARENT.include?(kind)
-
-            # if there's any stragglers
-            false
           end
 
           def should_not_create_telemetry?(parent_otel_context, kind)


lib/new_relic/agent/opentelemetry/trace/tracer.rb
```diff
@@ -15,19 +19,20 @@ def initialize(name = nil, version = nil)
           def start_span(name, with_parent: nil, attributes: nil, links: nil, start_timestamp: nil, kind: nil)
             parent_otel_context = ::OpenTelemetry::Trace.current_span(with_parent).context
 
-            finishable = if can_start_transaction?(parent_otel_context)
-              return if internal_span_kind_with_invalid_parent?(kind, parent_otel_context)
-
-              nr_item = NewRelic::Agent::Tracer.start_transaction_or_segment(name: name, category: :otel)
+            # is nil correct here? or should it be an invalid span?
```
kaylareopelle: question for you all. I think nil is correct, because nothing should be made or used, but I could also see an invalid span possibly being helpful to protect against methods being called on nil?

lib/new_relic/agent/opentelemetry/trace/tracer.rb
```diff
@@ -15,19 +19,20 @@ def initialize(name = nil, version = nil)
...
 
-              add_remote_context_to_txn(nr_item, parent_otel_context)
-
-              nr_item
+            finishable = if can_start_transaction?(parent_otel_context, kind)
+              start_transaction_from_otel(name, parent_otel_context, kind)
             else
+              # this should only be run if we are in an existing transaction
+              # TODO: Expand to include special segment handling by type (ex. DB, External, etc)
+              # We may need to remove the add_attributes line below when we implement
```
kaylareopelle: This will get handled in the segment-specific PR related to translating OTel attributes to NR

lib/new_relic/agent/opentelemetry/trace/tracer.rb
```diff
@@ -7,6 +7,10 @@ module Agent
     module OpenTelemetry
       module Trace
         class Tracer < ::OpenTelemetry::Trace::Tracer
+          VALID_KINDS = [:server, :client, :consumer, :producer, :internal, nil].freeze
```
kaylareopelle, line 10: I'm treating `nil` kind as an unspecified kind according to the spec.

lib/new_relic/agent/opentelemetry/trace/tracer.rb
```diff
@@ -62,12 +82,32 @@ def get_otel_span_from_finishable(finishable)
...
 
-          def can_start_transaction?(parent_otel_context)
-            parent_otel_context.remote? || !parent_otel_context.valid?
+          def can_start_transaction?(parent_otel_context, kind)
+            # if there's a current transaction, we add segments instead of making a new one
+            return false if NewRelic::Agent::Tracer.current_transaction
+
+            # can start a transaction if it has a remote parent
+            # and the kind is :server, :client, :internal, :consumer, :producer
+            return true if parent_otel_context.remote? && VALID_KINDS.include?(kind)
```
kaylareopelle, line 89: This may be higher precedence than the current_transaction. I'm not even sure if we should take current transaction into account. The `should_not_create_telemetry?` clause might take care of this.

Re-reading the spec makes it sound like [a remote parent trumps all else and forces a new transaction](https://source.datanerd.us/agents/agent-specs/blob/main/otel_bridge/Tracing-API.md#when-the-span-has-a-remote-parent). I don't know if we would ever have a current transaction if we have a remote parent? How do you interpret it?

lib/new_relic/agent/opentelemetry/trace/tracer.rb
```diff
@@ -62,12 +82,32 @@ def get_otel_span_from_finishable(finishable)
...
+            # if there's any stragglers
+            false
           end
 
-          def internal_span_kind_with_invalid_parent?(kind, parent_otel_context)
-            !parent_otel_context.valid? && kind == :internal
+          def should_not_create_telemetry?(parent_otel_context, kind)
+            !parent_otel_context.valid? &&
+              KINDS_THAT_DO_NOT_START_TXNS_WITHOUT_REMOTE_PARENT.include?(kind) &&
+              NewRelic::Agent::Tracer.current_transaction.nil?
```
kaylareopelle, line 105: If there isn't a current transaction, we don't want to even try to make a segment for these span kinds. No telemetry should be created.

test/multiverse/suites/hybrid_agent/span_test.rb
```diff
@@ -29,7 +29,7 @@ def test_finish_does_not_fail_if_no_finishable_present
           end
 
           def test_finishable_can_finish_transactions
-            span = @tracer.start_span('test')
+            span = @tracer.start_span('test', kind: :server)
```
kaylareopelle, line 32: This is to accommodate for the spec updates. A span with a `:server` kind should always start a new transaction.

test/multiverse/suites/hybrid_agent/tracer_test.rb
```diff
@@ -123,9 +123,6 @@ def test_set_otel_tracestate_with_newrelic_entry
           end
 
           def test_set_otel_tracestate_without_newrelic_entry
-            # TODO: Need to address trace context work for setting empty tracestate
-            skip 'need to make trace context changes to support this'
-
```
kaylareopelle, line 128: We already fixed this, just forgot to remove the skip

lib/new_relic/agent/opentelemetry/trace/tracer.rb
tannalynn: I think most situations where there is a remote parent would have a current_transaction already, right? But i'm not very sure because the otel of it all. But like in the regular agent, Because the rack instrumentation that starts the transaction would have already started the transaction before we get to the point that we are accepting the DT headers, the transaction would be already created when we set all the remote parent stuff. Is that process very different with the otel stuff included?
But also, that is not like in conflict with the spec saying a transaction needs to be started right? Hmm. maybe i need to think about that a little more.

lib/new_relic/agent/opentelemetry/trace/tracer.rb
tannalynn: I think logically nil makes sense if nothing was started. What places are calling this method though, could we check them to see if they expect a non nil answer?

hannahramadan: if `start_span` calls our `start_transaction_or_segment`, it seems well protected against `nils` (although another check can probably be done). But could a user theoretically call `start_span` and then try to do something like add an attribute, and then error out from the `nil`?

kaylareopelle: Good points all around. The return value in the API (which is the no-op version of the SDK) returns either a non-recording span or an invalid span, so we should probably do that too. 

We can't really know what places will call this method since it's an implementation of an OTel API, so we should probably be prepared for methods to be called on the span after the fact. I'll fix this.

lib/new_relic/agent/opentelemetry/trace/tracer.rb
kaylareopelle: Good questions. I'm really not sure -- I decided [to post](https://newrelic.slack.com/archives/C07PUH00WBZ/p1768414809823059) in Slack (internal) to see what the other agents are doing.

hannahramadan: Assuming a transaction has already started - the point about rack being quick - but we see remote parent, would it satisfy the spec to overwrite the IDs of our previously created NR txn with the new ones coming in from the otel span? Idk if thats weird lol

lib/new_relic/agent/opentelemetry/trace/tracer.rb
kaylareopelle: Updated in 4ff194033b35a97563a3e38b80ada66e0e8922aa to use an invalid span, mirroring [what OTel's API does](https://github.com/open-telemetry/opentelemetry-ruby/blob/a6ce85bc55e06437814c5c13c477b6273d8b9566/api/lib/opentelemetry/trace/tracer.rb#L50-L68)

lib/new_relic/agent/opentelemetry/trace/tracer.rb
kaylareopelle: The response from Slack was "it should be the same behavior that your agent has when someone tries to start a new transaction in another transaction using custom instrumentation."

So, I think we're covered with the use of `start_transaction_or_segment`, where it'll start a transaction if one isn't happening, but otherwise it'll make a segment. 

Does that match your expectation of what the `start_transaction_or_segment` API does?

lib/new_relic/agent/opentelemetry/trace/tracer.rb
```diff
@@ -62,12 +80,32 @@ def get_otel_span_from_finishable(finishable)
...
+            return true if !parent_otel_context.valid? &&
+              KINDS_THAT_START_TRANSACTIONS.include?(kind)
+
+            # cannot start a transaction if it has an invalid context
+            # and has a kind of :client, :producer, :internal or nil
+            return false if !parent_otel_context.valid? &&
+              KINDS_THAT_DO_NOT_START_TXNS_WITHOUT_REMOTE_PARENT.include?(kind)
+
+            # if there's any stragglers
+            false
```
tannalynn: Is this like this to make future changes/expansion easier? Because line 98 doesn't need to exist if we always return false at the end anyways

lib/new_relic/agent/opentelemetry/trace/tracer.rb
```diff
@@ -62,12 +80,32 @@ def get_otel_span_from_finishable(finishable)
...
           end
 
-          def internal_span_kind_with_invalid_parent?(kind, parent_otel_context)
-            !parent_otel_context.valid? && kind == :internal
+          def should_not_create_telemetry?(parent_otel_context, kind)
+            !parent_otel_context.valid? &&
+              KINDS_THAT_DO_NOT_START_TXNS_WITHOUT_REMOTE_PARENT.include?(kind) &&
+              NewRelic::Agent::Tracer.current_transaction.nil?
```
tannalynn, line 105: Is it accurate the we do not want to create telemetry only when all 3 of these are true? I noticed it's using `&&` instead of `||`. 
For example, What happens when there is no current transaction, and the kind is client, but it does have a remote parent? Perhaps I'm misunderstanding what this is really doing. But it seems like that scenario would return an invalid span bc of this check and would never even get to the starting new transaction part?  Please let me know if i'm missing something or following the logic wrong

test/multiverse/suites/hybrid_agent/transaction_starting_test.rb
```diff
@@ -0,0 +1,213 @@
...
+            # just to be extra safe to make sure finishable is the correct class
+            NewRelic::Agent::Tracer.current_transaction&.finish
+          end
+
+          def teardown
+            NewRelic::Agent.instance.transaction_event_aggregator.reset!
+            NewRelic::Agent.instance.span_event_aggregator.reset!
+          end
+
+          def test_span_with_remote_parent_makes_web_transaction_when_kind_client
```
tannalynn, line 23: This test seems like it's testing the same scenario i was asking about in my other comment, so if this works then I must be missing something in my understanding.

lib/new_relic/agent/opentelemetry/trace/tracer.rb
kaylareopelle: Yeah, good point. I think I like Line 98 being explicit so that we can see what the logic is that correlates with the spec, but maybe that isn't necessary. I could also put those two conditions into a method since they're also used in the should_create_telemetry? method..

How about I remove 102? I was just feeling a little uncertain about what conditions could be hit, but probably better to get that feedback. It'll return nil without it, which is still falsey.

lib/new_relic/agent/opentelemetry/trace/tracer.rb
kaylareopelle: > For example, What happens when there is no current transaction, and the kind is client, but it does have a remote parent?

So in that case, we should start a transaction. The remote parent shows it's part of a distributed trace, and we want to keep that trace alive. The first condition (`!parent_otel_context.valid?`) would return false, because remote parents are valid otel contexts. Then, we'd jump to the `can_start_transaction?` condition, which should return true, and then call `start_transaction_or_segment`. 

In your comment later on, you found the test that's doing this. Does that help clear things up?

tannalynn: oooh okay, i think i was just not understanding what the `parent_otel_context.valid?` actually was doing and stuff, that totally makes sense, ty. I figured it must be working if theres literally a test for it but i was confused.

tannalynn (approved):

PR #3405, opened 2026-01-15
Parallel instrumentation

### LINKED ISSUE #3268: Add instrumentation for Parallel
Add instrumentation for the [Parallel gem](https://github.com/grosser/parallel) to support forking processes. Will likely need to be similar to the resque instrumentation and use the PipeChannelManager class for communicating between forked processes

tannalynn:
Adds instrumentation for the parallel gem. We aren't creating any spans or anything, all we need to do is set up the pipe channel manager to communicate data from the forked child processes back to the parent process that is what will send the data to the collector.

closes https://github.com/newrelic/newrelic-ruby-agent/issues/3268


test/multiverse/suites/parallel/parallel_marshalling_test.rb
```diff
@@ -0,0 +1,37 @@
...
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require File.join(File.dirname(__FILE__), '..', '..', '..', 'new_relic', 'marshalling_test_cases')
+
+if NewRelic::LanguageSupport.can_fork?
+
+  class ParallelMarshallingTest < Minitest::Test
+    include MultiverseHelpers
+    include MarshallingTestCases
```
tannalynn, line 11: the tests being run are these marshalling test cases. This is the same way we test the resque instrumentation that deals with forked processes. I did have to rewrite one of the tests (test_sends_metrics) to change the expectation because with parallel we expect to see some slightly different behavior than resque.

lib/new_relic/agent/instrumentation/parallel.rb
```diff
@@ -0,0 +1,26 @@
...
+DependencyDetection.defer do
+  @name = :parallel
+
+  depends_on do
+    defined?(Parallel) &&
+      NewRelic::LanguageSupport.can_fork?
+  end
+
+  executes do
+    NewRelic::Agent.logger.info('Installing Parallel instrumentation')
```
hannahramadan: I think our `log_and_instrument` method will take care of logging this message

```suggestion
```

lib/new_relic/agent/instrumentation/parallel.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    if use_prepend?
+      prepend_instrument Parallel.singleton_class, NewRelic::Agent::Instrumentation::Parallel::Prepend
+    else
+      NewRelic::Agent::Instrumentation::Parallel::Chain.instrument!
```
hannahramadan: Would the chain_instrument method make sense to use here? The chain and prepend methods are what trigger our info logging messaged about instrumentation

lib/new_relic/agent/pipe_service.rb
```diff
@@ -69,6 +69,12 @@ def shutdown
         @pipe.close if @pipe # rubocop:disable Style/SafeNavigation
       end
 
+      # Validates that data can be marshalled. For PipeService, we always
+      # use Marshal, so this always returns true.
+      def valid_to_marshal?(data)
```
hannahramadan, line 74: What is this new method doing/where is it being called? I see we have another `valid_to_marshal?` method in `new_relic_service.rb`. Are they related?

lib/new_relic/agent/instrumentation/parallel/instrumentation.rb
```diff
@@ -0,0 +1,49 @@
...
+
+        setup_for_txn_metric_merge_at_exit
+
+        yield
+      end
+
+      def setup_for_txn_metric_merge_at_exit
+        # Clear out any existing transaction metrics to prevent duplicates
+        # when merging metrics back in at the end of the forked process
+        if (txn = NewRelic::Agent::Tracer.current_transaction)
```
hannahramadan, line 23: So we can be confident parallel will never need to start a transaction/it will always be running inside a current?

lib/new_relic/agent/pipe_service.rb
tannalynn: At one point I was seeing an error about this method not existing, so I added it.

lib/new_relic/agent/instrumentation/parallel.rb
tannalynn: good point! changed

kaylareopelle (dismissed):

hannahramadan (dismissed):

hannahramadan (approved):

Issue #3279, opened 2025-09-30
Hybrid Agent: Spike on Transformation Rules


kaylareopelle:
OpenTelemetry Semantic Conventions for attributes and span names are different than the standards adopted by New Relic language agents. When the OpenTelemetry Bridge is in use, these attributes and span names should be converted to match New Relic standards. 

The full list of transformations is available [in the agent spec (internal)](https://source.datanerd.us/agents/agent-specs/blob/main/otel_bridge/Tracing-API.md#transformation-rules).

Spike on the best approach to implement this in the Ruby agent. 

Here are some example implementations in other languages:

.NET: 
* [RPC](https://github.com/newrelic/newrelic-dotnet-agent/pull/3253)
* [HTTP](https://github.com/newrelic/newrelic-dotnet-agent/pull/3246)
* [Message Queues](https://github.com/newrelic/newrelic-dotnet-agent/pull/3245)

Node: 
* newrelic/node-newrelic#3125
* newrelic/node-newrelic#3340

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-471806

PR #3410, opened 2026-01-23
Create an external request segment using OTel attributes

kaylareopelle:
When a :client type span is processed in an existing transaction, translate the OTel attributes to the required attrs for an ExternalRequestSegment. Update the segment before finish to make sure it has the http response code.

Translate v1.17 and v1.23 semantic conventions into the right fields.

This does not cover DB client spans and makes an assumption that all client spans are currently for external requests. 
In addition, it doesn't attempt to add external request segments in the case that a client span was received with a remote parent. In that case, a transaction would be started and the attributes need to be handled separately.

Relates to: #3279

### DIFF
--- a/lib/new_relic/agent/opentelemetry/trace/tracer.rb
+++ b/lib/new_relic/agent/opentelemetry/trace/tracer.rb
@@ -76,6 +76,10 @@ module NewRelic
           def start_otel_client_segment(name:, attributes: nil, start_timestamp: nil, kind: nil)
             attributes = NewRelic::EMPTY_HASH if attributes.nil?
             uri = create_uri(attributes)
+
+            # We need to have a URI to create an External Request Segment
+            return NewRelic::Agent::Tracer.start_segment(name: name) if uri.nil? || uri.empty?
+
             procedure = attributes['http.request.method'] || attributes['http.method']
 
             NewRelic::Agent::Tracer.start_external_request_segment(


lib/new_relic/agent/opentelemetry/trace/span.rb
```diff
@@ -11,9 +11,20 @@ class Span < ::OpenTelemetry::Trace::Span
...
 
           def finish(end_timestamp: nil)
+            update_client_span
             finishable&.finish
           end
 
+          def update_client_span
+            if finishable.is_a?(NewRelic::Agent::Transaction::ExternalRequestSegment) &&
+                finishable.http_status_code.nil?
+              finishable_attrs = finishable.attributes.custom_attributes
```
tannalynn, line 21: How come the status code is stored in the custom attributes there? Just curious, because that seemed unexpected.

tannalynn (approved):

lib/new_relic/agent/opentelemetry/trace/span.rb
kaylareopelle: Ha, I'm glad you noticed this! I meant to bring it up. (apologies in advance for the novel 📚 ) 

Right now, all the attributes from OTel are added as custom attributes on the segment/transaction on initialization ([here](https://github.com/newrelic/newrelic-ruby-agent/blob/bb1fa59f2ef297e0441169f6e44ebddd8d755c1d/lib/new_relic/agent/opentelemetry/trace/tracer.rb#L45)) and when the `add_attributes` or `set_attribute` OTel APIs are called ([defined here](https://github.com/newrelic/newrelic-ruby-agent/blob/c82fc54ffd07d8a157b24895586fc500b8bbeb62/lib/new_relic/agent/opentelemetry/trace/span.rb#L28-L34)).

NR's `add_custom_span_attributes` API was the most readily available to send attributes through to NR. Now that we're putting attributes in other places with specialized segment creation, this doesn't really make sense long term.

My current idea is to:
* Temporarily store attributes on the OTel span object when the OTel APIs are called
* Assign them using our NR-specific methods (ex. `start_external_request_segment`)
* Remove them from the attrs hash after they're assigned to avoid double sending data 
* When the span finishes, do any NR-specific final assignment (ex. http status code), and attach whatever we haven't assigned as custom attributes

This seemed like a bigger refactor than made sense in this PR, so I decided to leave it for another day. If you think it makes more sense to be part of this PR so that we establish a pattern before adding more attribute translation; I'm fine with that too.

Also open to other ideas to handle attribute storage/assignments and happy to chat/brainstorm together.

tannalynn: Ok cool that you for explaining that. I'm perfectly fine leaving it as is since that is a planned change you'll be making in the future, no need to force it in now!

kaylareopelle: Sounds good! I opened an issue for this to keep track of it: https://github.com/newrelic/newrelic-ruby-agent/issues/3422

lib/new_relic/agent/opentelemetry/segments/http_external.rb
```diff
@@ -0,0 +1,29 @@
...
+          def create_uri(attributes)
+            scheme = attributes['url.scheme'] || attributes['http.scheme']
+            host = attributes['server.address'] || attributes['net.peer.name']
+            port = attributes['server.port'] || attributes['net.peer.port']
+            path = attributes['url.path'] || attributes['http.target'] || NewRelic::SLASH
+
+            # if we don't have all the pieces we need to build a full URI,
+            # fall back to the full URL representations;
+            # though these don't usually have a port
+            if [scheme, host, port, path].any?(&:nil?)
```
hannahramadan, line 19: Cool syntax for checking if any are nil!

lib/new_relic/agent/opentelemetry/trace/tracer.rb
```diff
@@ -2,57 +2,95 @@
...
+            # tracer_provider = tracer_provider
             @name = name || ''
             @version = version || ''
           end
 
           def start_span(name, with_parent: nil, attributes: nil, links: nil, start_timestamp: nil, kind: nil)
             parent_otel_context = ::OpenTelemetry::Trace.current_span(with_parent).context
 
-            return Span::INVALID if should_not_create_telemetry?(parent_otel_context, kind)
+            return ::OpenTelemetry::Trace::Span::INVALID if should_not_create_telemetry?(parent_otel_context, kind)
```
hannahramadan, line 30: Why did we remove our version of an invalid span in `span.rb` and use Otel's directly?

kaylareopelle: Great question! Some of the tests started to fail because they looked specifically for the OpenTelemetry API version of the invalid span, which got me curious about the usefulness of our unique constant.

The OpenTelemetry API code has a few conditions that check specifically for equality with their `Span::INVALID` constant. This also extends to code in the OpenTelemetry Logs SDK. Those conditions would all fail with our new constant, so I thought it might be more beneficial to use OTel's constant instead.

One potential downside is that the OTel API invalid span won't have our custom instance variables, like finishable. So far, this hasn't posed a problem in our tests, but it may be that we need to add more testing to be sure.

Happy to take a closer look at this and/or remove this change from this PR and address it in a more specific PR.

hannahramadan: Hmm would invalid spans ever need finishable? or other NR things? From what I understand, invalid spans don't really do anything anyways so I think leaning on Otels invalid span directly sounds great to me! Thank you for clarifying ◡̈

PR #3413, opened 2026-01-27
Support Grape 3.1.0

### LINKED ISSUE #3412: Agent instrumentation incompatible with Grape 3.1.0
Grape 3.1.0 breaks current agent instrumentation. The agent is not seeing a calls class name, and therefore not able to build the expected transaction name.

hannahramadan:
Grape v3.1.0 made a [change](https://github.com/ruby-grape/grape/pull/2633): Removed attr_readers: Replaced instance and base attr_readers with direct @instance and @base instance variable access

Causing our check to get the API class, `(context.class.respond_to?(:base) && context.class.base)`, to fail. 

We now access @base though the instance variable.

Full CI Run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/21405695803

closes #3412

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+- **Feature: Add support for Grape v3.1.0**
+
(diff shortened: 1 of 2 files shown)


kaylareopelle (approved):
It's awesome that this change is backwards-compatible!

PR #3420, opened 2026-01-28
Logging gem instrumentation

### LINKED ISSUE #3301: Add log forwarding support for the `logging` gem
[`logging`](https://github.com/TwP/logging) is a popular Ruby logging library that's similar to Java's log4j. Currently, the Ruby agent does not support automatic forwarding for this library. It translates mdc (mapped diagnostic context) key value pairs into attributes on log events. See the [agent-spec (internal)](https://source.datanerd.us/agents/agent-specs/blob/main/Application-Logging.md#context-attributes) for details.

The work in https://github.com/newrelic/newrelic-ruby-agent/issues/2399 may be relevant here, since both libraries have user-defined attributes on log events.

hannahramadan:
Adds Logging gem instrumentation. 

closes #3301


lib/new_relic/agent/local_log_decorator.rb
```diff
@@ -29,8 +29,6 @@ def decorate(message)
 
       def decorating_enabled?
         NewRelic::Agent.config[:'application_logging.enabled'] &&
-          (NewRelic::Agent::Instrumentation::Logger.enabled? ||
```
hannahramadan, line 32: Checking for if logging instrumentations are enabled doesn't seem like it belongs in this method check. The LocalLogDecorator module is only ever used in logging-related instrumentations. It seems reasonable that a logging instrumentation would have to be enabled for us to get here.

lib/new_relic/agent/instrumentation/logging.rb
```diff
@@ -0,0 +1,23 @@
...
+  depends_on do
+    defined?(Logging::Logger) && NewRelic::Agent.config[:'application_logging.enabled']
+  end
+
+  executes do
+    require_relative 'logging/instrumentation'
+
+    if use_prepend?
+      require_relative 'logging/prepend'
+      prepend_instrument Logging::Logger, NewRelic::Agent::Instrumentation::Logging::Logger::Prepend, 'Logging'
```
hannahramadan, line 17: The Gem is called "logging", but somehow recording a supportability metric this feels maybe confusing?  I don't know what the alt would be? LoggingGem?  I am okay with it as is if other are, but just wanted to call this out.

test/new_relic/agent/local_log_decorator_test.rb
```diff
@@ -45,19 +45,6 @@ def test_does_not_decorate_if_local_decoration_disabled
         end
       end
 
-      def test_does_not_decorate_if_instrumentation_logger_and_logastasher_disabled
```
hannahramadan, line 48: Inline with my change to remove logger and logstasher checks from the decorating_enabled? method

lib/new_relic/agent/instrumentation/logging.rb
tannalynn: lmao some gems do be named like that. I'm good with whatever, Logging is fine to me because that is just what its called, but i'm totally fine with LoggingGem if that makes more sense to you.

lib/new_relic/agent/local_log_decorator.rb
tannalynn: Hmm, interesting. I wonder why we had it like that originally. 
It does seem odd that we would check if one of them is enabled with an || like that, right? because what if the log is coming from the other one? idk, curious. 
I agree that it makes sense to remove it, but to be safe, have you checked to make sure it works normally in like a test app?

hannahramadan: I think I added that `||` check awhile ago with LogStasher instrumentation, which was not the thing to do ☠️ idk probably pattern matching Logger instrumentation at the time

Just double checked a test app, and log decoration works as expected with this change.

lib/new_relic/agent/instrumentation/logging/instrumentation.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      def log_event_with_new_relic(event)
+        # If sending to multiple loggers, decorate each log
+        event.data = NewRelic::Agent::LocalLogDecorator.decorate(event.data)
+
+        # Prevents duplicate NR events when the same log goes through multiple loggers
+        if event.logger == @name
+          begin
+            mdc_data = capture_mdc_data
+            NewRelic::Agent.record_instrumentation_invocation(INSTRUMENTATION_NAME)
```
kaylareopelle, line 22: I'm curious, why put this down here rather than when the method is first called?

lib/new_relic/agent/instrumentation/logging/instrumentation.rb
```diff
@@ -0,0 +1,42 @@
...
+            NewRelic::Agent.agent.log_event_aggregator.record_logging_event(event, mdc_data)
+          rescue => e
+            NewRelic::Agent.logger.debug("Failed to capture Logging event: #{e.message}")
```
kaylareopelle, line 25: When we rescue the error here, will the log event still get emitted for the user?

lib/new_relic/agent/instrumentation/logging.rb
kaylareopelle: I'm also fine either way. It's a bit weird, but so is "logger". Having the instrumentation in a standard place makes me feel like it'll be easy to resolve any confusion about the name when someone looks at our repo.

lib/new_relic/agent/local_log_decorator.rb
kaylareopelle: I think this is here to make sure that decorating is allowed on the instrumentation level. It checks for the overall configs, but there could be a scenario where those are on, but the individual instrumentations are disabled. We can probably check this in a better way. Would this code even be reached if the instrumentation was disabled using `instrumentation.logger: disabled` for example?

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -22,6 +22,7 @@ class LogEventAggregator < EventAggregator
       SENT_METRIC = 'Supportability/Logging/Forwarding/Sent'.freeze
       LOGGER_SUPPORTABILITY_FORMAT = 'Supportability/Logging/Ruby/Logger/%s'.freeze
       LOGSTASHER_SUPPORTABILITY_FORMAT = 'Supportability/Logging/Ruby/LogStasher/%s'.freeze
+      LOGGING_SUPPORTABILITY_FORMAT = 'Supportability/Logging/Ruby/Logging/%s'.freeze
```
kaylareopelle, line 25: Now that we're getting more of these, I wonder if it makes sense to refactor so that we're passing two variables; one for the library and then one for the enabled/disabled? sprintf's are kind of hard to read when they have more than one variable, so it might not be worth it.

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -110,6 +111,29 @@ def record_logstasher_event(log)
         nil
       end
 
+      def record_logging_event(log, mdc_data)
```
kaylareopelle, line 114: Probs not for this PR, but now that we've reached the magic three instrumentations, I think we should find another way to organize these since we have a minimum of two methods that are unique to whatever library is being instrumented.

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -185,6 +209,33 @@ def add_logstasher_event_attributes(event, log)
...
+        event['method_name'] = log.method_name unless log.method_name.nil? || log.method_name.empty?
+        event['logger'] = log.logger unless log.logger.empty?
+        add_mdc_data_to_event(event, mdc_data) if !mdc_data.empty?
+
+        event
+      end
+
+      def add_mdc_data_to_event(event, mdc_data)
+        mdc_data.each do |key, value|
+          event["mdc.#{key}"] = value
```
kaylareopelle: If these are going to be attributes on the log event, I think [the spec](https://source.datanerd.us/agents/agent-specs/blob/main/Application-Logging.md#context-attributes) expects us to prefix these with `context.*`

test/multiverse/suites/logging/Envfile
```diff
@@ -0,0 +1,18 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+instrumentation_methods :chain, :prepend
+
+logging_versions = [
+  [nil, 2.7],
+  ['2.3.0', 2.6] # logging 2.3.0 is the earliest version to support Ruby 2.6
```
kaylareopelle: ❤️

test/new_relic/agent/local_log_decorator_test.rb
kaylareopelle: I think we might still want a test like this to pass even without the checks in the decorator? But since this is a unit test, maybe that's too difficult. Moving the logic to the multiverse might make more sense if we want to leverage the instrumentation being installed.

lib/new_relic/agent/instrumentation/logging/instrumentation.rb
hannahramadan: Ahh because if the instrumentation invocation was outside this `if`, it would be called per appender, so ie if one log is being sent 4 places, we would call the invocation 4 times. That didn't seem as accurate as calling it once per log.

lib/new_relic/agent/instrumentation/logging/instrumentation.rb
hannahramadan: Yes! The yield below still catches. Did the rescue here look off to you?

lib/new_relic/agent/instrumentation/logging.rb
hannahramadan: Thanks both! I will leave it as 'logging'

lib/new_relic/agent/local_log_decorator.rb
hannahramadan: I'm not sure I understand the scenario this was built for. But right now, the `LocalLogDecorator` class is only used in 3 spots - `logger/instrumentation.rb`, `logstasher/instrumentation.rb`, and now `logging/instrumentation.rb`. So if any of those instrumentations were disabled, those files would never run and this code would never be reached.

lib/new_relic/agent/log_event_aggregator.rb
hannahramadan: I agree! You mentioned this in another comment about how we've reached the magic 3, and a clean up organization is in order ✨ What do you think about doing that refactor (+ this change) in a separate PR once this is merged?

lib/new_relic/agent/log_event_aggregator.rb
hannahramadan: Thank you! Great eye 🦅 https://github.com/newrelic/newrelic-ruby-agent/pull/3420/commits/29463e8cf734c35736505f545a1f53946c16e1b1

lib/new_relic/agent/instrumentation/logging/instrumentation.rb
kaylareopelle: Oh interesting! That makes sense. Thank you.

lib/new_relic/agent/instrumentation/logging/instrumentation.rb
kaylareopelle: It did, but I think that's just because I'm so used to the method-level rescues now that I forgot this one was nested within its own begin block. Thank you!

lib/new_relic/agent/local_log_decorator.rb
kaylareopelle: Got it. I should've looked up where it was called 🤦 Thank you for clarifying and for fixing this up!

lib/new_relic/agent/log_event_aggregator.rb
kaylareopelle: That sounds great! Could you make a ticket for that work?

kaylareopelle (approved):

PR #3430, opened 2026-02-03
Add instrumentation registry to Tracer Provider

### LINKED ISSUE #3407: Hybrid Agent: Support include/exclude rules
[Spec](https://source.datanerd.us/agents/agent-specs/blob/main/otel_bridge/Tracing-API.md#includeexclude-rules) (Internal)

The New Relic agent needs to handle cases where both OpenTelemetry and New Relic have instrumented the same library/method. This can be done with include/exclude configuration documented in the agent spec. 

* Implement the include / exclude configs
* Determine a default exclude list for duplicate instrumentation 
* Make sure the spec'd precedence order is applied
* The include/exclude configs should only work with exact matches

kaylareopelle:
An instrumentation registry is needed to keep track of the tracers we're instrumenting:
* We need to make sure unique tracers are created for every unique name provided
* We need to access an already existing tracer for a given name rather than create a new one
* The `opentelemetry.traces.include` / `exclude` configuration options also determine what tracers we will instrument
* We skip excluded tracers when we install instrumentation to prevent unnecessary prepending

In addition, this PR also updates the default for opentelemetry.traces.enabled to match the current spec.

See [Tracing API Spec - Include/Exclude Rules](https://source.datanerd.us/agents/agent-specs/blob/main/otel_bridge/Tracing-API.md#includeexclude-rules) (Internal) for details.

Closes #3407

### DIFF
--- a/lib/new_relic/agent/opentelemetry_bridge.rb
+++ b/lib/new_relic/agent/opentelemetry_bridge.rb
@@ -5,6 +5,8 @@
 module NewRelic
   module Agent
     class OpenTelemetryBridge
+      DEFAULT_EXCLUDED_TRACERS = %w[elasticsearch-api dalli].freeze
+
       def initialize
         # currently, we only have support for traces
         # this method should change when we add support for metrics and logs.
@@ -33,13 +35,21 @@ module NewRelic
         install_instrumentation
       end
 
+      def self.calculate_excluded_tracers
+        excluded_names = NewRelic::Agent.config[:'opentelemetry.traces.exclude'].split(',').map(&:strip)
+        included_names = NewRelic::Agent.config[:'opentelemetry.traces.include'].split(',').map(&:strip)
+
+        # Priority order (highest to lowest):
+        # 1. Configured exclude list - always excluded
+        # 2. Configured include list - overrides default exclude
+        # 3. Default exclude list - excluded unless overridden by configured include
+        excluded_names + (DEFAULT_EXCLUDED_TRACERS - included_names)
+      end
+
       def self.install_instrumentation
         return unless defined?(::OpenTelemetry::Instrumentation::Registry)
 
-        # Cache the exclude list computation - do string splits once
-        excluded_names = NewRelic::Agent.config[:'opentelemetry.traces.exclude'].split(',').map(&:strip)
-        included_names = NewRelic::Agent.config[:'opentelemetry.traces.include'].split(',').map(&:strip)
-        excluded_set = (excluded_names - included_names).to_set
+        excluded_set = calculate_excluded_tracers.to_set
 
         return ::OpenTelemetry::Instrumentation.registry.install_all if excluded_set.empty?
 
--- a/lib/new_relic/agent/opentelemetry_bridge.rb
+++ b/lib/new_relic/agent/opentelemetry_bridge.rb
@@ -5,7 +5,11 @@
 module NewRelic
   module Agent
     class OpenTelemetryBridge
-      DEFAULT_EXCLUDED_TRACERS = %w[elasticsearch-api dalli].freeze
+            # Exclude tracers from native OTel instrumentation by default to prevent double-reporting.
+            # Affects libraries that the Ruby agent already instruments automatically.
+            # Can be overridden via the opentelemetry.traces.include configuration.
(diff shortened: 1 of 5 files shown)


lib/new_relic/agent/opentelemetry_bridge.rb
```diff
@@ -30,6 +30,34 @@ def self.install
...
+        install_instrumentation
+      end
+
+      def self.install_instrumentation
+        return unless defined?(::OpenTelemetry::Instrumentation::Registry)
+
+        # Cache the exclude list computation - do string splits once
+        excluded_names = NewRelic::Agent.config[:'opentelemetry.traces.exclude'].split(',').map(&:strip)
+        included_names = NewRelic::Agent.config[:'opentelemetry.traces.include'].split(',').map(&:strip)
+        excluded_set = (excluded_names - included_names).to_set
```
tannalynn: I was looking at the spec for this and it says that the order of precedence for exclude/include depends on whether it is the default exclude list (which is bypassed by include, like its doing here), but the highest precedence is supposed to be the customer configured exclude list. 
From the spec (lower is higher precedence for some reason)
```
Built-in exclude list
Customer configured include list
Customer configured exclude list
```

But with the logic I see here, we aren't following that, since we always will remove it from include even if the customer puts it on both. 

Would anyone actually be doing that? they probably shouldn't be, but I did notice it was different from the spec, so I wanted to point it out. 

I could be wrong though, let me know if I am misunderstanding anything.

kaylareopelle: We talked about this sync. I'm going to adjust the exclude list creation process to better adhere to the spec.

kaylareopelle: @tannalynn - fixed this in [6bcb15d](https://github.com/newrelic/newrelic-ruby-agent/pull/3430/commits/6bcb15db8cfd14829ef7d77b297c7760ef8e6140)!

tannalynn (dismissed):
Looks good, thanks for looking into it! 
I am curious though about the things included in the default list, why were those chosen? No need to change them or anything, just wanted to know!

kaylareopelle:
> Looks good, thanks for looking into it! I am curious though about the things included in the default list, why were those chosen? No need to change them or anything, just wanted to know!

@tannalynn - Those are on the default list because they're the names of tracers created by native instrumentation. In OTel, native instrumentation is when a library includes calls to the OTel APIs within their code, instead of needing a secondary library to instrument. Native instrumentation is normally set up to try to send the data as long as the `opentelemetry-api` is present. Since our OTel bridge expects the same thing, the default list tries to protect users from accidental double instrumentation.

lib/new_relic/agent/opentelemetry_bridge.rb
```diff
@@ -5,6 +5,8 @@
 module NewRelic
   module Agent
     class OpenTelemetryBridge
+      DEFAULT_EXCLUDED_TRACERS = %w[elasticsearch-api dalli].freeze
```
hannahramadan: I saw this in a discussion comment, what do you think about adding a little info on what this list is as a comment?
```suggestion
      # Remove tracers created by native instrumentations. This helps prevent double reporting. 
      # https://opentelemetry.io/ecosystem/registry/?language=ruby&flag=native
      DEFAULT_EXCLUDED_TRACERS = %w[elasticsearch-api dalli].freeze
```

hannahramadan (dismissed):

lib/new_relic/agent/opentelemetry_bridge.rb
kaylareopelle: Great idea! What do you think about adding this in as well? It might be overkill on the details?
```suggestion
            # Exclude tracers from native OTel instrumentation by default to prevent double-reporting.
            # Affects libraries that the Ruby agent already instruments automatically.
            # Can be overridden via the opentelemetry.traces.include configuration.
            # https://opentelemetry.io/ecosystem/registry/?language=ruby&flag=native
            DEFAULT_EXCLUDED_TRACERS = %w[elasticsearch-api dalli].freeze
```

hannahramadan: I love details! This looks great ◡̈

tannalynn (approved):

PR #3433, opened 2026-02-04
Translate server attributes

### LINKED ISSUE #3435: Hybrid Agent: Translate server span attributes
When an OpenTelemetry Span is created with a server kind, there are certain attributes we want to apply to the agent attributes within the New Relic-created telemetry. Translate these attributes for transactions and segments. See the agent spec for which attributes should be translated.

kaylareopelle:
This is the final attribute translation PR before our initial release.

This one maps server attributes to the correct New Relic fields and also updates transaction names to include those attributes when the right values are present. If the right values are not available, the name provided through the OTel API will be used instead.

Closes: #3435

### DIFF
--- a/lib/new_relic/agent/opentelemetry/segments/server.rb
+++ b/lib/new_relic/agent/opentelemetry/segments/server.rb
@@ -7,16 +7,22 @@ module NewRelic
     module OpenTelemetry
       module Segments
         module Server
-          def set_server_transaction_name(original_name, tracer_name, attributes)
+          def create_server_transaction_name(original_name, tracer_name, attributes)
             attributes ||= NewRelic::EMPTY_HASH
-            host = attributes['server.address'] || attributes['http.host']
             method = attributes['http.request.method'] || attributes['http.method']
             path = attributes['url.path'] || attributes['http.target']
 
-            if [host, method, path].any?(&:nil?)
-              original_name
+            if method && path
+              # TransactionNamer.name_for is used in ControllerInstrumentation
+              # This will produce a name that looks something like:
+              # "Controller/OpenTelemetry::Instrumentation::Rack/GET /path"
+              # "method /path" is roughly what the semantic conventions have for
+              # a server span name in the stable conventions, but the old
+              # conventions prefix with HTTP; so we create the string
+              # ourselves until only the stable conventions are used
+              Instrumentation::ControllerInstrumentation::TransactionNamer.name_for(nil, nil, :web, {class_name: tracer_name, name: "#{method} #{path}"})
             else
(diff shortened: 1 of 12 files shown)


lib/new_relic/agent/opentelemetry/segments/server.rb
```diff
@@ -0,0 +1,51 @@
...
+              # conventions prefix with HTTP; so we create the string
+              # ourselves until only the stable conventions are used
+              Instrumentation::ControllerInstrumentation::TransactionNamer.name_for(nil, nil, :web, {class_name: tracer_name, name: "#{method} #{path}"})
+            else
+              original_name
+            end
+          end
+
+          def update_request_attributes(nr_item, attributes)
+            return unless nr_item.is_a?(Transaction)
```
kaylareopelle, line 30: I did some digging and I couldn't find any scenario where a segment would update request attributes after a transaction is started. It seems to only attempt to assign things to the RequestAttributes class on initialization when the `:request` key is present in the options hash.

lib/new_relic/agent/opentelemetry/segments/server.rb
```diff
@@ -0,0 +1,51 @@
...
+
+            if method && path
+              # TransactionNamer.name_for is used in ControllerInstrumentation
+              # This will produce a name that looks something like:
+              # "Controller/OpenTelemetry::Instrumentation::Rack/GET /path"
+              # "method /path" is roughly what the semantic conventions have for
+              # a server span name in the stable conventions, but the old
+              # conventions prefix with HTTP; so we create the string
+              # ourselves until only the stable conventions are used
+              Instrumentation::ControllerInstrumentation::TransactionNamer.name_for(nil, nil, :web, {class_name: tracer_name, name: "#{method} #{path}"})
```
kaylareopelle, line 23: Hannah, I know we talked about using the original provided name, but that had some side effects that ultimately broke cross-agent tests. This strategy is a little safer to keep those passing.

test/multiverse/suites/hybrid_agent/server_mapping_test.rb
kaylareopelle, line 1: I think I may have missed stubbing the `sampled?` method somewhere in these test files during either an `in_transaction` call or a `start_span` call. My hunch is that's why we're getting some intermittent failures when I run this locally. Please keep an eye out for any missing `sampled?` stubs!

kaylareopelle:
CI Failure looks like it's Dalli-related?

tannalynn (approved):

hannahramadan (approved):

PR #3434, opened 2026-02-04
Fix context bug

kaylareopelle:
The agent would log a warning and in some cases log an error if trace context was nil and this method was called. Trace context can be nil if we're looking at a root span, so we allow that and just return the passed-in context. This leaves the error message for unexpected behavior.

It also allows the OTel propagation method to more closely match the behavior in our `TraceContext#parse` method to handle `nil` trace parents.

### DIFF
--- a/test/multiverse/suites/hybrid_agent/trace_propagator_test.rb
+++ b/test/multiverse/suites/hybrid_agent/trace_propagator_test.rb
@@ -54,6 +54,33 @@ module NewRelic
               assert_instance_of(::OpenTelemetry::Context, result)
             end
 
+            def test_extract_returns_context_when_trace_context_parse_returns_nil
+              carrier = {'traceparent' => 'invalid-trace-parent'}
+              context = ::OpenTelemetry::Context.current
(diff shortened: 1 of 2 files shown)


test/multiverse/suites/hybrid_agent/trace_propagator_test.rb
```diff
@@ -54,6 +54,33 @@ def test_extract_with_empty_carrier_returns_context
...
+                assert_equal(context, result, 'Should return the original context when trace_context is nil')
+              end
+            end
+
+            def test_extract_does_not_log_error_when_trace_context_is_nil
+              carrier = {'traceparent' => 'malformed-data'}
+              result = nil
+
+              NewRelic::Agent::DistributedTracing::TraceContext.stub(:parse, nil) do
+                log = with_array_logger(:debug) do
```
kaylareopelle, line 75: it's supposed to be an error-level log, so could change this one to error?

tannalynn (approved):

PR #3439, opened 2026-02-05
Vendor detect parallel config

tannalynn:
Implements a customer request to opt in to sequential vendor metadata detection to avoid creating extra threads during the agent start up. 
Adds the config `utilization.detect_in_parallel` that when disables, uses the old logic of sequentially checking for vendor metadata.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
   A new configuration option has been added, `error_collector.backtrace_truncate_location`, which allows the user to specify where in the backtrace to truncate when the number of frames exceeds `error_collector.max_backtrace_frames`. Options are `'top'` (removes frames from the beginning), `'middle'` (removes frames from the middle, preserving the beginning and end), or `'end'` (removes frames from the end). The default is `'middle'`. [PR#3424](https://github.com/newrelic/newrelic-ruby-agent/pull/3424)
 
+- **Feature: Add configuration option utilization.detect_in_parallel**
+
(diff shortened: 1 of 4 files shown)


lib/new_relic/agent/utilization_data.rb
hannahramadan, line 120: I feel like its chill because this is how our original code was, but any benefit to wrapping the collector_has stuff in a mutex.synchronize block?

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
   A new configuration option has been added, `error_collector.backtrace_truncate_location`, which allows the user to specify where in the backtrace to truncate when the number of frames exceeds `error_collector.max_backtrace_frames`. Options are `'top'` (removes frames from the beginning), `'middle'` (removes frames from the middle, preserving the beginning and end), or `'end'` (removes frames from the end). The default is `'middle'`. [PR#3424](https://github.com/newrelic/newrelic-ruby-agent/pull/3424)
 
+- **Feature: Add configuration option utilization.detect_in_parallel**
+
+  A new configuration option has been added, `utilization.detect_in_parallel`, which controls whether the agent uses threads when detecting cloud vendor information to speed up agent startup. When set to `false`, vendor detection runs sequentially without creating threads. The default is `true`. [PR#3439](https://github.com/newrelic/newrelic-ruby-agent/pull/3439)
```
kaylareopelle, line 11: What do you think about including something about the glibc bug?

lib/new_relic/agent/utilization_data.rb
tannalynn: I don't think we need it. mostly because even though this gets split into multiple threads, only one would ever succeed, so the if will succeed and modify the collector_hash only one time (or none times). I dont think there could be a situation where both the aws and azure metadata would exist at the same time.

hannahramadan (approved):

Issue #3140, opened 2025-04-21
Verify behavior when both the `opentelemetry-sdk` and `newrelic_rpm` gem are installed and enabled


kaylareopelle:
The `OpenTelemetryBridge` installation process uses the same strategy as the `opetnelemetry-sdk` to initialize the tracer provider:
 
https://github.com/open-telemetry/opentelemetry-ruby/blob/b2cac915e2662f91e032097c4e0c6e9395128d64/sdk/lib/opentelemetry/sdk/configurator.rb#L150

https://github.com/newrelic/newrelic-ruby-agent/blob/813f7eeac10e9fff635225fb43e0e5dea0a6c6b3/lib/new_relic/agent/opentelemetry_bridge.rb#L22

Investigate whether this will cause problems when: 
* both gems are installed, but neither enabled
* both gems are installed, but only `OpenTelemetry::SDK.configure` is called
* both gems are installed, but only `opentelemtry_bridge.enabled` is true in `newrelic_rpm`
* both gems are installed, and both `OpenTelemetry::SDK.configure` is called and `opentelemetry_bridge.enabled` is true
* both gems are installed and a tracer provider is manually applied using the `OpenTelemetry::SDK` gem and `opentelemetry_bridge.enabled` is also true

(as well as any other scenarios the person working on this sees fit)

hannahramadan:
One issue resolved: https://github.com/newrelic/newrelic-ruby-agent/pull/3452

PR #3452, opened 2026-02-12
Fix context sharing between NR and OTel

hannahramadan:
Previously, OTel would log the following error, suggesting a context sharing mismatch:
```
ERROR -- : OpenTelemetry error: calls to detach should match corresponding calls to attach.
```
By patching OTel's `current_span` method we allow OTel to context manage when a context is available or return the current new relic span when available. 

Relates to #3140


lib/new_relic/agent/opentelemetry_bridge.rb
```diff
@@ -68,12 +33,14 @@ def self.install
...
         require_relative 'opentelemetry/context'
+        require_relative 'opentelemetry/trace_patch'
 
-        # TODO: Add a warning if SDK gem is installed
+        NewRelic::Agent.logger.warn('OpenTelemetry SDK gem is installed. This may interfere with New Relic instrumentation.') if defined?(OpenTelemetry::SDK)
 
         ::OpenTelemetry.tracer_provider = OpenTelemetry::Trace::TracerProvider.new
         Transaction.prepend(OpenTelemetry::TransactionPatch)
         ::OpenTelemetry.propagation = OpenTelemetry::Context::Propagation::TracePropagator.new
+        ::OpenTelemetry::Trace.singleton_class.prepend(NewRelic::Agent::OpenTelemetry::TracePatch)
```
kaylareopelle: Small thing -- wdyt about putting the prepends together on top and then the tracer_provider/propagation calls after?

hannahramadan: Great idea! Added some spacing too https://github.com/newrelic/newrelic-ruby-agent/pull/3452/commits/be8ec906297f4b0bb290a8ae1f1c9f6dcfcd5d3d

kaylareopelle (approved):

PR #3463, opened 2026-02-18
Reduce ActionCable broadcast metric cardinality

### LINKED ISSUE #3466: Create an option to avoid high cardinality problems for Action Cable broadcast metrics
Previously, the broadcast metrics used the `:broadcasting` value from the Active Support Notifications payload. However, this metric has much higher cardinality than we realized when we initially rolled out this instrumentation. 

To get around this, we will create a configuration option enables simpler metric names and attach the `:broadcasting` value as an attribute on the span event.

kaylareopelle:
Previously, the metrics for Action Cable broadcast included the value for the `broadcasting` from the Active Support Notification payload. This value can have very high cardinality. 

Now, we provide a configuration option, `:simplify_action_cable_broadcast_metrics` to allow users to remove the `broadcasting` value from the metric and add it to the span's agent attributes instead. The new metric, using this config, will look like: `"Ruby/ActionCable/broadcast"`

Closes #3466


lib/new_relic/agent/instrumentation/action_cable_subscriber.rb
```diff
@@ -29,11 +32,26 @@ def transaction_name_from_payload(payload)
...
+          else
+            (payload[:broadcasting] || payload[:channel_class]) + '/'
+          end
+        end
+
+        def add_broadcasting_attribute(segment, payload)
+          return unless NewRelic::Agent.config[:simplify_action_cable_broadcast_metrics]
+          return unless payload.key?(:broadcasting)
+
+          segment.transaction.add_agent_attribute(
```
hannahramadan, line 53: Really like the idea to add as an attribute if we are simplifying the metric!

lib/new_relic/agent/instrumentation/action_cable_subscriber.rb
```diff
@@ -29,11 +32,26 @@ def transaction_name_from_payload(payload)
         end
 
         def metric_name_from_payload(name, payload)
-          "Ruby/ActionCable/#{metric_name(payload)}/#{action_name(name)}"
+          "Ruby/ActionCable/#{metric_name(payload)}#{action_name(name)}"
```
hannahramadan, line 35: You explained this change well in the tests! Thank you. I wondered at first if we'd want a comment in the code, but can't decide. In 2 years, would you want that noted in the code or should the test cases be enough? I'm torn lol

hannahramadan (dismissed):

lib/new_relic/agent/instrumentation/action_cable_subscriber.rb
kaylareopelle: Haha I am also torn, but since you had the same thought, I'm down to add it to the code.

lib/new_relic/agent/instrumentation/action_cable_subscriber.rb
```diff
@@ -29,11 +32,29 @@ def transaction_name_from_payload(payload)
...
         end
 
         def metric_name(payload)
-          payload[:broadcasting] || payload[:channel_class]
+          # The trailing / is added in the metric_name method to protect against
+          # double / characters in the name because there are some cases where
+          # metric_name will return nil
```
kaylareopelle, line 41: @hannahramadan - Just realized I put this under a different method than you original comment. Would you rather this live within `metric_name_from_payload`?

hannahramadan: I think this is fine! Thank you for adding ◡̈

hannahramadan (dismissed):

hannahramadan (approved):

PR #3472, opened 2026-02-24
Add API `add_transaction_log_attributes`

### LINKED ISSUE #3423: Add custom attributes to log per transaction
Create an api method that would allow a customer to add custom log attributes for all logs created in the scope of the current transaction.

[Original FR ](https://new-relic.atlassian.net/browse/NR-475183)

hannahramadan:
New API to add transaction-scoped custom attributes to log events for the current transaction.

`NewRelic::Agent#add_transaction_log_attributes`

closes #3423

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+- **Feature: Add new `NewRelic::Agent#add_custom_transaction_log_attributes` API**
+
+  A new API, `NewRelic::Agent#add_custom_transaction_log_attributes`, allows users to add transaction-scoped custom attributes to log events for the current transaction. These attributes will only be applied to logs created within the scope of the current transaction. [PR#3472](https://github.com/newrelic/newrelic-ruby-agent/pull/3472)
+
 ## v10.2.0
 
 - **Feature: Introduce Hybrid Agent for OpenTelemetry Tracing Support**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## dev
 
-- **Feature: Add new `NewRelic::Agent#add_custom_transaction_log_attributes` API**
+- **Feature: Add new `NewRelic::Agent#add_transaction_log_attributes` API**
 
-  A new API, `NewRelic::Agent#add_custom_transaction_log_attributes`, allows users to add transaction-scoped custom attributes to log events for the current transaction. These attributes will only be applied to logs created within the scope of the current transaction. [PR#3472](https://github.com/newrelic/newrelic-ruby-agent/pull/3472)
+  A new API, `NewRelic::Agent#add_transaction_log_attributes`, allows users to add transaction-scoped custom attributes to log events for the current transaction. These attributes will only be applied to logs created within the scope of the current transaction. [PR#3472](https://github.com/newrelic/newrelic-ruby-agent/pull/3472)
 
 ## v10.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 ## dev
 
-- **Feature: Add new `NewRelic::Agent#add_transaction_log_attributes` API**
-
 - **Feature: Add new `NewRelic::Agent.add_transaction_log_attributes` API**
 
   A new API, `NewRelic::Agent.add_transaction_log_attributes`, allows users to add transaction-scoped custom attributes to log events for the current transaction. These attributes will only be applied to logs created within the scope of the current transaction. [PR#3472](https://github.com/newrelic/newrelic-ruby-agent/pull/3472)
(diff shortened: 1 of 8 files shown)


lib/new_relic/agent.rb
```diff
@@ -921,6 +921,55 @@ def add_custom_log_attributes(params)
...
+    #
+    #                         Attribute pairs with empty or nil contents
+    #                         will be dropped.
+    #
+    #                         If called outside of a transaction context,
+    #                         the method will log a warning and return without
+    #                         adding attributes.
+    # @!scope class
+    # @api public
+    def add_custom_transaction_log_attributes(params)
```
hannahramadan: This name feels a touch long to me, could also do `add_transaction_log_attributes` or open to other ideas

tannalynn: hmmm. yeah it does seem long but all the other ideas i can come up with are just as long, or longer lol. i think add_transaction_log_attributes is probably better but they are basically the same, the word custom probably doesn't really need to be in it i think, since it's specifically about log attributes. what do you think

hannahramadan: I think you're right, custom doesn't need to be there and it sounds so much better without it haha. Updating!

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -154,8 +154,11 @@ def increment_event_counters(severity)
       def record_batch(txn, logs)
         # Ensure we have the same shared priority
         priority = LogPriority.priority_for(txn)
+        txn_attrs = txn.log_attributes&.custom_attributes
+
         logs.each do |log|
           log.first[PRIORITY_KEY] = priority
+          log.last.merge!(txn_attrs) if txn_attrs && !txn_attrs.empty?
```
kaylareopelle, line 161: What is `log.last` here?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+- **Feature: Add new `NewRelic::Agent#add_transaction_log_attributes` API**
+
+  A new API, `NewRelic::Agent#add_transaction_log_attributes`, allows users to add transaction-scoped custom attributes to log events for the current transaction. These attributes will only be applied to logs created within the scope of the current transaction. [PR#3472](https://github.com/newrelic/newrelic-ruby-agent/pull/3472)
```
kaylareopelle: The new API is a class method rather than an instance method, so we should swap the `#` with `.`.

```suggestion
- **Feature: Add new `NewRelic::Agent.add_transaction_log_attributes` API**

  A new API, `NewRelic::Agent.add_transaction_log_attributes`, allows users to add transaction-scoped custom attributes to log events for the current transaction. These attributes will only be applied to logs created within the scope of the current transaction. [PR#3472](https://github.com/newrelic/newrelic-ruby-agent/pull/3472)
```

test/new_relic/agent/log_event_aggregator_test.rb
```diff
@@ -784,6 +784,29 @@ def test_labels_excluded_only_for_case_insensitive_match
       assert_labels(config, expected_label_attributes)
     end
 
+    def test_record_batch_with_transaction_attributes
+      in_transaction do |txn|
+        txn.add_transaction_log_attributes({'txn1_key' => 'txn1_value', 'txn1_another_key' => 'txn1_another_value'})
+        @aggregator.record('Just some general information', 'INFO')
+      end
```
kaylareopelle, line 791: What do you think about adding one more test that makes sure multiple logs within the same transaction all get the same attributes?

lib/new_relic/agent/log_event_aggregator.rb
hannahramadan: This is a great question! Each log is a two item array that looks something like this:

```ruby
[{"priority" => 1.276657},
 {"level" => "INFO",
  "timestamp" => 1772490700928.16,
  "message" => "Test message",
  "trace.id" => "c1d9f0f2081bbd486f03a9099c911667",
  "span.id" => "ad5edb55e8671f8f",
  "level_number" => 1]
```

So we add any transaction attributes with the rest of the attributes!

kaylareopelle: Thank you!

CHANGELOG.md
kaylareopelle: Whoops, I may have done this incorrectly. I think my suggestion missed the first line of the original entry.

test/new_relic/agent/log_event_aggregator_test.rb
hannahramadan: added! https://github.com/newrelic/newrelic-ruby-agent/pull/3472/changes/b2af78160635739760fa1bccf1243d41d86331f6

CHANGELOG.md
hannahramadan: Thank you ◡̈ Updated this

kaylareopelle (approved):

tannalynn (approved):

PR #3481, opened 2026-03-06
Add ignored_middleware_classes config

### LINKED ISSUE #1814: Allow to ignore certain middleware, e.g. Rack::Cors
We would like to ignore events coming from CORS requests that are handled in our Rails app via `Rack::Cors`. They are polluting our statistics in NewRelic and they also generate higher cost expenses as the number of transactions is massive!

### Feature Description
We would like to be able to explicitly ignore this middleware or this type of request transactions.

### Describe Alternatives
Alternatively, we would like to know how the same could be achieved with the existing ruby agent and a special type of configuration or initialization.

### Priority
Please help us better understand this feature request by choosing a priority from the following options:
**Must Have**

kaylareopelle:
This configuration option will allow users to supply a list of Rack middlewares for the agent to ignore. When the agent starts, it wraps middlewares that we want to instrument. Now, the new configuration will be taken into account during that wrapping process.

closes #1814


hannahramadan (approved):
✨

tannalynn (approved):

PR #3483, opened 2026-03-09
Use Fiber ID for Falcon

### LINKED ISSUE #3190: Per-fiber (rather that per-thread) transactions
I've added instrumentation to a simple reverse proxy that uses the falcon web server, which uses separate fibers (rather than threads) to handle requests. I have data coming into the dashboard, and it looks reasonable, but I also have a lot of warnings in my logs like: "Unable to calculate elapsed transaction time for <transaction name>."

Now, I'm not certain of the cause (other than that some transactions do not have start times), but based on what I'm seeing in the code, it appears that in`NewRelic::Agent::TransactionTimeAggregator`, the `@stats` variable is a hash that is keyed on thread id. Since multiple fibers running in the same thread will share the same thread, I'm thinking that the stats for multiple requests may be conflated and that the missing start times may be caused by certain requests finishing and clearing out their own data in the hash -- which would then affect other ongoing requests. If that's the case, then my transaction timings are pretty much meaningless.

Of course, I understand that falcon is not a supported framework, but this is less about supporting a particular framework than about the assumption that a single thread can only have a single transaction at any point in time.

Also, I'm not entirely sure my analysis above is correct. The most important thing for me, right now, is to know whether I can trust the transaction timings I see, at all. The thing is, they really do look reasonable.

### Feature Description
An option (maybe similar to `ActiveSupport::IsolatedExecutionState`) to specify whether transaction data should be stored per-thread or per-fiber.

### Describe Alternatives
Aside from explicitly passing transactions around, which would intertwine instrumentation with application code, I can't think of any.

### Additional context
None

### Priority
Please help us better understand this feature request by choosing a priority from the following options:
[Nice to Have, Really Want, Must Have, Blocker]

Depends on whether I'm actually right about what's happening, but if so, then... really want, I suppose.

### LINKED ISSUE #3482: errors logged under falcon server
## Description

Recently migrated to using falcon from puma on Heroku and noticing there are errors logged like this:

Image

~~Note these only appear to show up for scheduler dynos, not for web or worker dynos. Only the web dyno is running under falcon. The rest are not running under async frameworks right now.~~ Sorry, my earlier observation was incorrect, these errors are showing up for all dyno types.

Note there are other warnings being reported for the web dyno as reported [here](https://github.com/newrelic/newrelic-ruby-agent/issues/3190#issuecomment-4009849157).

## Expected Behavior

There should be no errors in the logs.

## [Troubleshooting](https://forum.newrelic.com/s/hubtopic/aAX8W0000008bSoWAI/troubleshooting-frameworks) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results

## Steps to Reproduce
TIP:  Link a sample application that demonstrates the issue. 

## Your Environment

Ruby 3.4.8
Rails 8.1.2
falcon 0.54.3
falcon-rails 0.2.4

## Additional context
See also https://github.com/newrelic/newrelic-ruby-agent/issues/3190#issuecomment-4009849157

hannahramadan:
Falcon uses fibers for concurrency. The agent's `TransactionTimeAggregator` was using `Thread.current.object_id` to track transaction state, causing collisions when multiple fibers on the same thread had overlapping transactions. Now when Falcon is the dispatcher, use `Fiber.current.object_id` instead.

Additionally, Falcon only loads `async/http/server`, not the full async-http gem, so `Async::HTTP::VERSION` may not be available. Added a `defined?` check to prevent the NameError.

closes #3482
closes #3190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@
 
 - **Bugfix: Fix "Unable to calculate elapsed transaction time" warnings when using Falcon web server**
 
-  The agent now uses `Fiber.current.object_id` instead of `Thread.current.object_id` to track transaction state when running under Falcon, preventing collisions from concurrent requests sharing the same thread. Also fixes a NameError: uninitialized constant `Async::HTTP::VERSION` when using Falcon. Thanks to [@97jaz](https://github.com/97jaz) and [@gsar](https://github.com/gsar) for bringing this to our attention. [PR#3483](https://github.com/newrelic/newrelic-ruby-agent/issues/3483)
+  The agent now uses `Fiber.current.object_id` instead of `Thread.current.object_id` to track transaction state when running under Falcon, preventing collisions from concurrent requests sharing the same thread. Also fixes a "NameError: uninitialized constant `Async::HTTP::VERSION`" when using Falcon. Thanks to [@97jaz](https://github.com/97jaz) and [@gsar](https://github.com/gsar) for bringing this to our attention. [PR#3483](https://github.com/newrelic/newrelic-ruby-agent/issues/3483)
 
 - **Bugfix: Fix typo in harvest.rb causing NoMethodError**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@
 
   Previously, the agent would cause a ~3 second shutdown delay when running the `rails test` command. The `Rails::Command::TestCommand` constant has been added to the default `autostart.denylisted_constants` list to prevent the agent from starting during Rails test runs. Thanks to [@varyform](https://github.com/varyform) for bringing this to our attention. [PR#3478](https://github.com/newrelic/newrelic-ruby-agent/issues/3478)
 
+- **Bugfix: Fix "Unable to calculate elapsed transaction time" warnings when using Falcon web server**
+
+  The agent now uses `Fiber.current.object_id` instead of `Thread.current.object_id` to track transaction state when running under Falcon, preventing collisions from concurrent requests sharing the same thread. Also fixes a NameError: uninitialized constant `Async::HTTP::VERSION` when using Falcon. Thanks to [@97jaz](https://github.com/97jaz) and [@gsar](https://github.com/gsar) for bringing this to our attention. [PR#3483](https://github.com/newrelic/newrelic-ruby-agent/issues/3483)
+
 ## v10.2.0
 
 - **Feature: Introduce Hybrid Agent for OpenTelemetry Tracing Support**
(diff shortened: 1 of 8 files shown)


lib/new_relic/agent/transaction_time_aggregator.rb
```diff
@@ -96,7 +105,7 @@ def in_transaction?(thread_id = current_thread)
         end
 
         def current_thread
-          Thread.current.object_id
+          TransactionTimeAggregator.current_execution_context_id
```
tannalynn, line 113: We also keep the thread id on the span, in abstract segment class. Do you think we should change that one to use this new method?

hannahramadan: Yes I think this is a good change so that each Falcon request shows a its sown fiber id vs all the same thread id.

CHANGELOG.md
```diff
@@ -28,6 +28,10 @@
 
   Previously, the agent would cause a ~3 second shutdown delay when running the `rails test` command. The `Rails::Command::TestCommand` constant has been added to the default `autostart.denylisted_constants` list to prevent the agent from starting during Rails test runs. Thanks to [@varyform](https://github.com/varyform) for bringing this to our attention. [PR#3478](https://github.com/newrelic/newrelic-ruby-agent/issues/3478)
 
+- **Bugfix: Fix "Unable to calculate elapsed transaction time" warnings when using Falcon web server**
+
+  The agent now uses `Fiber.current.object_id` instead of `Thread.current.object_id` to track transaction state when running under Falcon, preventing collisions from concurrent requests sharing the same thread. Also fixes a NameError: uninitialized constant `Async::HTTP::VERSION` when using Falcon. Thanks to [@97jaz](https://github.com/97jaz) and [@gsar](https://github.com/gsar) for bringing this to our attention. [PR#3483](https://github.com/newrelic/newrelic-ruby-agent/issues/3483)
```
```suggestion
  The agent now uses `Fiber.current.object_id` instead of `Thread.current.object_id` to track transaction state when running under Falcon, preventing collisions from concurrent requests sharing the same thread. Also fixes a "NameError: uninitialized constant `Async::HTTP::VERSION`" when using Falcon. Thanks to [@97jaz](https://github.com/97jaz) and [@gsar](https://github.com/gsar) for bringing this to our attention. [PR#3483](https://github.com/newrelic/newrelic-ruby-agent/issues/3483)
```

tannalynn (dismissed):

kaylareopelle (dismissed):
nit on the changelog, but feel free to dismiss!

tannalynn (approved):

kaylareopelle (approved):

PR #3489, opened 2026-03-12
WIP: Sidekiq force new transaction

### LINKED ISSUE #3364: Prevent background job execution from inclusion in web transactions
We have reports of Sidekiq job execution getting included in Web Transaction total times, causing web transaction timing to be thrown off. 

That the Sidekiq job was kicked off by the Web Transaction should be recorded, but the execution of the job should live in a separate transaction and segments related to the job execution should not be included in the web transaction.

This would be **a breaking chang**e for our existing instrumentation and we would need to consider whether it's appropriate to extend to other background job libraries like Resque and DelayedJob.

To resolve this, one approach would be to:
* update sidekiq client instrumentation to make sure the job queuing is recorded with a method tracer
* update sidekiq server instrumentation to finish a current web transaction if present before starting a new sidekiq transaction

In addition, we need to make sure that method calls for a controller action that occur after a job is queued are still reported by the web transaction.

Related FR (Internal): https://new-relic.atlassian.net/browse/NR-497134

kaylareopelle:
This configuration option ends the current transaction if there is one, which forces a new transaction to start when
`perform_action_with_newrelic_trace` is called.

In addition, the `sidekiq.ignore_retry_errors` code has been refactored to toggle based on a key in the trace_options hash and removes the duplicated `perform_action_*` method from the Sidekiq server instrumentation.

Closes #3364

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 
 - **Feature: Add sidekiq.force_new_transaction configuration**
 
-  The `sidekiq.force_new_transaction` configuration option forces a new transaction to start when Sidekiq's server instrumentation is invoked to make sure every job starts its own transaction. This prevents Sidekiq jobs from being nested within web transactions that initiatied them. [PR#3489](https://github.com/newrelic/newrelic-ruby-agent/pull/3489)
+  The `sidekiq.force_new_transaction` configuration option forces a new transaction to start when Sidekiq's server instrumentation is invoked to make sure every job starts its own transaction. This prevents the duration of Sidekiq jobs from being included in the duration of the web transactions that started them. [PR#3489](https://github.com/newrelic/newrelic-ruby-agent/pull/3489)
 
 - **Bugfix: Provide config option to reduce cardinality of ActionCable broadcast metrics**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 
 - **Feature: Add sidekiq.force_new_transaction configuration**
 
-  The `sidekiq.force_new_transaction` configuration option forces a new transaction to start when Sidekiq's server instrumentation is invoked to make sure every job starts its own transaction. This prevents the duration of Sidekiq jobs from being included in the duration of the web transactions that started them. [PR#3489](https://github.com/newrelic/newrelic-ruby-agent/pull/3489)
+  The `sidekiq.force_new_transaction` configuration option forces a new transaction to start when Sidekiq's server instrumentation is invoked to make sure every job starts its own transaction. This prevents the duration of Sidekiq jobs from being included in the duration of the web transactions that started them. This is disabled by default. [PR#3489](https://github.com/newrelic/newrelic-ruby-agent/pull/3489)
 
 - **Bugfix: Provide config option to reduce cardinality of ActionCable broadcast metrics**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,13 +16,13 @@
 
   Thanks to [@jdelStrother](https://github.com/jdelStrother) for providing valuable feedback that helped shape this instrumentation.
 
-- **Feature: Add new 'ignored_middleware_classes' configuration**
-
-  A new configuration option, `ignored_middleware_classes`, allows users to exclude specific middlewares from instrumentation (ex. Rack::Cors). It defaults to an empty array. [Issue#1814](https://github.com/newrelic/newrelic-ruby-agent/issues/1814) [PR#3481](https://github.com/newrelic/newrelic-ruby-agent/pull/3481)
-
 - **Feature: Add new `NewRelic::Agent.add_transaction_log_attributes` API**
 
   A new API, `NewRelic::Agent.add_transaction_log_attributes`, allows users to add transaction-scoped custom attributes to log events for the current transaction. These attributes will only be applied to logs created within the scope of the current transaction. [PR#3472](https://github.com/newrelic/newrelic-ruby-agent/pull/3472)
+
+- **Feature: Add sidekiq.force_new_transaction configuration**
+
+  The `sidekiq.force_new_transaction` configuration option forces a new transaction to start when Sidekiq's server instrumentation is invoked to make sure every job starts its own transaction. This prevents Sidekiq jobs from being nested within web transactions that initiatied them. [PR#3489](https://github.com/newrelic/newrelic-ruby-agent/pull/3489)
 
 - **Bugfix: Provide config option to reduce cardinality of ActionCable broadcast metrics**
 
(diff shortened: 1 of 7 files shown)


test/multiverse/suites/sidekiq/sidekiq_force_new_transaction_test.rb
```diff
@@ -0,0 +1,168 @@
...
+        'Expected sidekiq.force_new_transaction to be true when configured'
+    end
+  end
+
+  def test_sidekiq_force_new_transaction_configuration_can_be_set_to_false
+    with_config(:'sidekiq.force_new_transaction' => false) do
+      refute NewRelic::Agent.config[:'sidekiq.force_new_transaction'],
+        'Expected sidekiq.force_new_transaction to be false when configured'
+    end
+  end
```
tannalynn, line 36: Is there a particular reason these couple of tests were added? Seems kind of like it's just testing to make sure our config works correctly in general, not really about how the config works for sidekiq?

test/multiverse/suites/sidekiq/sidekiq_force_new_transaction_test.rb
```diff
@@ -0,0 +1,168 @@
...
+
+  def test_separate_transactions_for_web_and_sidekiq_when_force_new_true
+    skip 'Test requires Sidekiq v7+' unless NewRelic::Helper.version_satisfied?(Sidekiq::VERSION, '>=', '7.0.0')
+
+    with_config(:'sidekiq.force_new_transaction' => true) do
+      run_web_transaction_with_inline_job
+
+      transactions = harvest_transaction_events!
+
+      # TODO: Skeptical -- is this the right call? Should we have three transactions?
```
tannalynn, line 116: What is the third txn? is one duplicated? i noticed you dedup txn names below

test/multiverse/suites/sidekiq/sidekiq_force_new_transaction_test.rb
```diff
@@ -0,0 +1,168 @@
...
+
+      sidekiq_span = spans.find { |s| s['name'].include?('SidekiqJob/NRDeadEndJob/perform') }
+
+      assert sidekiq_span, 'Expected to find sidekiq job span'
+      assert sidekiq_span['nr.entryPoint'],
+        'Expected sidekiq span to be marked as entry point (separate transaction)'
+
+      # TODO: Suspicious
+      # The web transaction may or may not create a span depending on sampling
+      # The key assertion is that the sidekiq job has its own span as a separate transaction
```
tannalynn, line 151: Im not sure what this means exactly? Shouldn't both txns be sampled in this test?

CHANGELOG.md
```diff
@@ -24,6 +24,10 @@
 
   A new API, `NewRelic::Agent.add_transaction_log_attributes`, allows users to add transaction-scoped custom attributes to log events for the current transaction. These attributes will only be applied to logs created within the scope of the current transaction. [PR#3472](https://github.com/newrelic/newrelic-ruby-agent/pull/3472)
 
+- **Feature: Add sidekiq.force_new_transaction configuration**
+
+  The `sidekiq.force_new_transaction` configuration option forces a new transaction to start when Sidekiq's server instrumentation is invoked to make sure every job starts its own transaction. This prevents the duration of Sidekiq jobs from being included in the duration of the web transactions that started them. [PR#3489](https://github.com/newrelic/newrelic-ruby-agent/pull/3489)
```
```suggestion
  The `sidekiq.force_new_transaction` configuration option forces a new transaction to start when Sidekiq's server instrumentation is invoked to make sure every job starts its own transaction. This prevents the duration of Sidekiq jobs from being included in the duration of the web transactions that started them. This is disabled by default. [PR#3489](https://github.com/newrelic/newrelic-ruby-agent/pull/3489)
```
or something like that, just to be clear it is not enabled by default?

kaylareopelle: Yes, great call!

test/multiverse/suites/sidekiq/sidekiq_force_new_transaction_test.rb
kaylareopelle: Oh, thank you! Glad my past self called this out. Meant to add a comment in the PR to start a discussion. 

So what I'm seeing now in the tests is that there's three transactions being created, but I think the first and third are the same transaction. 

Here's an example: 
```
{reservoir_size: 1200, events_seen: 3}
{"timestamp" => 1773953003.163938, "name" => "web", "duration" => 0.0001900196075439453, "type" => "Transaction", "error" => false, "priority" => 1.406131, "nr.apdexPerfZone" => "S", "parent.transportType" => "Unknown", "guid" => "58d55b8c2bb52b82", "traceId" => "b896a01373951493bec7b149e2768715", "sampled" => true}
{}
{"request.uri": "/"}
{"timestamp" => 1773953003.1642728, "name" => "OtherTransaction/SidekiqJob/NRDeadEndJob/perform", "duration" => 2.8371810913085938e-05, "type" => "Transaction", "error" => false, "priority" => 1.405819, "parent.transportType" => "Unknown", "guid" => "944701dfa66d2196", "traceId" => "30a7738cc90e688ad1fd594a5e23248e", "sampled" => true}
{}
{}
{"timestamp" => 1773953003.163938, "name" => "web", "duration" => 0.0004417896270751953, "type" => "Transaction", "error" => false, "priority" => 1.406131, "nr.apdexPerfZone" => "S", "parent.transportType" => "Unknown", "guid" => "58d55b8c2bb52b82", "traceId" => "b896a01373951493bec7b149e2768715", "sampled" => true}
{}
{"request.uri": "/"}
```

So here, the first and third transactions are named "web". Their durations are different, but their guids are both 
`58d55b8c2bb52b82`. Trace IDs and priorities match too. This makes me think it's just really one transaction, but I don't know for sure.

I didn't see anything strange in the UI when I ran this code in an example, but I also didn't specifically look at the data exported for transactions in this scenario to see how a transaction would be handled in this case. I'll take a look at that and report back here.

tannalynn: Hmm, yeah  that interesting. I'm curious why that would be happening, I'm not sure if it's actually a problem, or just a side effect of the way the tests are set up, but I think it would be good for us to figure out exactly why it's duplicated. Do you know if we see similar duplication when we aren't forcing new txns?

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2138,6 +2138,13 @@ def self.convert_to_constant_list(string_array)
           :allowed_from_server => false,
           :description => %Q(If `true`, the agent will ignore exceptions raised during Sidekiq's retry attempts and will only report the error if the job permanently fails.)
         },
+        :'sidekiq.force_new_transaction' => {
```
hannahramadan, line 2141: The name `force_new_transaction` sounds like something we wouldn't recommend. Maybe its the word "force"? What about `sidekiq.separate_transactions` or `sidekiq.start_new_transaction`

kaylareopelle:
Closing in favor of #3514

PR #3492, opened 2026-03-17
Fix httprb test compatibility with HTTP 6.0.1

hannahramadan:
In our tests, we create HTTP `Request.new` and `Response.new`, which were updated in 6.0.1 to require keyword arguments. Our test helpers needed updating to accommodate this change, which is backwards compatible for older versions. Actual instrumentation is unaffected.


kaylareopelle (approved):
🚀

PR #3500, opened 2026-03-30
Rubocop updates

### LINKED ISSUE #3389: Evaluate rubocop todos
Our rubocop version and `.rubocop_todo.yml` was updated with #3385, so we have some new opportunities to improve our code. Let's evaluate our code based on the newly generated todo file. Once issues are addressed, regenerate the todo file with `bundle exec rubocop --regenerate-todo`

hannahramadan:
Takes care of all the safe auto-correctable Rubocops in our TODO minus a few that either won't work for us or I'm not stoked on the style of, but open to adding if others would like.

full ci: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/23862993474

Cops addressed:
- Layout/EmptyLineAfterMagicComment
- Layout/EmptyLinesAfterModuleInclusion
- Lint/RedundantCopDisableDirective
- Lint/RedundantTypeConversion
- Minitest/AssertInstanceOf
- Minitest/AssertKindOf
- Minitest/AssertSame
- Minitest/LifecycleHooksOrder
- Minitest/RefuteEqual
- Minitest/RefuteOperator
- Minitest/ReturnInTestMethod
- Performance/CompareWithBlock
- Style/EmptyStringInsideInterpolation
- Style/RedundantLineContinuation
- Style/SuperArguments
- Style/SuperWithArgsParentheses

Cops now disabled via rubocop.yml:
- Style/CombinableDefined - Disabling for we need the specificity in instrumentations ie depends_on { defined?(Rake) && defined?(Rake::VERSION) }
- Style/RedundantLineContinuation - will enable when support for 2.6 is dropped
- Minitest/TestMethodName - will incorrectly raise on test helper methods
- Minitest/AssertOperator - prefer style of this cop disabled
- Layout/EmptyLinesAroundMethodBody - team prefers the look of empty lines around method body

closes #3389


.rubocop_todo.yml
```diff
@@ -81,40 +42,11 @@ Lint/UselessConstantScoping:
...
-    - 'test/multiverse/suites/sequel/sequel_safety_test.rb'
-    - 'test/multiverse/suites/sidekiq/sidekiq_instrumentation_test.rb'
-    - 'test/new_relic/agent/attribute_pre_filtering_test.rb'
-    - 'test/new_relic/agent/sampler_test.rb'
-    - 'test/new_relic/agent/span_event_primitive_test.rb'
-    - 'test/new_relic/agent/transaction_event_primitive_test.rb'
-    - 'test/new_relic/http_client_test_cases.rb'
-
 # Offense count: 23
 # This cop supports safe autocorrection (--autocorrect).
```
hannahramadan: I think this cop reads weird. No other reason for not fixing this lol. I'm down to if others want. The benefit is better debugging because you see actual values in the error.

Before:
`assert gc_time > 0.0, "GC Time wasn't recorded!"`

After:
`assert_operator gc_time, :>, 0.0, "GC Time wasn't recorded!"`

.rubocop_todo.yml
```diff
@@ -141,58 +73,21 @@ Minitest/AssertRaisesCompoundBody:
...
-    - 'test/new_relic/language_support_test.rb'
-
 # Offense count: 19
 Minitest/TestFileName:
   Enabled: false
 
-# Offense count: 19
+# Offense count: 18
 # This cop supports safe autocorrection (--autocorrect).
 Minitest/TestMethodName:
```
hannahramadan: These are false positives, they are counting helper methods

.rubocop_todo.yml
```diff
@@ -263,7 +151,7 @@ Style/BitwisePredicate:
   Exclude:
     - 'test/new_relic/agent/attribute_filter_test.rb'
 
-# Offense count: 30
+# Offense count: 31
 # This cop supports safe autocorrection (--autocorrect).
 Style/CombinableDefined:
```
hannahramadan: Wasn't stoked to disable this since it was so close to instrumentations. 

Before
`depends_on { defined?(Rake) && defined?(Rake::VERSION) }`

After
`depends_on { defined?(Rake::VERSION) }`

.rubocop_todo.yml
hannahramadan: This needs to stick. Fails Ruby 2.6 when removed

.rubocop_todo.yml
```diff
@@ -14,19 +14,6 @@ Gemspec/DevelopmentDependencies:
...
-    - 'test/performance/rails_app/db/seeds.rb'
-
-# Offense count: 50
-# This cop supports safe autocorrection (--autocorrect).
-Layout/EmptyLinesAfterModuleInclusion:
-  Enabled: false
-
 # Offense count: 25
 # This cop supports safe autocorrection (--autocorrect).
 Layout/EmptyLinesAroundMethodBody:
```
hannahramadan: I also left this one, just because I like the look of the before better. I find it easier to read.

Before
```
def in_transaction(name: nil,
  partial_name: nil,
  category:,
  options: {})

  finishable = start_transaction_or_segment(
    name: name,
    partial_name: partial_name,
    category: category,
    options: options
  )
```

After
```
def in_transaction(name: nil,
  partial_name: nil,
  category:,
  options: {})
  finishable = start_transaction_or_segment(
    name: name,
    partial_name: partial_name,
    category: category,
    options: options
  )
```

tannalynn: I also prefer the readability in the before, personally.

lib/new_relic/agent/transaction/transaction_sample_buffer.rb
```diff
@@ -82,7 +82,7 @@ def max_capacity
         # This doesn't use the more convenient #last and #sort_by to avoid
         # additional array allocations (and abundant alliteration)
         def truncate_samples
-          @samples.sort! { |a, b| a.duration <=> b.duration }
+          @samples.sort_by!(&:duration)
```
tannalynn: We have a comment about why we aren't using sort_by here, for performance reasons, so it seems to be an intentional choice. Not sure how old this code is, so maybe we should run some micro benchmarking about it to see if this is still something we want to keep doing the way we were previously. (it's possible it mattered more on older ruby versions)

lib/new_relic/agent/error_trace_aggregator.rb
```diff
@@ -94,7 +94,7 @@ def notice_agent_error(exception)
       def register_config_callbacks
         Agent.config.register_callback(:'error_collector.enabled') do |enabled|
           reset! if enabled == false
-          ::NewRelic::Agent.logger.debug("Error traces will #{enabled ? '' : 'not '}be sent to the New Relic service.")
+          ::NewRelic::Agent.logger.debug("Error traces will #{'not ' unless enabled}be sent to the New Relic service.")
```
kaylareopelle, line 97: looking ahead to the code in the next file, does it make any sense to standardize on `enabled` vs. `@enabled`?

lib/new_relic/agent/log_event_aggregator.rb
```diff
@@ -472,7 +472,7 @@ def format_log_level_constant(log_level)
       def severity_too_low?(severity)
         severity_constant = format_log_level_constant(severity)
         # always record custom log levels
-        return false unless Logger::Severity.constants.include?(severity_constant)
+        return false unless Logger::Severity.const_defined?(severity_constant)
```
kaylareopelle: This looks fine here, but got curious about the cop's details and found an issue that notes it as potentially unsafe: https://github.com/rubocop/rubocop/issues/15040

.rubocop_todo.yml
kaylareopelle: Same! Would it make sense to take this out of the TODO and specify we want the cop off in .rubocop.yml?

.rubocop_todo.yml
kaylareopelle: What do you think about moving this one out of the TODO file and into `.rubocop.yml`?

.rubocop_todo.yml
tannalynn: yeah thats a good idea, i didn't think about that

lib/new_relic/agent/log_event_aggregator.rb
tannalynn: Hmm, that does seem concerning in general imo. Maybe we should just disable this cop as well? 
thanks for looking into that!

lib/new_relic/agent/transaction/transaction_sample_buffer.rb
hannahramadan: Seems like a fine change- allocations are constant and about 2x faster. What do you think?

```
  ALLOCATIONS (single sort operation)                                             
  ----------------------------------------------------------------------          
  Size       | sort! { a <=> b }         | sort_by!(&:duration)      | Diff       
  ----------------------------------------------------------------------          
  100        | 5                         | 10                        | +5         
  1,000      | 2                         | 7                         | +5         
  10,000     | 2                         | 7                         | +5         
                                                                           
```

```
  ----------------------------------------------------------------------          
  Size       | sort!           | sort_by!        | Speedup                        
  ----------------------------------------------------------------------          
  100        | 0.277           | 0.231           | 1.20x faster                   
  1,000      | 3.528           | 1.716           | 2.06x faster                   
  10,000     | 47.808          | 22.579          | 2.12x faster  
```

lib/new_relic/agent/error_trace_aggregator.rb
hannahramadan: We could! Although I think there are a few other places where we aren't caching @enabled and instead doing a method call/config look up. Would you like me to make that edit as a part of this PR, or I can also make a new ticket. Maybe a good first issue.

lib/new_relic/agent/log_event_aggregator.rb
hannahramadan: I reverted this! Nice catch

lib/new_relic/agent/error_trace_aggregator.rb
kaylareopelle: Whoops, totally thought I responded to this. I think you can make a new ticket. Love the idea of tagging it as a good first issue.

kaylareopelle (approved):

PR #3508, opened 2026-04-01
Slack Notification tests

### LINKED ISSUE #3269: Fix slack notification tests
Test files under the `slack_notification_tests` directory were incorrectly named the plural `_tests`, instead of `_test`, which prevented them from regularly running. Many of them are now failing- we should rename these files and investigate the failing tests. 

Also [noted](https://github.com/newrelic/newrelic-ruby-agent/pull/3267#discussion_r2380183838) - we'll have to add HTTParty as a dependency to every Rails env if we want to run the Slack Notifier tests with the main unit_tests CI. We could eventually hit a problem with feedjira too since it has a similar condition and is also used for Gabby. Is there a way to exclude a test from the Rails environments without moving it to multiverse?

hannahramadan:
We would have to add HTTParty as a dependency to every Rails env in order to run these with the main unit_tests CI. To avoid that, there is now separate CI job to run Slack notification tests in the scheduled cron workflow. There is some conditional loading so that tests are skipped without error during regular test runs.

closes https://github.com/newrelic/newrelic-ruby-agent/issues/3269

Slack notification tests from scheduled CI run: https://github.com/newrelic/newrelic-ruby-agent/actions/runs/24096981943/job/70297995468


.github/workflows/ci_cron.yml
```diff
@@ -307,7 +323,16 @@ jobs:
 
   notify_slack_fail:
     name: Notify slack fail
-    needs: [unit_tests, multiverse, infinite_tracing, multiverse_services_1, multiverse_services_2, multiverse_elasticsearch, multiverse_services_kafka, multiverse_services_mysql_pg]
+    needs:
```
kaylareopelle, line 326: Nice!

.github/workflows/ci_cron.yml
```diff
@@ -285,6 +285,22 @@ jobs:
...
+        run: 'git config --global init.defaultBranch main'
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # tag v6.0.2
+      - uses: ruby/setup-ruby@09a7688d3b55cf0e976497ff046b70949eeaccfd # tag v1.288.0
+        with:
+          ruby-version: 4.0.0
+      - run: gem install httparty minitest minitest-mock
+      - name: Run Slack Notifications Tests
+        run: |
+          ruby test/new_relic/slack_notifications_test/slack_notifier_test.rb
+          ruby test/new_relic/slack_notifications_test/gem_notifier_test.rb
```
kaylareopelle, line 303: If you wanted to make this a one-liner, you could try: 
`TEST='test/new_relic/slack_notifications_test/*' bundle exec rake test`
(but you might have to do a full bundle step)

kaylareopelle (approved):

.github/workflows/ci_cron.yml
hannahramadan: I tried throwing in bundle at one point, but that brought in the dev dependencies from gemspec, which excludes httparty in the CI and it didn't like it :( This step just does the minimal gem installs on lineo 299

kaylareopelle: Ah got it! Nice job finding a solution!

PR #3512, opened 2026-04-06
Refactor sidekiq.ignore_retry_errors

kaylareopelle:
Now, we use trace_args to pass a flag when errors should not be noticed. This could also be applied to other frameworks, but it currently is only assigned in Sidekiq server instrumentation.


hannahramadan (approved):
Clean refactor 🧼

tannalynn (approved):

PR #3513, opened 2026-04-06
Make Transaction#finish idempotent

kaylareopelle:
Previously, if `Transaction#finish` was called multiple times on the same transaction, multiple transactions could be created for one operation. Now, the `finish` method is protected by a mutex and an instance variable to make sure transactions are only finished once.

### DIFF
--- a/test/new_relic/agent/transaction_test.rb
+++ b/test/new_relic/agent/transaction_test.rb
@@ -1900,5 +1900,80 @@ module NewRelic::Agent
       assert_nil txn.initial_segment
       refute_predicate txn, :finished?
     end
+
+    def test_finish_is_idempotent
+      txn = Tracer.start_transaction(name: 'test', category: :controller)
+      state = Tracer.state
+
+      # First finish should work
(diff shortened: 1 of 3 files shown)


test/new_relic/agent/transaction_test.rb
```diff
@@ -1900,5 +1900,89 @@ def test_transaction_finished_works_when_inital_segment_nil
       assert_nil txn.initial_segment
       refute_predicate txn, :finished?
     end
+
+    def test_finish_is_idempotent
```
kaylareopelle, line 1904: These tests don't include harvesting tests because they were too flaky. #3514 has tests related to harvesting for multiple transactions that originally revealed this bug. They are stable with this change. 

I think we can rely on those tests covering the multiple transaction behavior, but I can also try to figure out how to get the harvesting tests to work here if you prefer. 

As a note -- none of the other tests in this file check to see if transactions are harvested.

tannalynn (approved):

PR #3347, opened 2025-11-21
Remove unused Datastores API args

kaylareopelle:
The arguments in these methods have not been in use for 9 years.

PR #3517, opened 2026-04-08
Remove deprecation warning for Datastores.wrap

### LINKED ISSUE #3516: Change deprecation logging from the Datastores.wrap method to only log once
in `lib/new_relic/agent/datastores.rb:108`, in the wrap method, we are logging a warning that the method will be changed in the next major version every time the method is called. This was included in 9.24 as preparation for our 10.0 release, but was not updated accordingly in 10.0. 
We should update the logging call to use our `log_once` method we have on our logger class, to ensure that this log message is not disruptive to customers who are using this method frequently.
We should also update this method to make the appropriate change in the next major release (v 11).

kaylareopelle:
This was missed in the cleanup for 10.0. The arguments were removed in https://github.com/newrelic/newrelic-ruby-agent/pull/3347.

Closes #3516 

(If I am wrong, there's a log_once fix in the `log_once_datastores` branch)

### DIFF
--- a/lib/new_relic/agent/datastores.rb
+++ b/lib/new_relic/agent/datastores.rb
@@ -106,11 +106,6 @@ module NewRelic
       # @api public
       #
       def self.wrap(product, operation, collection = nil, callback = nil)
-        NewRelic::Agent.logger.warn('The NewRelic::Agent::Datastores.wrap method is changing. ' \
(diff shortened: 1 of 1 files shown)


kaylareopelle:
The warning is still needed. Closing in favor of #3519

PR #3519, opened 2026-04-08
Log Datastores.wrap deprecation warning once

### LINKED ISSUE #3516: Change deprecation logging from the Datastores.wrap method to only log once

kaylareopelle:
Previously, the deprecation warning was being logged every time wrap was called. Also, the major version release was incomplete. It updated the documentation without changing the args passed to the callback method. The documentation matches the current method and we now have a TODO: MAJOR VERSION flag to catch it next time.

Closes: #3516


hannahramadan (dismissed):
🚀

tannalynn (approved):

PR #3526, opened 2026-04-10
Rails81 structured events as log events

tannalynn:
Add instrumentation to automatically instrument all rails events as log events. Also includes a config option for users to specify if they only want certain events captured by the agent.


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -1784,6 +1784,27 @@ def self.convert_to_constant_list(string_array)
           :allowed_from_server => false,
           :description => 'Controls auto-instrumentation of the LogStasher library at start-up. May be one of: `auto`, `prepend`, `chain`, `disabled`.'
         },
+        :'instrumentation.rails_event_logger' => {
+          :default => value_of(:'application_logging.enabled'),
+          :documentation_default => 'true',
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => 'Controls instrumentation of Rails.event as structured logs (Rails 8.1+)'
```
kaylareopelle: Since we removed 8.1+ from the continuations PR, should we do that in this one too?

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 
 ## dev
 
+- **Feature: Add Rails.event instrumentation for structured logging (Rails 8.1+)**
+
+  The agent now supports Rails.event (introduced in Rails 8.1) as structured log events. When enabled, events published via `Rails.event.notify` are captured and forwarded to New Relic as log events. Event payloads, tags, context, timestamps, and source locations are automatically captured as log attributes.
```
kaylareopelle: Same question about the direct 8.1 reference here too.

Also, do you think it's worth calling out that by debug events are captured by default? And sharing the setting if they want to change it?

tannalynn: Ah right, debug events. I double checked and it looks like there is no way for us to know if something is a debug event or not, it will look the same as other events (unless someone adds info to it on their own, but rails doesn't add anything that makes it clear), it just only shows up if they turn on debug mode. The only way they could change that via the config would be by setting the filter on the names of events they want to capture.

hannahramadan (approved):

CHANGELOG.md
kaylareopelle: Ah okay. I thought the debug status would be captured and those events could be included/excluded based on the log level config. Good to know! Thanks for clarifying!

kaylareopelle (approved):

PR #3539, opened 2026-04-20
Refactor hybrid agent attribute translation

### LINKED ISSUE #3422: Hybrid Agent: Refactor attribute handling
Refactor the way attributes are applied from OTel APIs to avoid duplication. 

Some of the provided attributes should be set as agent attributes or intrinsic attributes. Attributes that aren't set in these other categories should be preserved as custom attributes.

My current idea is to: 
- Temporarily store attributes on the NR/OTel Span object when the OTel APIs are called
- Assign relevant attributes using our NR-specific methods (ex. start_external_request_segment)
- Remove them from the attrs hash after they're assigned to avoid double sending data
- When the span finishes, do any NR-specific final assignment (ex. http status code), and attach whatever we haven't assigned as custom attributes

kaylareopelle:
Now, the hybrid agent attribute translation is performed using a series of mapping constants and translators for each category of semantic conventions to translate into New Relic attributes.

This refactors the existing translation for HTTP Client, HTTP Server, and Datastore translation. It also adds stable
attributes to the Datastore translation options.

Attributes are initially translated when a span starts. At that point a Translator is assigned to the span. During the span's lifecycle, if attributes are added using APIs, they will be translated using the translator to correctly dispatch the attributes to the New Relic fields.

Attributes can be translated for segment APIs, instance variable assignment, and also direct application as intrinsic and agent attributes. If there is an attribute without a New Relic translation, then the attributes will be sent to custom
attributes.

See the [attribute_translation.md](https://github.com/newrelic/newrelic-ruby-agent/pull/3539/changes#diff-eaa87444fe269e0fc80bb863f933722dd37010aa4f6d86a0f76fb6f0a6c5c368R1) file for details.

Closes #3422


tannalynn (dismissed):

hannahramadan (approved):
🚀

PR #3540, opened 2026-04-21
Add config `active_record_use_table_name`

### LINKED ISSUE #3465: Create option for alternative Active Record metric naming approach
We received a report that customers who use single-table inheritance may encounter problems with high cardinality for their Active Record metric names. The reporter provided a solution that used table names instead of class names to solve the problem.

As a response, we're creating a configuration option that will allow users to choose whether they want to use table names or class names within their Active Record metrics.

**Update**: this can also be seen in other AR-related values when STIs are heavily used. Check more than just the metric name while working on this issue.

hannahramadan:
Introduce new config `active_record_use_table_name` to use the Active Record model's table name instead of the class name when naming Active Record metrics, which also impacts span and segment names.

closes https://github.com/newrelic/newrelic-ruby-agent/issues/3465

### DIFF
--- a/lib/new_relic/agent/instrumentation/active_record_helper.rb
+++ b/lib/new_relic/agent/instrumentation/active_record_helper.rb
@@ -20,7 +20,7 @@ module NewRelic
         def instrument_save_methods
           ::ActiveRecord::Base.class_eval do
             def newrelic_model_collection_name
-              NewRelic::Agent::Instrumentation::ActiveRecordHelper.use_table_name? ? self.class.table_name : self.class.name
+              ::NewRelic::Agent.config[:active_record_use_table_name] ? self.class.table_name : self.class.name
             end
             private(:newrelic_model_collection_name)
 
@@ -45,7 +45,7 @@ module NewRelic
         def instrument_relation_methods
           ::ActiveRecord::Relation.class_eval do
             def newrelic_relation_collection_name
-              NewRelic::Agent::Instrumentation::ActiveRecordHelper.use_table_name? ? self.klass.table_name : self.name
+              ::NewRelic::Agent.config[:active_record_use_table_name] ? self.klass.table_name : self.name
             end
             private(:newrelic_relation_collection_name)
 
@@ -108,12 +108,6 @@ module NewRelic
         TABLE_NAME_CACHE = {}
         @table_name_cache_lock = Mutex.new
 
-        def use_table_name?
-          return @use_table_name if defined?(@use_table_name)
-
-          @use_table_name = NewRelic::Agent.config[:active_record_use_table_name]
-        end
-
         # convert vendor (makara, etc.) wrapper names to their bare names
         # ex: postgresql_makara -> postgresql
         def bare_adapter_name(adapter_name)
@@ -144,7 +138,7 @@ module NewRelic
           return unless splits.length == 2
 
           model_name = splits.first
-          return model_name unless use_table_name?
+          return model_name unless NewRelic::Agent.config[:active_record_use_table_name]
 
           table_name_for(model_name) || model_name
         end
--- a/lib/new_relic/agent/instrumentation/active_record_helper.rb
+++ b/lib/new_relic/agent/instrumentation/active_record_helper.rb
@@ -19,10 +19,15 @@ module NewRelic
 
         def instrument_save_methods
           ::ActiveRecord::Base.class_eval do
+            def newrelic_collection_name
+              ::NewRelic::Agent.config[:active_record_metrics_use_table_name] ? self.class.table_name : self.class.name
+            end
+            private :newrelic_collection_name
+
             alias_method(:save_without_newrelic, :save)
 
             def save(*args, &blk)
-              ::NewRelic::Agent.with_database_metric_name(self.class.name, nil, ACTIVE_RECORD) do
+              ::NewRelic::Agent.with_database_metric_name(newrelic_collection_name, nil, ACTIVE_RECORD) do
                 save_without_newrelic(*args, &blk)
               end
             end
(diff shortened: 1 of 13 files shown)


tannalynn (dismissed):

lib/new_relic/agent/instrumentation/active_record_helper.rb
```diff
@@ -19,18 +19,23 @@ def instrument_additional_methods
 
         def instrument_save_methods
           ::ActiveRecord::Base.class_eval do
+            def newrelic_model_collection_name
+              ::NewRelic::Agent.config[:active_record_use_table_name] ? self.class.table_name : self.class.name
```
kaylareopelle, line 23: I think these conditions will run every time we make a database metric. They should be quick, but since the `active_record_use_table_name` config shouldn't change, what do you think about trying to find a way to set this once?

test/multiverse/suites/active_record/active_record_test.rb
```diff
@@ -505,6 +505,74 @@ def test_with_database_metric_name
     )
   end
 
+  def test_active_record_use_table_name_save_uses_table_name
+    with_config(active_record_use_table_name: true) do
+      in_web_transaction do
+        Animals::Dog.create!(name: 'Oliver')
```
kaylareopelle: ❤️

lib/new_relic/agent/instrumentation/active_record_helper.rb
```diff
@@ -123,11 +135,33 @@ def split_name(name)
         end
 
         def model_from_splits(splits)
-          if splits.length == 2
-            splits.first
+          return unless splits.length == 2
```
kaylareopelle, line 138: What do you think about adding specific tests for some of the new logic between this line and line 163?

lib/new_relic/agent/instrumentation/active_record_helper.rb
```diff
@@ -123,11 +135,33 @@ def split_name(name)
...
+          return model_name unless NewRelic::Agent.config[:active_record_use_table_name]
+
+          table_name_for(model_name) || model_name
+        end
+
+        def table_name_for(model_name)
+          return TABLE_NAME_CACHE[model_name] if TABLE_NAME_CACHE.key?(model_name)
+
+          @table_name_cache_lock.synchronize do
+            TABLE_NAME_CACHE[model_name] = resolve_table_name(model_name) unless TABLE_NAME_CACHE.key?(model_name)
```
kaylareopelle, line 150: Why is the `unless` check here? It seems like this would already be known by line 147.

test/multiverse/suites/active_record/active_record_test.rb
```diff
@@ -505,6 +505,74 @@ def test_with_database_metric_name
     )
   end
 
+  def test_active_record_use_table_name_save_uses_table_name
+    with_config(active_record_use_table_name: true) do
```
kaylareopelle: The title here is `save` but the operation is `create!`. I don't think `create!` actually calls `save`. It instead calls `save!`, which also sadly doesn't call `save`. 

Please check me on this though. [Here's the source for `create!`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/persistence.rb#L50). The other methods are also in that file.

Since we don't directly instrument `create!`, what do you think about changing the title to `save!` and adding a note?

lib/new_relic/agent/instrumentation/active_record_helper.rb
```diff
@@ -19,18 +19,23 @@ def instrument_additional_methods
...
+            def newrelic_model_collection_name
+              ::NewRelic::Agent.config[:active_record_use_table_name] ? self.class.table_name : self.class.name
+            end
+            private(:newrelic_model_collection_name)
+
             alias_method(:save_without_newrelic, :save)
 
             def save(*args, &blk)
-              ::NewRelic::Agent.with_database_metric_name(self.class.name, nil, ACTIVE_RECORD) do
+              ::NewRelic::Agent.with_database_metric_name(newrelic_model_collection_name, nil, ACTIVE_RECORD) do
```
kaylareopelle, line 30: I noticed some of these updated methods don't test the config functionality: 
* save (w/o bang) (see note in the test for create!)
* touch
* delete_all
* destroy_all
* calculate
* pluck

I'm a little worried that if we don't test them all explicitly and something needs to change, we'll forget to update one of those methods. 

Would you be up for adding tests for them?

kaylareopelle (changes requested):
Nice work on this! Requesting changes since there's one approval and I want to chat about a few things before merge.

lib/new_relic/agent/instrumentation/active_record_helper.rb
hannahramadan: What do you think of this change https://github.com/newrelic/newrelic-ruby-agent/pull/3540/changes/db7583f0c60ca1cf7217e39fed8e94e1a2a28200

lib/new_relic/agent/instrumentation/active_record_helper.rb
hannahramadan: This was an addition Mr. Claude recommended, suggesting that in the time between the first `if TABLE_NAME_CACHE.key?(model_name)` check and entering the lock, a thread could slip in and populate the cache. Apparently this is known as a "double-checked locking pattern"- learned something new! It seems safer so I accepted the suggestion, but if we feel this is overkill, I am happy to remove.

lib/new_relic/agent/instrumentation/active_record_helper.rb
hannahramadan: Reverted this after discussing live

lib/new_relic/agent/instrumentation/active_record_helper.rb
hannahramadan: tests added! https://github.com/newrelic/newrelic-ruby-agent/pull/3540/changes/b4bfe49a789eff771cc096aae9524baf69adaad4

test/multiverse/suites/active_record/active_record_test.rb
hannahramadan: Done ◡̈

lib/new_relic/agent/instrumentation/active_record_helper.rb
hannahramadan: New tests here: https://github.com/newrelic/newrelic-ruby-agent/pull/3540/changes/bc7d1173dec32837f97b178011a35e3b45f287ee

kaylareopelle (approved):

PR #3548, opened 2026-05-01
Bugfix for semantic logger rails_semantic_logger

### LINKED ISSUE #3546: SemanticLogger instrumentation incompatible with `log(level, msg)`
Thanks for the SemanticLogger integration! I did run into one bug though -

## Description

If an exception reaches ActionDispatch::DebugExceptions, it throws 

```
wrong number of arguments (given 2, expected 1) (ArgumentError)

    caller: /usr/local/bundle/ruby/4.0.0/gems/rails_semantic_logger-4.20.0/lib/rails_semantic_logger/extensions/action_dispatch/debug_exceptions.rb:15
    |           ActionController::Base.logger.log(level, wrapper.exception)
                                             ^^^^
    callee: /usr/local/bundle/ruby/4.0.0/gems/newrelic_rpm-10.4.0/lib/new_relic/agent/instrumentation/semantic_logger/prepend.rb:11
    |     def log(log)
              ^^^
```

## Expected Behavior

It logs the exception

## Steps to Reproduce
Install rails_semantic_logger, throw an exception during a request that reaches ActionDispatch::DebugExceptions

## Your Environment
ruby 4.0.3
rails_semantic_logger 4.20.0
semantic_logger 4.18.0
newrelic_rpm 10.4.0
rails 8.1.3

## Additional context

rails_semantic_logger modifies ActionDispatch::DebugExceptions#log_error, which ends up calling `ActionController::Base.logger.log(level, exception)` here - https://github.com/reidmorrison/rails_semantic_logger/blob/7b05bf1b3460c04308155b3873b11d75078b9515/lib/rails_semantic_logger/extensions/action_dispatch/debug_exceptions.rb#L15

This ought to then call `add` here - https://github.com/reidmorrison/semantic_logger/blob/777caacebf1e81d2aeabdebed1b0351c1e3816d9/lib/semantic_logger/logger.rb#L70

but the NewRelic instrumentation only accepts a single argument to `log` - https://github.com/newrelic/newrelic-ruby-agent/blob/5490d8371666183693604ee56f4725a295b47287/lib/new_relic/agent/instrumentation/semantic_logger/prepend.rb#L11

---

It is the only use of `log(level, msg, ...)` that I can find in SemanticLogger - everywhere else uses `log(log)`. There's possibly an argument to be made that rails_semantic_logger ought to be calling `add(level, exception)` instead of `log(level, exception)`, but given any application code might also be using the latter form, I guess it ought to be fixed in NewRelic's instrumentation.

tannalynn:
Now we check the logic the same way semantic logger does https://github.com/reidmorrison/semantic_logger/blob/master/lib/semantic_logger/logger.rb#L70

And we just skip instrumentation if it's not the right object, because semantic logger will turn it into the object and run it back through here later.

resolves https://github.com/newrelic/newrelic-ruby-agent/issues/3546


lib/new_relic/agent/instrumentation/semantic_logger/chain.rb
```diff
@@ -12,7 +12,9 @@ def self.instrument!
 
         alias_method(:log_without_new_relic, :log)
 
-        def log(log)
+        def log(log, *args)
+          return log_without_new_relic(log, *args) unless log.is_a?(::SemanticLogger::Log)
+
```
tannalynn, line 17: Now we check the logic the same way semantic logger does https://github.com/reidmorrison/semantic_logger/blob/master/lib/semantic_logger/logger.rb#L70

And we just skip instrumentation if it's not the right object, because semantic logger will turn it into the object and run it back through here later.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
   The agent now supports Dalli 5.0+, which removed `Dalli::Protocol::Binary` in favor of the meta protocol exclusively. For Dalli 3.2.0+, `pipelined_get` instrumentation now correctly targets `Dalli::Protocol::Base` (where the method is defined) rather than `Dalli::Protocol::Binary`, fixing a gap where `get_multi` calls went uninstrumented when using the meta protocol. For Dalli 5.0+, the agent additionally instruments `Dalli::Protocol::Meta#read_multi_req`, which is invoked by Dalli's single-server `get_multi` optimization. [PR#3541](https://github.com/newrelic/newrelic-ruby-agent/pull/3541)
 
+- **Bugfix: Fix SemanticLogger instrumentation incompatibility with `rails_semantic_logger`**
```
```suggestion
- **Bugfix: Fix Semantic Logger instrumentation incompatibility with `rails_semantic_logger`**
```

hannahramadan (dismissed):
🚀

kaylareopelle (approved):

hannahramadan (approved):

Issue #3437, opened 2026-02-05
Hybrid Agent: Translate attributes for specialty database conventions


kaylareopelle:
When an OpenTelemetry Span is created with a :client kind, and has a `db.system` or `db.system.name` attribute for Redis, Mongo, or DynamoDB libraries, there are certain attributes we want to apply to the agent attributes within the New Relic-created telemetry. Translate these attributes for datastore segments. See the agent spec for which attributes should be translated.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-520995

kaylareopelle:
@hannahramadan - This could be a good one for you to pick up!

hannahramadan:
Redis and Mongo complete. The DynamoDB attributes are either covered by the default translator or don't emit the attributes we'd want to translate or build. It is _possible_ to build the `host`, but could be unreliable. Datastore segments can be started without the host, so we will defer building `host` until the OTel instrumentation evolves or we see customer demand.

PR #3562, opened 2026-05-13
Add Redis Datastore Translator

hannahramadan:
Add Redis Datastore Translator. Redis closely mirrors the default datastore translator with the exception of it's mappings creating a nosql_statement vs a sql. Because of that, tracer code was also edit to allow for that difference. 

Part of https://github.com/newrelic/newrelic-ruby-agent/issues/3437

### DIFF
--- a/lib/new_relic/agent/opentelemetry/attribute_translator.rb
+++ b/lib/new_relic/agent/opentelemetry/attribute_translator.rb
@@ -24,15 +24,9 @@ module NewRelic
             'opentelemetry-instrumentation-redis' => RedisDatastoreTranslator
           },
           discriminating_attribute: {
-            'db.system' => {
-              'redis' => RedisDatastoreTranslator,
-              default: DatastoreTranslator
-            },
-            'db.system.name' => {
-              'redis' => RedisDatastoreTranslator,
-              default: DatastoreTranslator
-            }
-            # 'rpc.system' => {default: RpcTranslator},
+            'db.system' => DatastoreTranslator,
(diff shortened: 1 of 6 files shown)


lib/new_relic/agent/opentelemetry/attribute_translator.rb
```diff
@@ -19,13 +20,19 @@ class AttributeTranslator
...
+            'opentelemetry-instrumentation-pg' => DatastoreTranslator,
+            'opentelemetry-instrumentation-redis' => RedisDatastoreTranslator
           },
           discriminating_attribute: {
-            'db.system' => DatastoreTranslator,
-            'db.system.name' => DatastoreTranslator
-            # 'rpc.system' => RpcTranslator,
+            'db.system' => {
+              'redis' => RedisDatastoreTranslator,
+              default: DatastoreTranslator
```
kaylareopelle: Since instrumentation scope is evaluated first, I don't think we need unique analysis for discriminating attributes. If the translator is working, it should always exit after the first scope check for Redis.

Am I missing something though? Did you discover something in your testing that means we need the extra attribute check?

hannahramadan: Good call out! I think I was just looking for places that we might want to differentiate between the redis and others, but you're right that instrumentation scope would pick it up first and the situation where we'd need this extra bit is probably never going to be an issue. I'll revert this (yay- it looks cleaner without this change too!)

kaylareopelle (approved):
This is great! Thank you!

PR #3564, opened 2026-05-14
Fix for instrumentation.__ configs when using boolean

### LINKED ISSUE #3558: Not able to disable Rails.event instrumentation
Tried to disable the recently added feature in https://github.com/newrelic/newrelic-ruby-agent/pull/3526, and it doesn't seem to disable the feature.

Did some investigation on the code and I think I found the problem.
This is the configuration I used locally to test the code:
```
development:
  agent_enabled: true
  instrumentation:
    rails_event_logger: false
```

Then started a Rails console with `NEW_RELIC_LICENSE_KEY=key rails c` and observed the following:
```ruby
require 'new_relic/agent/instrumentation/rails_event_log_subscriber'
NewRelic::Agent::Instrumentation::RailsEventLogSubscriber.enabled?
=> :unsatisfied
```
Notice the `:unsatisfied`.

Looking at the code here https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/log_event_aggregator.rb
```ruby
def rails_event_logger_enabled?
  application_logging_and_instrumentation_enabled?(NewRelic::Agent::Instrumentation::RailsEventLogSubscriber.enabled?)
end

def record_rails_event(event)
  return unless rails_event_logger_enabled?
  ...
end
```
It seems that the `:unsatisfied` value will actually make `record_rails_event` work as if the feature was enabled.

I also tracked the reason for the `:unsatisfied`, which comes from https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/dependency_detection.rb when checking the dependencies for https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/rails_event_logger.rb
```
depends_on do
  NewRelic::Agent.config[:'application_logging.enabled'] &&
    NewRelic::Agent.config[:'instrumentation.rails_event_logger']
end
```

So disabling `rails_event_logger` will actually mark it as `unsatisfied`, and then the code listening to the event will treat that as feature enabled.

tannalynn:
If you create an `instrumentation._` config with a boolean type instead of string, it becomes un-disable-able because the config cache didn't know that was allowed. I changed the config to be a string (will still work with ppl typing in false in their yml file though), and I also made a change so it correctly sets it if we create a new  `instrumentation._` config thats a boolean.

resolves https://github.com/newrelic/newrelic-ruby-agent/issues/3558


lib/new_relic/dependency_detection.rb
```diff
@@ -70,7 +70,8 @@ def configure_as_unsatisfied
       #       https://github.com/newrelic/newrelic-ruby-agent/issues/2912
       return if name == :padrino
 
-      NewRelic::Agent.config.instance_variable_get(:@cache)[config_key] = :unsatisfied
+      cache = NewRelic::Agent.config.instance_variable_get(:@cache)
+      cache[config_key] = config_key && cache[config_key] == false ? :disabled : :unsatisfied
```
tannalynn, line 74: Since we also changed the config to be a string like other `instrumentation._` configs,  this change is more of a defensive change in case we accidentally create an `instrumentation._` config as a boolean again in the future, this will make sure it will still work as expected and allows the instrumentation to be disabled.

test/new_relic/dependency_detection_test.rb
```diff
@@ -469,6 +469,27 @@ def test_prepend_is_replaced_by_unsatisfied_when_appropriate
...
+    end
+
+    with_config(:"instrumentation.#{name}" => false) do
+      DependencyDetection.detect!
+
+      assert_equal :disabled, dd.config_value
+
+      DependencyDetection.detect!
+
+      refute executed
```
tannalynn, line 490: There's a second detect w/ assertion bc with the way the bug worked, it would only run the executes block the second time detect! was called.

lib/new_relic/dependency_detection.rb
hannahramadan: I'm wondering about adding the defensiveness in code here vs like a test that stops us from accidentally doing it to start. Wdyt?

tannalynn: I thought about doing that, but i thought it made more sense to just make sure dependency detection works better in general rather than just forcing us to not do a specific thing. In the future maybe we'll want to be able to do it, so this way we won't have to fight with our past selves to do it lol.

test/new_relic/dependency_detection_test.rb
kaylareopelle: Would you mind adding something like this as a comment before the second detect? or update the name to note two detects? I feel like the context could be easy to forget

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+- **Bugfix: Fix `instrumentation.rails_event_logger: false` not disabling the instrumentation**
+
+  Setting `instrumentation.rails_event_logger` to `false` was not disabling the Rails.event instrumentation as expected. The instrumentation would still be installed during Rails boot. This has been fixed. [PR#3564](https://github.com/newrelic/newrelic-ruby-agent/pull/3564)
```
kaylareopelle, line 7: Do you think we should include anything about the `enabled`/`disabled` switch instead of `true`/`false`? If people configured the `instrumentation.rails_event_logger` to be false, will their configs just work now?

tannalynn: yeah, false gets mapped to disabled, so it should just work. i remember i even checked it last week to make sure it would work as it should, so we should be good. I didn't add anything here about the enabled/disabled thing just because i changed the config docs desc to mention it, and it'll work even if they do false. but we can add more if you think it would be helpful for ppl

test/new_relic/dependency_detection_test.rb
tannalynn: good point, added the comment

CHANGELOG.md
kaylareopelle: Got it. That makes sense! I like the entry as it is in that case. 

What do you think about adding a comment on the original bug report that explains it instead? Just so the reporter knows that their current config is fine, but they could update it?

kaylareopelle (approved):

hannahramadan (approved):

CHANGELOG.md
tannalynn: wait, i was wrong, i must have done something weird when i was checking whether that worked bc i definitely thought it had worked correctly, but it thinks it should be auto if it says false. i'll see if i can change that actually

PR #3571, opened 2026-05-21
bugfix: Per-log library supportability metrics

### LINKED ISSUE #3535: Logging library supportability metrics don't reflect instrumentation enablement
The supportability metrics for logging libraries `Supportability/Logging/Ruby/{library}/{enabled|disabled}` currently use `application_logging.enabled` for all libraries, rather than checking whether each specific library's instrumentation is actually enabled. 

Each library's supportability metric should reflect its actual instrumentation, not whether `application_logging` is enabled or not. 

Changing this, we probably still want an overarching supportability metric to see the enabled/disabled status of `application_logging.enabled` - perhaps something like `Supportability/Logging/ApplicationLogging/Ruby/{enabled|disabled}`, but should consult the spec.

hannahramadan:
Fix `Supportability/Logging/Ruby/{library}` to reflect per-library instrumentation state.

closes #3535


test/new_relic/agent/log_event_aggregator_test.rb
```diff
@@ -48,6 +47,15 @@ def populate_container(container, n)
     include NewRelic::DataContainerTests
 
     def test_records_enabled_metrics_on_startup
+      # manually set cache values that with_config can't override
```
hannahramadan, line 50: This part is a little funky. These tests pass locally, but not in the CI. The browser_monitoring_test does a similar thing in the method `setup_nonce` that also hacks the cache to a certain value. 

I tested with an application and confirmed that the enabled/disabled metrics work.

test/multiverse/suites/logging/logging_instrumentation_test.rb
```diff
@@ -271,4 +271,10 @@ def test_forwarding_threshold_allows_custom_levels
       assert_equal 'ERROR', events[1][1]['level']
     end
   end
+
+  def test_records_logging_supportability_metric
+    NewRelic::Agent.config.notify_server_source_added
+
+    assert_metrics_recorded('Supportability/Logging/Ruby/Logging/enabled')
+  end
```
kaylareopelle, line 279: What about adding the inverse test too? Maybe something using `with_config` to set the instrumentation to `:disabled` and make sure the disabled metric is recorded?

test/new_relic/agent/log_event_aggregator_test.rb
kaylareopelle, line 50: I like adding tests specifically to look at the `:unsatisfied` setting. What do you think about updating the name of the test to include `unsatisfied`?

hannahramadan: This test is testing a few things, what do you think of the rename to `test_records_supportability_metrics_on_startup`

kaylareopelle: Yes, that's a good swap. Thanks!

kaylareopelle (dismissed):

kaylareopelle (approved):

PR #3574, opened 2026-05-27
RPC Attribute Mapping

### LINKED ISSUE #3438: Hybrid Agent: Translate RPC attributes
When an OpenTelemetry Span is created with certain attributes related to RPC server or client calls, those attributes should be translated to behave the way they would in our standard GRPC server / client instrumentation. In addition, the name for those transactions/segments should follow New Relic's the GRPC naming processes we use for telemetry emitted from our own instrumentation. See the agent spec for details about which attributes should be translated.

kaylareopelle:
This PR provides support for OpenTelemetry spans that use RPC semantic conventions (specifically looking at gRPC and gruf instrumentations) to translate these conventions into their corresponding New Relic attributes. 

In addition, it adds the `kind` instance variable to spans. This should help us eventually add span kind to every OTel span.

Resolves #3438

### DIFF
--- a/lib/new_relic/agent/opentelemetry/translators/attribute_mappings.rb
+++ b/lib/new_relic/agent/opentelemetry/translators/attribute_mappings.rb
@@ -151,7 +151,24 @@ module NewRelic
           }
         }.freeze
 
-        RPC_MAPPINGS = { # v1.23, v1.17 client
+        RPC_SERVER_MAPPINGS = { # v1.20
+          'http_response_code' => {
+            otel_keys: ['rpc.grpc.status_code'],
+            category: :instance_variable
+          },
+          'request.headers.host' => {
+            otel_keys: ['net.sock.peer.addr', 'server.address', 'net.peer.name'],
+            category: :agent,
+            destinations: DEFAULT_DESTINATIONS
+          },
+          'request.method' => {
+            otel_keys: ['rpc.method'],
+            category: :agent,
+            destinations: DEFAULT_DESTINATIONS
+          }
+        }.freeze
+
+        RPC_CLIENT_MAPPINGS = { # v1.17
           'http_status_code' => {
             otel_keys: ['rpc.grpc.status_code'],
             category: :instance_variable
--- a/lib/new_relic/agent/opentelemetry/translators/attribute_mappings.rb
+++ b/lib/new_relic/agent/opentelemetry/translators/attribute_mappings.rb
@@ -152,7 +152,7 @@ module NewRelic
         }.freeze
 
         RPC_SERVER_MAPPINGS = { # v1.20
-          'http_response_code' => {
+          'response_status' => {
             otel_keys: ['rpc.grpc.status_code'],
             category: :instance_variable
           },
@@ -169,7 +169,7 @@ module NewRelic
         }.freeze
 
         RPC_CLIENT_MAPPINGS = { # v1.17
-          'http_status_code' => {
+          'grpc_status_code' => {
             otel_keys: ['rpc.grpc.status_code'],
             category: :instance_variable
           },
--- a/lib/new_relic/agent/opentelemetry/translators/attribute_mappings.rb
+++ b/lib/new_relic/agent/opentelemetry/translators/attribute_mappings.rb
@@ -150,6 +150,36 @@ module NewRelic
             destinations: DEFAULT_DESTINATIONS
           }
         }.freeze
+
+        RPC_MAPPINGS = { # v1.23, v1.17 client
+          'http_status_code' => {
+            otel_keys: ['rpc.grpc.status_code'],
+            category: :instance_variable
+            # tbd
+          },
+          # the value closest to library in intention is rpc.system. however,
+          # to get the correct segment name, we need to set this to rpc.service
+          'library' => {
+            otel_keys: ['rpc.service'],
+            segment_field: :library
+          },
+          'procedure' => {
+            otel_keys: ['rpc.method'],
+            segment_field: :procedure
+          },
+          'host' => {
+            # traditional host keys aren't sent by otel grpc instrumentation
+            # so we use net.sock.peer.address to fill the host field
+            otel_keys: ['net.sock.peer.addr', 'server.address', 'net.peer.name'],
+            segment_field: :host
+          },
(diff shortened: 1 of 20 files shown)


lib/new_relic/agent/opentelemetry/translators/attribute_mappings.rb
```diff
@@ -150,6 +150,52 @@ module AttributeMappings
             destinations: DEFAULT_DESTINATIONS
           }
         }.freeze
+
+        RPC_SERVER_MAPPINGS = { # v1.20
+          'http_response_code' => {
```
hannahramadan: Should the v1.20 NR attribute be `response.status` instead of `http_response_code`

lib/new_relic/agent/opentelemetry/translators/attribute_mappings.rb
```diff
@@ -150,6 +150,52 @@ module AttributeMappings
...
+          },
+          'request.method' => {
+            otel_keys: ['rpc.method'],
+            category: :agent,
+            destinations: DEFAULT_DESTINATIONS
+          }
+        }.freeze
+
+        RPC_CLIENT_MAPPINGS = { # v1.17
+          'http_status_code' => {
```
hannahramadan: Should the NR one be `grpc.statusCode`?

lib/new_relic/agent/opentelemetry/translators/rpc_translator.rb
```diff
@@ -0,0 +1,59 @@
...
+
+          def create_server_transaction_name(name: nil, attributes: nil, instrumentation_scope: nil)
+            txn_name = name&.delete_prefix(NewRelic::SLASH)
+            Instrumentation::ControllerInstrumentation::TransactionNamer.name_for(nil, nil, :web, {class_name: instrumentation_scope, name: txn_name})
+          end
+
+          def build_uri(attributes)
+            host = attributes['server.address'] || attributes['net.peer.name'] || attributes['net.sock.peer.addr']
+            # in otel contrib's grpc instrumentation, the test example starts
+            # with "dns:///", which can't be accurately parsed using ::URI.parse
```
hannahramadan: I didn't know about this and looked it up! [This guide](https://grpc.io/docs/guides/custom-name-resolution/) says DNS name resolution is used by default but there are other options and even custom ones. May be overkill to do anything about now esp since the default is covered, so maybe just something to note.

lib/new_relic/agent/opentelemetry/translators/attribute_mappings.rb
kaylareopelle: Great question. I'm not sure. I leaned toward `http_response_code` because that's what we already have in our agent as the closest equivalent. However, the agent spec gRPC implementation guide expects `response.status` too. I wasn't feeling great about this approach, and since you mentioned something, I'd like to shore this up.

lib/new_relic/agent/opentelemetry/translators/attribute_mappings.rb
kaylareopelle: Similar thinking to above -- I decided to go with what our agent already has, but it's not really a great fit since the status code isn't an HTTP code. I can fix this.

lib/new_relic/agent/opentelemetry/translators/rpc_translator.rb
kaylareopelle: Great find with that guide! It seems like the thing Ruby's `URI.parse` method is most upset about is the `///`. Since the `///` shows up in more than one location, what do you think about just trying to strip out the third `///` if it's present?

hannahramadan: Oooo maybe that's it! That approach makes it feel more catch all. I like

kaylareopelle: Unfortunately that didn't work the way I hoped it to. URI.parse won't strip off the `dns` part. I'll keep truckin', but wanted to update you.

kaylareopelle: Took another stab at this with [b78c95c](https://github.com/newrelic/newrelic-ruby-agent/pull/3574/commits/b78c95c652c94b7dd988392806ccae6368bcfade).

* sub vs. gsub - sub stops after the first match, gsub looks at the whole line
* \A - looks at beginning of string
* regex should match dns, unix, xds, etc.

lib/new_relic/agent/opentelemetry/translators/rpc_translator.rb
```diff
@@ -38,12 +38,16 @@ def create_server_transaction_name(name: nil, attributes: nil, instrumentation_s
...
             host = attributes['server.address'] || attributes['net.peer.name'] || attributes['net.sock.peer.addr']
-            # in otel contrib's grpc instrumentation, the test example starts
-            # with "dns:///", which can't be accurately parsed using ::URI.parse
-            host = host&.delete_prefix('dns:///')
             service = attributes['rpc.service']
             method = attributes['rpc.method']
 
+            # return early if this method is called with an attributes
+            # hash that doesn't have the required values
+            return unless host || service || method
```
hannahramadan, line 46: nice

lib/new_relic/agent/opentelemetry/translators/rpc_translator.rb
hannahramadan: Nice I like that - so it basically strips anything that comes before `///` + the backslashes. I think that is nice esp since custom prefixes are possible!

test/multiverse/suites/grpc/Envfile
```diff
@@ -10,7 +10,7 @@ instrumentation_methods(:chain, :prepend)
 
 # TODO: permit testing of the nil (latest) version against Ruby 3.3+
```
tannalynn: should we remove this comment? Since below it changes nil to run on 3.1+

tannalynn (dismissed):

test/multiverse/suites/grpc/Envfile
kaylareopelle: Good catch! Removed! a478bf791

tannalynn (approved):

hannahramadan (approved):

Issue #3577, opened 2026-06-01
Feature request: first-party Puma server-statistics plugin (`Puma/*` metrics)


ashleyboehs:
### Is your feature request related to a problem? Please describe.

There is no first-party way to report Puma's server statistics (thread-pool backlog, running threads, spare capacity, configured max threads, worker count) to New Relic. The community [`puma-newrelic`](https://github.com/benoist/puma-newrelic) gem attempts this but records the metrics from the Puma master process, where the agent defers its reporting thread under forking dispatchers, so the metrics are buffered and never delivered in clustered deployments.

### Feature Description

Add an opt-in Puma plugin to the agent that samples `Puma::Launcher#stats` from the master process and records `Puma/*` timeslice metrics, working correctly under clustered (forking) Puma and under High Security Mode.

Key design points:

- Runs as a Puma plugin (`plugin 'newrelic'`) in the master process, since only the master exposes cluster-wide stats.
- Restarts the agent's reporting thread in the master via `NewRelic::Agent.after_fork(force_reconnect: true)` so the sampled metrics are actually delivered. Gated by a config option.
- Uses only `record_metric` (integer gauges, no custom events/attributes), so it is functional under High Security Mode.
- Configurable sample rate.

I have a working implementation with unit tests and would like to open a PR for it. Opening this issue first per the contributing guidelines.

### Describe Alternatives

- The [`puma-newrelic`](https://github.com/benoist/puma-newrelic) community gem — does not deliver metrics from a clustered master (reporting thread deferred under forking dispatchers).
- A harvest-driven `NewRelic::Agent::Sampler` — its `poll` runs only on the per-worker harvest thread, which has no reliable handle to cluster-wide stats.

### Additional context

Open questions for maintainers (also noted in the PR): preferred metric namespace (`Puma/*` proposed, consistent with `CPU/`, `Memory/`, `GC/`); the master reporting as its own instance; and behavior on single-mode / non-forking platforms (e.g. JRuby).

### Priority

Nice to Have

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-571150

hannahramadan:
Closed by #3578

PR #3578, opened 2026-06-01
Add an opt-in Puma plugin for server statistics

ashleyboehs:
# Overview

Adds an opt-in Puma plugin that reports Puma server statistics — thread-pool backlog, running threads, spare capacity, configured max threads, and worker count — as `Puma/*` timeslice metrics. Enable it by adding `plugin 'newrelic'` to `config/puma.rb`. It is opt-in, works under forking dispatchers, and is safe to run under High Security Mode.

This started as a proof of concept to close a gap in first-party Puma server-stats visibility and is now proposed as a shippable, opt-in feature.

Implements #3577.

## What it adds

- `lib/puma/plugin/newrelic.rb` — a Puma plugin (`plugin 'newrelic'`) that runs in the Puma master and starts the sampler on a background thread, stopping cleanly on Puma's `halt`/`restart`/`stop` lifecycle events.
- `lib/new_relic/agent/puma_stats_sampler.rb` — `NewRelic::Agent::PumaStatsSampler`, which samples `Puma::Launcher#stats` and records `Puma/*` metrics. Handles both clustered mode (summing across `worker_status`) and single mode.
- Config options `puma.sample_rate` (default 15s) and `puma.start_reporting_thread_in_master` (default true).

## Design notes

**Why it runs in the master process.** Only the Puma master exposes cluster-wide statistics through `Puma::Launcher#stats`. Worker processes have no reliable handle to their own server statistics from a background thread (`Puma.stats_object` is set only in the master, and `Puma::Server.current` is request-thread-local). A harvest-driven `NewRelic::Agent::Sampler` would `poll` only on the per-worker harvest thread, where the cluster view is unavailable, so sampling is implemented as a Puma plugin in the master instead.

**Why the harvest thread is restarted in the master.** The agent intentionally defers its reporting thread under forking dispatchers (`using_forking_dispatcher?`), so metrics recorded from the master are buffered but never sent. To deliver the sampled metrics, the sampler restarts the reporting thread in the master via `NewRelic::Agent.after_fork(force_reconnect: true)` — the same entry point the agent's `Harvester` uses to recover the thread after a fork. With this enabled, the master reports to New Relic as its own instance. The behavior is gated by `puma.start_reporting_thread_in_master` so it can be turned off.

**High Security Mode.** Only `NewRelic::Agent.record_metric` is used — no custom events or custom attributes (which HSM drops) — and the values are integer gauges with no request, query, or user data. The plugin is therefore fully functional under HSM.

# Testing

Unit tests are added in `test/new_relic/agent/puma_stats_sampler_test.rb` (8 tests) covering:

- Clustered-mode aggregation (summing thread-pool stats across workers; reporting worker count).
- Single-mode aggregation (top-level stats; no worker count).
- End-to-end metric recording (asserting `Puma/*` values reach the metric store).
- JSON-string stats parsing (older Puma).
- Error isolation (a failing `stats` call must not crash the master thread).
- The forced-reporting decision being honored and respecting its config opt-out.

A multiverse functional test that boots a real clustered Puma and verifies metrics end to end is a sensible follow-up; flagging it here rather than including it in this initial PR.

## Open questions for maintainers

- **Metric namespace.** Proposed `Puma/*`, consistent with first-party samplers (`CPU/`, `Memory/`, `GC/`, `Workers/DelayedJob/`). Open to a different reserved namespace if preferred.
- **Master reporting as its own instance.** Restarting the harvest thread in the master means the master appears as a reporting instance. This is the only place with the cluster-wide view; happy to discuss alternatives.
- **Single-mode and non-forking platforms (e.g. JRuby).** Behavior and any guards worth adding.


hannahramadan:
Hi @ashleyboehs! 👋 Thank you so much for such a well thought-out and documented PR! Laying out your process, tradeoffs, and open questions made it easy for us to discuss this internally. A few thoughts:

**Design direction.** Our preference is that instrumentations are fully managed by the agent. Asking users to add `plugin 'newrelic'` to their puma config is somewhat of an anti-pattern for us since it puts the discoverability of the feature on the customer. For comparison, the [agent’s Sequel instrumentation](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/sequel.rb) detects the presence of Sequel and registers itself using Sequel's own extension and plugin APIs. We think Puma can get to a similar place. Would you be open to reworking the PR in that direction?

**Configuration.** When `puma.start_reporting_thread_in_master` is set to `false`, is it essentially disabling the instrumentation?

**Metric namespace.** Great question! We’d prefer metrics to be prefixed with `Ruby/Puma/*`

**Master reporting as its own instance.** Since master is the only place with the cluster-wide view, I think we will have to accept master reporting as an instance. We’d just want this side effect to be documented.

**Single-mode and non-forking platforms (e.g. JRuby).** We can consider this out of scope for this PR, but it would still be good to have the instrumentation no-op in these environments.

Thanks again for the thoughtful work here. Looking forward to your reply ◡̈

ashleyboehs:
Thanks Hannah, this is really helpful feedback and I'm glad the write-up made the discussion easier.

**Design direction.** Yes, I'm happy to rework this toward agent-managed registration rather than an opt-in Puma plugin. The Sequel comparison makes sense as the model. One constraint I want to flag before I rebuild it, since it shapes the approach: the Sequel instrumentation works because the agent and Sequel are loaded in the same process. The Puma stats only exist cluster-wide in the **master** process, and the agent is only loaded in the master when the app runs with `preload_app!` (or when newrelic is required directly in `puma.rb`). In a non-preload deployment the master never loads `newrelic_rpm`, so there's nothing there to self-register.

So "fully managed" would realistically mean: detect Puma via `DependencyDetection` and register the sampler against the launcher lifecycle when the agent is present in the master, document that this requires `preload_app!`, and no-op cleanly otherwise. Does that tradeoff work for you, or would you want to keep a manual fallback for non-preload setups? Happy to go either way, just want to align before reworking.

**Configuration.** Good question, and it exposes something confusing I'd like to remove. `start_reporting_thread_in_master: false` doesn't stop sampling — it skips the harvest-thread restart, so the master keeps recording metrics into a buffer that the deferred reporting thread never flushes. The net effect is no data, just wasted work. I'd rather drop that option entirely and gate the feature on a standard `disable_puma_instrumentation` config to match the other instrumentations.

**Metric namespace.** Will switch to `Ruby/Puma/*`.

**Master reporting as its own instance.** Agreed, I'll keep that behavior and call it out clearly in the docs.

**Single-mode and non-forking platforms.** Will add an explicit no-op for JRuby/non-forking environments so the instrumentation stays inert there.

I'll start on the rework once you confirm the preload tradeoff above. Thanks again.

hannahramadan:
@ashleyboehs Thanks for flagging the preload constraint. The tradeoff you've outline works. Detecting Puma via `DependencyDetection`, registering against the launcher lifecycle when the agent is present in the master, documenting the `preload_app!` requirement, and no-oping otherwise sounds right. Preloading the app seems to be a recommended setting, so hopefully the affected subset is small. I found in the cluster mode [documentation](https://github.com/puma/puma#cluster-mode):

> If the number of workers is greater than 1 (and --prune-bundler has not been specified), preloading will be enabled by default.

No need to keep a manual fallback for non-preload setups, documenting the requirement is should be enough. And everything else sounds great! Please let us know if you have any other questions that come up.

ashleyboehs:
I've pushed the rework from the opt-in plugin to agent-managed instrumentation. Dependency detection reads the runner that `Puma::Launcher#initialize` registers through `Puma.stats_object=`, starts the sampler in the master (single mode, or clustered with preload), and declines in workers. The multiverse suite now builds real launchers to cover the detection gate and lifecycle wiring on 6.6 and latest.

One tradeoff worth a decision: the plugin version didn't care who booted first, because Puma called `start(launcher)` itself at the right moment. Auto-detection inverts that. The agent looks for the runner once, at startup, which works whenever Puma boots the app (`puma`/`pumactl`, all modes) but misses entry points that build the app before constructing the launcher, like `rails server`, where detection runs too early and nothing re-runs it. The file header documents the limitation.

If you want `rails server` covered, the cleanest fix I see is a one-shot prepend on `Puma.stats_object=`, armed only when detection finds Puma loaded with no runner yet. That setter fires exactly when a launcher registers its runner, so the late path installs through the same `master?` gate, and the two paths cover each other's blind spots. Happy to add it here or in a follow-up, whichever you prefer.

hannahramadan:
@ashleyboehs I think `rails server` is a case worth covering in this PR. Thanks for being so on top of tradeoffs and solutions. 

Separately, it looks like there are some CI failures! Please let me know if you want help on either. For the unit test failure, take a look at `lib/new_relic/agent/instrumentation/sequel.rb` as the model, it uses `@name = :sequel` paired with a `disable_sequel_instrumentation` config.

ashleyboehs:
Thanks Hannah! Both are handled now.

**rails server boot order:** I pulled the master decision out of dependency detection. The `executes` block now checks whether Puma has already registered its runner. If it has (the `puma`/`pumactl` boot order), it installs in the master right away. If it hasn't (the `rails server` order, where the app loads before the launcher is built), it arms a one-shot interceptor on `Puma.stats_object=` and installs through the same `master?` gate the moment the launcher registers. The interceptor wraps only the agent reaction, so `super` runs unguarded and an agent-side failure can't escape into Puma's boot. There's a test that forces that failure and asserts Puma still records its runner.

**unit test failure:** Fixed using sequel as the model, like you suggested. I set `@name = :puma` directly instead of calling `named`, paired with a dedicated `disable_puma_instrumentation` config that the `depends_on` gate reads. That keeps the orphan-config test happy without inventing `disable_puma`/`instrumentation.puma` keys this instrumentation doesn't define.

The multiverse suite disables the instrumentation by default now so the repeated agent starts during setup stay inert, and the tests opt in with `with_config`. They cover both boot orders, the one-shot disarm on a second registration (hot restart), the non-master decline, and the failure isolation above. Verified against Puma 6.6.1 and 8.0.2.

Appreciate the pointer on sequel, that made the config side straightforward.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2080,6 +2080,33 @@ def self.convert_to_constant_list(string_array)
...
+        },
+        :'puma.sample_rate' => {
+          :default => 15,
+          :public => true,
+          :type => Integer,
+          :allowed_from_server => false,
+          :description => 'Number of seconds between samples of Puma server statistics taken in the ' \
+            'Puma master process.'
+        },
+        :'puma.start_reporting_thread_in_master' => {
```
hannahramadan: Do we still need this config? If disabling `puma.start_reporting_thread_in_master` means no `Puma/*` metrics are sent, could we just rely on `disable_puma_instrumentation` to achieve the same result and move some of the helpful description there about the master reporting as its own instance

lib/new_relic/agent/puma_stats_sampler.rb
```diff
@@ -0,0 +1,345 @@
...
+    # request, query, or user data, so the instrumentation is fully functional
+    # under High Security Mode.
+    class PumaStatsSampler
+      # Raised when Puma's +stats+ returns a Hash whose shape we don't
+      # recognize (no +:worker_status+ and no top-level worker keys). Lets
+      # +fetch_metrics+'s existing rescue route the case through the
+      # throttled sample-failure path instead of silently recording nothing.
+      class UnrecognizedStatsError < StandardError; end
+
+      METRIC_NAMESPACE = 'Puma'
```
hannahramadan:
```suggestion
      METRIC_NAMESPACE = 'Ruby/Puma'
```

lib/new_relic/agent/puma_stats_sampler.rb
```diff
@@ -0,0 +1,345 @@
...
+      # gauges.
+      WORKER_STAT_KEYS = %i[backlog running pool_capacity max_threads requests_count].freeze
+      # Runner metadata reported by +Puma::Runner#stats+ alongside (or, before
+      # the runner's +Puma::Server+ exists, instead of) the per-worker keys.
+      RUNNER_INFO_KEYS = %i[started_at versions].freeze
+      # Puma's +pool_capacity+ is the number of additional requests the pool
+      # can accept right now: idle (waiting) threads plus unspawned threads
+      # still allowed up to +max_threads+ (i.e. +waiting + (max - spawned)+).
+      # Reporting it as +spare_thread_capacity+ makes that meaning explicit.
+      METRIC_NAME_OVERRIDES = {pool_capacity: 'spare_thread_capacity'}.freeze
```
hannahramadan: While I agree `spare_thread_capacity` is intuitive, I wonder if we should preserve Puma's `pool_capacity` name. I searched Puma docs for pool_capacity and found the definition there. Generally, we try to stick to names that libraries have defined.

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2080,6 +2080,33 @@ def self.convert_to_constant_list(string_array)
...
+        # Puma
+        :disable_puma_instrumentation => {
+          :default => false,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => 'If `true`, disables the Puma server-statistics instrumentation that samples `Puma/*` metrics from the Puma master process.'
+        },
+        :'puma.sample_rate' => {
+          :default => 15,
```
hannahramadan: What is behind the decision to use a 15s default? The agent typically uses 60s

test/multiverse/suites/puma/Envfile
```diff
@@ -0,0 +1,21 @@
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+# The instrumentation's launcher-event registrations differ between major
+# Puma versions: 7+ fires :before_restart / :after_stopped while 6.x fires
+# :on_restart / :on_stopped. Pin a 6.x variant alongside the latest so the
+# suite exercises both paths.
+PUMA_VERSIONS = [
+  [nil],
```
hannahramadan: Can we add a version 7.x to test against here as well?

lib/new_relic/agent/instrumentation/puma.rb
```diff
@@ -0,0 +1,173 @@
...
+DependencyDetection.defer do
+  # Set +@name+ directly, as +sequel+ does, rather than calling +named+. The two
+  # are equivalent at runtime, but the orphan-config test treats a +named+ entry
+  # as a promise that +disable_<name>+ / +instrumentation.<name>+ config keys
+  # exist. This instrumentation defines neither; it gates on a dedicated
+  # +disable_puma_instrumentation+ via the explicit +depends_on+ below.
+  @name = :puma
+
+  depends_on do
+    defined?(Puma) && defined?(Puma::Launcher)
```
hannahramadan, line 160: Do we know if instrumentation works/ has any side effects on versions less than 6? I'm wondering if it is worth only installing on supported versions, we have a method `NewRelic::Helper.version_satisfied?` to help with this

lib/new_relic/agent/configuration/default_source.rb
ashleyboehs: Good point! I removed it.

lib/new_relic/agent/puma_stats_sampler.rb
ashleyboehs: Done. Added this across the board.

lib/new_relic/agent/puma_stats_sampler.rb
ashleyboehs: Agreed.

lib/new_relic/agent/configuration/default_source.rb
ashleyboehs: No reason. Changed to 60 to match the agent.

test/multiverse/suites/puma/Envfile
ashleyboehs: Added

lib/new_relic/agent/instrumentation/puma.rb
ashleyboehs: I haven't verified below 6, so I gated rather than guess. 

The stats we read come from `Puma.stats_hash` (the `@get_stats` ivar that `stats_object=` writes), which didn't exist before Puma 5.0. On anything older than 5 there's no runner to read, so the eager path finds nothing and the late path just prepends a `stats_object=` interceptor that never fires. You'd get a pointless monkeypatch on Puma's singleton and no metrics.

The launcher events we hook for stop/restart also got reworked in 6.0 when `LogWriter` was pulled out of `Events`, and the hook names kept moving around inside 6.x too (`on_stopped` didn't show up until 6.5). So even on 5.x where the stats are there, I can't vouch for the shutdown handling.

That's why the gate is `version_satisfied?(Puma::Const::VERSION, '>=', '6.0.0')`. One thing to flag: the suite only goes down to 6.6, so 6.0–6.5 are allowed in but not actually tested. If you'd rather the gate match, I can bump it to `'6.6.0'`.

test/multiverse/suites/puma/Envfile
```diff
@@ -0,0 +1,22 @@
...
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+# The instrumentation's launcher-event registrations differ between major
+# Puma versions: 7+ fires :before_restart / :after_stopped while 6.x fires
+# :on_restart / :on_stopped. Pin a 6.x and a 7.x variant alongside the latest
+# so the suite exercises both event-name paths and the current release.
+PUMA_VERSIONS = [
+  [nil],
+  ['7.0'],
```
hannahramadan: This is just for the CI to pass. Puma 7,0+ requires ruby 3.0+
```suggestion
  [nil, 3.0],
  ['7.0', 3.0],
```

lib/new_relic/agent/instrumentation/puma.rb
hannahramadan: Thank you for the clear explanation here! Let's bump to 6.6.0 for now to ship with the gate matching what we actually verify in CI

test/multiverse/suites/puma/Envfile
ashleyboehs: Done — applied the suggestion. Constraining the latest and 7.0 variants to Ruby 3.0+ skips them on the 2.6.10 lane, which is what the `rest` suite was failing on.

lib/new_relic/agent/instrumentation/puma.rb
ashleyboehs: Bumped the gate to `6.6.0` so it matches what we verify in CI.

hannahramadan (approved):
Thank you for this contribution 🚀

hannahramadan:
Hi @ashleyboehs! Thank you again for all your work here.

Since Puma is a popular tool with our users and this is a new instrumentation pattern, we opted to disable Puma instrumentation by default. We added a metric to track its install and will revisit enabling by default at a later time. Puma instrumentation will go in Ruby agent version 10.7.0. I'll let you know here when it's available.

PR #3579, opened 2026-06-02
instrumentation._ string configs also check for false to disable

tannalynn:
if someone puts "false" in the config for a instrumentation.___ config, now it will be treated as disabled


kaylareopelle (approved):
💀 🦄

hannahramadan (approved):

PR #3589, opened 2026-06-10
Set span.kind as intrinsic attribute on OTel spans

### LINKED ISSUE #3543: Make sure span kind is assigned on every OTel span
The New Relic agent correctly applies the client span kind as an attribute on datastore and external request segments. However, server spans created by the hybrid agent do not have the span kind attribute applied. 

Update the logic to make sure span kind is added as an intrinsic attribute for all OTel spans.

kaylareopelle:
Previously, `span.kind` would only be set on external request segments and datastore segments. Now, generic segments that pass through `SpanEventPrimitive#for_segment` will also have span kind assigned when there is an associated OTel span with the kind attribute.

Closes #3543

### DIFF
--- a/lib/new_relic/agent/opentelemetry/span_event_primitive_patch.rb
+++ b/lib/new_relic/agent/opentelemetry/span_event_primitive_patch.rb
@@ -15,7 +15,7 @@ module NewRelic
 
         private
 
-        def get_span_kind(span_data, segment)
+        def attach_span_kind(span_data, segment)
           if segment.instance_variable_defined?(:@otel_span)
             otel_span = segment.instance_variable_get(:@otel_span)
             kind = otel_span.kind&.to_s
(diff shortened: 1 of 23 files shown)


hannahramadan (dismissed):

tannalynn (dismissed):

lib/new_relic/agent/opentelemetry/span_event_primitive_patch.rb
```diff
@@ -59,7 +72,7 @@ def for_span_links(segment)
         end
 
         def attach_span_event_events(span_data, segment)
-          span_events = for_span_events(segment)
+          span_events = for_span_event_events(segment)
```
kaylareopelle, line 75: straggler method name. happy to revert too.

tannalynn (dismissed):

tannalynn (approved):

PR #3591, opened 2026-06-11
Hybrid Agent: Translate Messaging Attributes

### LINKED ISSUE #3436: Hybrid Agent: Translate Messaging attributes
When an OpenTelemetry Span is created with a :consumer kind, there are certain attributes we want to apply to the agent attributes within the New Relic-created telemetry. Translate these attributes for transactions and segments. See the agent spec for which attributes should be translated.

hannahramadan:
Adds attribute translation for consumer and producer spans.

### DIFF
--- a/lib/new_relic/agent/opentelemetry/trace/tracer.rb
+++ b/lib/new_relic/agent/opentelemetry/trace/tracer.rb
@@ -100,6 +100,7 @@ module NewRelic
                   library: segment_api_params[:library],
                   destination_type: segment_api_params[:destination_type],
                   destination_name: segment_api_params[:destination_name],
+                  parameters: messaging_segment_parameters(segment_api_params),
                   start_time: start_timestamp
                 )
               else
@@ -131,6 +132,12 @@ module NewRelic
             add_remote_context_to_txn(nr_item, parent_otel_context) if nr_item
 
             nr_item
+          end
+
+          def messaging_segment_parameters(segment_api_params)
+            return nil unless NewRelic::Agent.config[:'message_tracer.segment_parameters.enabled']
+
+            segment_api_params[:parameters]
           end
 
           def start_messaging_transaction(name, segment_api_params)
--- a/lib/new_relic/agent/opentelemetry/trace/tracer.rb
+++ b/lib/new_relic/agent/opentelemetry/trace/tracer.rb
@@ -2,7 +2,6 @@
 # See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
 # frozen_string_literal: true
 
-require_relative '../../messaging'
 require_relative '../attribute_translator'
 
 module NewRelic
@@ -141,15 +140,10 @@ module NewRelic
           end
 
           def start_messaging_transaction(name, segment_api_params)
-            return NewRelic::Agent::Tracer.start_transaction_or_segment(name: name, category: :task) unless segment_api_params[:library]
+            return NewRelic::Agent::Tracer.start_transaction_or_segment(name: name, category: :task) unless segment_api_params[:name]
 
             NewRelic::Agent::Tracer.start_transaction_or_segment(
-              name: NewRelic::Agent::Messaging.transaction_name(
-                segment_api_params[:library],
-                segment_api_params[:destination_type],
-                segment_api_params[:destination_name],
-                segment_api_params[:action]
-              ),
+              name: segment_api_params[:name],
               category: :message
             )
           end
--- a/lib/new_relic/agent/opentelemetry/trace/tracer.rb
+++ b/lib/new_relic/agent/opentelemetry/trace/tracer.rb
@@ -66,7 +66,7 @@ module NewRelic
             case kind
             when :client
               if segment_api_params[:uri] # HTTP Client and gRPC Client
-                return NewRelic::Agent::Tracer.start_external_request_segment(
+                NewRelic::Agent::Tracer.start_external_request_segment(
                   library: segment_api_params[:library] || @name,
                   uri: segment_api_params[:uri],
                   procedure: segment_api_params[:procedure],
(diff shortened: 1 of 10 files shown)


lib/new_relic/agent/opentelemetry/trace/tracer.rb
```diff
@@ -118,6 +134,26 @@ def start_transaction_from_otel(name, parent_otel_context, kind, translated: {})
...
+          def start_messaging_transaction(name, segment_api_params)
+            return NewRelic::Agent::Tracer.start_transaction_or_segment(name: name, category: :task) unless segment_api_params[:library]
+
+            NewRelic::Agent::Tracer.start_transaction_or_segment(
+              name: NewRelic::Agent::Messaging.transaction_name(
+                segment_api_params[:library],
+                segment_api_params[:destination_type],
+                segment_api_params[:destination_name],
+                segment_api_params[:action]
+              ),
```
kaylareopelle: What do you think about moving this logic out of the OTel Tracer and into the messaging translator as a `add_specialized_attributes` operation? [That's what I did for server spans](https://github.com/newrelic/newrelic-ruby-agent/blob/e20ce51d944ed6fee8050764fea46f4351978ad3/lib/new_relic/agent/opentelemetry/translators/http_server_translator.rb#L35).

lib/new_relic/agent/opentelemetry/translators/attribute_mappings.rb
```diff
@@ -196,6 +196,71 @@ module AttributeMappings
...
+          # MessagingTranslator#add_producer_segment_params.
+          'routingKey' => {
+            otel_keys: [
+              'messaging.kafka.message.key',
+              'messaging.rabbitmq.destination.routing_key'
+            ]
+          },
+          'correlation_id' => {
+            otel_keys: ['messaging.message.conversation_id']
+          }
```
kaylareopelle: ooh interesting. I don't think I always put keys in the translators if they weren't getting translated even if they were in the spec documentation. Is this used by the translator in some way?

lib/new_relic/agent/opentelemetry/translators/messaging_translator.rb
```diff
@@ -0,0 +1,113 @@
...
+            'kafka' => :topic,
+            'rabbitmq' => :exchange,
+            'aws.sqs' => :queue,
+            'aws_sqs' => :queue,
+            'aws.sns' => :topic,
+            'aws_sns' => :topic,
+            'aws.kinesis' => :stream,
+            'aws_kinesis' => :stream
+          }.freeze
+        }.freeze
```
kaylareopelle, line 35: Do we need freezes on every sub-hash?

test/multiverse/suites/hybrid_agent/messaging_mapping_test.rb
```diff
@@ -0,0 +1,643 @@
...
+            harvest_span_events!
+            reset_adaptive_samplers!
+          end
+
+          def teardown
+            mocha_teardown
+            # Tests that exercise remote-parent paths call into the
+            # AdaptiveSampler (which has a 10-call-per-period budget). Reset
+            # to avoid bleeding state into later tests in the suite.
+            reset_adaptive_samplers!
```
kaylareopelle: Interesting add!

So the `txn.stubs(:sampled?).returns(true)` wasn't taking care of everything?

test/multiverse/suites/hybrid_agent/messaging_mapping_test.rb
```diff
@@ -0,0 +1,643 @@
...
+            assert_metrics_recorded([
+              'OtherTransaction/Message/RabbitMQ/Queue/Consume/Named/orders.queue',
+              'OtherTransactionTotalTime'
+            ])
+          end
+
+          # ---------- consumer with remote parent ----------
+          # Per spec: a consumer span with a remote parent MUST create a new
+          # Other transaction. NR creates it with category :message and uses
+          # Messaging.transaction_name to produce the proper metric name.
```
kaylareopelle, line 500: Could you add this comment to the `producer with remote parent` section too?

kaylareopelle (commented):
You definitely got the trickiest translator! Thanks for your work juggling all the things on this one. I have a few small comments.

lib/new_relic/agent/opentelemetry/translators/messaging_translator.rb
hannahramadan: No I dont think so, I think that was leftover - I originally had producer and consumer as separate translators before taking inspo from you RPC translator. Will remove ◡̈

test/multiverse/suites/hybrid_agent/messaging_mapping_test.rb
hannahramadan: [done](https://github.com/newrelic/newrelic-ruby-agent/pull/3591/commits/71b690b1cf80b1ef9674089239957b6f8004d5c5)

lib/new_relic/agent/opentelemetry/translators/messaging_translator.rb
hannahramadan: [done](https://github.com/newrelic/newrelic-ruby-agent/pull/3591/commits/71b690b1cf80b1ef9674089239957b6f8004d5c5)

lib/new_relic/agent/opentelemetry/translators/attribute_mappings.rb
hannahramadan: Good call! I was laser focused on getting everything in the spec docs here and didn't through this. Here, it just creates and extra step associating an otel key with an NR key and then pulls it back out again in the translator. The translator can take care of all of that. https://github.com/newrelic/newrelic-ruby-agent/pull/3591/commits/65c10096d5bdb4f476768ae3742ee29aae1eb442

lib/new_relic/agent/opentelemetry/trace/tracer.rb
hannahramadan: Great refactor idea https://github.com/newrelic/newrelic-ruby-agent/pull/3591/commits/65c10096d5bdb4f476768ae3742ee29aae1eb442

test/multiverse/suites/hybrid_agent/messaging_mapping_test.rb
hannahramadan: It wasn't! I was getting flakes. I had Claude run through this and:

The remote-parent tests call `@tracer.start_span(with_parent: remote_context, ...)` directly, which creates a new `:message` Other transaction inside the tracer. That path bypasses `Transaction#sampled?` entirely; instead `trace_context.rb#determine_sampling_decision` writes `transaction.sampled = adaptive_sampler_remote_parent_sampled.sampled?` directly via the writer, before `start_span` returns to the test. There's no instance to stub at the right time, so the only way to control that outcome is at the AdaptiveSampler level.

Better explanation than I could give haha

kaylareopelle: Fascinating. I wonder if that's a bug? Because users (and contrib instrumentation) might feasibly call `@tracer.start_span` approach, and we could run into issues where `.sampled?` isn't returned correctly for the current transaction. Maybe we can look at this together sync?

kaylareopelle (commented):
thanks for all the updates! 🚀

test/multiverse/suites/hybrid_agent/messaging_mapping_test.rb
hannahramadan: Chatted about this live and decided to change the failing test to remove `sampled?` check, since it wasn't  required not to be per spec. Also removed `reset_adaptive_samplers!`

tannalynn (approved):

kaylareopelle (approved):

PR #3601, opened 2026-06-26
Prototype: OpenTelemetry config support

kaylareopelle:
Allow some OpenTelemetry environment variables to be applied to New Relic configuration values at the lowest precedence, above only defaults.


kaylareopelle:
Using a different precedence paradigm.

PR #3602, opened 2026-06-26
Add server-side config for AI

hannahramadan:
Allow `ai_monitoring.enabled` and `ai_monitoring.record_content.enabled` to be set server-side.

### DIFF
--- a/lib/new_relic/agent/configuration/server_source.rb
+++ b/lib/new_relic/agent/configuration/server_source.rb
@@ -42,6 +42,7 @@ module NewRelic
           filter_keys(merged_settings)
 
           apply_feature_gates(merged_settings, connect_reply, existing_config)
+          apply_ai_monitoring_gate(merged_settings, connect_reply, existing_config)
 
           # The value under this key is a hash mapping transaction name strings
           # to apdex_t values. We don't want the nested hash to be flattened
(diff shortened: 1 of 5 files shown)


lib/new_relic/agent/configuration/server_source.rb
```diff
@@ -154,6 +155,22 @@ def apply_feature_gates(merged_settings, connect_reply, existing_config)
         def ungated_value(key, merged_settings, existing_config)
           merged_settings.has_key?(key) ? merged_settings[key] : existing_config[key.to_sym]
         end
+
+        # `ai_monitoring.enabled` is gated by the `collect_ai` flag, but a value
+        # sent directly in the agent_config takes precedence over that gate.
+        def apply_ai_monitoring_gate(merged_settings, connect_reply, existing_config)
```
hannahramadan: This method and tests were inspired by the method `apply_feature_gates` and those tests

kaylareopelle: Can you tell me more about the `collect_ai` flag? What does it do? Who sets it? Is it like logs where there's an account-level toggle to control it?

lib/new_relic/agent/configuration/server_source.rb
```diff
@@ -154,6 +155,22 @@ def apply_feature_gates(merged_settings, connect_reply, existing_config)
         def ungated_value(key, merged_settings, existing_config)
           merged_settings.has_key?(key) ? merged_settings[key] : existing_config[key.to_sym]
         end
+
+        # `ai_monitoring.enabled` is gated by the `collect_ai` flag, but a value
+        # sent directly in the agent_config takes precedence over that gate.
```
kaylareopelle: for "sent directly in the agent config", is that if someone specifically sets it using server-side config? could any config source set it?

lib/new_relic/agent/instrumentation/ruby_openai.rb
```diff
@@ -33,3 +33,11 @@
     end
   end
 end
+
+# `ai_monitoring.enabled` gates this instrumentation at install time, but can be
+# enabled by server-side config after startup detection has run. Re-check the
+# dependency when the server source is applied.
```
kaylareopelle, line 39: This surprises me! Does the agent install instrumentation before all the config sources have been applied?

tannalynn: Oh yeah, so it does all the other config sources before it installs instrumentation, just not the ssc/connect payload stuff since that can take a while. That's why our instrumentation install related configs (like instrumentation.whatever and disable_whatever) are not allowed from server.

lib/new_relic/agent/configuration/server_source.rb
tannalynn: Also, how does apply_feature_gates work? and what is it actually doing? sorry, i don't think i've ever seen/looked at it before.

kaylareopelle (approved):

PR #3607, opened 2026-07-01
Fix CI for rails edge

tannalynn:
Rails edge changed something and now they reference refute_pattern from minitest but that's not in the minitest version we're running, so we need to bump minitest when we run rails edge. And then we needed to do other stuff to cope with the minitest version change

[ci cron run](https://github.com/newrelic/newrelic-ruby-agent/actions/runs/28539601901)

### DIFF
--- a/test/multiverse/lib/multiverse/suite.rb
+++ b/test/multiverse/lib/multiverse/suite.rb
@@ -612,6 +612,7 @@ module Multiverse
     # Rails and minitest_tu_shim both want to do MiniTest::Unit.autorun for us
     # We can't sidestep, so just gut the method to avoid doubled test runs
     def prevent_minitest_auto_run
+      # TODO: OLD RUBIES - 2.6 - Remove this method when we drop support
       # MiniTest 4.x (absent on modern Minitest, which has no MiniTest::Unit)
       if defined?(::MiniTest::Unit)
         ::MiniTest::Unit.class_eval do
--- a/test/multiverse/lib/multiverse/suite.rb
+++ b/test/multiverse/lib/multiverse/suite.rb
@@ -328,6 +328,11 @@ module Multiverse
         require 'minitest/unit'
       end
       require 'minitest/mock'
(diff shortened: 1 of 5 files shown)


test/multiverse/suites/rails/rails_event_logger_test.rb
```diff
@@ -13,6 +13,7 @@ def setup
     skip 'Rails.event requires Rails 8.1+' unless rails_event_available?
 
     @aggregator = NewRelic::Agent.agent.log_event_aggregator
+    @aggregator.reset!
```
tannalynn, line 16: I can't imagine this failure was _actually_ related to the minitest bump, it was probably just a regular intermittent failure I got unlucky enough to get while trying to fix all the minitest stuff, so it gets to be in this PR too.

test/multiverse/lib/multiverse/suite.rb
```diff
@@ -607,10 +612,12 @@ def patch_minitest_base_for_old_versions
     # Rails and minitest_tu_shim both want to do MiniTest::Unit.autorun for us
     # We can't sidestep, so just gut the method to avoid doubled test runs
     def prevent_minitest_auto_run
-      # MiniTest 4.x
-      ::MiniTest::Unit.class_eval do
-        def self.autorun
-          # NO-OP
+      # MiniTest 4.x (absent on modern Minitest, which has no MiniTest::Unit)
```
kaylareopelle, line 616: I was curious about how many versions still needed minitest 4. It looks like just Ruby 2.6. What do you think about adding a comment to remove the method when we drop support?
```suggestion
      # TODO: OLD RUBIES - 2.6 - Remove this method when we drop support
      # MiniTest 4.x (absent on modern Minitest, which has no MiniTest::Unit)
```

kaylareopelle (dismissed):
This looks great! One suggestion, but we might not need it.

kaylareopelle (approved):

PR #3608, opened 2026-07-06
Fix intermittent rails suite failures

tannalynn:
So if the tests ran in a certain order, it would trigger rails to reload our test apps routes. But we don't define the routes normally (because test app), so instead of reloading, it just removes all the routes. Not ideal. There is a config to make it never do the route reloading though, so i added that config (only on rails 7.2+, which is the only versions that have this)


kaylareopelle (approved):
Thanks for your sleuthing on this one! 🕵️

hannahramadan (approved):

PR #3609, opened 2026-07-06
Add GCR hostname configs

### LINKED ISSUE #3295: Support unique instance identifiers for Google Cloud Run containers
When running the New Relic Ruby agent on Google Cloud Run (fully managed), the agent reports the hostname as "localhost" for every container instance. As a result, all Cloud Run instances are collapsed into a single localhost entry in New Relic APM.
This makes it impossible to:
- Identify which specific instance is experiencing performance issues or errors
- Correlate performance spikes or exceptions to individual containers
- Debug production issues effectively when multiple instances are running concurrently
- Track instance-level metrics across different container revisions
This severely limits the observability benefits of New Relic in Cloud Run environments, especially during production incident response.

### Feature Description
Support distinct host/instance display names on Cloud Run by either auto-detecting the platform or opting in via an env var, and then deriving a per-instance identifier from Cloud Run instance metadata.

### Describe Alternatives
Manual override attempted: I looked for Ruby agent configuration options to override the hostname/instance identifier, but either:
No suitable configuration option exists for this use case, or
Workaround considered: Setting a unique hostname via environment variable per instance, but Cloud Run doesn't provide a straightforward way to inject unique per-instance environment variables at deployment time (environment variables are set at the service/revision level, not per instance).
Documentation request: If a configuration option already exists that can solve this, clear documentation with a Cloud Run-specific example would be extremely helpful.

### Additional context
Environment examples:
Platform: Google Cloud Run (fully managed)
Ruby: 3.4.2
Framework: Rails 7.2.2.2
Agent: newrelic_rpm 9.21.0

### Priority
Nice to Have

hannahramadan:
Add option to report the GCP instance id as the hostname. Two new configs:
- `utilization.gcp_cloud_run.use_instance_as_host` (default `true`) - reports instance id
- `utilization.gcp_cloud_run.include_revision_in_host` (default `false`) reports {K_REVISION}-{instance id} 

Falls back to socket hostname on failure.

Closes #3295.

Example from playground application:
Screenshot 2026-07-07 at 5 57 01 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - **Feature: Report a unique hostname for Google Cloud Run instances**
 
-  The agent now detects Cloud Run and reports the GCP instance id as the hostname so individual instances can be distinguished. This is controlled by the new `utilization.gcp_cloud_run.use_instance_as_host` configuration option (default `true`). Set `utilization.gcp_cloud_run.include_revision_in_host` (default `false`) to `true` to report the hostname as `{K_REVISION}-{instance id}` instead. [Issue#3295](https://github.com/newrelic/newrelic-ruby-agent/issues/3295)
+  The agent now detects Cloud Run and reports the GCP instance id as the hostname so individual instances can be distinguished. This is controlled by the new `utilization.gcp_cloud_run.use_instance_as_host` configuration option (default `true`). Set `utilization.gcp_cloud_run.include_revision_in_host` (default `false`) to `true` to report the hostname as `{K_REVISION}-{instance id}` instead, where [`K_REVISION`](https://docs.cloud.google.com/run/docs/container-contract#env-vars) is the Cloud Run revision name. [Issue#3295](https://github.com/newrelic/newrelic-ruby-agent/issues/3295)
 
 ## v10.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - **Feature: Report a unique hostname for Google Cloud Run instances**
 
-  The agent now detects Cloud Run and reports the GCP instance id as the hostname so individual instances can be distinguished. This is controlled by the new `utilization.gcp_cloud_run.use_instance_as_host` configuration option (default `true`). Set `utilization.gcp_cloud_run.include_revision_in_host` (default `false`) to `true` to report the hostname as `{K_REVISION}-{instance id}` instead, where [`K_REVISION`](https://docs.cloud.google.com/run/docs/container-contract#env-vars) is the Cloud Run revision name. [Issue#3295](https://github.com/newrelic/newrelic-ruby-agent/issues/3295)
+  The agent now detects Cloud Run and reports the GCP instance id as the hostname so individual instances can be distinguished. Before this change, all Google Cloud Run hostnames were `localhost`. This feature is controlled by the new `utilization.gcp_cloud_run.use_instance_as_host` configuration option (default `true`). Set `utilization.gcp_cloud_run.include_revision_in_host` (default `false`) to `true` to report the hostname as `{K_REVISION}-{instance id}` instead, where [`K_REVISION`](https://docs.cloud.google.com/run/docs/container-contract#env-vars) is the Cloud Run revision name. Thank you to [@kawa-onushi](https://github.com/kawa-onushi) for suggesting this improvement. [Issue#3295](https://github.com/newrelic/newrelic-ruby-agent/issues/3295) [PR#3609](https://github.com/newrelic/newrelic-ruby-agent/pull/3609/)
 
 ## v10.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+- **Feature: Report a unique hostname for Google Cloud Run instances**
+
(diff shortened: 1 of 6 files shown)


lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2351,6 +2351,20 @@ def self.convert_to_constant_list(string_array)
...
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => 'If `true`, the agent reports the GCP instance id as the hostname when running on Google Cloud Run.'
+        },
+        :'utilization.gcp_cloud_run.include_revision_in_host' => {
+          :default => false,
+          :public => true,
+          :type => Boolean,
+          :allowed_from_server => false,
+          :description => 'If `true`, the agent prepends the `K_REVISION` value to the GCP instance id to form the hostname (`{K_REVISION}-{instance id}`) on Google Cloud Run. Has no effect unless `utilization.gcp_cloud_run.use_instance_as_host` is also `true`.'
```
kaylareopelle, line 2366: wdyt about adding a link to the K_REVISION documentation here? https://docs.cloud.google.com/run/docs/container-contract#env-vars

lib/new_relic/agent/hostname.rb
```diff
@@ -3,22 +3,60 @@
 # frozen_string_literal: true
 
 require 'socket'
+require 'net/http'
 require 'new_relic/helper'
 
 module NewRelic
   module Agent
     module Hostname
+      CLOUD_RUN_REVISION = 'K_REVISION'
```
kaylareopelle, line 11: I'm not sure we need this one as a separate constant. Do you find it more helpful to write `CLOUD_RUN_REVISION` than `K_REVISION`?

lib/new_relic/agent/hostname.rb
```diff
@@ -3,22 +3,60 @@
...
+          instance_id
+        end
+      end
+
+      def self.gcp_instance_id
+        uri = URI(GCP_INSTANCE_ID_URI)
+        response = nil
+        Net::HTTP.start(uri.host, uri.port, open_timeout: 1, read_timeout: 1) do |http|
+          response = http.request(Net::HTTP::Get.new(uri, GCP_METADATA_HEADERS))
+        end
```
kaylareopelle: This looks really similar to the call we have in `request_metadata`. Is it possible use  or abstract that method and pass in the URI and GCP_METADATA_HEADERS as arguments?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+- **Feature: Report a unique hostname for Google Cloud Run instances**
+
+  The agent now detects Cloud Run and reports the GCP instance id as the hostname so individual instances can be distinguished. This is controlled by the new `utilization.gcp_cloud_run.use_instance_as_host` configuration option (default `true`). Set `utilization.gcp_cloud_run.include_revision_in_host` (default `false`) to `true` to report the hostname as `{K_REVISION}-{instance id}` instead. [Issue#3295](https://github.com/newrelic/newrelic-ruby-agent/issues/3295)
```
```suggestion
  The agent now detects Cloud Run and reports the GCP instance id as the hostname so individual instances can be distinguished. Before this change, all Google Cloud Run hostnames were `localhost`. This feature is controlled by the new `utilization.gcp_cloud_run.use_instance_as_host` configuration option (default `true`). Set `utilization.gcp_cloud_run.include_revision_in_host` (default `false`) to `true` to report the hostname as `{K_REVISION}-{instance id}` instead. Thank you to [@kawa-onushi](https://github.com/kawa-onushi) for suggesting this improvement. [Issue#3295](https://github.com/newrelic/newrelic-ruby-agent/issues/3295) [PR#3609](https://github.com/newrelic/newrelic-ruby-agent/pull/3609/)
```

lib/new_relic/agent/hostname.rb
hannahramadan: I didn't know what the `K` in `K_REVISION` stands for (still don't lol) so I thought it would be more clear. But def happy to take this out and just directly call ENV['K_REVISION'] where we need it ◡̈

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: Great idea! https://github.com/newrelic/newrelic-ruby-agent/pull/3609/changes/ac2adeb5e32ae28dfb300fd4e5d8bffa21619fa7

lib/new_relic/agent/hostname.rb
kaylareopelle: haha you know what, I don't either! on second thought, I like this exactly as it is. makes more sense when reading the code/don't need to document things.

CHANGELOG.md
hannahramadan: I couldn't apply this suggestion directly since it became outdated after I already updated the CHANGELOG. So I added it here https://github.com/newrelic/newrelic-ruby-agent/pull/3609/changes/eccc9bbe35e132e11462677820d43d29baf124d6

lib/new_relic/agent/hostname.rb
hannahramadan: Can create a shared helper https://github.com/newrelic/newrelic-ruby-agent/pull/3609/changes/d48f883752575bc31a28811f44dbe76a5f972332 - what do you think?

kaylareopelle (dismissed):

tannalynn (dismissed):

kaylareopelle (approved):

lib/new_relic/agent/hostname.rb
choznerol: > I didn't know what the `K` in `K_REVISION` stands for (still don't lol)

I believe the `K` is for Knative because [Cloud Run is built on top of Knative](https://knative.dev/docs/install/knative-offerings/) 😄.

PR #3611, opened 2026-07-09
chore: Initial work to add agent config schema

mvicknr:
_Before contributing, please read our [contributing guidelines](https://github.com/newrelic/newrelic-ruby-agent/blob/main/CONTRIBUTING.md) and [code of conduct](https://github.com/newrelic/.github/blob/main/CODE_OF_CONDUCT.md)._

# Overview
Adds a script to automatically generate and update a configuration schema for the agent. This PR has the initial schema generated from the script.

The script will automatically run on local build and on pushes (except main/dev). It will determine if there are config changes and if so, it will update the schema. On local, the dev should include these changes in their commit. In the PR build, if the changes aren't already made in the schema, it will make the changes and add them to the PR.

The schema version will be updated during the release process.

Both GH workflows were tested locally with act.

[schemaGeneration/README.md provides more details](https://github.com/mvicknr/newrelic-ruby-agent/blob/84fc502594a00415deb498e376579a18515064ed/.fleetControl/schemaGeneration/README.md)

Submitter Checklist:
- [x] Add new tests for your change, if applicable

# Testing
exercise your changes. 

# Reviewer Checklist
- [ ] Perform code review
- [ ] Open a separate PR to add a CHANGELOG entry

### DIFF
--- a/.fleetControl/schemaGeneration/generate_schema.rb
+++ b/.fleetControl/schemaGeneration/generate_schema.rb
@@ -1,0 +1,180 @@
+#!/usr/bin/env ruby
+# This file is distributed under New Relic's license terms.
+# See https://github.com/newrelic/newrelic-ruby-agent/blob/main/LICENSE for complete details.
+# frozen_string_literal: true
+
+require 'json'
+require 'fileutils'
+require_relative '../../lib/new_relic/agent/configuration/default_source'
+# Reused for description cleanup so the schema, the config docs, and the
+# generated newrelic.yml all strip markdown the same way (no drift).
+require_relative '../../lib/tasks/helpers/newrelicyml'
+
+# Generates a JSON Schema (Draft 2020-12) describing the New Relic Ruby agent's
+# configuration, read straight from NewRelic::Agent::Configuration::DEFAULTS.
+# This is the Ruby parallel to the Java agent's GenerateSchema.groovy, and feeds
+# the same Fleet Control consumption path (../schemas/config.json).
+#
+# Run it: `ruby .fleetControl/schemaGeneration/generate_schema.rb`
+module GenerateSchema
+  DEFAULTS = NewRelic::Agent::Configuration::DEFAULTS
+  SCHEMA_URI = 'https://json-schema.org/draft/2020-12/schema'
+  OUTPUT_PATH = File.expand_path('../schemas/config.json', __dir__)
+  Boolean = NewRelic::Agent::Configuration::Boolean
+
+  # Public settings that aren't configurable via newrelic.yml (env-only, read
+  # before the YAML loads). Reuse the existing authoritative list so this
+  # generator and the newrelic.yml generator can't drift.
+  EXCLUDE_KEYS = NewRelicYML::SKIP
+
+  # The agent accepts Array-typed settings as either a real YAML array or a
+  # single delimited string (a :transform splits it), so the schema allows both.
+  ARRAY_OR_STRING = {
+    'anyOf' => [
+      {'type' => 'array', 'items' => {'type' => 'string'}},
+      {'type' => 'string'}
+    ]
+  }.freeze
+
+  # Enum values for settings that are effectively enums but carry no :allowlist
+  # in DEFAULTS. The Java agent keeps an equivalent ENUM_OVERRIDES map. Values
+  # are verified against the code that consumes each setting, NOT the prose
+  # descriptions (which omit log_level's `fatal` and mislabel `to_app`):
+  #   - log_level:  agent_logger.rb LOG_LEVELS
+  #   - record_sql: database.rb#record_sql_method
+  #
+  # LONG-TERM FIX: these maps are a second source of truth that can drift from
+  # the code. The better fix is to add an :allowlist to each of these settings
+  # in default_source.rb. The generator already turns :allowlist into an enum
+  # automatically (see build_property), and the agent enforces :allowlist at
+  # runtime (manager.rb#enforce_allowlist) — so doing that gives real runtime
+  # validation these settings currently lack AND lets these override maps be
+  # deleted entirely.
+  BASE_ENUM_OVERRIDES = {
+    'log_level' => %w[debug info warn error fatal],
+    'transaction_tracer.record_sql' => %w[off none raw obfuscated],
+    'slow_sql.record_sql' => %w[off none raw obfuscated]
+  }.freeze
+
+  # instrumentation.* toggles. Most accept the standard set
+  # (dependency_detection.rb VALID_CONFIG_VALUES); a few are simple on/off.
+  INSTRUMENTATION_STANDARD_VALUES = %w[auto prepend chain disabled].freeze
+  INSTRUMENTATION_ONOFF_VALUES = %w[enabled disabled].freeze
+  INSTRUMENTATION_ONOFF_KEYS = %w[
+    instrumentation.excon
+    instrumentation.mongo
+    instrumentation.stripe
+    instrumentation.rails_event_logger
+  ].freeze
+
+  module_function
+
+  def generate(defaults = DEFAULTS)
+    properties = {}
+    defaults.sort_by { |key, _| key.to_s }.each do |key, spec|
+      next unless public?(spec) && !deprecated?(spec)
+      next if EXCLUDE_KEYS.include?(key)
+
+      properties[key.to_s] = build_property(spec, enum_override_for(key, spec))
+    end
+
+    {
+      '$schema' => SCHEMA_URI,
+      'title' => 'New Relic Ruby Agent Configuration',
+      'description' => 'Configuration settings for the New Relic Ruby agent. ' \
+        'Generated from DEFAULTS in lib/new_relic/agent/configuration/default_source.rb.',
+      'type' => 'object',
+      # true (not false), matching the Java agent: the agent ships new config
+      # keys every release, and a Fleet Control deployment may validate against
+      # a schema generated from an older agent. Strict validation would reject
+      # any newer key and break users who upgrade before the schema is
+      # republished. Flip to false only alongside a lockstep republish process.
+      'additionalProperties' => true,
+      'properties' => properties
+    }
+  end
+
+  def build_property(spec, enum_override = nil)
+    property = spec[:type] == Array ? ARRAY_OR_STRING.dup : {'type' => json_type(spec[:type])}
+
+    description = description_for(spec)
+    property['description'] = description if description
+
+    default = default_for(spec)
+    property['default'] = default unless default == OMIT
+
+    enum = spec[:allowlist] ? spec[:allowlist].map(&:to_s) : enum_override
+    property['enum'] = enum if enum
+    property['writeOnly'] = true if spec[:exclude_from_reported_settings]
+
+    property
+  end
+
+  # Enum values for a setting that has no :allowlist. Returns nil when the
+  # setting isn't an enum. Only applies to string/symbol settings; :allowlist
+  # always wins over an override.
+  def enum_override_for(key, spec)
+    return nil if spec[:allowlist]
+    return nil unless [String, Symbol].include?(spec[:type])
+
+    key = key.to_s
+    return BASE_ENUM_OVERRIDES[key] if BASE_ENUM_OVERRIDES.key?(key)
+    return nil unless key.start_with?('instrumentation.')
+
+    INSTRUMENTATION_ONOFF_KEYS.include?(key) ? INSTRUMENTATION_ONOFF_VALUES : INSTRUMENTATION_STANDARD_VALUES
+  end
+
+  def to_json_string(defaults = DEFAULTS)
+    # JSON.pretty_generate emits the array/object newline separators even for
+    # empty containers, leaving an ugly blank indented line between the
+    # brackets (e.g. an empty default []). Collapse those back to [] / {}.
+    JSON.pretty_generate(generate(defaults))
+      .gsub(/\[\s*\n\s*\]/, '[]')
+      .gsub(/\{\s*\n\s*\}/, '{}')
+  end
+
+  def write_file(defaults = DEFAULTS, path = OUTPUT_PATH)
+    FileUtils.mkdir_p(File.dirname(path))
+    File.write(path, "#{to_json_string(defaults)}\n")
+    path
+  end
+
+  # --- helpers ----------------------------------------------------------------
+
+  def public?(spec)
+    spec[:public] == true
+  end
+
+  def deprecated?(spec)
+    spec[:deprecated] == true
+  end
+
+  def json_type(type)
+    case type.to_s
+    when Boolean.to_s then 'boolean'
+    when 'Integer' then 'integer'
+    when 'Float' then 'number'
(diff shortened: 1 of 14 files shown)


.fleetControl/schemas/config.json
```diff
@@ -0,0 +1,2043 @@
...
+        "chain",
+        "disabled"
+      ]
+    },
+    "labels": {
+      "type": "string",
+      "description": "A dictionary of label names and values that will be applied to the data sent from this agent. May also be expressed as a semicolon-delimited ; string of colon-separated : pairs. For example, Server:One;Data Center:Primary.",
+      "default": ""
+    },
+    "license_key": {
```
mvicknr, line 1307: The java agent has this as required, should we do that here as well?

.fleetControl/schemas/config.json
```diff
@@ -0,0 +1,2043 @@
...
+      "type": "boolean",
+      "description": "If false, LLM instrumentation (OpenAI only for now) will not capture input and output content on specific LLM events.\n\n\tThe excluded attributes include:\n\t- content from LlmChatCompletionMessage events\n\t- input from LlmEmbedding events\n\n\tThis is an optional security setting to prevent recording sensitive data sent to and received from your LLMs.",
+      "default": true
+    },
+    "allow_all_headers": {
+      "type": "boolean",
+      "description": "If true, enables capture of all HTTP request headers for all destinations.",
+      "default": false
+    },
+    "app_name": {
```
mvicknr, line 48: The java agent has this as required, should we do that here as well?

.fleetControl/schemaGeneration/generate_schema.rb
```diff
@@ -0,0 +1,239 @@
...
+  }.freeze
+
+  # Enum values for settings that are effectively enums but carry no :allowlist
+  # in DEFAULTS. Values are verified against the code that consumes each setting,
+  # NOT the prose descriptions (which omit log_level's `fatal` and mislabel
+  # `to_app`):
+  #   - log_level:  agent_logger.rb LOG_LEVELS
+  #   - record_sql: database.rb#record_sql_method
+  #
+  # LONG-TERM FIX: these maps are a second source of truth that can drift from
```
hannahramadan: Thank you for this! I'll leave this alone for now, but have made [a note](https://github.com/newrelic/newrelic-ruby-agent/issues/3613) about potentially adding an allowlist for these configs and then updating this file.

.fleetControl/schemas/config.json
hannahramadan: The agent will resolve to the env as the app_name if one isn't provided. If there is no env somehow, the agent won't start and the health check file will send error 'NR-APM-005' and write 'Missing application name in agent configuration'. Do we think that is enough?

.fleetControl/schemas/config.json
```diff
@@ -0,0 +1,2043 @@
...
...Image class and a notify class method for a User class, both within a MyCompany module namespace:\n\n\t\t\tmodule MyCompany\n\t\t\tclass Image\n\t\t\t\tdef render_png\n\t\t\t\t\t# code to render a PNG\n\t\t\t\tend\n\t\t\tend\n\n\t\t\tclass User\n\t\t\t\tdef self.notify\n\t\t\t\t\t# code to notify users\n\t\t\t\tend\n\t\t\tend\n\t\tend\n\t\n\n\tGiven that source code, the newrelic.yml config file might request instrumentation for both of these methods like so:\n\n\t\t\tautomatic_custom_instrumentation_method_list:\n\t\t\t- MyCompany::Image#render_png\n\t\t\t- MyCompany::User.notify\n\t\n\n\tThat configuration example uses YAML array syntax to specify both methods. Alternatively, you can use a comma-delimited string:\n\n\t\t\tautomatic_custom_instrumentation_method_list: 'MyCompany::Image#render_png, MyCompany::User.notify'\n\t\n\n\tWhitespace around the comma(s) in the list is optional. When configuring the agent with a list of methods via the NEW_RELIC_AUTOMATIC_CUSTOM_INSTRUMENTATION_METHOD_LIST environment variable, use this comma-delimited string format:\n\n\t\t\texport NEW_RELIC_AUTOMATIC_CUSTOM_INSTRUMENTATION_METHOD_LIST='MyCompany::Image#render_png, MyCompany::User.notify'",
```
hannahramadan: How does Fleet Control display descriptions? will `\n\t` be translated into a newline + tab or render as is?

.fleetControl/schemas/config.json
hannahramadan: We don't actually block the agent from starting when the license_key is missing, we just of course don't send data to New Relic. So I don't think it is a must to make it required.

lib/tasks/config.rake
```diff
@@ -42,5 +42,18 @@ namespace :newrelic do
       format = args[:format] || 'text'
       output(format)
     end
+
+    desc 'Generate the Fleet Control JSON Schema for agent configuration'
+    task :schema => [] do
+      require File.expand_path(File.join(File.dirname(__FILE__), '..', '..', '.fleetControl', 'schemaGeneration', 'generate_schema.rb'))
+      GenerateSchema.write_file
```
hannahramadan: should this be `GenerateSchema.run` instead? The README sounds like running the file and rake task are the same, and the file calls `GenerateSchema.run`

.fleetControl/schemas/config.json
mvicknr: Good call - I'll check into this

.fleetControl/schemas/config.json
mvicknr: Yes, I think this is enough. I also looked into the k8s fleet scenario and the operator uses the deployment name if the app name is not set in the CR.

lib/tasks/config.rake
mvicknr: Yes, thank you for catching that - updated

.fleetControl/schemas/config.json
mvicknr: The UI won't use the formatting but when looking at it, it wouldn't be very useful without the newlines and tabs. I am chatting with dev and product about it. If we don't get a consensus in a reasonable timeframe, I can add a ticket to revisit it once we have a plan.

mvicknr: @hannahramadan take a look at the most recent changes and see if it is a good solution - it narrows it down to a schema field description instead of having the extra detail that is more akin to docs

.fleetControl/schemaGeneration/tests/bump_schema_version_test.rb
```diff
@@ -0,0 +1,80 @@
+# This file is distributed under New Relic's license terms.
```
kaylareopelle, line 1: I'm not sure that these tests are running as part of the CI. If they aren't, could you move them somewhere so that they do? A good example could be: https://github.com/newrelic/newrelic-ruby-agent/pull/3508

.fleetControl/schemaGeneration/schema_diff.rb
```diff
@@ -0,0 +1,199 @@
...
+  def type_changes(old_prop, new_prop, path)
+    old_sig = type_signature(old_prop)
+    new_sig = type_signature(new_prop)
+    return [] if old_sig == new_sig
+
+    [change(path, 'type_changed', "#{describe_type(old_sig)} -> #{describe_type(new_sig)}")]
+  end
+
+  def describe_type(signature)
+    signature.is_a?(String) ? signature : 'array-or-string'
```
kaylareopelle: Why is this only `array-or-string` as the fallback? We have other datatypes in the configs too. Is that a JSON requirement?

.fleetControl/schemaGeneration/schema_diff.rb
```diff
@@ -0,0 +1,199 @@
...
+    changes.concat(description_changes(old_prop, new_prop, path))
+    changes
+  end
+
+  # Highest-severity bump implied by the changes.
+  def recommend_bump(changes)
+    severities = changes.map { |c| c[:severity] }
+    return 'major' if severities.include?('breaking')
+    return 'minor' if severities.include?('additive')
+    return 'patch' if severities.include?('cosmetic')
```
kaylareopelle: More than `cosmetic`, I think `fix` probably makes more sense for us. If we have a release with only fixes and no new features, it'll be a patch.

.fleetControl/schemas/config.json
```diff
@@ -0,0 +1,2047 @@
...
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        {
+          "type": "string"
+        }
+      ],
+      "description": "An array of ActiveSupport custom event names to subscribe to and instrument. For example, - one.custom.event - another.event - a.third.event",
```
kaylareopelle: Hi @mvicknr! The Ruby team was chatting about the descriptions today. We found this one, and are curious about how it will show up for customers in the Fleet Control UI. Would something like this make sense in the Fleet Control UI even though the dashes aren't on newlines? 

We're not sure if we need to clean up our descriptions further to fit in this new environment. We also came across [this](https://github.com/newrelic/newrelic-ruby-agent/pull/3611/changes#diff-cca8b94a86025150d504c81e219a2225ea10b5c027225076ef8c02e5c2f54e65R271-R277) and are wondering if just the first lines need to fit for the new environment.

mvicknr: I shared a video of the UI in your team channel. For the 2nd question - the schema generator takes the first part since it is the description itself. Subsequent lines tend to have code examples or other formatted things that are in addition to the description. This was a tweak I added for your agent specifically - here is the discussion with product: https://newrelic.slack.com/archives/C0B7QF1DG5P/p1784291759415399

mvicknr: @kaylareopelle once the team gets the descriptions updated, I can remove/tweak the code to read more than just the first line

kaylareopelle: Hi @mvicknr, our current plan is to update the descriptions so that the first line is clean for the fleet control UI. We'll use the later lines for our code examples, etc. @hannahramadan is getting started on this. We'll have more for you soon!

.fleetControl/schemas/config.json
```diff
@@ -0,0 +1,2047 @@
...
+      "type": "string",
+      "description": "Sets the level of detail of log messages. Possible log levels, in increasing verbosity, are: error, warn, info or debug.",
+      "default": "info",
+      "enum": [
+        "debug",
+        "info",
+        "warn",
+        "error",
+        "fatal"
+      ]
```
kaylareopelle, line 1333: @mvicknr - Are the enums case-sensitive for fleet control? For example, the agent will downcase any log level received, so `DEBUG` and `Debug` are also both valid.

mvicknr: The UI will display a dropdown choice of the values in the enum array. For now, this is all based on a clean install so it will be fine but when we get to brownfield scenarios this will be more important. Do you know if all APM agents are case-insensitive? Is it just this field or all?

kaylareopelle: Good to know. I imagine each APM agent is different. 

For the Ruby agent, it's inconsistent about whether the agent treats the value as case-insensitive. Log levels have a specific workaround. Booleans are case-insensitive. Most enums have specific casing requirements (ex. `transaction_tracer.record_sql` requires exact strings).

mvicknr: dotnet purposefully doesn't use an enum here for this reason: https://github.com/newrelic/newrelic-dotnet-agent/blob/main/src/Agent/NewRelic/Agent/Core/Config/Configuration.xsd#L395

we could do the same here and just leave the levels in the comments for the user to reference

kaylareopelle: Ah that's helpful context. @hannahramadan, when we opened #3613 we planned to include `:log_level`, but after taking this into account, we might want to leave that config out and only add the allowlist for exact string matches like the `record_sql` options. What do you think?

hannahramadan: Chatted offline and we will make sure the allowlist is case insensitive ◡̈

mvicknr: @kaylareopelle Forgive me, I'm not sure I follow. Do you need me to change this schema to use a string here instead of the enum as it is now?

kaylareopelle: @mvicknr, yes, the log_level should be a string type now and not an enum so that we can support the current behavior of case-insensitive levels.

.fleetControl/schemas/config.json
```diff
@@ -7,7 +7,7 @@
   "properties": {
     "active_record_use_table_name": {
       "type": "boolean",
-      "description": "If true, the agent will use the Active Record model's table name instead of the class name when naming Active Record metrics, spans, and transaction trace segments. This can reduce cardinality when multiple models share a database table. Defaults to false.",
```
mvicknr: @hannahramadan  the config.json is generated so the changes have to be made in the source file: lib/new_relic/agent/configuration/default_source.rb

mvicknr: nevermind, I see they were made in the dev branch

.fleetControl/schemas/config.json
mvicknr: fixed

mvicknr:
@kaylareopelle @hannahramadan I think everything is done on this PR. One caveat is that the `agent_config_schema.yml` workflow will not work for forks. I think this is ok for your team since you will be making the config updates.

kaylareopelle:
> I think everything is done on this PR. One caveat is that the `agent_config_schema.yml` workflow will not work for forks. I think this is ok for your team since you will be making the config updates.

Hi @mvick, it looks good to me. @hannahramadan is working on #3613 and we were planning to get that merged, pull out the enum workarounds in this PR, and then merge this PR. Does that work for you?

hannahramadan:
@mvicknr! We merged #3613 and I updated this PR to remove ENUM_OVERRIDES and re-generate the schema file https://github.com/newrelic/newrelic-ruby-agent/pull/3611/changes/0384a1771b9626287ec04b96b43e4fb80f60d377. 

Question - since we edited all allowlist configs to now be case-insensitive, if someone types in a valid value, but in a case that isn't on that generated JSON enum list, will it still work with fleet control?

mvicknr:
> @mvicknr! We merged #3613 and I updated this PR to remove ENUM_OVERRIDES and re-generate the schema file [0384a17](https://github.com/newrelic/newrelic-ruby-agent/commit/0384a1771b9626287ec04b96b43e4fb80f60d377).
> 
> Question - since we edited all allowlist configs to now be case-insensitive, if someone types in a valid value, but in a case that isn't on that generated JSON enum list, will it still work with fleet control?

If an enum is given it must match the case in the enum so for case sensitive items. See my convo with Kayla: https://github.com/newrelic/newrelic-ruby-agent/pull/3611#discussion_r3760217557

hannahramadan:
Thanks @mvicknr! That makes sense and we are leaving the enums in the json file. We are reviewing one more time!

kaylareopelle (dismissed):

kaylareopelle (approved):

hannahramadan (approved):

PR #3612, opened 2026-07-09
fix for 7.2+ multi db explain plans

### LINKED ISSUE #3610: Explain plans use the default connection pool for all databases in multi-DB Rails apps (Rails >= 7.2), and can leave pooled connections in aborted transactions
## Description

Since the Rails 7.2 explain-plan repair (#2922 / PR #2940), `NewRelic::Agent::Database.explain_this` collects every explain plan via:

```ruby
def explain_this_using_with_connection(statement)
  ::ActiveRecord::Base.with_connection do |conn|
    conn.exec_query("EXPLAIN #{statement.sql}", "Explain #{statement.name}", statement.binds)
  end
end
```

`ActiveRecord::Base.with_connection` always checks out from the **default (primary) pool**. In a multi-database Rails app, queries that ran on a secondary database (via an abstract class + `connects_to`) get EXPLAINed against the primary database, where their tables don't exist.

The pre-7.2 path (`explain_this_using_adapter_connection`) resolved the connection from `statement.config`, i.e. the correct database. That routing was lost in the 7.2+ path.

## Impact we observed (production)

Rails 8.0.5, two PostgreSQL databases (primary + a read-only secondary on a separate server).

1. **Every explain of a secondary-DB query fails.** Agent log, at harvest time:
   ```
   ERROR : Couldn't fetch the explain plan for statement: PG::UndefinedTable: ERROR:  relation "<secondary_db_table>" does not exist
   ```
   So for multi-DB apps the feature silently produces no plans for secondary-DB queries (and burns a primary-pool checkout per attempt).

2. **Much worse: primary-pool connections ended up stuck in aborted transactions.** We had storms of
   `PG::InFailedSqlTransaction: ERROR: current transaction is aborted, commands ignored until end of transaction block`
   on requests completely unrelated to the explained queries — peaking at ~13,000 errors/hour. Affected connections stayed poisoned until process restart — over a ~217k-line log window we counted 10,365 `BEGIN` vs 2 `ROLLBACK`. We also saw the agent's own explain attempts hitting already-poisoned checkouts:
   ```
   ERROR : Couldn't fetch the explain plan for statement: PG::InFailedSqlTransaction: ERROR:  current transaction is aborted...
   ```

   Timing correlation from two initially-healthy Puma workers: each logged its first wrong-database explain error, and began failing primary-DB writes with `PG::InFailedSqlTransaction` minutes later (worker A: first explain error 12:57:29 → first poisoned write 12:58:57; worker B: 12:56:59 → 13:02:20). Workers with no explain errors in the window showed no poisoning onset.

   We have not pinned the exact low-level mechanism by which the failed EXPLAIN leaves an open aborted transaction on the checked-out connection (Rails' pool checkin does not roll back transactions it doesn't know about, so once it happens the connection stays poisoned indefinitely). We're confident in the correlation and in the wrong-pool routing; happy to run diagnostics if it helps.

## Expected behavior

- Explain plans should run against the database the query actually executed on, or be skipped when that database can't be resolved.
- A failed EXPLAIN should never be able to leave a pooled app connection in a bad transaction state (e.g. verify/reset connection state in an `ensure`).

## Environment

- newrelic_rpm 9.18.0
- Rails 8.0.5 (multi-DB: `ApplicationRecord` primary + a second abstract base with `connects_to`)
- Ruby 3.3.x, PostgreSQL (both DBs), Puma (cluster mode), Heroku

## Workaround

```yaml
transaction_tracer:
  explain_enabled: false
slow_sql:
  explain_enabled: false
```

Verified in our production: deploying this dropped both error classes (`UndefinedTable` explain failures and the `InFailedSqlTransaction` storms) from ~13k/hour peak to zero, flat for 24+ hours under the same traffic that previously triggered them.

tannalynn:
changes the rails 7.2 explain plan logic to work similar to how it does on < 7.2. 

Prior to 7.2 we created our own connection for the explain plan, but in 7.2+ we used a connection from the shared connection pool (which, if there is multiple dbs, these are only to the primary db). So the explain plan would fail on multi db apps when explaining something from a different db, and then the connection just gets returned to the pool, all messed up. And then it gets reused, causing problems.

So now we don't use the connection pool, we do what we used to do and make our own connection (also we make the connection to the correct db if there are multiple). 
And I added a safety net for errors on the connection. Even though we aren't using the pool anymore so the connection shouldn't be shared anyways, still good to make sure we close things out well.

resolves https://github.com/newrelic/newrelic-ruby-agent/issues/3610

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## dev
 
-- **Bugfix: Explain plans could target the wrong database and leave connections in a bad state in multi-database Rails apps (Rails >= 7.2)**
+- **Bugfix: Explain plans could target the wrong database in multi-database Rails apps (Rails >= 7.2)**
 
   On Rails 7.2+, the agent gathered explain plans using a connection from the app's default/shared pool rather than a dedicated one. This primarily affected multi-database apps. Explain plans could be generated against the wrong database, and a failed explain could leave a shared connection in a bad state, affecting unrelated requests. The agent now uses its own dedicated connection for explain plans, as it did before Rails 7.2, and resets or discards that connection whenever an explain attempt fails, so a bad connection is never reused. Thank you, [@masiafrest](https://github.com/masiafrest) for the detailed report! [Issue#3610](https://github.com/newrelic/newrelic-ruby-agent/issues/3610) [PR#3612](https://github.com/newrelic/newrelic-ruby-agent/pull/3612)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - **Bugfix: Explain plans could target the wrong database and leave connections in a bad state in multi-database Rails apps (Rails >= 7.2)**
 
-  On Rails 7.2+, the agent gathered explain plans using a connection from the app's default/shared pool rather than a dedicated one. This primarily affected multi-database apps, where explain plans could be generated against the wrong database, and a failed explain could leave a shared connection in a bad state, affecting unrelated requests. The agent now uses its own dedicated connection for explain plans, as it did before Rails 7.2, and resets or discards that connection whenever an explain attempt fails, so a bad connection is never reused. Thank you, [@masiafrest](https://github.com/masiafrest) for the detailed report! [Issue#3610](https://github.com/newrelic/newrelic-ruby-agent/issues/3610) [PR#3612](https://github.com/newrelic/newrelic-ruby-agent/pull/3612)
+  On Rails 7.2+, the agent gathered explain plans using a connection from the app's default/shared pool rather than a dedicated one. This primarily affected multi-database apps. Explain plans could be generated against the wrong database, and a failed explain could leave a shared connection in a bad state, affecting unrelated requests. The agent now uses its own dedicated connection for explain plans, as it did before Rails 7.2, and resets or discards that connection whenever an explain attempt fails, so a bad connection is never reused. Thank you, [@masiafrest](https://github.com/masiafrest) for the detailed report! [Issue#3610](https://github.com/newrelic/newrelic-ruby-agent/issues/3610) [PR#3612](https://github.com/newrelic/newrelic-ruby-agent/pull/3612)
 
 ## v10.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - **Bugfix: Explain plans could target the wrong database and leave connections in a bad state in multi-database Rails apps (Rails >= 7.2)**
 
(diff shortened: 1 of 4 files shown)


lib/new_relic/agent/database.rb
```diff
@@ -91,38 +91,42 @@ def get_connection(config, &connector)
...
-        if use_execute
-          connection.execute("EXPLAIN #{statement.sql}")
-        else
-          connection.exec_query("EXPLAIN #{statement.sql}", "Explain #{statement.name}", statement.binds)
-        end
+      # Resets the connection on error so it isn't reused while in a bad state
+      def run_explain(config, connection)
+        yield
+      rescue
+        ConnectionManager.instance.reset_or_discard_connection(config, connection)
```
hannahramadan, line 122: I like this new safety net

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # New Relic Ruby Agent Release Notes
 
+## dev
+
+- **Bugfix: Explain plans could target the wrong database and leave connections in a bad state in multi-database Rails apps (Rails >= 7.2)**
```
```suggestion
- **Bugfix: Explain plans could target the wrong database in multi-database Rails apps (Rails >= 7.2)**
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
...
+
+  On Rails 7.2+, the agent gathered explain plans using a connection from the app's default/shared pool rather than a dedicated one. This primarily affected multi-database apps, where explain plans could be generated against the wrong database, and a failed explain could leave a shared connection in a bad state, affecting unrelated requests. The agent now uses its own dedicated connection for explain plans, as it did before Rails 7.2, and resets or discards that connection whenever an explain attempt fails, so a bad connection is never reused. Thank you, [@masiafrest](https://github.com/masiafrest) for the detailed report! [Issue#3610](https://github.com/newrelic/newrelic-ruby-agent/issues/3610) [PR#3612](https://github.com/newrelic/newrelic-ruby-agent/pull/3612)
```
```suggestion
  On Rails 7.2+, the agent gathered explain plans using a connection from the app's default/shared pool rather than a dedicated one. This primarily affected multi-database apps. Explain plans could be generated against the wrong database, and a failed explain could leave a shared connection in a bad state, affecting unrelated requests. The agent now uses its own dedicated connection for explain plans, as it did before Rails 7.2, and resets or discards that connection whenever an explain attempt fails, so a bad connection is never reused. Thank you, [@masiafrest](https://github.com/masiafrest) for the detailed report! [Issue#3610](https://github.com/newrelic/newrelic-ruby-agent/issues/3610) [PR#3612](https://github.com/newrelic/newrelic-ruby-agent/pull/3612)
```

kaylareopelle (dismissed):

hannahramadan (approved):

kaylareopelle (approved):

PR #3615, opened 2026-07-17
Cap segment artifacts on very long running transactions

tannalynn:


lib/new_relic/agent/transaction/datastore_segment.rb
```diff
@@ -134,6 +134,8 @@ def add_backtrace_parameter
         end
 
         def notice_sql_statement
+          return if record_on_finish?
```
tannalynn, line 137: this one doesn't use the config because actually, the agent specs say we shouldn't have been doing this at all anyways (when the segment limit is reached), so this is actually just the correct behavior.

test/new_relic/agent/transaction/abstract_segment_test.rb
```diff
@@ -295,6 +295,81 @@ def test_children_time_ranges_do_exist
...
+
+        # Regression test: gating must be per-segment (record_on_finish?), not
+        # transaction-wide, or a within-limit sibling would lose accuracy just
+        # because another segment tripped the limit.
+        def test_sibling_within_limit_keeps_accurate_exclusive_duration_despite_sibling_over_limit
+          # limit=3 accounts for in_transaction's root segment + segment_a + segment_b.
+          with_config(:'transaction_tracer.limit_segments' => 3, :'transaction_tracer.cap_segment_artifacts' => true) do
+            segment_a = segment_b = segment_c = nil
+            in_transaction do |txn|
+              segment_a = BasicSegment.new('segment_a')
```
kaylareopelle, line 344: TIL BasicSegment!

test/new_relic/agent/transaction/datastore_segment_test.rb
```diff
@@ -578,6 +578,54 @@ def test_notice_sql
           end
         end
 
+        def test_notice_sql_not_recorded_when_segment_over_limit
```
kaylareopelle, line 581: I'm not sure I'm following this test. How are the segments over the limit here?

tannalynn: oh it's setting segment.record_on_finish = true forcefully (on 588) instead. Functionally it's the same because that is the same check we do in the code to make the decision, since that uses the segment limit. If you think it's weird or too confusing i'm cool with changing it to do the config way.

kaylareopelle: Ahh gotcha. Thanks for clarifying! I think I'd rather see it done the config way like the test below. Or, possibly change the title to be `when_record_on_finish_true` so it's more aligned to the content. What do you think?

kaylareopelle (changes requested):
Just thought of something else -- we might want a changelog entry for this.

test/new_relic/agent/transaction/abstract_segment_test.rb
```diff
@@ -295,6 +295,81 @@ def test_children_time_ranges_do_exist
...
+          assert_equal [[1.0, 2.0]], segment.instance_variable_get(:@children_timings)
+        end
+
+        # Regression test: gating must be per-segment (record_on_finish?), not
+        # transaction-wide, or a within-limit sibling would lose accuracy just
+        # because another segment tripped the limit.
+        def test_sibling_within_limit_keeps_accurate_exclusive_duration_despite_sibling_over_limit
+          # limit=3 accounts for in_transaction's root segment + segment_a + segment_b.
+          with_config(:'transaction_tracer.limit_segments' => 3, :'transaction_tracer.cap_segment_artifacts' => true) do
+            segment_a = segment_b = segment_c = nil
```
hannahramadan, line 342: TIL this way to assign multiple variables! V cool

hannahramadan (commented):
Should we also add a CHANGELOG for this?

test/new_relic/agent/transaction/datastore_segment_test.rb
tannalynn: updated!

tannalynn:
> Just thought of something else -- we might want a changelog entry for this.

oh yeah! Added now! I did 2 since one part was more of a bugfix for something we were supposed to always be doing.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## dev
 
+- **Feature: Add transaction_tracer.cap_segment_artifacts configuration option**
+
+  Long-running transactions with many segments can cause continuously increasing memory usage for the lifetime of the transaction. The agent now offers an opt-in `transaction_tracer.cap_segment_artifacts` configuration option (default `false`, local-only). When enabled, once [`transaction_tracer.limit_segments`](https://docs.newrelic.com/docs/apm/agents/ruby-agent/configuration/ruby-agent-configuration/#transaction_tracer-limit_segments) is reached, the agent also stops recording exclusive time for any segments created afterward in that transaction, reducing memory usage at the cost of less accurate timing data for the transaction.[PR#3615](https://github.com/newrelic/newrelic-ruby-agent/pull/3615)
```
kaylareopelle: What do you mean by `local-only`?

CHANGELOG.md
```diff
@@ -12,6 +16,10 @@
 
   The agent now detects Cloud Run and reports the GCP instance id as the hostname so individual instances can be distinguished. Before this change, all Google Cloud Run hostnames were `localhost`. This feature is controlled by the new `utilization.gcp_cloud_run.use_instance_as_host` configuration option (default `true`). Set `utilization.gcp_cloud_run.include_revision_in_host` (default `false`) to `true` to report the hostname as `{K_REVISION}-{instance id}` instead, where [`K_REVISION`](https://docs.cloud.google.com/run/docs/container-contract#env-vars) is the Cloud Run revision name. Thank you to [@kawa-onushi](https://github.com/kawa-onushi) for suggesting this improvement. [Issue#3295](https://github.com/newrelic/newrelic-ruby-agent/issues/3295) [PR#3609](https://github.com/newrelic/newrelic-ruby-agent/pull/3609/)
 
+- **Bugfix: Slow SQL recorded after transaction_tracer.limit_segments exceeded**
```
kaylareopelle: I think I understand that it's fixing the statement after the colon, but maybe being more explicit here would help? 

```suggestion
- **Bugfix: Slow SQL no longer recorded after transaction_tracer.limit_segments exceeded**
```

CHANGELOG.md
tannalynn: I meant to remove that part (bc its weird claude stuff) when i was editing it, but i must have accidentally pressed undo one too many times at some point lol. I'll fix that. (0ef2b0e7fe3cd8d5eee955189150721225a03a76)

```suggestion
  Long-running transactions with many segments can cause continuously increasing memory usage for the lifetime of the transaction. The agent now offers an opt-in `transaction_tracer.cap_segment_artifacts` configuration option (defaults to `false`). When enabled, once [`transaction_tracer.limit_segments`](https://docs.newrelic.com/docs/apm/agents/ruby-agent/configuration/ruby-agent-configuration/#transaction_tracer-limit_segments) is reached, the agent also stops recording exclusive time for any segments created afterward in that transaction, reducing memory usage at the cost of less accurate timing data for the transaction.[PR#3615](https://github.com/newrelic/newrelic-ruby-agent/pull/3615)
```

kaylareopelle (approved):

PR #3616, opened 2026-07-17
fix intermittent CI failures

tannalynn:
it started with the bunny fix, but i saw like idk 3 intermittent failures before 11am so i just needed to do something. so I went through a bunch of our old failed jobs and decided to just try and fix a all of the recent intermittent failures i could find. lol. I found some of these several times (sidekiq i'm looking at you), but i know there are other ones that are less common that i didn't see today.

Fixed:
- the bunny/vendor log one
- the mock socket one
- the sidekiq timeout one
- one where the thread count changes but its not actually related to the harvest thread
- one that just needed the harvesters cleaned out before the test ran

### DIFF
--- a/lib/new_relic/agent/utilization/pcf.rb
+++ b/lib/new_relic/agent/utilization/pcf.rb
@@ -17,7 +17,7 @@ module NewRelic
             return false unless pcf_keys_present?
 
             process_response(ENV)
-          rescue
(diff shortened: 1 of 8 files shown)


kaylareopelle (approved):
Let's give it a try! Thanks for the investigation.

PR #3617, opened 2026-07-21
Continuous Profiling

tannalynn:
Implements continuous profiling

- We're doing the otel format for profiles, not pprof. Also, we have to convert the stackprof output into the otel profile, so that's a lot. 

- We're using dependency detection to control the profilers ability to start, so it won't ever accidentally start if the required gems aren't there (stackprof and protobuf). 

- Agent commands: Left this in but since the command doesn't exist yet it's unreachable. Also, not sure what they will end up actually making the name of the command since it's going to be done later, but I have it as (start/stop)_continuous_profiling. 

- Perfverse changes: adding some logging stuff for profiling, (also a couple of weird things changed because of how i at one point had done perfverse changes on this branch and then had to move everything off of it.)

[agent spec](https://source.datanerd.us/agents/agent-specs/blob/main/Profiling.md)


test/new_relic/agent/pipe_channel_manager_test.rb
```diff
@@ -11,7 +11,7 @@ class NewRelic::Agent::PipeChannelManagerTest < Minitest::Test
   include TransactionSampleTestHelper
 
   def setup
-    @test_config = {:monitor_mode => true}
+    @test_config = {:monitor_mode => true, :disable_harvest_thread => true}
```
tannalynn, line 14: this is just bc these tests fail locally if you have a license key env var set, because then the harvest thread tries to do stuff. (but also they only fail if you run just the file, not all tests)

kaylareopelle (dismissed):

hannahramadan (dismissed):

CHANGELOG.md
```diff
@@ -1,5 +1,32 @@
...
+
+  | Configuration name | Default | Behavior |
+  | ------------------ | ------- | -------- |
+  | profiling.enabled | `false` | If `true`, the agent collects and reports continuous profiling data |
+  | profiling.include | `cpu` | What to sample: `cpu` for CPU time, or `object` for object allocations |
+  | profiling.sample_period | `0.01` | Seconds between stack samples. Only used when `profiling.include` is `cpu`. Must be between 0.000001 and 0.999999 |
+  | profiling.object_allocation_interval | `10000` | Object allocations between stack samples. Only used when `profiling.include` is `object`. Must be between 1000 and 999999 |
+  | profiling.delay | `0` | Milliseconds to wait before profiling starts. `0` starts immediately |
+  | profiling.duration | `0` | Milliseconds to profile before stopping automatically. `0` profiles until the process exits |
+
```
```suggestion
  | profiling.enabled | `false` | If `true`, the agent collects and reports continuous profiling data. |
  | profiling.include | `cpu` | What to sample: `cpu` for CPU time, or `object` for object allocations. |
  | profiling.sample_period | `0.01` | Seconds between stack samples. Only used when `profiling.include` is `cpu`. Must be between 0.000001 and 0.999999. |
  | profiling.object_allocation_interval | `10000` | Object allocations between stack samples. Only used when `profiling.include` is `object`. Must be between 1000 and 999999. |
  | profiling.delay | `0` | Milliseconds to wait before profiling starts. `0` starts immediately. |
  | profiling.duration | `0` | Milliseconds to profile before stopping automatically. `0` profiles until the process exits. |

```

test/perfverse/newrelic.yml
```diff
@@ -3,6 +3,7 @@ common: &default_settings
   log_level: info
   log_file_path: STDOUT
 
+
```
kaylareopelle:
```suggestion
```

PR #3619, opened 2026-07-21
Prototype: Add OpenTelemetry env var config source

kaylareopelle:
This allows certain OpenTelemetry environment variables to be treated with a precedence just below New Relic environment variables.

There is also a log that will print out a debug statement for all environment variables that begin with `OTEL_` that were not explicitly applied.

### DIFF
--- a/lib/new_relic/agent/configuration/open_telemetry_source.rb
+++ b/lib/new_relic/agent/configuration/open_telemetry_source.rb
@@ -31,7 +31,7 @@ module NewRelic
         end
 
         def has_key?(key)
-          @otel_values.key?(key)
+          @otel_values.key?(key) && otel_enabled?
         end
 
         def [](key)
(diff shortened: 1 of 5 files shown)


lib/new_relic/agent/configuration/open_telemetry_source.rb
```diff
@@ -0,0 +1,65 @@
...
+            raw = ENV[env_var]
+            next unless raw
+
+            @otel_values[config_key] = transform(config_key, raw)
+          end
+
+          # optional in the spec
+          log_ignored_keys
+
+          super({})
```
tannalynn, line 30: how come is this `{}` instead of the configs?

kaylareopelle: We have to check to make sure the hybrid agent is enabled before we apply the OpenTelemetry Source configuration values, so we save the values in `@otel_values` and bring them into the stack during the `has_key?` checks that are part of the `fetch` in `Configuration::Manager`. Since the class inherits from DottedHash > Hash, we have to make sure a hash-like object is returned on initialization.

Down to refactor or look at another approach too. It does look kind of strange, so maybe a comment is in order?

tannalynn: Yeah maybe a comment would be good, just to explain why its a little weird. but i think that makes sense and seems reasonable. ty!

PR #3621, opened 2026-07-23
Update config docs workflow

### LINKED ISSUE #2145: Exit config docs workflow early if there are no changes
Currently, when the config docs workflow runs and the generated configuration document is the same as the published configuration document, the workflow will fail. 

Ideally, we'd like the workflow to exit early without failing if the generated configuration document is the same as the published document.

Slack thread for context: https://newrelic.slack.com/archives/GGXADQ9F1/p1689203589622659 

Some things to consider:
* The `delete_branch_on_fail` job will fail if the config doc has no changes because a branch will not be created

kaylareopelle:
* Add a trigger to run on Release, just like Lambda
* Add a step to compare the generated doc with the existing doc to look for a diff before creating a branch
* Gate branch and PR creation to the doc diff result

Resolves #2145

### DIFF
--- a/.github/workflows/config_docs.yml
+++ b/.github/workflows/config_docs.yml
@@ -1,6 +1,9 @@
 name: Update Config Docs
 
 on:
+  workflows: ["Release"]
+  types:
+    - completed
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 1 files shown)


tannalynn (approved):
😎

PR #3626, opened 2026-07-31
Fix test failures

tannalynn:
- redis 6 came out and does an extra thing now, so we needed to change our expectations for redis 6+
- fixed 2 intermittent failures I've seen recently

### DIFF
--- a/test/multiverse/suites/redis/Envfile
+++ b/test/multiverse/suites/redis/Envfile
@@ -6,6 +6,7 @@ instrumentation_methods :chain, :prepend
 
 REDIS_VERSIONS = [
   [nil],
+  ['5.4.1'],
   ['4.8.0'],
   ['3.3.0']
 ]
--- a/test/multiverse/suites/redis/Envfile
+++ b/test/multiverse/suites/redis/Envfile
@@ -4,13 +4,6 @@
 
 instrumentation_methods :chain, :prepend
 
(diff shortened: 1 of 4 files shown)


test/new_relic/agent/instrumentation/rails/action_dispatch_subscriber.rb
```diff
@@ -23,11 +23,11 @@ class ActionDispatchSubscriberTest < Minitest::Test
 
     def test_start
       in_transaction do |txn|
-        time = Time.now.to_f
```
kaylareopelle, line 26: So surprised we still had some of these in tests!

test/multiverse/suites/redis/Envfile
```diff
@@ -4,13 +4,6 @@
...
 
-# TODO: Check for redis-clustering gem compatibility with Ruby 4.0
-# Error in the redis-clustering gem
-# Gem Load Error is: Proc's self is not shareable: #<Proc:0x00007f3adcd46840 (lambda)>
-suite_condition('The redis-clustering gem is not compatible with Ruby 4.0') do
-  RUBY_VERSION < '4.0.0'
-end
-
 REDIS_VERSIONS = [
   [nil],
```
kaylareopelle, line 8: Since 6.0 is out, I think we should add a 5.0 testing column here

```suggestion
  [nil],
  ['5.4.1'],
```

hannahramadan (approved):

PR #3633, opened 2026-08-04
Use case-insensitive boolean config validation

### LINKED ISSUE #3632: 10.x silently discards uppercase boolean env var values (e.g. NEW_RELIC_DISTRIBUTED_TRACING_ENABLED=FALSE), reverting the setting to its default
## Description

Agent 9.x parsed boolean environment variables case-insensitively. In `environment_source.rb`:

```ruby
elsif type == NewRelic::Agent::Configuration::Boolean
  if /false|off|no/i.match?(value)   # /i — "FALSE", "Off", "NO" all matched
    self[config_key] = false
```

Agent 10.0 moved coercion into `Manager#type_coerce` ([PR#3341](https://github.com/newrelic/newrelic-ruby-agent/pull/3341)), which matches against a lowercase-only map with a case-sensitive pattern (`manager.rb`):

```ruby
BOOLEAN_MAP = { 'true'=>true, 'yes'=>true, 'on'=>true,
                'false'=>false, 'no'=>false, 'off'=>false }.freeze
# pattern: /^(?:true|yes|on|false|no|off)$/   ← no /i
```

A value like `NEW_RELIC_DISTRIBUTED_TRACING_ENABLED=FALSE` — which disabled distributed tracing on every 9.x release — now fails the pattern, and `default_with_warning` replaces it with the setting's **default**, which for `distributed_tracing.enabled` is `true`. The only signal is one log line. The v10.0.0 changelog entry ("Improve configuration validation and coercion") does not mention that boolean parsing became case-sensitive.

The failure mode is nasty for this particular setting: an explicit opt-out silently becomes an opt-in on upgrade, with direct data-ingest/billing consequences. In our case a routine Ruby upgrade pulled 10.2.0 via an unpinned gem and a production service began emitting ~300 GB/day of span data despite `NEW_RELIC_DISTRIBUTED_TRACING_ENABLED=FALSE` in its deployment manifests. A second sharp edge: the invalid env value does not fall through to the next configuration layer — it shadows a correct `newrelic.yml` value with the built-in default, so even a `distributed_tracing.enabled: false` in the config file cannot rescue an uppercase env var.

## Expected Behavior

Either of:
1. Boolean env var parsing remains case-insensitive (downcase before the map lookup), matching 9.x and matching how most tooling treats boolean-ish env vars; or
2. The case-sensitivity is documented as a breaking change in the v10.0.0 changelog/migration guide, and the warning is escalated for settings where falling back to the default inverts an explicit opt-out.

Option 1 seems strictly better: the accepted token set is unchanged, so downcasing cannot change the meaning of any currently-valid value.

## Troubleshooting or NR Diag results

Agent log on 10.2.0:

```
Received an invalid 'FALSE' value for the 'distributed_tracing.enabled' configuration parameter! Expected to receive a value of type NewRelic::Agent::Configuration::Boolean matching pattern '(?-mix:^(?:true|yes|on|false|no|off)$)'. Using the default value of 'true'.
```

## Steps to Reproduce

```ruby
# gem install newrelic_rpm -v 10.2.0 ; gem install newrelic_rpm -v 9.13.0
ENV['NEW_RELIC_AGENT_ENABLED'] = 'false'
ENV['NEW_RELIC_DISTRIBUTED_TRACING_ENABLED'] = 'FALSE'
gem 'newrelic_rpm', ARGV[0]
require 'newrelic_rpm'
puts NewRelic::Agent.config[:'distributed_tracing.enabled']
```

| agent | result |
|---|---|
| 9.13.0 | `false` |
| 10.2.0 | `true` |

Also reproducible for any Boolean setting with any of `TRUE/FALSE/YES/NO/ON/OFF` uppercase or mixed case.

## Your Environment

- newrelic_rpm 10.2.0 (regression introduced in 10.0.0 via PR#3341; 9.13.0 and 9.14.0 behave as documented)
- Observed under Ruby 3.3.6 in production and reproduced under Ruby 4.0.5; the coercion path is pure Ruby and version-independent
- Env vars supplied via Kubernetes deployment manifests

## Additional context

Uppercase boolean env values are common in the wild (Kubernetes YAML authors frequently write `"TRUE"`/`"FALSE"`), and they were valid for the entire 9.x line, so fleets carrying them will regress one service at a time as upgrades land.

kaylareopelle:
In version 9.x, the agent accepted capitalized boolean values, like "FALSE", and mixed-case values like "True". Version 10.0.0 included #3341, which unintentionally removed the case-insensitive requirement. This caused users who had any casing besides all lowercase to have their configuration options fall back to the defaults. Now, the agent uses case-insensitive checks again. 

Resolves #3632

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@
 
 - **Bugfix: Normalize boolean configuration values to allow all casing**
 
-  In version 9.x, the agent accepted capitalized boolean values, like "FALSE", and mixed-case values like "True". Version 10.0.0 included [PR#3341](https://github.com/newrelic/newrelic-ruby-agent/pull/3341), which unfortunately removed the case insensitive requirement. This caused users who had any casing besides all lowercase to have their configuration options fall back to the defaults. Now, the agent uses case-insensitive checks again. Our thanks go to [@willie](https://github.com/willie) for bringing this to our attention. [Issue#3632](https://github.com/newrelic/newrelic-ruby-agent/issues/3632) [PR#3633](https://github.com/newrelic/newrelic-ruby-agent/issues/3632)
+  In version 9.x, the agent accepted capitalized boolean values, like "FALSE", and mixed-case values like "True". Version 10.0.0 included [PR#3341](https://github.com/newrelic/newrelic-ruby-agent/pull/3341), which unintentionally removed the case-insensitive requirement. This caused users who had any casing besides all lowercase to have their configuration options fall back to the defaults. Now, the agent uses case-insensitive checks again. Our thanks go to [@willie](https://github.com/willie) for bringing this to our attention. [Issue#3632](https://github.com/newrelic/newrelic-ruby-agent/issues/3632) [PR#3633](https://github.com/newrelic/newrelic-ruby-agent/issues/3632)
 
 ## v10.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@
 
   When a response body's first fragment was a frozen `String` and there were multiple fragments, browser instrumentation hit a `FrozenError` and the browser timing header was never injected. This began appearing with `ERB` 6.0.3+, which started freezing more of its compiled strings. This issue is now fixed. Thank you to [@md5](https://github.com/md5) for reporting this issue! [Issue#3624](https://github.com/newrelic/newrelic-ruby-agent/issues/3624)[PR#3625](https://github.com/newrelic/newrelic-ruby-agent/pull/3625)
 
+- **Bugfix: Normalize boolean configuration values to allow all casing**
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,10 @@
...
 
+- **Bugfix: Normalize boolean configuration values to allow all casing**
+
+  In version 9.x, the agent accepted capitalized boolean values, like "FALSE", and mixed-case values like "True". Version 10.0.0 included [PR#3341](https://github.com/newrelic/newrelic-ruby-agent/pull/3341), which unfortunately removed the case insensitive requirement. This caused users who had any casing besides all lowercase to have their configuration options fall back to the defaults. Now, the agent uses case-insensitive checks again. Our thanks go to [@willie](https://github.com/willie) for bringing this to our attention. [Issue#3632](https://github.com/newrelic/newrelic-ruby-agent/issues/3632) [PR#3633](https://github.com/newrelic/newrelic-ruby-agent/issues/3632)
```
```suggestion
  In version 9.x, the agent accepted capitalized boolean values, like "FALSE", and mixed-case values like "True". Version 10.0.0 included [PR#3341](https://github.com/newrelic/newrelic-ruby-agent/pull/3341), which unintentionally removed the case-insensitive requirement. This caused users who had any casing besides all lowercase to have their configuration options fall back to the defaults. Now, the agent uses case-insensitive checks again. Our thanks go to [@willie](https://github.com/willie) for bringing this to our attention. [Issue#3632](https://github.com/newrelic/newrelic-ruby-agent/issues/3632) [PR#3633](https://github.com/newrelic/newrelic-ruby-agent/issues/3632)
```

hannahramadan (approved):

PR #3637, opened 2026-08-07
Fix intermittently failing `test_segment_can_disable_scoped_metric_recording`

### LINKED ISSUE #2429: Fix intermittently failing segment unit test, `test_segment_can_disable_scoped_metric_recording`
For a few months, we've had overnight CI failures due to `NewRelic::Agent::Transaction::SegmentTest#test_segment_can_disable_scoped_metric_recording`. In the past week, this test has failed on Ruby versions as old as 2.5.9 and as new as 3.2.2. Rerunning the failed jobs usually makes the test pass on the subsequent run.

The test was initially skipped in #2428.

hannahramadan:
Reset metrics between tests to avoid bleed over from other tests. Ran a few times and I think this should work, but we can keep an eye on the CI in case this starts failing again.

closes #2429

### DIFF
--- a/test/new_relic/agent/transaction/segment_test.rb
+++ b/test/new_relic/agent/transaction/segment_test.rb
@@ -14,6 +14,7 @@ module NewRelic
           NewRelic::Agent.config.add_config_for_testing(@additional_config)
           NewRelic::Agent.config.notify_server_source_added
           nr_freeze_process_time
+          clear_metrics!
           reset_buffers_and_caches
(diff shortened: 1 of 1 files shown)


test/new_relic/agent/transaction/segment_test.rb
hannahramadan, line 61: "segementland" lol

tannalynn (approved):

PR #3642, opened 2026-08-17
Fix ArgumentError on multi-key operations with dalli 5.1.0

dbackeus:
# Overview

Dalli 5.1.0 added an optional request options argument to `Dalli::Protocol::Meta#read_multi_req` and `Dalli::Protocol::Base#pipelined_get` and now always calls them with two arguments (petergoldstein/dalli#1147, petergoldstein/dalli#1154). The agent redefines both methods with a single `keys` parameter, so on dalli 5.1.0 every multi-key operation (`get_multi`, `get_multi_cas`, `get_multi_with_metadata`) raises:

```
ArgumentError: wrong number of arguments (given 2, expected 1)
  dalli-5.1.0/lib/dalli/protocol/base.rb:45:in 'request'
  newrelic_rpm-10.7.0/lib/new_relic/agent/instrumentation/memcache/prepend.rb:80:in 'read_multi_req'
```

This changes the four affected method definitions (the prepend and chain variants of `read_multi_req` and `pipelined_get`) to accept and forward any additional arguments. The tracing itself only uses `keys`, so behavior is unchanged on older dalli versions.

We hit this in production after upgrading to newrelic_rpm 10.7.0 alongside dalli 5.1.0.

Submitter Checklist:
- [x] Add new tests for your change, if applicable

# Testing

Without the fix, the memcache multiverse suite fails with 7 errors on dalli 5.1.0, reproducing the ArgumentError above. Added a `get_multi_cas` test to cover the `pipelined_get` path, which dalli 5.1.0 routes through the pipelined getter even with a single server (the existing tests cover the `read_multi_req` path).

Ran the memcache multiverse suite locally against dalli 5.1.0, 4.3.3, and 3.1.0 in both prepend and chain modes; all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/new_relic/agent/instrumentation/memcache/dalli.rb
+++ b/lib/new_relic/agent/instrumentation/memcache/dalli.rb
@@ -57,8 +57,9 @@ module NewRelic
 
               alias_method(:read_multi_req_without_newrelic_trace, :read_multi_req)
 
-              def read_multi_req(keys)
-                send_multiget_with_newrelic_tracing(keys) { read_multi_req_without_newrelic_trace(keys) }
+              # Dalli 5.1.0 added an optional options argument
+              def read_multi_req(keys, *args)
+                send_multiget_with_newrelic_tracing(keys) { read_multi_req_without_newrelic_trace(keys, *args) }
               end
             end
           end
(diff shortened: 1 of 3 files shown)


tannalynn (approved):
Thank you very much for submitting a fix for this issue!

dbackeus:
Thank you very much for the quick merge 🙏 

Looking forward to release 🚀

PR #3645, opened 2026-08-19
Allowlists don't need to be case sensitive

### LINKED ISSUE #3613: Add :allowlist for relevant configs AND case de-sensitize allowlist
A part of this ticket is allowing the allowlist configs to be case insensitive.

As pointed out in #3611, we have 3 configs that only accept fixed values, but lack the helpful `:allowlist` option:

- `log_level`
- `transaction_tracer.record_sql`
- `slow_sql.record_sql`

For each of these, an invalid value doesn't error or warn. It falls back to the default:

- `log_level` does `LOG_LEVELS.fetch(level.to_s.downcase, ::Logger::INFO)`, so an unknown becomes `INFO`
- `transaction_tracer.record_sql` and `slow_sql.record_sql` both call `Database.record_sql_method` which has a `case` statement with an `else` branch that returns `:obfuscated`

Adding an :allowlist gives a warning for invalid values and still just falls back to the default. We then can also delete the `BASE_ENUM_OVERRIDES` value in `generate_schema.rb` (yet to be merged as of this ticket writing).

Note: Make sure to handle aliases.

`application_logging.forwarding.log_level` has an allowlist of 
`:allowlist => %w[debug info warn error fatal unknown DEBUG INFO WARN ERROR FATAL UNKNOWN]`. We should make sure any aliases for `log_level` are captured like there. The sql methods need to include `false`, which the case statement in `Database.record_sql_method` call already handles.

hannahramadan:
Configs with allowlists now can pass through an additional screening that ignores case, making allowlists case-insensitive.

Adds allowlists for:
- log_level
- transaction_tracer.record_sql
- slow_sql.record_sql

closes #3613

### DIFF
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -576,7 +576,7 @@ module NewRelic
           :public => true,
           :type => String,
           :allowed_from_server => true,
-          :allowlist => %w[obfuscated raw none off false],
+          :allowlist => ['obfuscated', 'raw', 'none', 'off', 'false', false],
           :description => <<~DESCRIPTION
             Obfuscation level for SQL queries reported in transaction trace nodes. By default, this is set to `obfuscated`, which strips out the numeric and string literals.
 
@@ -2197,7 +2197,7 @@ module NewRelic
           :public => true,
           :type => String,
           :allowed_from_server => true,
-          :allowlist => %w[obfuscated raw none off false],
+          :allowlist => ['obfuscated', 'raw', 'none', 'off', 'false', false],
           :description => 'Defines an obfuscation level for slow SQL queries. Valid options are `obfuscated`, `raw`, or `none`.'
         },
         :'slow_sql.use_longer_sql_id' => {
--- a/lib/new_relic/agent/configuration/default_source.rb
+++ b/lib/new_relic/agent/configuration/default_source.rb
@@ -303,6 +303,7 @@ module NewRelic
           :public => true,
           :type => String,
           :allowed_from_server => false,
+          :allowlist => %w[debug info warn error],
           :description => 'Sets the level of detail of log messages. Possible log levels, in increasing verbosity, are: `error`, `warn`, `info` or `debug`.'
         },
         # General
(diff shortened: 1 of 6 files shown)


lib/new_relic/agent/configuration/manager.rb
```diff
@@ -274,10 +274,19 @@ def enforce_allowlist(key, value)
           return value unless allowlist = default_source.allowlist_for(key)
           return value if allowlist.include?(value)
 
+          value = normalized_case_match(allowlist, value)
+          return value unless value.nil?
+
           default_with_warning(key, value, 'Expected to receive a value found on the following list: ' \
                                ">>#{allowlist}<<, but received '#{value}'.")
```
tannalynn: by the time it reaches `default_with_warning` here, I think the value always be nil, bc of the return on 278 (and the original `value` is being overwritten on 277)

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -575,6 +576,7 @@ def self.convert_to_constant_list(string_array)
           :public => true,
           :type => String,
           :allowed_from_server => true,
+          :allowlist => ['obfuscated', 'raw', 'none', 'off', 'false', false],
```
tannalynn: since this config is a String type, would having a bool `false` on the allowlist cause issues?

lib/new_relic/agent/configuration/default_source.rb
```diff
@@ -2195,6 +2197,7 @@ def self.convert_to_constant_list(string_array)
...
```
tannalynn: Also a boolean `false` here too

lib/new_relic/agent/configuration/manager.rb
hannahramadan: Ah great catch! Fixed this https://github.com/newrelic/newrelic-ruby-agent/pull/3645/changes/17325e19a388feb4fdda00e9e202788d5380b5f0

lib/new_relic/agent/configuration/default_source.rb
hannahramadan: I didn't see any issues with it, but also, I wouldn't want to release something that messes up this config since it could carry sensitive data. Just accepting strings for this like before https://github.com/newrelic/newrelic-ruby-agent/pull/3645/changes/5c6fa6608ef46b2c5ed32365d74af0d22431eecd

hannahramadan: ooo okay I reverted this. Looks like we have [a test](https://github.com/newrelic/newrelic-ruby-agent/blob/e5f3218c1e8755b0a7c8f0d939c8ce2c540c3e41/test/new_relic/agent/agent/connect_test.rb#L136) that allows for a `false` boolean to be treated as if turning .record_sql off.

tannalynn (approved):

kaylareopelle (approved):

PR #3663, opened 2026-09-01
Add support for `browser_monitoring.version` config

hannahramadan:
Adds support for new configuration option, `browser_monitoring.version`. 

Customers may specificy which version of the browser agent to use. If the specified version cannot be resolved (out of support, malformed, etc.), browser monitoring will not be installed. See the [browser agent EOL policy](https://docs.newrelic.com/docs/browser/browser-monitoring/getting-started/browser-agent-eol-policy/) for the latest supported versions. 

If unset, the latest browser monitoring version will be used (default).

### DIFF
--- a/lib/new_relic/agent/javascript_instrumentor.rb
+++ b/lib/new_relic/agent/javascript_instrumentor.rb
@@ -29,7 +29,6 @@ module NewRelic
       end
 
       def warn_if_requested_version_not_used
-        binding.irb
         return if missing_config?(:'browser_monitoring.version')
         return unless missing_config?(:js_agent_loader)
 
--- a/lib/new_relic/agent/javascript_instrumentor.rb
+++ b/lib/new_relic/agent/javascript_instrumentor.rb
@@ -24,6 +24,19 @@ module NewRelic
         if !NewRelic::Agent.config[:'rum.enabled']
           NewRelic::Agent.logger.debug('Real User Monitoring is disabled for this agent. Edit your configuration to change this.')
         end
+
+        warn_if_requested_version_not_used
+      end
+
+      def warn_if_requested_version_not_used
+        binding.irb
+        return if missing_config?(:'browser_monitoring.version')
+        return unless missing_config?(:js_agent_loader)
+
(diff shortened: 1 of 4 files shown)


lib/new_relic/agent/javascript_instrumentor.rb
```diff
@@ -24,6 +24,18 @@ def log_configuration
...
+
+      def warn_if_requested_version_not_used
+        return if missing_config?(:'browser_monitoring.version')
+        return unless missing_config?(:js_agent_loader)
+
+        requested_version = NewRelic::Agent.config[:'browser_monitoring.version']
+        NewRelic::Agent.logger.warn("Requested browser_monitoring.version '#{requested_version}' did not return a " \
+          "browser agent loader; please ensure '#{requested_version}' is available and supported. No browser " \
+          'monitoring will be injected for this connect.')
       end
```
kaylareopelle: So we don't actually need to apply the version anywhere because the connect response will still send the version to us in the configs we're already checking?

lib/new_relic/agent/javascript_instrumentor.rb
```diff
@@ -24,6 +24,18 @@ def log_configuration
         if !NewRelic::Agent.config[:'rum.enabled']
           NewRelic::Agent.logger.debug('Real User Monitoring is disabled for this agent. Edit your configuration to change this.')
         end
+
+        warn_if_requested_version_not_used
+      end
+
+      def warn_if_requested_version_not_used
+        return if missing_config?(:'browser_monitoring.version')
+        return unless missing_config?(:js_agent_loader)
```
tannalynn, line 33: What if browser monitoring just isn't turned on at all? wouldnt this still do the warning in that case?

hannahramadan: This would hit only if someone set `browser_monitoring.version`. Its forsure not full-proof bc yeah there could be a number of reasons this fails. 

My hope is that is someone did add the config to set and then it also failed, that the reason if failed is more likely a versioning issue with their configuration vs them like disabling it.

tannalynn (approved):

kaylareopelle (approved):

PR #3668, opened 2026-09-08
Fix the ci

tannalynn:
[full ci run](https://github.com/newrelic/newrelic-ruby-agent/actions/runs/34260922455)


test/fixtures/cross_agent_tests/distributed_tracing/trace_context.json
```diff
@@ -159,7 +159,6 @@
     "inbound_headers": [
       {
         "traceparent": "00-0af7651916cd43dd8448eb211c80319c-00f067aa0ba902b7-00",
-        // "tracestate": "33@nr=0-0-33-2827902-7d3efb1b173fecfa--0-1.2-1518469636035",
```
tannalynn, line 162: so apparently this isn't actually valid json lol. the old json gem was just less strict about it than 3.0 is. So it would pretend it was fine and treated it like a comment, so it ignored that line.

test/multiverse/suites/config_file_loading/Envfile
```diff
@@ -4,41 +4,19 @@
...
-  ['5.2.0', 2.7, 3.4],
-  ['5.1.0', 2.6, 3.3],
-  ['5.0.0', 2.6, 3.3],
-  ['4.0.0', 2.6, 3.3],
-  ['3.3.0', 2.6, 3.3]
-]
+# Testing multiple pinned psych/stringio versions in the same job used to cause
+# Bundler "already activated" errors, since they're default gems and only one
+# version can be active per process. Rather than fight that, just test whatever
+# psych/stringio veresions ship by default
```
tannalynn: with old rubygems versions, bundler was less angry about this and it was fine. Newer versions though have bundler being much stricter about default gems. Changing this to just test whatever ruby versions default is imo the best solution at this point.

hannahramadan:
```suggestion
# psych/stringio versions ship by default
```

hannahramadan (dismissed):

hannahramadan (approved):

